PR #4, opened 2019-04-11
Drop envoy-perf/nighthawk (unmodified, won't build yet)

oschaaf:


oschaaf:
So -- trying to get this to build for a follow-up PR -- now that we have moved `nighthawk/` into `/`, there seems to be a problem. It looks like some `.cc` files cannot find includes residing in the same directory. `envoy_include_prefix()` looks suspect, as it specializes treatment of `source/` and `include/` which now  matches `nighthawk` code. Will look into to this some more, but pending a resolution, it might make sense to hold off review for this.

oschaaf:
This is ready for review. Follow up to this is staged in https://github.com/envoyproxy/nighthawk/pull/5

PR #8, opened 2019-04-15
Simple test origin

oschaaf:
Introduces a test origin for Nighthawk, based on Envoy.

Current functionality is simple: the request headers can be used to indicate how many bytes the response body should have, and response headers can be added/appended. The response consist of repeating the character `a`.

Deprecates envoyproxy/envoy-perf#70

### DIFF
--- a/source/server/BUILD
+++ b/source/server/BUILD
@@ -1,0 +1,27 @@
+load(
+    "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+envoy_cc_library(
+    name = "http_test_server_filter_lib",
+    srcs = ["http_test_server_filter.cc"],
+    hdrs = ["http_test_server_filter.h"],
+    repository = "@envoy",
+    deps = [
+        "//api/server:response_options_proto_cc",
+        "@envoy//source/exe:envoy_common_lib",
+    ],
+)
+
+envoy_cc_library(
+    name = "http_test_server_filter_config",
+    srcs = ["http_test_server_filter_config.cc"],
+    repository = "@envoy",
+    deps = [
+        ":http_test_server_filter_lib",
+        "@envoy//include/envoy/server:filter_config_interface",
+    ],
+)
--- a/source/server/http_test_server_filter.cc
+++ b/source/server/http_test_server_filter.cc
@@ -22,35 +22,46 @@ HttpTestServerDecoderFilter::HttpTestServerDecoderFilter(
 
 void HttpTestServerDecoderFilter::onDestroy() {}
 
+bool HttpTestServerDecoderFilter::mergeJsonConfig(std::string json,
+                                                  nighthawk::server::ResponseOptions& config,
+                                                  std::string& error_message) {
+  error_message = "";
+  try {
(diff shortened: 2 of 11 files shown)


source/server/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    "envoy_cc_library",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+envoy_cc_library(
+    name = "http_test_server_filter_lib",
+    srcs = ["http_test_server_filter.cc"],
+    hdrs = ["http_test_server_filter.h"],
+    repository = "@envoy",
```
htuch: Why do we need this `@envoy` business?

source/server/http_test_server_filter.cc
```diff
@@ -0,0 +1,92 @@
...
+  error_message = "";
+  try {
+    nighthawk::server::ResponseOptions json_config;
+    Envoy::MessageUtil::loadFromJson(std::string(json), json_config);
+    config.MergeFrom(json_config);
+    Envoy::MessageUtil::validate(config);
+  } catch (Envoy::EnvoyException exception) {
+    error_message = fmt::format("Error merging json config: {}", exception.what());
+  }
+  return error_message.empty();
```
htuch: Seems possibly fragile, since sometimes exceptions might have an empty string message. Can we make `error_message` an `absl::optional<std::string>`, or something like this?

source/server/http_test_server_filter.cc
```diff
@@ -0,0 +1,92 @@
...
+}
+
+Envoy::Http::FilterHeadersStatus
+HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool) {
+  const auto* request_config_header = headers.get(TestServer::HeaderNames::get().TestServerConfig);
+  nighthawk::server::ResponseOptions base_config = config_->server_config();
+  std::string error_message;
+
+  if (!request_config_header ||
+      mergeJsonConfig(request_config_header->value().c_str(), base_config, error_message)) {
```
htuch: Is there a TODO for being able to clear fields as well?

source/server/BUILD
oschaaf: Without it we get a batch of errors. the first one looks like
```
ERROR: /home/oschaaf/code/envoy-perf-vscode/nighthawk/test/server/BUILD:8:1: no such target '//bazel:windows_x86_64': target 'windows_x86_64' n
ot declared in package 'bazel' defined by /home/oschaaf/code/envoy-perf-vscode/nighthawk/bazel/BUILD and referenced by '//test/server:http_test_server_filter_integration_test'
```

source/server/http_test_server_filter.cc
oschaaf: TLDR; done

(Right now, it's not possible to get an emptry string I think, but absl::optional looks like it would be the better api and more resilient towards later breakage when maintaining..)

source/server/http_test_server_filter.cc
oschaaf: Noted

oschaaf:
@htuch Ugh, clang tidy flakes again, it's almost structural now :-(
this PR is ready for another pass though

PR #9, opened 2019-04-15
 Connection prefetching feature

oschaaf:
Enabling connection prefetching will cause the client to open the requested amount of connections to the benchmark target, before performing the actual benchmark. The prefetching is only performed intially.

### DIFF
--- a/source/client/benchmark_client_impl.h
+++ b/source/client/benchmark_client_impl.h
@@ -12,8 +12,10 @@
 #include "envoy/upstream/upstream.h"
 
 #include "nighthawk/client/benchmark_client.h"
+#include "nighthawk/client/prefetchable_pool.h"
 #include "nighthawk/common/sequencer.h"
 #include "nighthawk/common/statistic.h"
+#include "nighthawk/common/uri.h"
 
 #include "common/common/logger.h"
 #include "common/http/header_map_impl.h"
@@ -21,7 +23,6 @@
 
 #include "client/stream_decoder.h"
 #include "common/ssl.h"
-#include "nighthawk/common/uri.h"
 
 namespace Nighthawk {
 namespace Client {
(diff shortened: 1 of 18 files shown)


oschaaf:
/assign @htuch - this is ready for another pass
(the pending comments from https://github.com/envoyproxy/envoy-perf/pull/69 have been addressed)

source/client/benchmark_client_impl.h
```diff
@@ -67,12 +69,15 @@ class BenchmarkClientHttpImpl : public BenchmarkClient,
...
   bool tryStartOne(std::function<void()> caller_completion_callback) override;
   Envoy::Stats::Store& store() const override { return store_; }
-
   // StreamDecoderCompletionCallback
   void onComplete(bool success, const Envoy::Http::HeaderMap& headers) override;
   void onPoolFailure(Envoy::Http::ConnectionPool::PoolFailureReason reason) override;
 
 private:
+  void prefetchPoolConnections() override {
+    dynamic_cast<PrefetchablePool*>(pool_.get())->prefetchConnections();
```
htuch: Can you just make `pool_` of type `PrefetchablePool` to avoid the dynamic cast?

test/options_test.cc
```diff
@@ -107,6 +109,34 @@ TEST_F(OptionsImplTest, BadH2FlagThrows) {
       MalformedArgvException, "Couldn't find match for argument");
 }
 
+TEST_F(OptionsImplTest, H2Flag) {
```
htuch: Please add `//` single line human comments above each `TEST_F`.

source/client/benchmark_client_impl.h
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/9/commits/ce508385831a7fb0fb319abf155a2b70011aae2a; does this look better?

test/options_test.cc
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/9/commits/b388e0fe4af9121aaa40c2c167fbd5d338bb4f41; also noticed that some duplication crept in unnoticed, probably because of me missing that while resolving merge conflicts.

PR #14, opened 2019-04-18
Add BurstingRateLimiter for simulating super-bursty traffic.  

oschaaf:
Introduces a new `--burst-size` option. Implemented via `BurstingRateLimiter`, which accumulates acquisitions by the underlying configured `RateLimiter`, and starts releasing when the burst-size threshold is met. For example, `--burst-size 100 --rps 1000` will make the client release 100 requests every 100 milliseconds. 

### DIFF
--- a/ci/run_coverage.sh
+++ b/ci/run_coverage.sh
@@ -32,7 +32,7 @@ if [ "$VALIDATE_COVERAGE" == "true" ]
 then
     COVERAGE_VALUE=$(grep -Po '.*lines[.]*: \K(\d|\.)*' "${COVERAGE_SUMMARY}")
     # TODO(oschaaf): The target is 97.5%, so up this whenever possible in follow ups.
-    COVERAGE_THRESHOLD=96.6
+    COVERAGE_THRESHOLD=97.2
     COVERAGE_FAILED=$(echo "${COVERAGE_VALUE}<${COVERAGE_THRESHOLD}" | bc)
     
     echo "HTML coverage report is in ${COVERAGE_DIR}/coverage.html"
--- a/ci/run_coverage.sh
+++ b/ci/run_coverage.sh
@@ -32,7 +32,7 @@ if [ "$VALIDATE_COVERAGE" == "true" ]
 then
     COVERAGE_VALUE=$(grep -Po '.*lines[.]*: \K(\d|\.)*' "${COVERAGE_SUMMARY}")
     # TODO(oschaaf): The target is 97.5%, so up this whenever possible in follow ups.
-    COVERAGE_THRESHOLD=97.2
+    COVERAGE_THRESHOLD=96.8
     COVERAGE_FAILED=$(echo "${COVERAGE_VALUE}<${COVERAGE_THRESHOLD}" | bc)
     
     echo "HTML coverage report is in ${COVERAGE_DIR}/coverage.html"
--- a/source/client/factories_impl.cc
+++ b/source/client/factories_impl.cc
@@ -41,6 +41,10 @@ SequencerPtr SequencerFactoryImpl::create(Envoy::TimeSource& time_source,
   StatisticFactoryImpl statistic_factory(options_);
   RateLimiterPtr rate_limiter =
       std::make_unique<LinearRateLimiter>(time_source, Frequency(options_.requestsPerSecond()));
+  const uint64_t burst_size = options_.burstSize();
+  if (burst_size) {
+    rate_limiter = std::make_unique<BurstingRateLimiter>(std::move(rate_limiter), burst_size);
+  }
   SequencerTarget sequencer_target = [&benchmark_client](std::function<void()> f) -> bool {
     return benchmark_client.tryStartOne(f);
   };
--- a/source/common/rate_limiter_impl.cc
+++ b/source/common/rate_limiter_impl.cc
@@ -12,7 +12,14 @@ BurstingRateLimiter::BurstingRateLimiter(RateLimiterPtr&& rate_limiter, const ui
 }
 
 bool BurstingRateLimiter::tryAcquireOne() {
+  // Callers may not be able to actually make use of a successfull acquisition, and
+  // call releaseOne() to indicate so. We remember state here so we can use that later when
+  // that happens to restore it.
+  previously_releasing_ = releasing_;
+
   if (releasing_) {
+    // When releasing_ we drain the accumulated total until there's nothing left. We'll transit to
+    // accumulating mode after that.
     accumulated_--;
     releasing_ = accumulated_ > 0;
     return true;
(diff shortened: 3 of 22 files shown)


oschaaf:
Rebased to master and squashed. Has an extra commit to fix the build that broke after the rebase:
- move to HeaderString::getStringView()

Force pushed the squashed version, but I'm leaving a note here to make sure I don't forget about the earlier ASAN failure in CI: 
https://circleci.com/gh/envoyproxy/nighthawk/209. This seems worth looking into, but it seems unlikely that it was related to this PR.

May need tweaking the coverage threshold in here, it may err because it didn't consider the test origin when I adjusted it earlier.

/assign @htuch

oschaaf:
It looks like the hack to avoid runtime conflicts in the integration test for the client came back to haunt us :(
That test now flakes with `assert failure: thread_advancing_time_->isCurrentThreadId(). Details: time should only be advanced on one thread in the context of a test`.

oschaaf:
Reading the code, there's a comment in the `BaseIntegrationTest` which remarks about a hack where it calls `sleep()` [1]. I think that is where the time system initializes its thread-id for further checking later on.
If it wouldn't do that we would be fine... Unfortunately, the constructor that accepts a time system ignores whatever we pass in, so it seems that we also cannot pass in a different time system that doesn't perform this check to avoid the assert. 

[1] https://github.com/envoyproxy/envoy/blob/b41ba5925a4e93d22a86c6501d63314ccf0d79f3/test/integration/integration.cc#L239

oschaaf:
I think we should land https://github.com/envoyproxy/nighthawk/pull/15 first, and then rebase this.

oschaaf:
Note: just merged master into this, which should clean up the diff now that #15 was merged over there.

htuch:
@oschaaf can you merge master again? Thanks

oschaaf:
@htuch done, merged master

ci/run_coverage.sh
```diff
@@ -32,7 +32,7 @@ if [ "$VALIDATE_COVERAGE" == "true" ]
 then
     COVERAGE_VALUE=$(grep -Po '.*lines[.]*: \K(\d|\.)*' "${COVERAGE_SUMMARY}")
     # TODO(oschaaf): The target is 97.5%, so up this whenever possible in follow ups.
-    COVERAGE_THRESHOLD=96.6
+    COVERAGE_THRESHOLD=96.8
```
htuch: Huzzah!

source/common/rate_limiter_impl.cc
```diff
@@ -6,6 +6,39 @@
...
+  if (releasing_) {
+    accumulated_--;
+    releasing_ = accumulated_ > 0;
+    return true;
+  } else {
+    // We are greedy here, which should help with timely releases of bursts when
+    // the system is lagging behind.
+    while (rate_limiter_->tryAcquireOne()) {
+      accumulated_++;
+      if ((accumulated_ % burst_size_) == 0) {
```
htuch: The use of `%` here together with both ++ on release and acquire is clever; it took me a minute to grok though, could you add a comment or two?

source/common/rate_limiter_impl.cc
```diff
@@ -6,6 +6,39 @@
...
+  }
+  return false;
+}
+
+void BurstingRateLimiter::releaseOne() {
+  if (releasing_) {
+    accumulated_++;
+    ASSERT(accumulated_ <= burst_size_);
+  } else {
+    rate_limiter_->releaseOne();
```
htuch: Is it ever possible to execute this line?

source/common/rate_limiter_impl.h
```diff
@@ -10,6 +10,28 @@
 
 namespace Nighthawk {
 
+/**
+ * BurstingRatelimiter can be wrapped around another rate limiter. It has two modes:
+ * 1. First it will be accumulating acquisitions by forwarding calls to the wrapped
+ *    rate limiter, until the accumulated acquisitions equals the specified burst size.
+ * 2. Release mode. In this mode, BatchingRateLimiter is in control and will be handling
+ *    acquisition calls (returning true and substracting from the accumulated total until
```
htuch: It cleverly uses increment to subtract! :)

test/rate_limiter_test.cc
```diff
@@ -39,4 +41,78 @@ TEST_F(RateLimiterTest, LinearRateLimiterInvalidArgumentTest) {
...
 }
 
+TEST_F(RateLimiterTest, BurstingRateLimiterTest) {
+  const uint64_t burst_size = 3;
+  std::unique_ptr<MockRateLimiter> mock_rate_limiter = std::make_unique<MockRateLimiter>();
+  MockRateLimiter& unsafe_mock_rate_limiter = *mock_rate_limiter;
+  EXPECT_CALL(unsafe_mock_rate_limiter, tryAcquireOne)
+      .Times(burst_size)
+      .WillRepeatedly(Return(true))
+      .RetiresOnSaturation();
```
htuch: What is `RetireOnSaturation`? Is this effectively the same as `InSequence`?

source/common/rate_limiter_impl.cc
oschaaf: Done

source/common/rate_limiter_impl.cc
oschaaf: No, good point, this didn't make sense. I tightened this up.

test/rate_limiter_test.cc
oschaaf: Thanks, `InSequence` specifies the intent here more clearly and is more compact, so I switched it over to use that.

oschaaf:
@htuch this is ready for another round

source/common/rate_limiter_impl.cc
```diff
@@ -6,6 +6,50 @@
...
+  }
+
+  previously_releasing_ = absl::nullopt;
+  return false;
+}
+
+void BurstingRateLimiter::releaseOne() {
+  ASSERT(accumulated_ < burst_size_);
+  ASSERT(previously_releasing_ != absl::nullopt && previously_releasing_ == true);
+  // The caller wasn't able to put its earlier successfull acquisition to good use, so we restore
```
htuch: Oh, this is actually different to what I thought you were doing. I thought you were doing:
* During acquisition, keep increasing accumulated until % burst_size == 0.
* During release, keep increasing accumulated until % burst_size ==0.

If you ++ and --, won't you run the risk of getting non-bursty behavior when released tokens get immediately reacquired and a thread spends a long period without requesting a burst?

source/client/factories_impl.cc
```diff
@@ -41,6 +41,10 @@ SequencerPtr SequencerFactoryImpl::create(Envoy::TimeSource& time_source,
   StatisticFactoryImpl statistic_factory(options_);
   RateLimiterPtr rate_limiter =
       std::make_unique<LinearRateLimiter>(time_source, Frequency(options_.requestsPerSecond()));
+  const uint64_t burst_size = options_.burstSize();
+  if (burst_size) {
```
htuch: Nit: prefer `burst_size > 0`

PR #16, opened 2019-04-22
Optionize ip-version selection in the client

oschaaf:
Adds a new `--address-family` option that controls behaviour of
the target host-to-address resolution process in the Nighthawk client.
Acceptable values are `[v4,v6,auto]`, defaults to `ipv4`.

https://github.com/envoyproxy/nighthawk/pull/15 needs to go in first.

### DIFF
--- a/test/client_test.cc
+++ b/test/client_test.cc
@@ -93,6 +93,14 @@ public:
     return fmt::format("http://{}:{}/", address, port_);
   }
 
+  const char* getAddressFamilyOptionString() {
+    auto ip_version = GetParam();
+    RELEASE_ASSERT(ip_version == Envoy::Network::Address::IpVersion::v4 ||
+                       ip_version == Envoy::Network::Address::IpVersion::v6,
+                   "bad ip version");
+    return ip_version == Envoy::Network::Address::IpVersion::v6 ? "v6" : "v4";
+  }
+
   int port_;
   pid_t pid_;
   int fd_port_[2];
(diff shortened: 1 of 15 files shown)


test/client_test.cc
```diff
@@ -93,6 +93,14 @@ class ClientTest : public Envoy::BaseIntegrationTest,
     return fmt::format("http://{}:{}/", address, port_);
   }
 
+  const char* getAddressFamilyOptionString() {
+    auto ip_version = GetParam();
+    RELEASE_ASSERT(ip_version == Envoy::Network::Address::IpVersion::v4 ||
```
htuch: Don't we want to `throw` and print the error in the main CLI driver in a user friendly way?

test/options_test.cc
```diff
@@ -183,6 +186,27 @@ TEST_F(OptionsImplTest, VerbosityValuesAreConstrained) {
                           MalformedArgvException, "Value 'foo' does not meet constraint");
 }
 
+/// -------
```
htuch: ?

test/client_test.cc
oschaaf: We do so, see `ParseAddressFamilyBadValues` below. This is just for test.

oschaaf: (this assert is primarily meant to catch when it is time for maintenance, `GetParam()` would be returning something we didn't expect.

oschaaf:
@htuch comments addressed, ready for another look.

PR #19, opened 2019-04-23
Update CI image

oschaaf:
Experimentally update the CI image we use to what Envoy uses, to see if our CI works with that.


oschaaf:
Looking at the details of the failures, it looks like it will require some work to get NH's CI to work with the latest Envoy build image. Closing this for now.

Issue #20, opened 2019-04-23
Request header attribute parameterization


oschaaf:
Allow control of request attributes for requests send by `Nighthawk`:
- Request method (POST, HEAD, etc)
- Headers
- Request entity bodies

PR #61, opened 2019-04-24
Static request parameterization

oschaaf:
Adds static request parameterization options to NH:
- Request method (`--request-method FOO`)
- Request headers (`--request-header foo:bar`)
- Request entity body size is adjustable (`--request-body-size 123`)

This fixes https://github.com/envoyproxy/nighthawk/issues/20

Moves some proto's into to where they belong: `/api/client` (tech-debt).
Fixes https://github.com/envoyproxy/nighthawk/issues/56

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -6,12 +6,11 @@ import "google/protobuf/duration.proto";
 import "validate/validate.proto";
 import "envoy/api/v2/core/base.proto";
 
-
 message RequestOptions {
-  string request_method = 1;
+  envoy.api.v2.core.RequestMethod request_method = 1;
   repeated envoy.api.v2.core.HeaderValueOption request_headers = 2;
-  oneof oneof_request_size { 
-    uint32 request_size = 3 [(validate.rules).uint32 = {gte: 0, lte: 4194304}];
+  oneof oneof_request_body_size { 
+    uint32 request_body_size = 3 [(validate.rules).uint32 = {gte: 0, lte: 4194304}];
   }
 }
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -9,10 +9,9 @@ import "envoy/api/v2/core/base.proto";
 message RequestOptions {
   envoy.api.v2.core.RequestMethod request_method = 1;
   repeated envoy.api.v2.core.HeaderValueOption request_headers = 2;
-  oneof oneof_request_body_size { 
-    uint32 request_body_size = 3 [(validate.rules).uint32 = {gte: 0, lte: 4194304}];
-  }
+  uint32 request_body_size = 3 [(validate.rules).uint32 = {gte: 0, lte: 4194304}];
 }
+
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
 message CommandLineOptions {
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -38,8 +38,10 @@ message CommandLineOptions {
   uint64 burst_size = 10;
   // See :option:`--address-family` for details.
   string address_family = 11;
-  // See :option:`--request_options` for details.
-  RequestOptions request_options = 12;
+  oneof oneof_request_options {
+    // See :option:`--request_options` for details.
+    RequestOptions request_options = 12;
+  }
   // See :option:`--uri` for details.
   string uri = 13;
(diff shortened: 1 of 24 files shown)


api/client/options.proto
```diff
@@ -3,7 +3,17 @@ syntax = "proto3";
 package nighthawk.client;
 
 import "google/protobuf/duration.proto";
+import "validate/validate.proto";
+import "envoy/api/v2/core/base.proto";
 
+
+message RequestOptions {
+  string request_method = 1;
```
htuch: Should this be an enum? See envoy.api.v2.core.RequestMethod, you can probably reuse that.

api/client/options.proto
```diff
@@ -3,7 +3,17 @@ syntax = "proto3";
...
+  repeated envoy.api.v2.core.HeaderValueOption request_headers = 2;
+  oneof oneof_request_size {
```
htuch: Why is this a `oneof`? I think you explained this before but I've forgotten.

api/client/options.proto
```diff
@@ -29,6 +39,8 @@ message CommandLineOptions {
   uint64 burst_size = 10;
   // See :option:`--address-family` for details.
   string address_family = 11;
+  // See :option:`--request_options` for details.
+  RequestOptions request_options = 12;
```
htuch: Is the idea that these will be delivered dynamically in the future in the same format as well? How do you think this will work with non-uniform, will this be `oneof` with some non-uniform proto config?

source/client/options_impl.cc
```diff
@@ -78,6 +80,24 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "default output format is 'v4'.",
       false, "v4", &address_families_allowed, cmd);
 
+  std::vector<std::string> request_methods = {"GET",    "HEAD",    "POST",    "PUT",
```
htuch: I'm almost wondering if we should have some way to automatically generated `tclap` (or other) human friendly flags from proto. We're doing a lot of this! :)

api/client/options.proto
oschaaf: Done

api/client/options.proto
oschaaf: Using `oneof` is bit of a hack that makes merging work well. So here we can have a default config that specifies we should upload a request body with a certain size by default (hypothetically), and one can merge in '0' to override that. 
I couldn't get this to work with a wrapped type.
We discussed something similar over at https://github.com/envoyproxy/envoy-perf/pull/70/files/3a8f22b235b87a8439b51047ee6e86a88ce442bf#diff-b183cebb7f1f309859fc6a3137456da8

api/client/options.proto
oschaaf: Changed this to have a `oneof` right off the bat.

oschaaf:
This is ready for review, tests passing

/assign @htuch

source/client/stream_decoder.cc
```diff
@@ -57,7 +57,15 @@ void StreamDecoder::onPoolFailure(Envoy::Http::ConnectionPool::PoolFailureReason
 
 void StreamDecoder::onPoolReady(Envoy::Http::StreamEncoder& encoder,
                                 Envoy::Upstream::HostDescriptionConstSharedPtr) {
-  encoder.encodeHeaders(request_headers_, true);
+  encoder.encodeHeaders(request_headers_, request_body_size_ == 0);
+  if (request_body_size_ > 0) {
+    // TODO(oschaaf): We can probably get away without allocating/copying here if we pregenerate
+    // potential request-body candidates up front. Revisit this when we have non-uniform request
+    // distributions and on-the-fly reconfiguration in place.
+    std::string body(request_body_size_, 'a');
```
htuch: One thing to keep in mind is that repeated `a` will compress very easily if we ever have compression on the data plane.

api/client/options.proto
htuch: This PR generally looks good. I'd like to chat tomorrow about how to use wrapped types here, it seems it should be possible.

oschaaf: I backed it out of here, we do not need it yet.
To concretely demonstrate the problem, I created https://github.com/envoyproxy/nighthawk/pull/63 for discussion. That demonstrates the problem clearly by failing a test when introducing the typed wrapper.

htuch: OK, let's continue there.

oschaaf:
Clang tidy is flaking afaict, I suggest we try https://github.com/envoyproxy/nighthawk/issues/62 to resolve that

PR #63, opened 2019-04-26
Back out using `oneoff`, add TODO to fix overriding with default value.

oschaaf:
This disables `HttpTestServerIntegrationTest.TestZeroLengthRequest` and backs out the `oneoff` proto hack we used to make overriding with `0` work.

The now disabled tests attempts to merge `{response_size:0}` into the default configuration, which will not take take effect, because the proto merge ignores the type-default (0) when merging. Conceptually proto doesn't make a distinction between "default" and "not specified". Tihs also adds a TODO to provide a way around this.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -44,8 +44,6 @@ build:clang-tsan --copt -fsanitize=thread
 build:clang-tsan --linkopt -fsanitize=thread
 build:clang-tsan --linkopt -fuse-ld=lld
 build:clang-tsan --define tcmalloc=disabled
-# The line below is Nighthawk-specific, and can be backed out when we no longer in the client integration test. 
-build:clang-tsan --test_env=TSAN_OPTIONS=die_after_fork=0
 # Needed due to https://github.com/libevent/libevent/issues/777
 build:clang-tsan --copt -DEVENT__DISABLE_DEBUG_MODE
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,70 @@
-import %workspace%/envoy/.bazelrc
+# The following .bazelrc content is forked from the main Envoy repository. This is necessary since
+# this needs to be available before we can access the Envoy repository contents via Bazel.
+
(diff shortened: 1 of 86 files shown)


htuch:
I think this might be a function of the JSON -> protobuf conversion for WKT. Can you try something closer to raw proto, e.g. `response_size: { value: 0 }`?

oschaaf:
@htuch I tried this:
```diff
--- a/test/server/http_test_server_filter_integration_test.cc
+++ b/test/server/http_test_server_filter_integration_test.cc
@@ -82,7 +82,8 @@ public:
     Envoy::BufferingStreamDecoderPtr response = makeSingleRequest(
         lookupPort("http"), "GET", "/", "", downstream_protocol_, version_, "foo.com", "",
         [response_size](Envoy::Http::HeaderMapImpl& request_headers) {
-          const std::string header_config = fmt::format("{{response_size:{}}}", response_size);
+          const std::string header_config =
+              fmt::format("{{ response_size: {{ value: {} }} }}", response_size);
           request_headers.addCopy(
               Nighthawk::Server::TestServer::HeaderNames::get().TestServerConfig, header_config);
         });
```

But that results in the exact same test failure.

oschaaf:
@htuch clang-tidy flaked again 
I've been running clang-tidy via docker in a loop on my dev-box for a couple of hours, I cannot reproduce it there. Comparing logs shows that the failing run wan't close to completing all the work.
So now my hypothesis is that we're running out of memory in CI. I added a118ec2 to this PR to reduce parallelism of `clang-tidy` in CI, to hopefully get it sorted.

oschaaf:
@htuch with 1897466 this now passes.

There may or may not still be a potential flake regarding calling `sleep()` from different threads in the client test.  I observed that happening in https://circleci.com/gh/envoyproxy/nighthawk/556, but that was before merging to the latest master. I can't reproduce that on my dev machine.

PR #71, opened 2019-05-03
check_format: import Envoy check/fix_format and various fixups.

htuch:
Signed-off-by: Harvey Tuch <htuch@google.com>

### DIFF
--- a/tools/check_format.sh
+++ b/tools/check_format.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+bazel run @envoy//tools:check_format.py -- \
+  --skip_envoy_build_rule_check  --namespace_check Nighthawk --include_dir_order envoy \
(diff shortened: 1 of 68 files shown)


tools/check_format.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+bazel run @envoy//tools:check_format.py -- \
+  --skip_envoy_build_rule_check  --namespace_check Nighthawk --include_dir_order envoy \
```
oschaaf: Do we want to treat `absl`, `gtest`, `gmock` and `external` special in here? E.g. move them higher up?

htuch: I prefer to have them in the final block, as with Envoy. These are all the misc. deps that don't get an explicit callout.

oschaaf: I see, sgtm

PR #60, opened 2019-04-24
Experiment: Mixer integration server for testing

oschaaf:
Putting this out here for discussion:
This PR absorbs int_server.cc/int_server.h from
https://github.com/duderino/proxy/blob/jblatt_mixer_fault_testing/test/integration/
(with a few tiny maintenance tweaks to get it to build with Envoy's latest version)

Then it goes on to update the end-to-end integration test for the NH client
to rely on this server for testing, swapping out Envoy's integration
server. This allows us to ditch the awkward fork() we have to do
to avoid runtime conflicts and deflake CI.

Also enhanced and added to the existing client tests while playing around with
the new server, to get a feel for ease of use (which seems to be pretty easy).

PR #72, opened 2019-05-04
TSAN: exclude test which relies on fork()

oschaaf:
Somehow the integration test server excepts when running under TSAN because of Envoy
not being able to find its configuration file (with TSANs `die_after_fork disabled`). 

The code below seems to suggest that TSAN doesn't support this scenario.
https://github.com/llvm-mirror/compiler-rt/blob/master/lib/tsan/rtl/tsan_interceptors.cc#L999

Further root cause analysis is not worth it at this point, because this code will be deprecated
soon in favor of either https://github.com/envoyproxy/nighthawk/pull/60 or moving this end-to-end testing to python.

So I propose we disable this test when we detect we are running under TSAN for now.

### DIFF
--- a/test/client_test.cc
+++ b/test/client_test.cc
@@ -6,7 +6,10 @@
 // soon in favor of either https://github.com/envoyproxy/nighthawk/pull/60 or moving this end-to-end
 // testing to python.
 // Thus we just disable this test when we detect we are running under TSAN.
-#ifndef ENVOY_CONFIG_TSAN
+#ifndef __has_feature
+#define __has_feature(x) 0 // Compatibility with non-clang compilers.
+#endif
+#if defined(__has_feature) && !__has_feature(thread_sanitizer)
 
 #include <chrono>
 
--- a/test/client_test.cc
+++ b/test/client_test.cc
@@ -1,3 +1,13 @@
+// This test relies on fork(). Somehow the integration test server excepts when running
+// under TSAN because of Envoy not being able to find its configuration file with die_after_fork
+// disabled. The code below seems to suggest that TSAN doesn't support this scenario.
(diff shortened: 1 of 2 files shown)


oschaaf:
/assign @htuch 
This just disables the end-to-end integration test when running under `TSAN`.
Hopefully we'll be able to deprecate that test soon in favor of a better way to do it.

test/client_test.cc
```diff
@@ -1,3 +1,16 @@
+// This test relies on fork(). Somehow the integration test server excepts when running
+// under TSAN because of Envoy not being able to find its configuration file with die_after_fork
+// disabled. The code below seems to suggest that TSAN doesn't support this scenario.
+// https://github.com/llvm-mirror/compiler-rt/blob/master/lib/tsan/rtl/tsan_interceptors.cc#L999
+// Further root cause analysis is not worth it at this point, because this code will be deprecated
+// soon in favor of either https://github.com/envoyproxy/nighthawk/pull/60 or moving this end-to-end
+// testing to python.
+// Thus we just disable this test when we detect we are running under TSAN.
+#ifndef __has_feature
+#define __has_feature(x) 0 // Compatibility with non-clang compilers.
```
htuch: Can you use the `ENVOY_CONFIG_TSAN` defined in the `.bazelrc` instead? This will a bit more robust and easier to handle during Google import. Thanks.

oschaaf: Done; but hmm, Envoy has a spot that might need a change for consistency [here](https://github.com/envoyproxy/envoy/blob/32e4d286668731594eb5c81ed664bd144d8d2d88/test/exe/main_common_test.cc#L119).

oschaaf: This worked, but broke after merging master. Reverting to use `__has_feature(...` made it work again, but it may be a good idea to figure out why `ENVOY_CONFIG_TSAN` no longer is defined.

PR #77, opened 2019-05-06
Improve human readable output.

### LINKED ISSUE #69: Improve human readable output for timings
Currently the output lists histograms in seconds, which is hard to read.
Defaulting to `milliseconds` would be better.

oschaaf:
Improves human-readable output on the CLI

Fixes #69

### DIFF
--- a/test/test_data/output_formatter.txt.gold
+++ b/test/test_data/output_formatter.txt.gold
@@ -1,7 +1,11 @@
 Nighthawk - A layer 7 protocol benchmarking tool.
 
-stat_id: 3 samples, mean: 0.002s, pstdev: 0.000816497s
-Percentile  Count       Latency        
+stat_id
+  samples: 3
+  mean:    0s 002 ms 000 us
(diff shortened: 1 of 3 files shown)


test/test_data/output_formatter.txt.gold
```diff
@@ -1,7 +1,11 @@
 Nighthawk - A layer 7 protocol benchmarking tool.
 
-stat_id: 3 samples, mean: 0.002s, pstdev: 0.000816497s
-Percentile  Count       Latency        
+stat_id
+  samples: 3
+  mean:    0s 002 ms 000 us
```
jmarantz: Why is there a space before "ms" and "us" but no space before "s"?

I think I'd prefer the no-space look but in any case can we be consistent?

oschaaf: Thanks, updated in https://github.com/envoyproxy/nighthawk/pull/77/commits/0d225c898f76822bac58ce0c7c6bf79c87f14262

PR #86, opened 2019-05-27
Refactor process context in preparation of running as a service.

oschaaf:
In preparation of Nighthawk as a service, refactor `ProcessContext` a bit so we can reuse it.
Also adds a new constructor to `OptionsImpl` that will take a proto as an argument.


include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,47 @@
...
+#include "common/event/real_time_system.h"
+
+namespace Nighthawk {
+namespace Client {
+
+/**
+ * Process context is shared between the CLI and grpc service. It is capable of executing
+ * a full Nighthawk test run. Only a single instance is allowed at a time.
+ */
+class ProcessContext {
```
htuch: I take it this isn't related to Envoy's recently introduced `ProcessContext`?

include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,47 @@
...
+#include "common/api/api_impl.h"
+#include "common/common/logger.h"
+#include "common/event/real_time_system.h"
+
+namespace Nighthawk {
+namespace Client {
+
+/**
+ * Process context is shared between the CLI and grpc service. It is capable of executing
+ * a full Nighthawk test run. Only a single instance is allowed at a time.
```
htuch: Why only a single instance at a time? Is this a singleton?

include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,47 @@
...
+class ProcessContext {
+public:
+  virtual ~ProcessContext() = default;
+
+  /**
+   * @return uint32_t the number of workers that will be used based on configuration
+   * and available hardware.
+   */
+  virtual uint32_t determineConcurrency() const PURE;
+  virtual Envoy::Event::TimeSystem& time_system() PURE;
```
htuch: Comments for the other interface methods missing.

include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,47 @@
...
+  virtual Envoy::Api::Impl& api() PURE;
+  virtual Envoy::Stats::Store& store() const PURE;
+
+  /**
+   * @param formatter used to transform output into the desired format.
+   * @return bool true iff execution was successfull.
+   */
+  virtual bool run(OutputFormatter& formatter) PURE;
```
htuch: Does the interface need a formatter, or can it return back some object  that is an neutral representation that can then be formatted?

include/nighthawk/client/process_context.h
oschaaf: Unrelated indeed, it's a coincidence

include/nighthawk/client/process_context.h
oschaaf: Fixed

include/nighthawk/client/process_context.h
oschaaf: The formatter was intended to be that neutral ground, but I think it's naming is less then ideal with this refactor. Maybe `OutputCollector` would be better; one can call `.toProto()` on it.

source/client/process_context_impl.h
```diff
@@ -25,6 +25,15 @@
 namespace Nighthawk {
 namespace Client {
 
+constexpr const char* ProcessContextLockFile = "/tmp/nighthawk.lock";
+
+/**
+ * Only a single instance is allowed at a time machine-wide in this implementation.
```
htuch: This one seems a bit strong; can't we have the end user make the call if we give them best practices? I agree we should prevent obviously bad things, but I can imagine situations in which you want to load test without latency measurement and want multiple NHs.

oschaaf: Yeah, so I was in two minds about this myself as well; would adding something like `--instance-name=unique-name` to allow distinct multiple concurrent instances resolve this in a way you would be comfortable with?

htuch: Yep, sounds good. Maybe in a separate PR?

include/nighthawk/client/process_context.h
htuch: Yes, `OutputCollector` is better, +1.

include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * @return Envoy::Event::TimeSystem& The time system implementation used by this instance.
+   */
+  virtual Envoy::Event::TimeSystem& time_system() PURE;
+
+  /**
+   * Envoy::Api::Impl& The api implementation used by this instance.
+   */
+  virtual Envoy::Api::Impl& api() PURE;
```
htuch: Why do we export these out of the `ProcessContext`? AFAICT, `ProcessContext` main goal is to encapsulate `run` and perform a full run of a workload. Why do we need to have these other features in the interface?

include/nighthawk/client/process_context.h
htuch: How come `ProcessContext` and not `Process`? Don't want to bike shed, just trying to understand if this is actually a context or an execution wrapper.

oschaaf: Renamed to `Process` as you are right in that it actually is an execution wrapper

include/nighthawk/client/process_context.h
oschaaf: Renamed to `OutputCollector`

source/client/process_context_impl.h
oschaaf: SG, created https://github.com/envoyproxy/nighthawk/issues/87 for tracking that new option.

include/nighthawk/client/process_context.h
oschaaf: You're right. Cleaned this up, and purged them from the interface.

source/client/process_context_impl.cc
```diff
@@ -0,0 +1,246 @@
...
+      dispatcher_(api_.allocateDispatcher()), cleanup_([this] { tls_.shutdownGlobalThreading(); }),
+      benchmark_client_factory_(options), sequencer_factory_(options), options_(options) {
+  machine_lock_ = open(ProcessContextLockFile, O_CREAT | O_RDWR, 0666);
+  if (flock(machine_lock_, LOCK_EX | LOCK_NB)) {
+    throw NighthawkException(
+        "Only a single ProcessContextImpl instance is allowed to exist at a time (machine wide).");
+  }
+  ares_library_init(ARES_LIB_INIT_ALL);
+  Envoy::Event::Libevent::Global::initialize();
+  configureComponentLogLevels(spdlog::level::from_str(options_.verbosity()));
```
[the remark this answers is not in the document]
oschaaf: Re: `ProcessWide::ProcessWide`:
Nice, sure. Updated to use it.

include/nighthawk/client/process.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Process {
+public:
+  virtual ~Process() = default;
+
+  /**
+   * @return uint32_t the number of workers that will be used based on configuration
+   * and available hardware.
+   */
+  virtual uint32_t determineConcurrency() const PURE;
```
htuch: Who uses this outside of `Process`?

include/nighthawk/client/process_context.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
```
htuch: Is this file stale?

source/client/process_context_impl.cc
```diff
@@ -0,0 +1,238 @@
+#include <sys/file.h>
```
htuch: `process_impl.cc`?

source/client/process_context_impl.cc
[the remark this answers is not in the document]
htuch: Should we remove the flock from this PR?

source/client/process_context_impl.h
```diff
@@ -0,0 +1,77 @@
+#pragma once
```
htuch: `process_impl.h`

source/client/process_context_impl.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "exe/process_wide.h"
+
+#include "client/benchmark_client_impl.h"
+#include "client/factories_impl.h"
+
+namespace Nighthawk {
+namespace Client {
+
+constexpr const char* ProcessLockFile = "/tmp/nighthawk.lock";
```
htuch: Maybe remove.

source/client/process_impl.cc
```diff
@@ -0,0 +1,239 @@
+#include "client/process_impl.h"
```
htuch: Oh, looks like the above file is stale..

oschaaf:
@htuch sorry about the stale `process_context*` files, I removed them.

source/client/process_impl.cc
```diff
@@ -0,0 +1,239 @@
...
+    throw NighthawkException(
+        "Only a single ProcessImpl instance is allowed to exist at a time (machine wide).");
+  }
+  configureComponentLogLevels(spdlog::level::from_str(options_.verbosity()));
+  tls_.registerThread(*dispatcher_, true);
+}
+
+ProcessImpl::~ProcessImpl() {
+  if (machine_lock_ > 0) {
+    flock(machine_lock_, LOCK_UN);
```
htuch: Can we remove and punt to next PR?

include/nighthawk/client/process.h
htuch: Ping on this one :)

oschaaf: Removed

source/client/process_impl.cc
oschaaf: Done

oschaaf:
/assign @htuch 
comments addressed

PR #88, opened 2019-05-30
Benchmark client configuration options

### LINKED ISSUE #32: Allow control of TLS ciphers and settings
Explicit control of TLS ciphers and session-reuse makes it easier when it comes to comparing server to server performance. Additionally, it would be nice to be able to test just the overhead of setting up these connections (and not perform any requests) and/or track times of specific milestones during the connection/tls setup process.

### LINKED ISSUE #45: Connection-pool configuration
Re-use and apply Envoy's configuration for tuning all available pool settings and limits

oschaaf:
Benchmark client configuration options

- Adds a the following fields to the benchmark client proto (and matching CLI args):

```
  envoy.api.v2.auth.UpstreamTlsContext tls_context = 13;
  uint32 max_pending_requests = 14 [(validate.rules).uint32 = {gte: 1}];
  uint32 max_active_requests = 15 [(validate.rules).uint32 = {gte: 1}];
  uint32 max_requests_per_connection = 16 [(validate.rules).uint32 = {gte: 1}];
```

This allows for cipher suite selection, tuning of connection re-use, tuning client side request queueing, and tuning the maximum amount of active requests.

When specifying `max_pending_requests > 1` the benchmark client will stop guarding the
exact pacing of requests, allowing open-loop test scenarios.

- Unifies validation in the CLI and service paths.
- Consolidates default values into a header and improves handling.
- Updates the options proto to use 32 bits where relevant, as well as
  switches fields to WKT's to enable us to make the distinction between
  `not-set` and type-default.

Resolves #45
Resolves #32


oschaaf:
@htuch this is still a work-in-progress, but it would be awesome if you could share thoughts on the general approach taken here

api/client/options.proto
```diff
@@ -41,6 +42,8 @@ message CommandLineOptions {
     // See :option:`--request_options` for details.
     RequestOptions request_options = 12;
   }
+  // See :option:`--cluster_config` for details.
+  envoy.api.v2.Cluster cluster_config = 13;
```
htuch: Yeah, the basic approach is good, but I'm wondering what we can say about which parts of the `Cluster` config will be honored vs. ignored?

oschaaf:
@htuch re: " I'm wondering what we can say about which parts of the Cluster config will be honored vs. ignored?" 
Maybe we can make the documentation/test coverage leading here. We verify settings work through tests, and when we find they do we document them as such and make a release note about it?

htuch:
@oschaaf if it's only a small subset, perhaps we should just explicitly pass in those fields inside `Cluster`, rather than the whole `Cluster`. If it's most of `Cluster`, yeah, docs and input validation could be the way to go.

oschaaf:
@htuch looking at a deserialized cluster proto over at https://gist.github.com/oschaaf/67f2f4c2451482df702b77774f205917, some thoughts

### We'll use these
-   circuit_breakers.thresholds.max_connections (default priority)
-   circuit_breakers.thresholds.max_pending_requests (default priority)
-   circuit_breakers.thresholds.max_retries (default priority)
-   tls_context.*

### We may not use these 
-   name
-   alt_stat_name

### Perhaps in the future?
-   lb_policy
-   hosts
-   health_checks
-   circuit_breakers.thresholds.track_remaining

### Ambigous, maybe we should move to use this field for our purpose and deprecate our own versions of it
-   dns_lookup_family
-   dns_resolvers

Personally I'm currently leaning towards:
- creating our own cluster-ish proto and then internally do a fairly mechanical field mapping to Envoy's cluster proto. The initial fields we expose would be just the circuit breaker fields, and then we
- fully re-use and expose Envoy's tls-context proto as a field on that

htuch:
@oschaaf thanks for putting together that list. I agree, if we have our own `Cluster`ish proto, and just put in two messages, `circuit_breakers` and `tls_context`, we probably get a nice solution.

oschaaf:
Best to land #88 to land first, but marking this as `ready for review` for allowing feedback.

api/client/options.proto
```diff
@@ -3,44 +3,59 @@ syntax = "proto3";
...
   // See :option:`--connections` for details.
-  uint64 connections = 2 [(validate.rules).uint64 = {gte: 1, lte: 1000000}];
+  google.protobuf.UInt32Value connections = 2 [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
   // See :option:`--duration` for details.
   google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 1];
   // See :option:`--timeout` for details.
   google.protobuf.Duration timeout = 4 [(validate.rules).duration.gte.seconds = 0];
   // See :option:`--h2` for details.
-  bool h2 = 5;
+  google.protobuf.BoolValue h2 = 5;
```
htuch: Generally in Envoy, these should be scalar if the default is empty/0/"" and wrapped if you need something else. Is this the case for all of these?

oschaaf: My 2p: While we may get away with scalar booleans today, in the future we may want to allow default configurations to override the ones we provide in code. When that happens, I think wrapping the booleans as well is warranted, because we can discern between "not provided" and 'false' values on incoming configurations, before we override the existing one.

source/client/benchmark_client_impl.h
```diff
@@ -51,13 +55,20 @@ class BenchmarkClientHttpImpl : public BenchmarkClient,
...
-                          bool prefetch_connections);
+                          bool prefetch_connections,
+                          envoy::api::v2::auth::UpstreamTlsContext tls_context);
 
   void setConnectionLimit(uint64_t connection_limit) { connection_limit_ = connection_limit; }
   void setConnectionTimeout(std::chrono::seconds timeout) { timeout_ = timeout; }
   void setMaxPendingRequests(uint64_t max_pending_requests) {
     max_pending_requests_ = max_pending_requests;
   }
+  void setMaxActiveRequests(uint64_t max_active_requests) {
```
mergeconflict: These are uint32_t in most other places; should the type match here?

source/client/options_impl.cc
```diff
@@ -121,40 +161,153 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
     throw NoServingException();
   }
 
-  requests_per_second_ = requests_per_second.getValue();
-  connections_ = connections.getValue();
-  duration_ = duration.getValue();
-  timeout_ = timeout.getValue();
+  if (requests_per_second.isSet()) {
```
mergeconflict: Maybe extract a helper function like

```c++
template <typename T> void getValueIfSet(const TCLAP::Arg<T>& src, T& dst) {
  if (src.isSet()) {
    dst = src.getValue();
  }
}
```

Or maybe better like

```c++
template <typename T> T getValueOrDefault(const TCLAP::Arg<T>& src, T&& default_value) {
  return src.isSet() ? src.getValue() : default_value
}
```

source/client/options_impl.cc
```diff
@@ -121,40 +161,153 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
+  }
+  if (max_active_requests.isSet()) {
+    max_active_requests_ = max_active_requests.getValue();
+  }
+  if (max_requests_per_connection.isSet()) {
+    max_requests_per_connection_ = max_requests_per_connection.getValue();
+  }
+
+  // CLI-specific tests.
+  if (requests_per_second_ > largest_acceptable_uint32_option_value) {
```
mergeconflict: Might be nice to accumulate these errors so the user can see everything they did wrong all at once.

source/client/options_impl.h
```diff
@@ -27,28 +27,44 @@ class OptionsImpl : public Options {
...
-  uint64_t requests_per_second_;
-  uint64_t connections_;
-  uint64_t duration_;
-  uint64_t timeout_;
+  void setNonTrivialDefaults();
+  void validate() const;
+
+  // We cap on negative values. TCLAP accepts negative values which we will get here as very
+  // large values. We just cap values, hoping we catch accidental wraparound to a reasonable extent.
+  static constexpr uint32_t largest_acceptable_uint32_option_value = UINT32_MAX - 30000;
```
mergeconflict: Hmm, if TCLAP accepts negative values, I wonder if that means it fails to accept positive values larger than `UINT16_MAX` for `uint32_t` args...

test/benchmark_http_client_test.cc
```diff
@@ -395,6 +398,63 @@ TEST_P(BenchmarkClientHttpTest, ConnectionPrefetching) {
...
 
+// TODO(oschaaf): Turns out this feature is h/2 only as of writing this.
+// Consider logging a warning/error when attempting to configure this with
+// H1 enabled.
+TEST_P(BenchmarkClientHttpTest, CapRequestConcurrency) {
+  setupBenchmarkClient("/lorem-ipsum-status-200", true, false);
+  const uint64_t requests = 4;
+  uint64_t inflight_response_count = requests;
+
+  // We configure so that max requests is the only thing that can  be throttling.
```
mergeconflict: Nit: extra space in "can be" here and below.

test/options_test.cc
```diff
@@ -272,10 +289,26 @@ TEST_F(OptionsImplTest, ProtoConstructorValidation) {
...
                           "CommandLineOptionsValidationError.RequestsPerSecond");
 }
 
+TEST_F(OptionsImplTest, BadTlsContextSpecification) {
+  // Bad JSON
+  EXPECT_THROW_WITH_REGEX(TestUtility::createOptionsImpl(fmt::format(
+                              "{} --tls-context {} http://foo/", client_name_, "{broken_json:")),
+                          MalformedArgvException, "Unable to parse JSON as proto");
+
+  // TODO(oschaaf): This should fail, but does not. Probably because we pass in a
```
mergeconflict: Will you correct that in this PR?

test/service_test.cc
```diff
@@ -89,7 +84,7 @@ TEST_P(ServiceTest, Basic) {
   r->Write(request_, {});
   r->WritesDone();
   EXPECT_TRUE(r->Read(&response_));
-  EXPECT_FALSE(response_.has_error_detail());
+  ASSERT_FALSE(response_.has_error_detail());
```
mergeconflict: Why change these from EXPECTs to ASSERTs?

source/client/options_impl.h
oschaaf: Added ParserIntRangeTest, which passes, I think we're good here?

oschaaf:
@mergeconflict thanks for the review!
I believe it was @htuch intended for me to be able to merge this myself during his absense
upon approval, but unfortunately the merge button is disabled: I suspect it is set up so that someone with write-access for this repo needs to approve before the PR becomes mergeable.

mergeconflict:
> I suspect it is set up so that someone with write-access for this repo needs to approve before the PR becomes mergeable.

Something like that, yeah, I think a maintainer has to perform the merge. I don't have permission either. @jmarantz, could you help?

PR #89, opened 2019-05-31
Update Envoy + python integration testing

oschaaf:
- Switch to Envoy@ec47fee6604468bc392937967415c736f19fb22129929881270a1635ad216d87
- Fixes to build with this version
- Remove C++ integration tests that broke unrepairably
- Add Python-based integration testing
- Sets us up with some basic orchestration to run servers
  (like the one from Nighthawk in these tests)
- Add python format checking (CI) / fixing
- Update grpc stub generation to a more sane approach


ci/do_ci.sh
```diff
@@ -91,29 +93,29 @@ if [ -n "$CIRCLECI" ]; then
...
+if grep 'docker\|lxc' /proc/1/cgroup; then
     # Create a fake home. Python site libs tries to do getpwuid(3) if we don't and the CI
     # Docker image gets confused as it has no passwd entry when running non-root
     # unless we do this.
     FAKE_HOME=/tmp/fake_home
     mkdir -p "${FAKE_HOME}"
     export HOME="${FAKE_HOME}"
     export PYTHONUSERBASE="${FAKE_HOME}"
-
+
```
htuch: Nit: gratuitous whitespace changes?

integration/integration_test_base.py
```diff
@@ -0,0 +1,157 @@
...
+            self.test_rundir, "nighthawk_client")
+        self.server_ip = "::1" if IntegrationTestBase.ipv6 else "127.0.0.1"
+        self.socket_type = socket.AF_INET6 if IntegrationTestBase.ipv6 else socket.AF_INET
+        self.test_server = None
+        self.server_port = -1
+        self.parameters = dict()
+
+    def getFreeListenerPortForAddress(self, address):
+        """
+        Determines a free port and returns that. Theoretically it is possible that another process
```
htuch: The clean way to do this is to start the server bound to port 0, then query its `/listeners` handle on the admin port. Then you need to know what the admin port is, which is ideally also bound to 0 in the config. To figure that out, you can do something like https://github.com/envoyproxy/envoy/blob/master/test/integration/hotrestart_test.sh#L72.

integration/integration_test.py
```diff
@@ -0,0 +1,49 @@
+"""@package integration_test.py
```
htuch: One high-level question on style nits; we have a bunch of new Python in this PR and it's going to grow going forwards. Should we have a CI Python format check? See https://github.com/envoyproxy/envoy/blob/master/tools/format_python_tools.sh for what Envoy has

integration/test_integration_basics.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python
```
htuch: Might be best to go straight to python3, as Envoy is mid-way through the slow march to Python 3 (Google import will also require this).

integration/integration_test_base.py
oschaaf: I agree that approach is cleaner. But, I think that soon(ish) we may also want to be running other proxies and services for running comparison test runs, which is what led me up to this approach.
Because when we want to add other server types, not having to implement the knowledge about how to get to the correct port(s) for each of them is a win? 
E.g. adding support for nginx in the current approach is easy (see (see `NighthawkTestServer(TestServerBase)`), whereas I wouldn't know how to query it for a port that got assigned for a specific listener.

Is the concern that the port we intended for a server to listen on gets bound to by another process else before our server gets around to that?

integration/test_integration_basics.py
oschaaf: I have had a fight with bazel/CI on this topic. I agree, and would love to move to Python 3, but are you OK with postponing that to a next PR? 
(I think we need another Envoy dependency upgrade / upgrade the CI image)

oschaaf:
@htuch added the fix-format/check-format for Python in CI here. You can see it fail in the `CircleCI` logs.
I would request postponing reformatting to fix that until we start working on landing https://github.com/envoyproxy/nighthawk/pull/88, because that contains a few small changes/additions as well as new tests; diffing is going to get harder if we reformat here.

oschaaf:
Never mind; python formatting fixes applied, will just apply it to #88 as well.

htuch:
We can postpone Python3 for a little bit, but probably will need to track it as high priority as it will become an import blocker. Format fix is also fine to relegate to an issue, but should be reasonable, I agree it makes sense as a standalone PR.

Re: port binding, I think it would be good to support best effort port picking when we grow to other proxies, but for shared test resources that are heavily contended, it can be flakey, so for the core NH tests we probably want to also handle zero port binding. We could move this to a later PR though.

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+Entry point for our integration testing
+"""
+
+import logging
+import os
+import sys
+import unittest
+from integration_test_base import IntegrationTestBase
```
htuch: Nit: prefer to have a separate block of imports for Python standard libs, then a block for the specific internal libs we have.

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+  versions = os.environ.get("ENVOY_IP_TEST_VERSIONS", "all")
+  if versions == "v4only":
+    ipv6 = False
+  elif versions == "v6only":
+    ipv4 = False
+  elif versions == "all":
+    pass
+  else:
+    raise Exception("Unknown ip version: '%s'" % versions)
+  return (ipv4, ipv6)
```
htuch: Nit: no need for parens here, just `return ipv4, ipv6`

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+import os
+import sys
+import unittest
+from integration_test_base import IntegrationTestBase
+
+import test_integration_basics
+
+
+def determineIpVersionsFromEnvironment():
+  ipv4 = True
```
htuch: I think you might want to model this as a `namedtuple`

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+  ipv6 = True
+  versions = os.environ.get("ENVOY_IP_TEST_VERSIONS", "all")
+  if versions == "v4only":
+    ipv6 = False
+  elif versions == "v6only":
+    ipv4 = False
+  elif versions == "all":
+    pass
+  else:
+    raise Exception("Unknown ip version: '%s'" % versions)
```
htuch: Nit: this should be a custom `NighthawkError` or more specific, subclassing `Exception`. This allows you to catch the more specific subclass.

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+  IntegrationTestBase.ipv6 = ipv6
+  res = unittest.TextTestRunner(verbosity=2).run(
+      unittest.TestLoader().loadTestsFromModule(test_integration_basics))
+
+  return res.wasSuccessful()
+
+
+if __name__ == '__main__':
+  logging.basicConfig(stream=sys.stderr, level=logging.DEBUG)
+  (v4, v6) = determineIpVersionsFromEnvironment()
```
htuch: Nit: superfluous parens

integration/integration_test.py
htuch: Actually, thinking about it, if you just have IPV4, IPv6 as enum-like constants and return a list here of the actual versions that are supported, it simplifies logic elsewhere.

integration/integration_test_base.py
```diff
@@ -0,0 +1,151 @@
+"""@package integration_test_base
+Base classes for Nighthawk integration tests
+"""
+
+import json
+import logging
+from nighthawk_test_server import NighthawkTestServer
```
htuch: Same point a ^^ re: import block ordering.

integration/integration_test_base.py
```diff
@@ -0,0 +1,151 @@
...
+    self.parameters = dict()
+
+  def getFreeListenerPortForAddress(self, address):
+    """
+        Determines a free port and returns that. Theoretically it is possible that another process
+        will steal the port before our caller is able to leverage it, but we take that chance.
+        The upside is that we can push the port upon the server we are about to start through configuration
+        which is compatible accross servers.
+        """
+    sock = socket.socket(self.socket_type, socket.SOCK_STREAM)
```
htuch: Is there a RAII-style `with` patterns that can autoclose the `socket` here?

integration/test_integration_basics.py
```diff
@@ -0,0 +1,84 @@
...
+    self.assertNighthawkCounterEqual(parsed_json, "upstream_cx_total", 1)
+    self.assertNighthawkCounterEqual(parsed_json, "upstream_cx_tx_bytes_total",
+                                     1400 if IntegrationTestBase.ipv6 else 1500)
+    self.assertNighthawkCounterEqual(parsed_json, "upstream_rq_pending_total", 1)
+    self.assertNighthawkCounterEqual(parsed_json, "upstream_rq_total", 25)
+    self.assertNumberOfCountersEqual(parsed_json, 9)
+
+  def test_h2(self):
+    parsed_json = self.runNighthawk(["--h2", self.getTestServerRootUri()])
+    self.assertNighthawkCounterEqual(parsed_json, "benchmark.http_2xx", 25)
```
htuch: Why not transform the parsed_json into another dictionary that you can then just index, e.g. `nighthawk_counters["upstream_cx_destroy"]`? That's probably a more Pythonic way to model the data.

integration/integration_test.py
oschaaf: Fixed

integration/integration_test.py
oschaaf: Fixed, PTAL

integration/integration_test.py
oschaaf: Done

integration/integration_test.py
oschaaf: Obsolete with earlier changes

integration/integration_test.py
oschaaf: Ditto

integration/integration_test_base.py
oschaaf: Fixed

integration/integration_test_base.py
oschaaf: Will add an item for tracking we can punt this to a follow up

integration/integration_test_base.py
oschaaf: When we move to python 3 we can do that. (see https://github.com/envoyproxy/nighthawk/pull/89/commits/0feec9a30c7fa3b7710f6876cd65e7ef29e145da, but had to back that out)

integration/test_integration_basics.py
oschaaf: I'll add an item to track this and attempt to move to Python3 in a follow up where we update the ci image and our Envoy dependency. 

With Python3 I think we would also have a fairly easy path to enhancing the integration tests with parallel execution 
& parameterized tests support.

integration/test_integration_basics.py
oschaaf: Indeed, done!

integration/integration_test_base.py
htuch: OK, please leave a TODO, I'm keen to get to Python 3 :)

integration/integration_test_base.py
```diff
@@ -0,0 +1,147 @@
...
+    sock.bind((address, 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    return port
+
+  def setUp(self):
+    """
+        Performs sanity checks and starts up the server. Upon exit the server is ready to accept connections.
+        """
```
htuch: Nit: weird indents. Also below with other doc comments. The text and closing """ are quite a bit off.

integration/integration_test_base.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+        Utility method to get the counters from the json indexed by name.
+        Note:the `client.` prefix will be stripped from the index.
+        """
+    counter_map = {}
+    counters = parsed_json["results"][0]["counters"]
+    for counter in counters:
+      counter_qualified_name = counter["name"]
+      counter_map[counter_qualified_name[len("client."):]] = int(counter["value"])
+    return counter_map
```
htuch: This could even be more concise as a Python dictionary comprehension.

integration/test_integration_basics.py
```diff
@@ -0,0 +1,84 @@
...
+from integration_test_base import (HttpIntegrationTestBase, HttpsIntegrationTestBase,
+                                   IntegrationTestBase)
+
+
+class TestHttp(HttpIntegrationTestBase):
+
+  def test_h1(self):
+    parsed_json = self.runNighthawk([self.getTestServerRootUri()])
+    counters = self.getNighthawkCounterMapFromJson(parsed_json)
+    self.assertEqual(counters["benchmark.http_2xx"], 25)
```
htuch: I'd like to figure out how to reduce boilerplate even further, e.g. by adding some method that takes a map of expected key/values, counters and then iterates over the former. It can be a TODO though as this is already fairly good.

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+  res = unittest.TextTestRunner(verbosity=2).run(
+      unittest.TestLoader().loadTestsFromModule(test_integration_basics))
+  return res.wasSuccessful()
+
+
+if __name__ == '__main__':
+  logging.basicConfig(stream=sys.stderr, level=logging.DEBUG)
+  ip_versions = determineIpVersionsFromEnvironment()
+  assert (len(ip_versions) > 0)
+  ok = True
```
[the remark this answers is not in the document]
oschaaf: nice, done

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+    supported.append(IpVersion.ipv6)
+  else:
+    raise NighthawkException("Unknown ip version: '%s'" % versions)
+  return supported
+
+
+def runTests(ip_version):
+  IntegrationTestBase.ip_version = ip_version
+  logging.info("Running %s integration tests." %
+               ("ipv6" if IntegrationTestBase.ip_version == IpVersion.ipv6 else "ipv4"))
```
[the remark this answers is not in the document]
oschaaf: Added a TODO

integration/integration_test_base.py
oschaaf: Fixed

integration/integration_test_base.py
oschaaf: done

integration/integration_test_base.py
oschaaf: done

integration/test_integration_basics.py
oschaaf: Added

integration/integration_test.py
```diff
@@ -0,0 +1,44 @@
...
+import test_integration_basics
+from common import IpVersion, NighthawkException
+from integration_test_base import IntegrationTestBase
+
+
+def determineIpVersionsFromEnvironment():
+  supported = []
+  versions = os.environ.get("ENVOY_IP_TEST_VERSIONS", "all")
+  if versions == "v4only":
+    supported.append(IpVersion.IPV4)
```
htuch: Nit: these can all be `supported = ` (and there's no need to empty initialize above, yay Python scoping).

integration/integration_test.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+import logging
+import os
+import sys
+import unittest
+
+import test_integration_basics
+from common import IpVersion, NighthawkException
+from integration_test_base import IntegrationTestBase
```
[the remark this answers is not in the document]
htuch: Maybe rename to `basic_integration_tests`?

integration/nighthawk_test_server.py
```diff
@@ -0,0 +1,90 @@
...
+    self.parameters["server_port"] = self.server_port
+
+    with open(self.config_template_path) as f:
+      config = Template(f.read())
+      config = config.substitute(self.parameters)
+
+    parameterized_config_path = None
+    with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".yaml") as tmp:
+      parameterized_config_path = tmp.name
+      tmp.write(config)
```
htuch: Arguably we could use the bootstrap CLI here to avoid needing to write files; I don't think it makes a huge difference, but throwing it out there.

integration/test_integration_basics.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+import os
+import sys
+import unittest
+
+from common import IpVersion
+from integration_test_base import (HttpIntegrationTestBase, HttpsIntegrationTestBase,
+                                   IntegrationTestBase)
+
+# TODO(oschaaf): rewrite the tests so we can just had a map of expected key values to it.
```
htuch: Nit: s/had/hand/

integration/integration_test.py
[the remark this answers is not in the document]
oschaaf: Renamed `integration_test_base` to `integration_test_fixtures`

integration/integration_test.py
oschaaf: done

integration/nighthawk_test_server.py
oschaaf: Actually in earlier revisions of this it worked via the bootstrap CLI; but like with the port, I leaned toward doing it like this with supporting for servers in the future in mind.

integration/test_integration_basics.py
oschaaf: fixed

oschaaf:
@htuch comments addressed. Performed a rename, `integration_test_base` is now known as `integration_test_fixtures` which makes things a bit less ambiguous, hopefully.

oschaaf:
@htuch squashed & force pushed, DCO fixed.

PR #90, opened 2019-06-03
Nighthawk-as-a-service

oschaaf:
This PR introduces a grpc service, which is able to accept a stream of load-test requests, and produces a stream of results.

Grpc stub generation is working, but the BUILD file which accomplishes that is a bit messy. A cleanup is staged in a follow up (we need an update to our Envoy dependency to be able to do that).


source/client/service_impl.cc
```diff
@@ -0,0 +1,104 @@
...
+  OutputCollectorFactoryImpl output_format_factory(time_system_, *options);
+  auto formatter = output_format_factory.create();
+  bool success = process_->run(*formatter);
+  nighthawk::client::SendCommandResponse response;
+  *(response.mutable_output()->Add()) = formatter->toProto();
+  response_queue_.Push(ServiceProcessResult(response, success ? "" : "Unkown failure"));
+  process_.reset();
+}
+
+void ServiceImpl::EmitResponses(
```
htuch: Nit: Can you switch to `camelCase` rather than `CamelCase` everywhere other than constructors to match Envoy style?

api/client/service.proto
```diff
@@ -0,0 +1,31 @@
...
+}
+
+message SendCommandResponse {
+  repeated Output output = 1;
+}
+
+service NighthawkService {
+
+  /*
+  Accepts a SendCommandRequest, which can be used to initiate, update, or cancel a test.
```
htuch: `SendCommand` sounds like a verb to me, whereas the request/response messages are nouns. I think maybe just calling them `ExecutionRequest` and `ExecutionResult` would be fine.

api/client/service.proto
```diff
@@ -0,0 +1,31 @@
...
+    kStart = 0;
+    kUpdate = 1;
+    UNSET = 2;
+  }
+  CommandType command_type = 1;
+  CommandLineOptions options = 2;
+}
+
+message SendCommandResponse {
+  repeated Output output = 1;
```
htuch: How come this is repeated?

api/client/service.proto
```diff
@@ -0,0 +1,31 @@
...
+package nighthawk.client;
+
+import "api/client/options.proto";
+import "api/client/output.proto";
+
+message SendCommandRequest {
+  enum CommandType {
+    kStart = 0;
+    kUpdate = 1;
+    UNSET = 2;
```
htuch: Nit: the style for proto here would be:
```
enum CommandType {
  UNSET = 0;
  START = 1;
  UPDATE = 2;
}
```
You want the unset/default to be field 0, enum values are always caps.

api/client/service.proto
```diff
@@ -0,0 +1,31 @@
...
+  }
+  CommandType command_type = 1;
+  CommandLineOptions options = 2;
```
htuch: Do you want to use a `oneof` to model the different command options here? I'm thinking that `CommandLineOptions` would only apply to start, whereas other options would later apply to update..

source/client/service_impl.cc

api/client/service.proto
oschaaf: Maybe one way to offer the update call could be to also take a CommandLineOptions, and then we emit an error when a change is attempted that could not be processed on the fly? Regardless the `oneof` allows for more flexibility - added `oneof`.

api/client/service.proto
oschaaf: Whoops, stale artifact. Cleaned it up, this no longer has `repeated`.

api/client/service.proto
oschaaf: Done

source/client/service_impl.cc
```diff
@@ -0,0 +1,104 @@
...
+      spdlog::level::from_str(options->verbosity()), "[%T.%f][%t][%L] %v", log_lock);
+  process_ = std::make_unique<ProcessImpl>(*options, time_system_);
+
+  // We perform this validation here because we need to rutime to be initialized
+  // for this, something that ProcessContext does for us.
+  try {
+    Envoy::MessageUtil::validate(request.options());
+  } catch (Envoy::EnvoyException exception) {
+    nighthawk::client::ExecutionResponse response;
+    response_queue_.push(ServiceProcessResult(response, exception.what()));
```
htuch: Why is it necessary to queue the response? Is it the case that gRPC stubs doesn't support cross-thread replies?

source/client/service_impl.cc
```diff
@@ -0,0 +1,104 @@
...
+namespace Nighthawk {
+namespace Client {
+
+void ServiceImpl::nighthawkRunner(const nighthawk::client::ExecutionRequest& request) {
+  Envoy::Thread::LockGuard lock(mutex_);
+  OptionsPtr options = std::make_unique<OptionsImpl>(request.options());
+  Envoy::Thread::MutexBasicLockable log_lock;
+  auto logging_context = std::make_unique<Envoy::Logger::Context>(
+      spdlog::level::from_str(options->verbosity()), "[%T.%f][%t][%L] %v", log_lock);
+  process_ = std::make_unique<ProcessImpl>(*options, time_system_);
```
htuch: Why do you need to lock process if it's local?

api/client/options.proto
```diff
@@ -16,11 +16,11 @@ message RequestOptions {
...
 message CommandLineOptions {
   // See :option:`--rps` for details.
-  uint64 requests_per_second = 1;
+  uint64 requests_per_second = 1 [(validate.rules).uint64 = {gte: 1, lte: 1000000}];
   // See :option:`--connections` for details.
-  uint64 connections = 2;
+  uint64 connections = 2 [(validate.rules).uint64 = {gte: 1, lte: 1000000}];
   // See :option:`--duration` for details.
-  google.protobuf.Duration duration = 3;
+  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.seconds = 1];
```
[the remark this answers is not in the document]
oschaaf: Polished the validations; fixed.

api/client/service.proto
oschaaf: Done

source/client/service_impl.cc
oschaaf: Sorry, cleaned this up. The locking was a remnant which is no longer needed.

source/client/service_impl.h
```diff
@@ -0,0 +1,91 @@
...
+private:
+  void nighthawkRunner(const nighthawk::client::ExecutionRequest& start_request);
+  void emitResponses(::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
+                                                ::nighthawk::client::ExecutionRequest>* stream,
+                     std::string& error_messages);
+
+  BlockingQueue<ServiceProcessResult> response_queue_;
+  std::thread nighthawk_runner_thread_;
+  Envoy::Event::RealTimeSystem time_system_; // NO_CHECK_FORMAT(real_time)
+  ProcessPtr process_ GUARDED_BY(mutex_);
```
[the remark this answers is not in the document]
oschaaf: The mutex is gone now, so this has gone stale.

api/client/options.proto
```diff
@@ -16,31 +16,36 @@ message RequestOptions {
...
+  string verbosity = 7
+      [(validate.rules).string = {in: ["trace", "debug", "info", "warn", "error", "critical"]}];
   // See :option:`--output-format` for details.
-  string output_format = 8;
+  string output_format = 8 [(validate.rules).string = {in: ["human", "yaml", "json"]}];
   // See :option:`--prefetch-connections` for details.
   bool prefetch_connections = 9;
   // See :option:`--burst-size` for details.
-  uint64 burst_size = 10;
+  uint64 burst_size = 10 [(validate.rules).uint64 = {gte: 0, lte: 1000000}];
```
htuch: `gte: 0` is redundant with the unsigned type.

api/client/options.proto
```diff
@@ -16,31 +16,36 @@ message RequestOptions {
...
+  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 1];
   // See :option:`--timeout` for details.
-  google.protobuf.Duration timeout = 4;
+  google.protobuf.Duration timeout = 4 [(validate.rules).duration.gte.seconds = 0];
   // See :option:`--h2` for details.
   bool h2 = 5;
   // See :option:`--concurrency` for details.
   string concurrency = 6;
+  // TODO(oschaaf): pvg throw NotImplementedException for the following rule:
+  // string concurrency = 6 [(validate.rules).string = {pattern: "^([0-9]*|auto)$"}];
```
htuch: Yes, I think regexes aren't well supported.

api/client/service.proto
```diff
@@ -0,0 +1,35 @@
...
+
+message ExecutionResponse {
+  Output output = 1;
+}
+
+service NighthawkService {
+
+  /*
+  Accepts a ExecutionRequest, which can be used to initiate, update, or cancel a test.
+  Returns a stream of ExecutionResponse objects which will contain results for all successfull
```
htuch: What is returned when an execution fails?

source/client/service_impl.cc
```diff
@@ -0,0 +1,89 @@
...
+    response_queue_.push_back(ServiceProcessResult({}, exception.what()));
+    return;
+  }
+
+  auto logging_context = std::make_unique<Envoy::Logger::Context>(
+      spdlog::level::from_str(options->verbosity()), "[%T.%f][%t][%L] %v", log_lock_);
+  process_ = std::make_unique<ProcessImpl>(*options, time_system_);
+  OutputCollectorFactoryImpl output_format_factory(time_system_, *options);
+  auto formatter = output_format_factory.create();
+  bool success = process_->run(*formatter);
```
htuch: Nit: `const bool`

source/client/service_impl.cc
```diff
@@ -0,0 +1,89 @@
...
+  nighthawk::client::ExecutionResponse response;
+  *(response.mutable_output()) = formatter->toProto();
+  response_queue_.emplace_back(ServiceProcessResult(response, success ? "" : "Unkown failure"));
```
htuch: Why are we queueing this up until the end of the request stream, vs. replying back immediately after the execution completes?

source/client/service_impl.cc
```diff
@@ -0,0 +1,89 @@
...
+  process_.reset();
```
htuch: Can `process_` be scoped to just this method; we won't need an explicit reset in that case.

source/client/service_impl.cc
```diff
@@ -0,0 +1,89 @@
...
+  nighthawk::client::ExecutionRequest request;
+
+  while (stream->Read(&request)) {
+    ENVOY_LOG(debug, "Read ExecutionRequest data: {}", request.DebugString());
+    if (request.has_start_request()) {
+      if (running_) {
+        error_message = "Only a single benchmark session is allowed at a time.";
+        break;
+      } else {
+        running_ = true;
```
htuch: When is this set `false` again?

source/client/service_impl.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic warning "-Wunused-parameter"
+#endif
+#include "api/client/service.grpc.pb.h"
+
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
```
htuch: Maybe just set `-Wunused-parameter` in the Bazel target `copts`..

api/client/service.proto
oschaaf: Changed the proto a bit & updated the docs.
Earlier, we would only support one full process context run per connection.
Passing errors in the grpc status trailers would suffice.
Now we support multiple back-to-back runs, and indicate success plus a message on each response we emit. 
If any response indicated failure, we return a a failure status at the grpc-level as well.
Let me know what you think.

source/client/service_impl.cc
oschaaf: Fixed.

source/client/service_impl.cc

source/client/service_impl.cc
oschaaf: It wasn't but it is now.

source/client/service_impl.h
oschaaf: Let me figure out why I tackled this like I did here, and ping you on Slack when I did that to discuss. I think this was CI specific.

api/client/options.proto
```diff
@@ -9,38 +9,43 @@ import "envoy/api/v2/core/base.proto";
...
-  string address_family = 11;
+  string address_family = 11 [(validate.rules).string = {in: ["v4", "v6"]}];
   oneof oneof_request_options {
     // See :option:`--request_options` for details.
     RequestOptions request_options = 12;
   }
   // See :option:`--uri` for details.
   string uri = 13;
+  // TODO(oschaaf): pvg throw NotImplementedException for the following rule:
+  // string uri = 13 [(validate.rules).string.uri = true];
```
htuch: I think for these fields with missing validators, you can include them but just comment out the validation rule.

api/client/service.proto
```diff
@@ -0,0 +1,50 @@
...
+
+message ExecutionRequest {
+  oneof command_specific_options {
+    StartRequest start_request = 1;
+    UpdateRequest update_request = 2;
+    CancellationRequest cancellation_request = 3;
+  }
+}
+
+/*
```
htuch: Nit, can you use `//` comments even for blocks, similar to Envoy style.

api/client/service.proto
```diff
@@ -0,0 +1,50 @@
...
+Output is guaranteed to be set when success is true. When success is false,
+output may or may not be set, and error message is guaranteed to be set.
+*/
+message ExecutionResponse {
+  Output output = 1;
```
htuch: This could be:
``
oneof status_specifier {
  Output output = 1;
  Error error = 2;
}
```

with the `error_message` a field in `Error`. I think there is something already defined like this in the WKT, see https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/discovery.proto#L190.
```

source/client/service_impl.cc
```diff
@@ -0,0 +1,103 @@
...
+#include <grpc++/grpc++.h>
+
+#include "client/client.h"
+#include "client/options_impl.h"
+
+namespace Nighthawk {
+namespace Client {
+
+void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionRequest& request) {
+  RELEASE_ASSERT(running_, "running_ ought to be set");
```
htuch: Is this an invariant? Should it just be `ASSERT`?

api/client/service.proto
```diff
@@ -0,0 +1,50 @@
...
+  string error_message = 3;
+}
+
+service NighthawkService {
+
+  /*
+  Accepts a stream of ExecutionRequests, which can be used to initiate, update, or cancel a test.
+  Returns a stream of ExecutionResponse objects which will contain results for all Nighthawk runs.
+  */
+  rpc sendCommand(stream ExecutionRequest) returns (stream ExecutionResponse) {
```
htuch: Nit: s/sendCommand/ExecutionStream; the RPC is a series of request/responses around an execution, rather than sending a single command. This can also be reflected in method names below.

source/client/service_impl.cc
```diff
@@ -0,0 +1,103 @@
...
+  if (error_messages.size() == 0) {
+    return grpc::Status::OK;
+  }
+
+  std::string error_message;
+  for (const auto& error : error_messages) {
+    error_message = fmt::format("{}\n", error);
+  }
+  ENVOY_LOG(error, "One or more errors processing grpc request stream: {}", error_message);
+  return grpc::Status(grpc::StatusCode::INTERNAL, fmt::format("Error: {}", error_message));
```
htuch: I think it would be fine to always reply `grpc::Status::OK`, since we have returned errors explicitly as responses on the stream. The way I think of this is that there are CLI execution errors and gRPC transport errors, in this case we have a good gRPC stream, just application level failure. This avoids having to accumulate all the errors, they can be logged as we go.

source/client/service_impl.cc
```diff
@@ -0,0 +1,103 @@
...
+      if (running_) {
+        error_messages.emplace_back("Only a single benchmark session is allowed at a time.");
+        break;
+      } else {
+        waitForRunnerThreadCompletion();
+        running_ = true;
+        runner_thread_ = std::thread(&ServiceImpl::handleExecutionRequest, this, request);
+      }
+    } else if (request.has_update_request() || request.has_cancellation_request()) {
+      error_messages.emplace_back("Request is not supported yet.");
```
htuch: In this case, you could return immediately with an internal error, as this is a stream violation.

source/client/service_impl.cc
```diff
@@ -0,0 +1,103 @@
...
+    ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
+                               ::nighthawk::client::ExecutionRequest>* stream) {
+  std::list<std::string> error_messages;
+  nighthawk::client::ExecutionRequest request;
+  stream_ = stream;
+  while (stream->Read(&request)) {
+    ENVOY_LOG(debug, "Read ExecutionRequest data: {}", request.DebugString());
+    if (request.has_start_request()) {
+      if (running_) {
+        error_messages.emplace_back("Only a single benchmark session is allowed at a time.");
```
htuch: It would be legit to cancel the stream here as well.

api/client/service.proto
oschaaf: Fixed

api/client/service.proto
oschaaf: Fixed

api/client/options.proto
oschaaf: Fixed

api/client/service.proto
oschaaf: Fixed, I think, please double check.

source/client/service_impl.cc
oschaaf: Fixed

source/client/service_impl.cc
oschaaf: Fixed

source/client/service_impl.cc
oschaaf: Fixed

source/client/service_impl.cc
oschaaf: makes a lot of sense, adjusted accordingly.

source/client/service_impl.h
oschaaf: As discussed out of band, `grpc_cc_proto` doesn't allow us to pass in `copts`, so we'd have to apply the suppression to a wider scope; I propose we keep this as-is.

api/client/service.proto
```diff
@@ -0,0 +1,43 @@
...
+message UpdateRequest {
+}
+
+// TODO(oschaaf): Not implemented yet.
+message CancellationRequest {
+}
+
+message ExecutionRequest {
+  oneof command_specific_options {
+    StartRequest start_request = 1;
```
htuch: You may want to set `option (validate.required) = true;` on this.

api/client/service.proto
```diff
@@ -0,0 +1,43 @@
...
+    UpdateRequest update_request = 2;
+    CancellationRequest cancellation_request = 3;
+  }
+}
+
+message ExecutionResponse {
+  oneof status_specifier {
```
htuch: Ditto

source/client/service_impl.cc
```diff
@@ -0,0 +1,91 @@
...
+  ProcessImpl process(*options, time_system_);
+  OutputCollectorFactoryImpl output_format_factory(time_system_, *options);
+  auto logging_context = std::make_unique<Envoy::Logger::Context>(
+      spdlog::level::from_str(options->verbosity()), "[%T.%f][%t][%L] %v", log_lock_);
+  auto formatter = output_format_factory.create();
+  if (process.run(*formatter)) {
+    response.clear_error_detail();
+    *(response.mutable_output()) = formatter->toProto();
+  } else {
+    response.mutable_error_detail()->set_message("Unknown failure");
```
htuch: Nit: This line is missing test coverage.

api/client/service.proto
```diff
@@ -0,0 +1,43 @@
...
+    Output output = 1;
+    google.rpc.Status error_detail = 7;
+  }
+}
+
+service NighthawkService {
+
```
htuch: Nit: superfluous blank line.

api/client/BUILD
```diff
@@ -20,3 +14,42 @@ api_proto_library(
         "@envoy_api//envoy/api/v2/core:base_export",
     ],
 )
+
+# TODO(oschaaf): This is a bit of a hack, we conjure the expected _base_cc_only ourselves here
+# because it doesn't get generated for us. Also see https://github.com/grpc/grpc/issues/14145
```
htuch: Did you try https://github.com/grpc/grpc/issues/14145#issuecomment-490809961?

source/client/service_impl.cc
oschaaf: Fixed

source/client/service_impl.cc
```diff
@@ -0,0 +1,91 @@
...
+
+// TODO(oschaaf): implement a way to cancel test runs, and update rps config on the fly.
+// TODO(oschaaf): unit-test Process, create MockProcess & use in service_test.cc / client_test.cc
+// TODO(oschaaf): should we merge incoming request options with defaults?
+// TODO(oschaaf): aggregate the client's logs and forward them in the grpc response.
+::grpc::Status ServiceImpl::ExecutionStream(
+    ::grpc::ServerContext* /*context*/,
+    ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
+                               ::nighthawk::client::ExecutionRequest>* stream) {
+  std::list<std::string> error_messages;
```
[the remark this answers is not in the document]
oschaaf: Removed

api/client/BUILD
oschaaf: Amongst a lot of things, I also tried that and failed; I remember our `cc_grpc_library` seems to expect different inputs. Maybe this dependency needs to be updated on our side.

api/client/service.proto
oschaaf: Done

api/client/service.proto
oschaaf: Done

api/client/service.proto
oschaaf: Removed

source/client/service_impl.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <grpc++/grpc++.h>
+
+#include "client/client.h"
+#include "client/options_impl.h"
+
+namespace Nighthawk {
+namespace Client {
+
+void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionRequest& request) {
+  ASSERT(running_, "running_ ought to be set");
```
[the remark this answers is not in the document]
oschaaf: I'm not sure I'm seeing it; could you ping me on slack to discuss if you get a chance?

api/client/BUILD
oschaaf: Good news on this one, this is fixed [here](https://github.com/envoyproxy/nighthawk/pull/88/files#diff-04dcaeb167b260b75a61e7f95b910e72) 
(The hack broke after updating the Envoy dependency, and instead the more sane way to generate the grpc stub works with that!)

source/common/sequencer_impl.cc
```diff
@@ -158,6 +157,8 @@ void SequencerImpl::waitForCompletion() {
   ASSERT(!running_);
 }
 
+void SequencerImpl::cancel() { cancelled_ = true; }
```
htuch: It seems like this PR has a partial implementation of request cancellation.. should this just be added as a supported request type or should these be removed?

test/service_test.cc
```diff
@@ -0,0 +1,171 @@
...
+                                   grpc::InsecureChannelCredentials());
+    stub_ = std::make_unique<nighthawk::client::NighthawkService::Stub>(channel_);
+    setBasicRequestOptions();
+  }
+
+  void TearDown() override { server_->Shutdown(); }
+
+  void setBasicRequestOptions() {
+    auto options = request_.mutable_start_request()->mutable_options();
+    options->set_uri("http://127.0.0.1:10001/");
```
htuch: Does this actually result in any loopback traffic being sent by NH? Or is this somehow mocked? I can't easily spot the later.

htuch: It's fine in a followup (maybe add a TODO), thanks. My main concern is in IPv6-only setups, which we need to support at Google.

htuch: If you can add a `TODO` we can resolve this one.

source/client/service_impl.cc
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+  writeResponseAndFinish(response);
+}
+
+void ServiceImpl::writeResponseAndFinish(const nighthawk::client::ExecutionResponse& response) {
+  if (!stream_->Write(response)) {
+    ENVOY_LOG(warn, "Stream write failed");
+  }
+  running_ = false;
```
htuch: The `ProcessImpl` fix looks good, the only other thing I'd suggest is moving this line above the `Write`.

source/client/service_impl.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void ServiceImpl::waitForRunnerThreadCompletion() {
+  if (runner_thread_.joinable()) {
```
htuch: I had one thought on a good pattern here. What if instead of an explicit thread, we use `std::async` and `std::future`? Could we avoid a shared `running_` flag with this pattern?

PR #91, opened 2019-06-27
Switch to Pyhon3 plus some groundwork for #88

oschaaf:
Switch to Pyhon3 plus some groundwork for #88

- Switches the integration tests to use python 3 
- Update the CI build image
- Use RAII in python socket handling as that now becomes possible
- Adds some things we'll need in https://github.com/envoyproxy/nighthawk/pull/88


integration/BUILD
```diff
@@ -12,4 +20,16 @@ py_test(
         "//:nighthawk_test_server",
         "@envoy//test/config/integration/certs",
     ],
+    # We enforce python 2, but I we may be able to stop doing that when we update
+    # bazel in CI. (Or we add python3 explicitly as a build dependency somehow).
```
htuch: Bazel is now at 0.27 on the latest hash, there is no PY2 here, stale comment?

integration/integration_test_fixtures.py
```diff
@@ -43,11 +45,10 @@ def getFreeListenerPortForAddress(self, address):
...
     """
-    # TODO(oschaaf): When we are on python 3 use RAII here.
-    sock = socket.socket(self.socket_type, socket.SOCK_STREAM)
-    sock.bind((address, 0))
-    port = sock.getsockname()[1]
-    sock.close()
+    with socket.socket(self.socket_type, socket.SOCK_STREAM) as sock:
+      sock.bind((address, 0))
+      port = sock.getsockname()[1]
+      sock.close()
```
htuch: No need for `close`.

integration/nighthawk_test_server.py
```diff
@@ -48,24 +48,24 @@ def serverThreadRunner(self):
...
-    return False
+  def waitUntilServerListening(self):
+    with socket.socket(self.socket_type, socket.SOCK_STREAM) as sock:
+      sock.settimeout(1)
+      tries = 10
+      while tries > 0:
+        time.sleep(0.5)
+        tries -= 1
+        if sock.connect_ex((self.server_ip, self.server_port)) == 0:
+          sock.close()
```
htuch: No need for any `close`.

integration/BUILD
oschaaf: Removed

Issue #22, opened 2019-04-23
Nighthawk as-a-service


oschaaf:
Currently Nighthawk supports a CLI which allows for quick and easy execution of single test runs.
But it would be really neat to have a service next to that, which would accept configuration (GRPC or CLI) and (re-)configure on the fly as requested.

oschaaf:
Update: this is in progress,  [design draft](https://docs.google.com/document/d/1LT04urVBF8msffS_ObPKLhzY6mxY7sspkm5iVSLkAMI/edit?usp=sharing)

mum4k:
@oschaaf is this issue also related to #137 and #229? Or is this tracking something else?

oschaaf:
> @oschaaf is this issue also related to #137 and #229? Or is this tracking something else?

This is tracking status of controlling load generation over gRPC, so #137 and #229 are related, yes

PR #92, opened 2019-06-27
Grpc service executable build

oschaaf:
Adds a target to the build for building the grpc service executable.

Final piece to the puzzle for resolving https://github.com/envoyproxy/nighthawk/issues/22


oschaaf:
While this still has rough edges, marking as `ready for review` as feedback is welcome at this stage.

source/client/service_main.cc
```diff
@@ -0,0 +1,58 @@
...
+
+  try {
+    listener_address_ =
+        Envoy::Network::Utility::parseInternetAddressAndPort(internet_address_and_port.getValue());
+  } catch (Envoy::EnvoyException e) {
+    throw NighthawkException(e.what());
+  }
+}
+
+void ServiceMain::Start() {
```
mergeconflict: Here's maybe an odd thought... Is it possible to implement "Nighthawk-as-a-service" as an Envoy filter, rather than a one-off service? That is, Envoy would be configured as an origin server rather than a proxy, and this Nighthawk gRPC service would just be a filter that Envoy would delegate to. WDYT?

source/client/BUILD
```diff
@@ -49,3 +47,20 @@ envoy_cc_library(
         "@envoy//source/exe:process_wide_lib",
     ],
 )
+
+envoy_cc_library(
```
mergeconflict: I'm still really unfamiliar with Nighthawk, so maybe this would make more sense to me with a quick walkthrough, but I'm not clear on why "service" stuff is living in /source/client. Similar with the class/namespace names, e.g. `Nighthawk::Client::ServiceMain`.

source/client/service_main.cc
```diff
@@ -0,0 +1,58 @@
...
+  grpc::ServerBuilder builder;
+  int grpc_server_port = listener_address_->ip()->port();
+  builder.AddListeningPort(listener_address_->asString(), grpc::InsecureServerCredentials(),
+                           &grpc_server_port);
+  builder.RegisterService(&service_);
+  server_ = builder.BuildAndStart();
+  if (server_ == nullptr) {
+    throw NighthawkException("Could not start the grpc service.");
+  }
+  std::cout << "Nighthawk grpc service listening: " << listener_address_->asString() << std::endl;
```
mergeconflict: It'd probably better if we could use the same logging mechanism that Envoy uses, rather than `cout` / `cerr`.

source/common/utility.cc
```diff
@@ -62,4 +64,23 @@ Envoy::Network::DnsLookupFamily Utility::parseAddressFamilyOptionString(absl::st
   throw NighthawkException("Invalid argument");
 }
 
+void Utility::parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* const* argv) {
+  cmd.setExceptionHandling(false);
+  try {
+    cmd.parse(argc, argv);
+  } catch (TCLAP::ArgException& e) {
```
mergeconflict: I don't really love how we're using exceptions so heavily for control-flow, which I guess means I don't really love TCLAP. I feel like it might be more Envoy-ish (particularly as a server, rather than a CLI tool) to just take a configuration file, similar to Envoy's bootstrap proto.

I guess I could say this is maybe a small argument in favor of making this service an Envoy filter, like I suggested above, but I'll definitely admit I don't know how much harder it'd be.

source/exe/service_main_entry.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "client/service_main.h"
+
+#include "absl/debugging/symbolize.h"
+
+// NOLINT(namespace-nighthawk)
+
+using namespace Nighthawk::Client;
+
+int main(int argc, const char** argv) {
+#ifndef __APPLE__
```
mergeconflict: @jmarantz, this looks like something you might be knowledgeable about as a reviewer...

test/service_main_test.cc
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Nighthawk {
+namespace Client {
+
+class ServiceMainTest : public Test {};
+
+// TODO(oschaaf): this gets us some coverage, but we need more functional testing.
+// See if we can add some python integration tests.
+TEST_F(ServiceMainTest, HelloWorld) {
+  std::vector<const char*> argv;
```
mergeconflict: I think you can use an initializer list rather than `push_back`.

oschaaf: Fixed

source/client/service_main.cc
oschaaf: Good point, fixed

ci/run_coverage.sh
```diff
@@ -32,7 +32,7 @@ if [ "$VALIDATE_COVERAGE" == "true" ]
 then
     COVERAGE_VALUE=$(grep -Po '.*lines[.]*: \K(\d|\.)*' "${COVERAGE_SUMMARY}")
     # TODO(oschaaf): The target is 97.5%, so up this whenever possible in follow ups.
-    COVERAGE_THRESHOLD=96.8
+    COVERAGE_THRESHOLD=97.0
```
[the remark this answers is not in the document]
jmarantz: Can you add a comment with a pointer to the coverage dashboard, if we have one?

If not, we should probably have one, like https://s3.amazonaws.com/lyft-envoy/coverage/report-master/coverage.html . In that case add another TODO?

source/client/BUILD
jmarantz: sgtm; add a TODO for that too?

source/client/service_main.cc
jmarantz: I assume there's a ton of weight in Nighthawk that people who just want production proxies might not want in their binary. It seems OK to me to have it as a compiled-out-by-default extension though; not sure. This strategic discussion needs @htuch  for resolution.

source/exe/service_main_entry.cc
jmarantz: this seems OK. Probably we should wrap this in a helper function to hide the ifdef, but not critical as it happens only once in this repo.

jmarantz:
Nope, my approval does not make it possible to merge.

ci/run_coverage.sh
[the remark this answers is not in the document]
oschaaf: Until recently I used to be able to access that type of report via the CircleCI [1]. 
However the interface elements in CircleCI to access that report recently disappeared somehow.
It's still uploading the artifacts, I'm not sure what is wrong.
Filed https://github.com/envoyproxy/nighthawk/issues/95

[1] E.g. https://circleci.com/gh/envoyproxy/nighthawk/1878#tests/containers/0

source/client/BUILD
oschaaf: Filed https://github.com/envoyproxy/nighthawk/issues/96

oschaaf:
> basically lgtm modulo nits. Going to approve to see if that gives me 'merge' power, but please fix the nit (adding TODOs mostly) and we'll see if I can merge.

Thanks for the review; I didn't push any code-changes, but I did file relevant issues.

jmarantz:
Re filling issues vs adding TODOs: fine with me. Will need to wait for
Harvey to merge and then add other approvers.

source/client/service_main.cc
htuch: It's an interesting idea. The main drawback I see is that filters are (at least by design) fairly limited in what they can do. Nighthawk reaches deep into things like connection pool management, request scheduling and configuration management, so it probably makes sense to have it be its own thing.

source/client/service_main.cc
```diff
@@ -0,0 +1,58 @@
...
+#include "common/utility.h"
+
+#include "client/service_impl.h"
+
+#include "tclap/CmdLine.h"
+
+namespace Nighthawk {
+namespace Client {
+
+ServiceMain::ServiceMain(int argc, const char** argv) {
```
htuch: `ServiceMain::serviceMain`, i.e. follow Envoy style here.

source/client/service_main.cc
```diff
@@ -0,0 +1,58 @@
...
+    listener_address_ =
+        Envoy::Network::Utility::parseInternetAddressAndPort(internet_address_and_port.getValue());
+  } catch (Envoy::EnvoyException e) {
+    throw NighthawkException(e.what());
+  }
+}
+
+void ServiceMain::Start() {
+  grpc::ServerBuilder builder;
+  int grpc_server_port = listener_address_->ip()->port();
```
htuch: Nit: `const int`

source/client/service_main.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "api/client/service.pb.h"
+#include "tclap/CmdLine.h"
+
+namespace Nighthawk {
+namespace Client {
+
+class ServiceMain : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  ServiceMain(int argc, const char** argv);
```
htuch: Nit: `lowerCase()` all these methods.

source/common/utility.cc
htuch: Ultimately we want any non-trivial config for this to be proto-driven. As long as it's as simple as it is today, CLI is fine. The NH CLI tool needs to have a decent CLI interface to be competitive with other tools, but NH-as-a-service is only used in automation most likely.

oschaaf: Ack

PR #93, opened 2019-07-03
Draft: (object) Pool & Stopwatch concepts

oschaaf:
- A generic object pool based on `unique_ptr`s with a custom deleter. 
- A stopwatch concept as an abstraction layer for latency measurement

Early draft PR, early feedback on the general idea is very welcome! :-)

Basically this is some groundwork which may be useful later.. 

The pool should help reducing per-request allocations by recycling certain concepts, like
streams, stopwatches, tracers, etc, as we can pre-allocate and also recycle used objects.
(I suspect we will end up seeing more rather then less of them in the future).

The stopwatch abstracts time measurements way, and may help the program perform
less calls to get the monotonic clock when we put it to use in favor of what we have today.


oschaaf:
@mergeconflict this is in draft, but if you do have time for it, high-level thoughts on the general approach and idea would be super-appreciated!

source/common/pool_impl.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include "nighthawk/common/poolable.h"
+
+#include "common/common/assert.h"
+
+namespace Nighthawk {
+
+template <typename Poolable> class PoolImpl {
+public:
+  using poolDeletionDelegate = std::function<void(Poolable*)>;
```
jmarantz: I think we should capitalize the "p" in "pool" here.

include/nighthawk/common/poolable.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "envoy/common/pure.h"
+
+namespace Nighthawk {
+
+class Poolable {
```
jmarantz: As you are using templates, I don't think you need this ABC at all. It's nice in a sense as documentation and compiler enforcement that your implementations implement your interface, but it costs you some memory and performance without adding functionality in this design, AFAICT.

How many Poolables do you expect to have lying around during operation? Maybe the above doesn't matter.

source/common/pool_impl.h
```diff
@@ -0,0 +1,64 @@
...
+#include <stack>
+#include <vector>
+
+#include "nighthawk/common/poolable.h"
+
+#include "common/common/assert.h"
+
+namespace Nighthawk {
+
+template <typename Poolable> class PoolImpl {
```
jmarantz: It is weird that you have Poolable both as an ABC and a template name. I think it means that no mehtods are actually called on the ABC; the compiler will know about the derived type during template expansion.

include/nighthawk/common/poolable.h
```diff
@@ -0,0 +1,14 @@
...
+public:
+  virtual ~Poolable() = default;
+  virtual void orphan() PURE;
```
jmarantz: can you say more about the orphan verb and the orphaned property? I'm not sure exactly what the implication is.

oschaaf: Improved naming, added doxygen comments.

source/common/pool_impl.h
oschaaf: done

include/nighthawk/common/milestone_tracker.h
```diff
@@ -40,8 +40,8 @@ class MilestoneTracker {
    * @return std::chrono::duration<double> elapsed duration between marking the
```
mergeconflict: Update return type comment.

source/client/benchmark_client_impl.cc
```diff
@@ -43,9 +43,10 @@ BenchmarkClientHttpImpl::BenchmarkClientHttpImpl(Envoy::Api::Api& api,
       benchmark_client_stats_({ALL_BENCHMARK_CLIENT_STATS(POOL_COUNTER(*scope_))}),
       stream_decoder_pool_(
           [this]() {
-            return new PoolableStreamDecoder(
+            PoolableStreamDecoder* decoder = new PoolableStreamDecoder(
```
mergeconflict: Seems you could just `return new ...`

source/client/stream_decoder.cc
```diff
@@ -19,27 +20,37 @@ void StreamDecoder::decodeHeaders(Envoy::Http::HeaderMapPtr&& headers, bool end_
...
   complete_ = true;
   onComplete(true);
 }
 
 void StreamDecoder::onComplete(bool success) {
-  if (success && measure_latencies_) {
-    latency_statistic_.addValue((time_source_.monotonicTime() - request_start_).count());
-  }
   ASSERT(!success || complete_);
+  milestone_tracker_.markMilestone(milestone_response_complete_);
```
mergeconflict: Maybe add some comments here and in `onPoolReady` below about what stats are being tracked?

source/client/stream_decoder.h
```diff
@@ -78,14 +88,21 @@ class StreamDecoder : public Envoy::Http::StreamDecoder,
     caller_completion_callback_ = std::move(caller_completion_callback);
   }
 
-  void reset() { complete_ = false; }
+  void reset() {
+    complete_ = false;
+    have_first_body_byte_ = false;
+    milestone_tracker_.reset();
+    // TODO(oschaaf): this is kind of a hack to mark this here.
```
mergeconflict: Can you expand the comment a bit to explain why it's a hack, and what would be better?

source/common/milestone_tracker_impl.h
```diff
@@ -16,13 +19,30 @@ class MilestoneTrackerImpl : public MilestoneTracker {
...
 
+// Compose a poolable MilestoneTrackerImpl.
+class PoolableMilestoneTrackerImpl : public MilestoneTrackerImpl, public PoolableImpl {
+public:
+  PoolableMilestoneTrackerImpl(Envoy::TimeSource& time_source)
+      : MilestoneTrackerImpl(time_source) {}
+};
+
+// Declare a pool for the poolable milestone
+class MilestoneTrackerPoolImpl : public PoolImpl<PoolableMilestoneTrackerImpl> {
```
mergeconflict: Same comment as for `StreamDecoderPoolImpl`, I don't think this needs to be a class.

include/nighthawk/common/poolable.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "envoy/common/pure.h"
+
+#include "common/common/non_copyable.h"
+
+namespace Nighthawk {
+
+class Poolable : Envoy::NonCopyable {
```
mergeconflict: Can you write some comments here about the intended usage? For example, what happens when a `Poolable` is orphaned? Who is responsible for deleting it?

source/client/stream_decoder.h
```diff
@@ -64,10 +75,41 @@ class StreamDecoder : public Envoy::Http::StreamDecoder,
...
+    milestone_tracker_.markMilestone(milestone_request_start_);
+  }
+
+  void setMeasureLatencies(const bool measure_latencies) { measure_latencies_ = measure_latencies; }
+  void freeSelf() {
```
mergeconflict: This feels scary to me. I don't think an object should generally care who owns it; it should behave uniformly regardless. If you want some specific ownership/lifetime semantics for an object, you wrap it in something like a `unique_ptr` or `shared_ptr` that captures those semantics.

The `delete this` below is especially scary. It may be technically correct right now (see https://isocpp.org/wiki/faq/freestore-mgmt#delete-this), but I've always been told it should be avoided (https://doc.castsoftware.com/pages/viewpage.action?pageId=211748357).

source/client/stream_decoder.h
```diff
@@ -78,8 +120,40 @@ class StreamDecoder : public Envoy::Http::StreamDecoder,
...
+                        std::function<void()> caller_completion_callback,
+                        Statistic& connect_statistic, Statistic& latency_statistic,
+                        const Envoy::Http::HeaderMap& request_headers, bool measure_latencies,
+                        uint32_t request_body_size)
+      : StreamDecoder(time_source, decoder_completion_callback,
+                      std::move(caller_completion_callback), connect_statistic, latency_statistic,
+                      request_headers, measure_latencies, request_body_size) {}
+};
+
+class StreamDecoderPoolImpl : public PoolImpl<PoolableStreamDecoder> {
```
mergeconflict: Does this need to be a class? I think it could just be a `using` alias or a `typedef`.

source/common/pool_impl.h
```diff
@@ -0,0 +1,84 @@
...
+           PoolInstanceResetDelegate&& reset_delegate = nullptr)
+      : construction_delegate_(construction_delegate), reset_delegate_(reset_delegate) {}
+
+  ~PoolImpl() {
+    while (!pool_.empty()) {
+      Poolable* poolable = pool_.back().get();
+      all_.erase(std::remove(all_.begin(), all_.end(), poolable), all_.end());
+      pool_.pop_back();
+    }
+    // Inform the in-flight poolables that they are own their own now.
```
mergeconflict: What are they supposed to do at this point?

source/common/pool_impl.h
```diff
@@ -0,0 +1,84 @@
...
+    for (auto poolable : all_) {
+      poolable->mark_orphaned();
+    }
+  }
+
+  void addPoolable(std::unique_ptr<Poolable>&& poolable) {
```
mergeconflict: I think the `&&` isn't needed; the `unique_ptr` argument has to be an rvalue anyway if it's a by-value parameter.

source/common/pool_impl.h
```diff
@@ -0,0 +1,84 @@
...
+    ASSERT(poolable.get() != nullptr);
+    all_.push_back(poolable.get());
+    pool_.emplace_back(std::move(poolable));
+  }
+
+  PoolablePtr get() {
```
mergeconflict: This is kind of interesting, if I'm reading it correctly. It seems like the intended semantics are kind of like borrowing a book from the library: only one person can have the book at a time, and if the library burns down, the book belongs to you. But I have to wonder if it's really needed, and whether there's a simpler way to get what we need.

source/common/poolable_impl.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "nighthawk/common/poolable.h"
+
+#include "common/common/assert.h"
+
+namespace Nighthawk {
+
+class PoolableImpl : public Poolable {
```
mergeconflict: I think we try to avoid mix-in (multiply inherited) classes like this, generally. Could it just be some kind of `Handle` class that is contained rather than inherited?

test/pool_test.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "test/mocks.h"
+#include "test/test_common/simulated_time_system.h"
+
+#include "gtest/gtest.h"
+
+using namespace testing;
+
+namespace Nighthawk {
+
+class MockPoolablePoolImpl : public PoolImpl<MockPoolable> {
```
mergeconflict: As above, probably doesn't need to be a class.

include/nighthawk/common/milestone_tracker.h
oschaaf: done

include/nighthawk/common/poolable.h
oschaaf: done, hope these comments are clarifying

source/client/benchmark_client_impl.cc
oschaaf: done

source/client/stream_decoder.h
oschaaf: comment updated

source/client/stream_decoder.h
oschaaf: Oh, yes, much nicer, done.

source/common/milestone_tracker_impl.h
oschaaf: fixed!

source/common/pool_impl.h
oschaaf: update comment

source/common/pool_impl.h
oschaaf: changed

source/common/pool_impl.h
oschaaf: Nice analogy, and correct. One TODO is that I still have to look into if we (still need) to use `dispatcher_.deferredDelete` here. It may be that we need to transfer ownership to the dispatcher instead of destructing directly here.

test/pool_test.cc
oschaaf: fixed

source/common/poolable_impl.h
oschaaf: Let me think about that, will revisit this comment once I did so

include/nighthawk/common/poolable.h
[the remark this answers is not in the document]
htuch: Also, with use of arenas etc.

include/nighthawk/common/milestone_tracker.h
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Call when a milestone is reached. Annotatess the milestone with a stimestamp.
+   */
+  virtual void markMilestone(const uint32_t milestone) PURE;
+
+  /**
+   * @return Envoy::MonotonicTime for when the milestone was reached.
+   */
+  virtual const Envoy::MonotonicTime getMilestone(const uint32_t milestone) const PURE;
```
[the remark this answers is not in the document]
htuch: Yeah, I don't think this does anything.

include/nighthawk/common/milestone_tracker.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
```
htuch: Yeah, I reckon it makes sense to split out the milestone tracker from the pool PR.

include/nighthawk/common/milestone_tracker.h
```diff
@@ -0,0 +1,49 @@
...
+#include <memory>
+
+#include "envoy/common/pure.h"
+#include "envoy/common/time.h"
+
+#include "absl/strings/string_view.h"
+
+namespace Nighthawk {
+
+class MilestoneTracker {
```
htuch: What this models is very similar to a trace span in something like Zipkin or OpenCensus. Envoy has some abstractions for this, see https://github.com/envoyproxy/envoy/blob/master/include/envoy/tracing/http_tracer.h#L69. I wonder if there is any useful overlap here?

Should we just use some tracing library like OpenCensus? Is the overhead and resolution of these sufficient for tracking time inside NH? This also reminds me a bit of VMware Kstats.

htuch:
@oschaaf for the pool concept, can you elaborate a bit on where you see the major wins being? Is this to reduce jitter and tail latency via a "zero allocation" model? Will we always be able to use the pool or will there still be a long tail where it is exhausted? Is this for objects that are expensive to build for reasons beyond memory allocation, e.g. compute intensive init/destructors?

include/nighthawk/common/poolable.h
mergeconflict: Yup, thanks! So, if my understanding is correct, I think `Poolable` isn't the right design for this. That is, I don't think you should need to implement `Poolable` for a class that you want to be manageable by a pool. As an analogy, you don't need to implement a `Shareable` interface to be managed by a `shared_ptr`.

Instead, I think you probably want to create a `PoolPtr<T>` class to capture the semantics you intend. It will probably behave a bit like a `unique_ptr` (in the sense that only one owner can borrow from the pool at a time, so it can only be moved and not copied), but might also have a `bool is_orphaned_` that the pool can set directly. If you can implement it just using `unique_ptr` with a custom deleter, the way you're already mostly doing, even better...

But, yeah, I agree with Josh and Harvey's comments: we should probably get some good evidence to support the need for this first.

oschaaf:
@mergeconflict @jmarantz @htuch 
thanks all for the feedback so far!

To help with converging this towards a decision: not super-nicely written, but [I have a doc here](https://docs.google.com/document/d/1_YNQFm-kRVknLU9PxyA_aSlvT3gD9ti8YvOztBzw2nk/edit?usp=sharing) which consolidates my notes, thoughts including a pro/con analysis and rationale in the first section.

include/nighthawk/common/poolable.h
oschaaf: Yes I agree the high level question should be addressed first. That aside, I do really like the idea you describe here, I feel following up on that would yield a huge conceptual improvement to this

oschaaf:
Closing this. I think https://docs.google.com/document/d/1_YNQFm-kRVknLU9PxyA_aSlvT3gD9ti8YvOztBzw2nk/edit?usp=sharing captures some value, and @mergeconflict came up with a [much nicer conceptual model](https://github.com/envoyproxy/nighthawk/pull/93/files#r301650830) during review  (`PoolPtr<T>`) .

Issue #80, opened 2019-05-06
Add an option to control how request pacing is maintained.


oschaaf:
Currently workers use a yield/spin-loop to determine when it is time for a request to be initiated. Add an option, that upon enabling, makes the client use the dispatcher to `sleep()` as an alternative.

PR #94, opened 2019-07-04
Optionize the sequencer idle strategy

oschaaf:
Choose between a busy (spin/yield) loop or `sleep()` while the sequencer waits for the next
scheduled request.

Needs https://github.com/envoyproxy/nighthawk/pull/88 to go in first.

Upon completion, this resolves https://github.com/envoyproxy/nighthawk/issues/80


oschaaf:
Assuming CI will be happy, I think this is ready for review, But to make reviewing easier, #88 would have to be merged to master first (and then I can rebase this, and we have a relevant diff).

oschaaf:
`clang-tidy` complains about something originating from TCLAP [1].

I am going to add another suppression, which will green light CI, but I'm a little confused; I feel we should have seen this earlier as the line clang-tidy points out is very similar to pre-existing code, which it flagged as OK.
So the warnings seem to pop up inconsistently, but when they do they are very reproducible. (In an earlier PR I added another similar suppression). 

I would appreciate it if someone can take a close look at if I didn't miss something important which triggers these seemingly arbitrary complaints, to ensure I'm not squelching a symptom to an actual problem. Subsequently, if consensus is that this is benign, I wonder if there's a better way to suppress this particular error locally in this file. Ideally we would then exclude this particular check from the code section that uses these TCLAP api's.

[1] 
```
clang-tidy-8 -header-filter=^/root/project/.* -extra-arg-before=-xc++ -p=/root/project -quiet /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/source/client/options_impl.cc
external/com_github_eile_tclap/include/tclap/Arg.h:491:8: error: Call to virtual function during construction [clang-analyzer-optin.cplusplus.VirtualCall,-warnings-as-errors]
                                                        toString() ) );
                                                        ^
source/client/options_impl.cc:140:29: note: Calling constructor for 'ValueArg<unsigned int>'
  TCLAP::ValueArg<uint32_t> max_requests_per_connection(
                            ^
external/com_github_eile_tclap/include/tclap/ValueArg.h:275:3: note: Calling constructor for 'Arg'
: Arg(flag, name, desc, req, true, v),
  ^
external/com_github_eile_tclap/include/tclap/Arg.h:480:7: note: Assuming the condition is false
        if ( _flag.length() > 1 )
             ^
external/com_github_eile_tclap/include/tclap/Arg.h:480:2: note: Taking false branch
        if ( _flag.length() > 1 )
        ^
external/com_github_eile_tclap/include/tclap/Arg.h:484:7: note: Left side of '&&' is true
        if ( _name != ignoreNameString() &&
             ^
external/com_github_eile_tclap/include/tclap/Arg.h:485:38: note: Left side of '||' is true
                 ( _flag == Arg::flagStartString() ||
                                                   ^
external/com_github_eile_tclap/include/tclap/Arg.h:484:2: note: Taking true branch
        if ( _name != ignoreNameString() &&
        ^
external/com_github_eile_tclap/include/tclap/Arg.h:491:8: note: This constructor of an object of type 'Arg' has not returned when the virtual method was called
                                                        toString() ) );
                                                        ^
```

htuch:
@oschaaf can you merge master and fix conflicts? Will take a look at the cleaned up diff. Thanks.

oschaaf:
@htuch merged master into this branch (& conflicts resolved)

source/client/options_impl.cc
```diff
@@ -136,12 +136,19 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "", "max-active-requests",
       fmt::format("Max active requests (default: {}).", max_active_requests_), false, 0, "uint32_t",
       cmd);
-
+  // NOLINTNEXTLINE
```
mergeconflict: Is this needed?

source/client/options_impl.cc
```diff
@@ -136,12 +136,19 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
   TCLAP::ValueArg<uint32_t> max_requests_per_connection(
       "", "max-requests-per-connection",
       fmt::format("Max requests per connection (default: {}).", max_requests_per_connection_),
       false, 0, "uint32_t", cmd);
 
+  TCLAP::ValueArg<std::string> sequencer_idle_strategy(
+      "", "sequencer-idle-strategy",
+      fmt::format("Choose between using a busy spin/yield loop or have the thread sleep while "
```
mergeconflict: Probably want to list the legal values here.

api/client/options.proto
```diff
@@ -56,6 +56,9 @@ message CommandLineOptions {
   google.protobuf.UInt32Value max_active_requests = 15 [(validate.rules).uint32 = {gte: 1}];
   // See :option:`--max-requests-per-connection` for details.
   google.protobuf.UInt32Value max_requests_per_connection = 16 [(validate.rules).uint32 = {gte: 1}];
+  // See :option:`--output-format` for details.
+  google.protobuf.StringValue sequencer_idle_strategy = 17
```
mergeconflict: I think this might be better as an enum.

include/nighthawk/client/options.h
```diff
@@ -40,6 +40,7 @@ class Options {
   virtual uint32_t maxPendingRequests() const PURE;
   virtual uint32_t maxActiveRequests() const PURE;
   virtual uint32_t maxRequestsPerConnection() const PURE;
+  virtual std::string sequencerIdleStrategy() const PURE;
```
mergeconflict: Same here, I think it'd be nicer to have an `enum class` to represent the legal values rather than a string.

source/common/sequencer_impl.cc
```diff
@@ -14,12 +14,12 @@ SequencerImpl::SequencerImpl(const PlatformUtil& platform_util,
                              Envoy::MonotonicTime start_time, RateLimiterPtr&& rate_limiter,
                              SequencerTarget target, StatisticPtr&& latency_statistic,
                              StatisticPtr&& blocked_statistic, std::chrono::microseconds duration,
-                             std::chrono::microseconds grace_timeout)
+                             std::chrono::microseconds grace_timeout, bool use_spin_loop)
```
mergeconflict: I think you'd want this to be the `enum class` type I suggested above, rather than a bool.

test/sequencer_test.cc
```diff
@@ -215,39 +218,45 @@ class SequencerIntegrationTest : public SequencerTestWithTimerEmulation {
...
-  EXPECT_EQ(0, sequencer.blockedStatistic().count());
+TEST_F(SequencerIntegrationTest, TheHappyFlow) { testRegularFlow(true); }
 
-  EXPECT_EQ(2, sequencer.statistics().size());
-}
+TEST_F(SequencerIntegrationTest, TheHappyFlowNoSpinning) { testRegularFlow(false); }
 
 // Test an always saturated sequencer target. A concrete example would be a http benchmark client
-// not being able to start any requests, for example due to misconfiguration or system conditions.
+// not being able to start a  ny requests, for example due to misconfiguration or system conditions.
```
mergeconflict: oops

api/client/options.proto
oschaaf: @mergeconflict 
So right now this is consistent with the other candidates that would be better represented as an enum.
I agree that an `enum` is the better option, so what I would propose here is leaving this as-is in this PR, and subsequently update all candidates in one go, in a single followup.

source/client/options_impl.cc
oschaaf: See https://github.com/envoyproxy/nighthawk/pull/94#issuecomment-508580132
This is annoying, but It would be great if someone can confirm this is a benign issue and I'm not squelching an actual problem that I failed to spot.

source/client/options_impl.cc
oschaaf: done

oschaaf:
@mergeconflict address comments, modulo the `enum` request. I would like to take on all subjects in one targeted follow up if you are ok with that.

api/client/options.proto
htuch: Yeah, I think a followup `enum` PR (before 0.2 release) makes sense.

source/client/factories_impl.cc
```diff
@@ -53,6 +53,8 @@ SequencerPtr SequencerFactoryImpl::create(Envoy::TimeSource& time_source,
   RateLimiterPtr rate_limiter =
       std::make_unique<LinearRateLimiter>(time_source, Frequency(options_.requestsPerSecond()));
   const uint64_t burst_size = options_.burstSize();
+  const bool useSpinLoop = options_.sequencerIdleStrategy() == "spin";
```
htuch: Nit: `use_spin_loop`

source/common/sequencer_impl.cc
```diff
@@ -135,7 +135,7 @@ void SequencerImpl::run(bool from_periodic_timer) {
     // Re-schedule the periodic timer if it was responsible for waking up this code.
     scheduleRun();
   } else {
-    if (targets_initiated_ == targets_completed_) {
+    if (use_spin_loop_ && (targets_initiated_ == targets_completed_)) {
```
htuch: Nit: superfluous parens.

htuch: Not sure, I follow the logic here; how do we `sleep()` if not spin looping?

htuch:
@oschaaf the clang-tidy error seems benign, not sure why we only see it now either.

api/client/options.proto
oschaaf: In progress over at https://github.com/oschaaf/nighthawk/tree/proto-enums

source/common/sequencer_impl.cc
oschaaf: Ugh, sorry you are right, this was polling instead of sleeping, I added sleep as well. So now we have three options:
- spin
- poll (at Envoy's minimum timer resolution of 1 ms)
- sleep

source/common/sequencer_impl.cc
```diff
@@ -144,7 +144,11 @@ void SequencerImpl::run(bool from_periodic_timer) {
       // TODO(oschaaf): Optionize performing this spin loop.
       platform_util_.yieldCurrentThread();
       spin_timer_->enableTimer(0ms);
-    }
+    } else if (idle_strategy_ == IdleStrategy::Sleep) {
+      // optionize sleep duration.
+      platform_util_.sleep(50us);
```
htuch: I wonder how different this is to spinning when the sleep duration is this low. It's fine to merge, but worth thinking about.

oschaaf: yes; I'm will write down some thoughts on this option (on why we have it, and when it makes sense to use it, and when one should not do so). And maybe the sleep time itself needs to be parameterized. But for, with the stock timing, it's an option to get us a higher precision compared to the 1 ms polling, while not maxing out cpu.

PR #97, opened 2019-07-08
Pending requests: measure time spend waiting on a full queue

oschaaf:
When allowing client-side queuing, this makes us report how much time we spend
being waiting on a full queue as blocking.
This allows us to get a sense of how much time we spend in closed loop mode
during a benchmark.

Addendum which fixes https://github.com/envoyproxy/nighthawk/pull/88


oschaaf:
This needs #88 to go in first.

source/client/benchmark_client_impl.cc
```diff
@@ -189,16 +189,18 @@ void BenchmarkClientHttpImpl::setRequestHeader(absl::string_view key, absl::stri
 }
 
 bool BenchmarkClientHttpImpl::tryStartOne(std::function<void()> caller_completion_callback) {
+  if (!cluster_->resourceManager(Envoy::Upstream::ResourcePriority::Default)
```
htuch: Not sure I see how this diff relates to the PR description..

oschaaf: Without this change, we would see the overflow stats rise excessively. With it we  will end up reporting the time we spend not being able to queue u a new request as time spend blocking (we indicate we could not by returning false). Hope that clarifies this?

htuch: I think so, but can you add comments and a test?
/wait

oschaaf: Done, and done. The test is implemented with python, wdyt?

integration/test_integration_basics.py
```diff
@@ -29,6 +29,31 @@ def test_h1(self):
...
+    self.assertGreater(counters["benchmark.http_2xx"], 1000)
+    self.assertEqual(counters["upstream_cx_http1_total"], 1)
+    global_histograms = self.getNighthawkGlobalHistogramsbyIdFromJson(parsed_json)
+    self.assertGreater(int(global_histograms["sequencer.blocking"]["count"]), 1000)
+    self.assertGreater(
+        int(global_histograms["benchmark_http_client.request_to_response"]["count"]), 1000)
+    return counters
+
+  def test_h1_mini_stress_test_with_client_side_queueing(self):
+    counters = self.mini_stress_test_h1([
```
htuch: For all the actual `test_*`, can you add Pydoc style comments describing what each one does? This is pretty standard style in Google Python style.

integration/test_integration_basics.py
```diff
@@ -29,6 +29,31 @@ def test_h1(self):
     self.assertEqual(counters["upstream_rq_total"], 25)
     self.assertEqual(len(counters), 9)
 
+  def mini_stress_test_h1(self, args):
+    # run a test with more rps then we can handle, and a very small client-side queue.
+    # we should observe both lots of successfull requests as well as time spend in blocking mode.,
+    parsed_json = self.runNighthawkClient(args)
+    counters = self.getNighthawkCounterMapFromJson(parsed_json)
+    self.assertGreater(counters["benchmark.http_2xx"], 1000)
```
htuch: This (and below) are coupled with the call site, maybe try and have one source of truth for the number of reqs.

oschaaf: Done

integration/test_integration_basics.py
oschaaf: Done

PR #98, opened 2019-07-10
Convert proto interface strings to enums

oschaaf:
Opening this PR which has the proposed approach for a single target.
If this looks good I'll push changes for the other candidates, which would follow the pattern here

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -55,7 +55,6 @@ message CommandLineOptions {
   // See :option:`--burst-size` for details.
   google.protobuf.UInt32Value burst_size = 10 [(validate.rules).uint32 = {lte: 1000000}];
 
-
   // See :option:`--address-family` for details.
   AddressFamily address_family = 11;
   oneof oneof_request_options {
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -14,6 +14,19 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+// We wrap all values so we can have a unified way of option handling with respect to
+// defaults, merging, etc. As there's no stock concept for enumerations, we manually
+// define custom wrappers for them. These used to be strings, which did provide the
+// wrapped type.
+message AddressFamily {
+  enum AddressFamilyOptions {
+    v4 = 0;
+    v6 = 1;
+    auto = 2;
+  }
+  AddressFamilyOptions value = 1;
+}
+
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
 message CommandLineOptions {
@@ -41,9 +54,10 @@ message CommandLineOptions {
   google.protobuf.BoolValue prefetch_connections = 9;
   // See :option:`--burst-size` for details.
   google.protobuf.UInt32Value burst_size = 10 [(validate.rules).uint32 = {lte: 1000000}];
+
+
   // See :option:`--address-family` for details.
-  google.protobuf.StringValue address_family = 11
-      [(validate.rules).string = {in: ["v4", "v6", "auto"]}];
+  AddressFamily address_family = 11;
   oneof oneof_request_options {
     // See :option:`--request_options` for details.
     RequestOptions request_options = 12;
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -27,6 +27,36 @@ message AddressFamily {
   AddressFamilyOptions value = 1;
 }
 
+message Verbosity {
+  enum VerbosityOptions {
+    INFO = 0;
+    TRACE = 1;
+    DEBUG = 2;
+    WARN = 3;
+    ERROR = 4;
+    CRITICAL = 5;
+  }
+  VerbosityOptions value = 1;
+}
+
+message OutputFormat {
+  enum OutputFormatOptions {
+    JSON = 0;
+    HUMAN = 1;
+    YAML = 2;
+  }
+  OutputFormatOptions value = 1;
+}
+
+message SequencerIdleStrategy {
+  enum SequencerIdleStrategyOptions {
+    SPIN = 0;
+    POLL = 1;
+    SLEEP = 2;
+  }
+  SequencerIdleStrategyOptions value = 1;
+}
+
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
 message CommandLineOptions {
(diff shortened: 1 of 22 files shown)


api/client/options.proto
```diff
@@ -14,6 +14,19 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+// We wrap all values so we can have a unified way of option handling with respect to
+// defaults, merging, etc. As there's no stock concept for enumerations, we manually
+// define custom wrappers for them. These used to be strings, which did provide the
+// wrapped type.
+message AddressFamily {
+  enum AddressFamilyOptions {
+    v4 = 0;
```
htuch: Please use all caps for enum constants, e.g. V4. Also, the default (with value 0) should be either UNSET, DEFAULT or something you know is always going to be a sensible default and never go away, e.g. AUTO in this case.

api/client/options.proto
```diff
@@ -14,6 +14,19 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+// We wrap all values so we can have a unified way of option handling with respect to
+// defaults, merging, etc. As there's no stock concept for enumerations, we manually
+// define custom wrappers for them. These used to be strings, which did provide the
+// wrapped type.
```
htuch: This is getting a bit clunky; can you remind me why we need to have this for all primitive types?

oschaaf: Without wrapping, we cannot make a distinction between ‘type-default’ and ‘not-set’. Which interferes with some things, like when we want to merge two proto’s and some would like to override something with the type default. Wrapping allows us to make that distinction. I must admit I have not tried if enum types behave exactly like this, I now realize I made an assumption that they would be treated similar to integers at the serialization level.

api/client/options.proto
oschaaf: Fixed

oschaaf:
@htuch converted all relevant proto fields to enums. We lost .1% coverage. I think the coverage loss can be attributed to having less actual code, so the ratio changed. I think could get the coverage percentage back up by working to get coverage for some other lines, unrelated to this PR, but that would pollute this change. So for now I reduced the coverage threshold.

oschaaf:
Staged https://github.com/envoyproxy/nighthawk/pull/99 as a follow up to this, which resolves some tech debt as well as bumps coverage to 97.3% by adding some tests

api/client/options.proto
htuch: Yeah, I think enums are effectively scalars. Hmm, Do we need full merge capabilities for client options here? I feel that there is some set that is CLI and fixed forever and others we want to dynamically update via header overrides etc.

api/client/options.proto
```diff
@@ -14,6 +14,49 @@ message RequestOptions {
...
+    AUTO = 0;
+    V4 = 1;
+    V6 = 2;
+  }
+  AddressFamilyOptions value = 1;
+}
+
+message Verbosity {
+  enum VerbosityOptions {
+    INFO = 0;
```
htuch: I think you want this and the others below to have an `UNSET` or `DEFAULT` value, since the currently assigned values are a policy decision.

source/client/options_impl.cc
```diff
@@ -166,18 +166,42 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
+        nighthawk::client::OutputFormat::OutputFormatOptions_Parse(upper_cased, &output_format_));
+  }
   TCLAP_SET_IF_SPECIFIED(prefetch_connections, prefetch_connections_);
   TCLAP_SET_IF_SPECIFIED(burst_size, burst_size_);
-  TCLAP_SET_IF_SPECIFIED(address_family, address_family_);
-  TCLAP_SET_IF_SPECIFIED(request_method, request_method_);
+  if (address_family.isSet()) {
+    std::string upper_cased = address_family.getValue();
+    absl::AsciiStrToUpper(&upper_cased);
+    ASSERT(nighthawk::client::AddressFamily::AddressFamilyOptions_Parse(upper_cased,
```
htuch: Can't use `ASSERT` here, as it gets removed in opt builds, use `RELEASE_ASSERT` (or preferably exceptions) instead.

api/client/options.proto
oschaaf: There is the test-server, which could read a base config from disk, but maybe that would also be nice for the grpc service (and even the CLI)? That would allow the end-user to override the hard-coded defaults, and subsequently tune via cli/grpc calls/headers. With uniform behavior across the tooling that we offer.

source/client/options_impl.cc
oschaaf: I will change these to `RELEASE_ASSERT`. With respect to exception, we shouldn't be able to get here at all, because the validation earlier should prevent that. Given that, do you still prefer the exception?

api/client/options.proto
oschaaf: Right, let me amend that

source/client/options_impl.cc
htuch: `RELEASE_ASSERT` is good.

api/client/options.proto
htuch: Trying to figure out which bits might never really change. E.g. could address family here change dynamically?

oschaaf: Perhaps; maybe far fetched, but a grpc server could be configured to auto hard-coded or on disk, and then overridden to v4 or v6 via calls to initiate benchmark runs.. Not sure if this is realistic, but I'm also not sure if we should prevent it?

source/client/options_impl.cc
oschaaf: Done

api/client/options.proto
oschaaf: Done

oschaaf:
Coverage failed on a timeout while fetching a dependency.
/retest

api/client/options.proto
oschaaf: After discussing this on Slack, I'm leaning towards keeping this as-is. I can see scenarios for any of the enum-fields where being able to override is going to be useful, and the way the proto is defined right now avoids having to change the api.

source/client/factories_impl.cc
```diff
@@ -94,15 +82,16 @@ OutputCollectorFactoryImpl::OutputCollectorFactoryImpl(Envoy::TimeSource& time_s
     : OptionBasedFactoryImpl(options), time_source_(time_source) {}
 
 OutputCollectorPtr OutputCollectorFactoryImpl::create() const {
-  const std::string format = options_.outputFormat();
-  if (format == "human") {
+  switch (options_.outputFormat()) {
+  case nighthawk::client::OutputFormat::HUMAN:
```
htuch: Do you handle the DEFAULT cases here and elsewhere?

oschaaf: `OptionImpl` will never return the `DEFAULT` enum default values here, hence the `NOT_REACHED_GCOVR_EXCL_LINE` below.  

Context: When we construct `OptionImpl`  from a `proto`, it uses `PROTOBUF_GET_WRAPPED_OR_DEFAULT` to either get the set value, or take the default value from the private field declared in the header (for POD types) when it is not set. [1] `PROTOBUF_GET_WRAPPED_OR_DEFAULT` considers the `DEFAULT` flag as not set as it doesn't get serialized.

[1] e.g. https://github.com/envoyproxy/nighthawk/pull/98/files/7d8ff35fcc49cb39e10e6acfefc5ecf2de5b48b5#diff-54af86e901ab4e3f97d4dcf19e326cc3R273

htuch: Got it, makes sense.

PR #100, opened 2019-07-16
Release notes for v0.2 + doc update

oschaaf:


README.md
```diff
@@ -25,13 +25,46 @@ For building the Nighthawk test server, see [here](source/server/README.md).
...
 ```bash
-bazel test -c dbg //test:nighthawk_test
+bazel test -c dbg //...
 ```
 
 ### Build it
 
 ```bash
-bazel build -c opt //:nighthawk_client
+bazel build -c opt //...
```
htuch: I think providing a concrete build target is a good idea.

README.md
```diff
@@ -25,13 +25,46 @@ For building the Nighthawk test server, see [here](source/server/README.md).
 ### Test it
 
 ```bash
-bazel test -c dbg //test:nighthawk_test
+bazel test -c dbg //...
```
htuch: `//test/...`?

README.md
```diff
@@ -25,13 +25,46 @@ For building the Nighthawk test server, see [here](source/server/README.md).
...
 
 ### Build it
 
 ```bash
-bazel build -c opt //:nighthawk_client
+bazel build -c opt //...
+```
+### Using the service
+
+The grpc service can be used to start a server which is able to perform back-to-back benchmark runs upon request. The service interface definition [can be found here.](https://github.com/envoyproxy/nighthawk/blob/59a37568783272a6438b5697277d4e56aa16ebbe/api/client/service.proto)
```
htuch: Nit: s/grpc/gRPC/ globally.

README.md
```diff
@@ -25,13 +25,46 @@ For building the Nighthawk test server, see [here](source/server/README.md).
...
+bazel test -c dbg //...
 ```
 
 ### Build it
 
 ```bash
-bazel build -c opt //:nighthawk_client
+bazel build -c opt //...
+```
+### Using the service
```
htuch: I would move this to after the CLI section and call it `### Nighthawk gRPC service`

README.md
```diff
@@ -25,13 +25,13 @@ For building the Nighthawk test server, see [here](source/server/README.md).
 ### Test it
 
 ```bash
-bazel test -c dbg //test:nighthawk_test
+bazel test -c dbg //test/... //integration/...
```
htuch: Can we move integration underneath test? It seems cleaner to have all tests in a single tree?

README.md
```diff
@@ -25,13 +25,13 @@ For building the Nighthawk test server, see [here](source/server/README.md).
...
 ```
 
 ### Build it
 
 ```bash
-bazel build -c opt //:nighthawk_client
+bazel build -c opt //:nighthawk_client nighthawk_test_server nighthawk_service
```
htuch: Does this work? Don't the latest binaries need a path? Also, should we just have a filegroup or something for this?

README.md
oschaaf: That is where they lived at first, but it caused problems with the coverage test:
https://circleci.com/gh/envoyproxy/nighthawk/1193?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

README.md
oschaaf: Yes this works. I will look into creating a filegroup.

oschaaf:
@htuch this now incorporates a filegroup `nighthawk` to target building all the tools. Also updated `README.md` to reflect that.

README.md
htuch: Can you update to use the new LLVM-style coverage script? See https://github.com/envoyproxy/envoy/pull/7552. This might not fix it, but it makes coverage much faster and easier to debug :)

PR #101, opened 2019-07-26
Update our Envoy dependency

oschaaf:
- Updates Envoy to https://github.com/envoyproxy/envoy/commit/8246167b9d238797cbc6c03dccc9e3921c37617d
- Swap out Envoy's integration test server  to use our own python integration wrappers to fire up a backend in the cpp benchmark client tests. This was needed to resolve the singleton runtime conflict with Envoy's integration test server rearing its head again. As an a-side this sets us up for testing against multiple types of backends later on.
- Coverage refresher: syncs the way we do coverage testing up with how Envoy pulls it off, greatly improving the time needed to generate a coverage report as well as improving accuracy tracking coverage of inlined code in headers. 
Notes:
- Fixed argument handling (as with the new coverage test method the test around it would fail).


test/ssl_test.cc
```diff
@@ -0,0 +1,90 @@
+#include <chrono>
```
htuch: Thanks @oschaaf . How much of this can be split into separate PRs? I.e. there is a ton of churn, around Python test infra, SSL objects, coverage, general upgrade? Ideally our Envoy upgrades are lightweight and some of these changes, e.g. test infra, are one-offs. Could you split out some of this easily?

oschaaf: Yes, this can be split up into multiple PR's easily; be we may have to temporarily allow a lower coverage threshold at some point. I'll carve off a more fun-sized piece right away into a separate PR.

oschaaf:
Closing this PR as all has been either split out into separate PR's or been superseded.

PR #104, opened 2019-08-01
Integration test update

oschaaf:
Split off from https://github.com/envoyproxy/nighthawk/pull/101

- As a step towards upgrading our Envoy dependency,
this swaps out Envoy's c++ integration test server with our own
python wrappers around our own test server. This permanently
avoids future runtime singleton conflicts.
- Also starts setting `--base-id` when starting up our test-server.
This is to support the upcoming test-sharding that will be performed
by updated coverage test methodoly which will land in a future update.


integration/BUILD
```diff
@@ -31,3 +31,24 @@ py_test(
         requirement("idna"),
     ],
 )
+
+py_binary(
+    name = "cpp_benchmark_client_server",
+    srcs = [
+        "cpp_benchmark_client_server.py",
+    ],
+    deps = [
```
htuch: Tiny nit: generally prefer singleton lists to be same line, e.g. `deps = ["foo"],`, if they fit.

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+import logging
+import os
+import sys
+import unittest
+
+from common import IpVersion, NighthawkException
+from integration_test_fixtures import (HttpIntegrationTestBase, HttpsIntegrationTestBase,
+                                       IntegrationTestBase)
+
+assert sys.version_info >= (3, 0)
```
htuch: Is this really needed for Python3 code?

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+
+httpbase = None
```
htuch: Can you avoid globals and just return the base, or make this a class structure?

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def main():
+  if len(sys.argv) != 3:
+    print("cpp_benchmark_client_server.py [ipv4|ipv6] [http|https]")
+    return -1
+  port = serverStartHook(
+      IpVersion.IPV6 if str.lower(sys.argv[1]) == "ipv6" else IpVersion.IPV4,
+      str.lower(sys.argv[2]) == "https")
+  print(str(port))
```
htuch: Nit: prefer `sys.stdout.write` to be consistent with the flush.

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+  print(str(getRunningServerPid()))
```
htuch: Can you make these prints more descriptive, rather than just integers on each line?

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+  return httpbase.getServerPid()
+
+
+def waitForExit():
+  return httpbase.waitForServerExit()
+
+
+def main():
+  if len(sys.argv) != 3:
+    print("cpp_benchmark_client_server.py [ipv4|ipv6] [http|https]")
```
htuch: Nit: prefer `sys.stderr.write` here. Also, use `%0`, e.g.
```
sys.stderr.write('%s [ipv4|ipv6] [http|https]" % sys.argv[0])
```

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+    return -1
```
htuch: sys.exit(1)

integration/cpp_benchmark_client_server.py
```diff
@@ -0,0 +1,58 @@
...
+  port = serverStartHook(
+      IpVersion.IPV6 if str.lower(sys.argv[1]) == "ipv6" else IpVersion.IPV4,
```
htuch: If you want to be more robust, suggest using `argparse`

test/test_data/benchmark_https_client_test_envoy.yaml
```diff
@@ -0,0 +1,50 @@
+admin:
```
htuch: Is this so different that it warrants a separate YAML? Also, have you considered using https://github.com/envoyproxy/envoy/blob/master/test/config/utility.h to generate these?

test/benchmark_http_client_test.cc
```diff
@@ -37,37 +38,81 @@ using namespace testing;
 
 namespace Nighthawk {
 
-class BenchmarkClientTestBase : public Envoy::BaseIntegrationTest,
-                                public TestWithParam<Envoy::Network::Address::IpVersion> {
+class PythonIntegrationTestBase : public TestWithParam<Envoy::Network::Address::IpVersion> {
```
htuch: Did you consider just running client, Envoy and test server all as independent processes under a  top-level Python script?

test/benchmark_http_client_test.cc
```diff
@@ -37,37 +38,81 @@ using namespace testing;
...
+    }
+    args.append(" http");
+    if (use_https) {
+      args.append("s");
+    }
+    pipe_ = popen(Envoy::TestEnvironment::runfilesPath("integration/cpp_benchmark_client_server")
+                      .append(args)
+                      .c_str(),
+                  "r");
+    RELEASE_ASSERT(pipe_ != nullptr, "Failed to open pipe");
```
htuch: So, we have C++ calling Python calling C++ here? How come we can't just launch the test server directly as a forked process?

test/benchmark_http_client_test.cc
```diff
@@ -37,37 +38,81 @@ using namespace testing;
...
+                   "couldn't understand server_pid");
+  }
+
+  void TearDown() override {
+    if (server_pid_ != 0) {
+      kill(server_pid_, SIGTERM);
+    }
+    if (pipe_ != nullptr) {
+      // We don't expect any output, lets print it when that happens
+      while (fgets(buffer_.data(), buffer_.max_size(), pipe_) != nullptr) {
```
htuch: Generally bad form to use `fgets`, can you use C++ streams or something a bit safer?

integration/cpp_benchmark_client_server.py
htuch: I see, these are for consumption by the parent process, please comment.

oschaaf:
Closing this one in favor of switching the benchmark client over to the mock-based testing and python integration testing offered over at https://github.com/envoyproxy/nighthawk/pull/105

PR #105, opened 2019-08-01
Update our Envoy dependency

oschaaf:
Updates Envoy to bdd6788f1e01787d015eabd9902f4b565e5dea98

- Has the required changes to make that work for NH
- Unexpectedly passes tests on my dev machine (the runtime conflict I ran into earlier seems to
have been resolved, possibly via https://github.com/envoyproxy/envoy/commit/9ea7ea21ac8212726f452810f1e8ad855a83c022))
- Introduces `FakeAdmin` / `FakeClusterManager`
- Adds some test coverage


source/common/ssl.cc
```diff
@@ -0,0 +1,159 @@
+#include "common/ssl.h"
+
+namespace Nighthawk {
```
htuch: This seems like it's getting really complicated. Do you think there is an upstream Envoy change we can make to simplify?

oschaaf:
@htuch re: " Do you think there is anything we can do structurally to reduce the likelihood that Nighthawk will break during import?"

Yeah I feel you. There's been a change in Envoy about a week ago that makes `ClusterInfoImpl`'s constructor pull things like `Envoy::Server::Admin` from our `MinimalTransportSocketFactoryContext`, which is what forced the `FakeXXX` implementations :(.

So, for NH to be able to create its connection pools, it must summon up full `Envoy::Upstream::ClusterInfoImpl` and `Envoy::Upstream::HostImpl`.

Auditing usage of those constructs in the actual pool implementations, it seems usage of the above concepts in there come down to:
- bumping (cluster) stats
- pulling options
- pulling resource limit numbers
- `HostImpl::createConnection` calls `cluster.createNetworkFilterChain`

So, intuitively, it seems like it should be possible to work towards a situation where it is possible to construct pools with statistics arguments, some options, a resourceManager and perhaps an interface / delegate for connection creation. Maybe all of these things could be combined into a separate interface which gets passed as a single argument to the constructor of the pools. If it would work with delegates, even less knowledge about the outside world would be needed from the pool's PoV.

source/common/ssl.cc
oschaaf: Re: "This seems like it's getting really complicated. Do you think there is an upstream Envoy change we can make to simplify?"

These fakes are gone now. I'll ping you on Slack to discuss changes made here in the commits I just pushed

oschaaf:
@htuch I just pushed a bunch of changes to this PR. Let me summarize PR contents as-is right now:

- Updates Envoy dependency + some changes to make that work for NH
- Switches to base us off of cluster manager. This allows us to deprecate code, and I also anticipate this helps with reducing future code churn while we track Envoy updates.
- Changes to properly initialize thread-local storage. 
- Switches us to use thread-local statistics instead of the isolated stores we had earlier. This makes sampling current aggregated stats values from the main thread during tests trivial, as well as paves the way for OpenTracing integration.
- Switches the benchmark client tests to mock-based testing for coverage, so we no longer need Envoy's integration test server.
- Adds functional python integration tests to make up for the lost c++ equivalents.
- Cleans up headers & deps. 
- Updates the coverage threshold to 98%

Important notes:
- The connection prefetching feature is disfuctional with this change, and should be amended in a follow-up. I think this won't be hard, but I don't want to bloat this PR.
- As the thread local stats perform automatic merging, we loose the per-worker cluster statistics. We still have per-worker hdr-histograms (because we track those ourselves) which is helpful, but I feel it would be nice to recover this information in the future, as its useful for observing inter-worker imbalances.

Changes are mostly straightforward, except for the cpp benchmarking which went through a big change.

oschaaf:
/cc @tonya11en This PR proposes changes which include a switch to thread-local statistics.
When we have that, sampling latency values from the main thread is going to be a fairly straightforward exercise.

PR #106, opened 2019-08-09
Update bazel coverage test

oschaaf:
- Update the CI build image
- Update bazel coverage test
- Tweak in the python orchestration to allow
  for multiple test server instances
- Tweak in rate limiter testing of invalid input values,  which would
  otherwise misbehave in the new coverage test as-is.
- Switch build/test ci tasks to use gcc so we keep testing with it.

Greatly speeds up coverage tests as well as  produces more accurate
coverage percentages.

Needs #105 to land first


oschaaf:
Just noticed in a local coverage run that this warning gets emitted:

```
warning: 15 functions have mismatched data
```

@htuch ^^ do you know if that is something we can safely ignore?

htuch:
Not sure, @lizan might be more familiar with the current LLVM based sanitizer warnings.

lizan:
should be OK, I didn't really dig it deeper than comparing coverage reports, though they seems to be from some dependencies.

oschaaf:
Thanks @lizan, this is ready for review then
/assign htuch

ci/do_ci.sh
```diff
@@ -5,11 +5,13 @@ set -e
 export BUILDIFIER_BIN="/usr/local/bin/buildifier"
 
 function do_build () {
+    setup_gcc_toolchain
     bazel build $BAZEL_BUILD_OPTIONS --verbose_failures=true //:nighthawk_client //:nighthawk_test_server \
-        //:nighthawk_service
+    //:nighthawk_service
```
htuch: Nit: I think the indent was clearer before

ci/do_ci.sh
```diff
@@ -5,11 +5,13 @@ set -e
...
 }
 
 function do_test() {
+    setup_gcc_toolchain
```
htuch: FWIW, if we move to only clang we will be in a good place when it comes to both coverage and normal test/build; Envoy basically only uses gcc tests as a regression now, life is faster/better with Clang.

ci/do_ci.sh
```diff
@@ -96,14 +107,11 @@ if [ -n "$CIRCLECI" ]; then
         mv "${HOME:-/root}/.gitconfig" "${HOME:-/root}/.gitconfig_save"
         echo 1
     fi
-
-    NUM_CPUS=8
-    if [ "$1" == "coverage" ]; then
-        NUM_CPUS=6
-    fi
+    # We constrain parallelism in CI to avoid running out of memory.
+    NUM_CPUS=6
```
htuch: We shouldn't need to do this if we switch to Clang I think.

test/run_nighthawk_bazel_coverage.sh
```diff
@@ -0,0 +1,69 @@
...
+  COVERAGE_THRESHOLD=98.3
+  COVERAGE_FAILED=$(echo "${COVERAGE_VALUE}<${COVERAGE_THRESHOLD}" | bc)
+  if test ${COVERAGE_FAILED} -eq 1; then
+      echo Code coverage ${COVERAGE_VALUE} is lower than limit of ${COVERAGE_THRESHOLD}
+      exit 1
+  else
+      echo Code coverage ${COVERAGE_VALUE} is good and higher than limit of ${COVERAGE_THRESHOLD}
+  fi
+fi
+echo "HTML coverage report is in ${COVERAGE_DIR}/index.html"
```
htuch: I'm going to assume this is the same as the Envoy repo variant unless you have any diff you want to call out.

.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 references:
   envoy-build-image: &envoy-build-image
-    envoyproxy/envoy-build:d0cefa7f071dbd4ef24399c2db8656c3a5d8c3ef
+    envoyproxy/envoy-build:8246167b9d238797cbc6c03dccc9e3921c37617d
```
htuch: @lizan do you think it would be easy/hard for us to move from CircleCI to Azure as well?

lizan: shouldn’t be hard especially if the project is not that large so you can do without RBE

ci/do_ci.sh
oschaaf: cool, done

ci/do_ci.sh
oschaaf: Fixed. It would be great to have check_format/fix for bash :)

ci/do_ci.sh
oschaaf: removed

test/run_nighthawk_bazel_coverage.sh
oschaaf: Yes, it's the same

oschaaf:
comments addressed /assign htuch

oschaaf:
@htuch I think we need to throttle parallelization even with clang, to avoid flakes in CI

lizan:
@oschaaf seems we don't have do_circle_ci.sh in this repo which set NUM_CPUS=8, so yeah you need it somewhere.

oschaaf:
Updated, we should now have `NUM_CPUS=8` set when we run in circleci

oschaaf:
Yup CI is happy now that we capped parallelism at 8. Thanks again, @lizan !

PR #107, opened 2019-08-10
Add back the connection prefetching feature

oschaaf:
As promised in https://github.com/envoyproxy/nighthawk/pull/105
This restores functionality of the connection prefetching feature

### DIFF
--- a/integration/test_integration_basics.py
+++ b/integration/test_integration_basics.py
@@ -207,3 +207,14 @@ class TestHttps(HttpsIntegrationTestBase):
     ])
     counters = self.getNighthawkCounterMapFromJson(parsed_json)
     self.assertEqual(counters["ssl.ciphers.ECDHE-RSA-AES256-GCM-SHA384"], 1)
+
+  def test_prefetching(self):
+    """
+    Test we prefetch connections.
+    """
+    parsed_json = self.runNighthawkClient([
+        "--duration 1", "--rps 1", "--prefetch-connections", "--connections 50",
+        self.getTestServerRootUri()
+    ])
+    counters = self.getNighthawkCounterMapFromJson(parsed_json)
+    self.assertEqual(counters["upstream_cx_http1_total"], 50)
--- a/integration/test_integration_basics.py
+++ b/integration/test_integration_basics.py
@@ -210,7 +210,9 @@ class TestHttps(HttpsIntegrationTestBase):
 
   def test_prefetching(self):
     """
-    Test we prefetch connections.
+    Test we prefetch connections. We test for 1 second at 1 rps, which should
+    result in 1 connection max without prefetching. However, we specify 50 connections
+    and the prefetching flag, so we ought to see 50 http1 connections created.
     """
     parsed_json = self.runNighthawkClient([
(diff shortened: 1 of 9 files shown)


source/client/process_impl.cc
```diff
@@ -46,6 +47,25 @@ using namespace std::chrono_literals;
...
+// http1 pool to the benchmark client, which allows us to offer connection prefetching.
+class ClusterManagerFactory : public Envoy::Upstream::ProdClusterManagerFactory {
+public:
+  using Envoy::Upstream::ProdClusterManagerFactory::ProdClusterManagerFactory;
+
+  Envoy::Http::ConnectionPool::InstancePtr
+  allocateConnPool(Envoy::Event::Dispatcher& dispatcher, Envoy::Upstream::HostConstSharedPtr host,
+                   Envoy::Upstream::ResourcePriority priority, Envoy::Http::Protocol protocol,
+                   const Envoy::Network::ConnectionSocket::OptionsSharedPtr& options) override {
+    if (protocol == Envoy::Http::Protocol::Http11 || protocol == Envoy::Http::Protocol::Http11) {
```
oschaaf: Note to self: one of these ought to test for http1.0

integration/test_integration_basics.py
```diff
@@ -207,3 +207,14 @@ def test_h2_tls_context_configuration(self):
...
+  def test_prefetching(self):
+    """
+    Test we prefetch connections.
+    """
+    parsed_json = self.runNighthawkClient([
+        "--duration 1", "--rps 1", "--prefetch-connections", "--connections 50",
+        self.getTestServerRootUri()
+    ])
+    counters = self.getNighthawkCounterMapFromJson(parsed_json)
+    self.assertEqual(counters["upstream_cx_http1_total"], 50)
```
htuch: How does this verify that connections were prefetched?

source/client/process_impl.cc
htuch: Yes :)

source/client/benchmark_client_impl.cc
```diff
@@ -19,6 +19,13 @@ using namespace std::chrono_literals;
 namespace Nighthawk {
 namespace Client {
 
+void Http1PoolImpl::createConnections(const uint32_t connection_limit) {
+  ENVOY_LOG(error, "Prefetching {} connections.", connection_limit);
+  for (uint32_t i = 0; i < connection_limit; i++) {
+    createNewConnection();
```
htuch: If I understand the code correctly, Envoy is already supporting multiple connections in each host's pool for HTTP/1.1, and we're just making sure it's fully populated, vs. HTTP/2 where there is only one?

integration/test_integration_basics.py
oschaaf: Added a comment to clarify; without prefetching we'd expect a single connection because we
test at 1 rps for 1 second.

source/client/benchmark_client_impl.cc
oschaaf: Correct. Once H/2 supports multiple connections I assume it's going to be easy to extend this to support it with our pool prefetching / prepopulation feature.

oschaaf:
comments addressed
/assign htuch

PR #112, opened 2019-08-14
Update HdrHistogram_c 0.9.8 -> 0.9.11

oschaaf:


oschaaf:
The coverage failure that was reported in CI is addressed in https://github.com/envoyproxy/nighthawk/pull/114

PR #114, opened 2019-08-14
Fix the test we have for respecting log verbosity

oschaaf:
The c++ test started failing when landing on master, because
it conflicts with the new test methology which uses sharding.
(The stderr capturing we used breaks when running in sharded mode)

Move the test to a python-based integration-test equivalent.
Also, reduce log noise in the tests.

### DIFF
--- a/integration/test_integration_basics.py
+++ b/integration/test_integration_basics.py
@@ -21,7 +21,7 @@ class TestHttp(HttpIntegrationTestBase):
     Runs the CLI configured to use plain HTTP/1 against our test server, and sanity
     checks statistics from both client and server.
     """
-    parsed_json, logs = self.runNighthawkClient([self.getTestServerRootUri()])
+    parsed_json, _ = self.runNighthawkClient([self.getTestServerRootUri()])
     counters = self.getNighthawkCounterMapFromJson(parsed_json)
     self.assertEqual(counters["benchmark.http_2xx"], 25)
     self.assertEqual(counters["upstream_cx_destroy"], 1)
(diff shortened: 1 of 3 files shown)


integration/test_integration_basics.py
```diff
@@ -214,9 +214,27 @@ def test_prefetching(self):
...
         "--duration 1", "--rps 1", "--prefetch-connections", "--connections 50",
         self.getTestServerRootUri()
     ])
     counters = self.getNighthawkCounterMapFromJson(parsed_json)
     self.assertEqual(counters["upstream_cx_http1_total"], 50)
+
+  def test_log_verbosity(self):
+    """
+    Test that that the specified log verbosity level is respected.
+    This tests for a sentinel we know is only writter when the level
```
htuch: Nit: writter --> right?

integration/test_integration_basics.py
```diff
@@ -192,15 +192,15 @@ def test_h2_tls_context_configuration(self):
...
+    parsed_json, logs = self.runNighthawkClient([
         "--duration 1",
         "--tls-context {common_tls_context:{tls_params:{cipher_suites:[\"-ALL:ECDHE-RSA-AES128-SHA\"]}}}",
         self.getTestServerRootUri()
     ])
     counters = self.getNighthawkCounterMapFromJson(parsed_json)
     self.assertEqual(counters["ssl.ciphers.ECDHE-RSA-AES128-SHA"], 1)
 
-    parsed_json = self.runNighthawkClient([
+    parsed_json, logs = self.runNighthawkClient([
```
htuch: If logs aren't going to be used, you make this `parsed_json, _ = self.runNight...`

oschaaf:
Comments addressed.

Furthermore, after swapping the c++ test for `LogVerbosity` in favor of a python integration test, the coverage percentage dipped a fraction, which caused the coverage test failure to persists. Added a tweak to the existing c++ tests to restore coverage.

/assign htuch

PR #118, opened 2019-08-19
Service: address back to back execution race

oschaaf:
Address a race discovered via repeated parallel execution of
`ServiceTest, BackToBackExecution` by adding a short timeout
before concluding that there's still a previous future active and handling
the earlier test.

Theoretically it's still possible to run into this, but  in practice I think this
is fair enough for actual usage, and it is sufficient to deflake the
test which detected it on my machine.

Thanks @htuch for discovering the issue


source/client/service_impl.cc
```diff
@@ -69,8 +69,12 @@ ::grpc::Status ServiceImpl::ExecutionStream(
   while (stream->Read(&request)) {
     ENVOY_LOG(debug, "Read ExecutionRequest data: {}", request.DebugString());
     if (request.has_start_request()) {
+      // It is possible to receive a back-to-back request here, while the future that is associated
+      // to our previous response is still active. To resolve this in practice we allow waiting here
+      // for a short period in favor of adding a more complex atomic bool to track if we are ready
+      // for the next request.
       if (future_.valid() &&
-          future_.wait_for(std::chrono::seconds(0)) != std::future_status::ready) {
+          future_.wait_for(std::chrono::seconds(1)) != std::future_status::ready) {
```
htuch: Is there any way to do proper synchronization here that is race free? This seems an improvement, but our build/test machines sometimes encounter heavy load and we can flake with this.

oschaaf: Yes there is, I just wanted to test the waters for this approach because it's the simplest one. I'll update it to eliminate the race altogether.

oschaaf:
Updated to eliminate the race altogether. Ready for another look!

/assign htuch

source/client/service_impl.cc
```diff
@@ -66,13 +67,18 @@ ::grpc::Status ServiceImpl::ExecutionStream(
...
+  busy_ = false;
   while (stream->Read(&request)) {
     ENVOY_LOG(debug, "Read ExecutionRequest data: {}", request.DebugString());
     if (request.has_start_request()) {
-      if (future_.valid() &&
-          future_.wait_for(std::chrono::seconds(0)) != std::future_status::ready) {
+      // It is possible to receive a back-to-back request here, while the future that is associated
+      // to our previous response is still active. We check the running_ flag to see if the previous
+      // future has progressed in a state where we can do another one. This avoids the odd flake in
+      // ServiceTest.BackToBackExecution.
```
htuch: I'm trying to figure out how this works.. it seems that if we had a previous request with an active future, won't we be still waiting on it in the previous requst's `finishGrpcStream`?

oschaaf: Well, we're testing servicing of back-to-back execution on the same stream in this test, `finishGrpcStream` isn't involved in the failure. The failure happens when the client queues up the second request on the same stream right after receiving a response to the first one. The future that was responsible for servicing the reply can still be working to tear down. We only want to ensure that we have a single process context active at any time, so the proposed fix here is to indicate we are ready for the next request from within the code associated to the future (right before putting the reply on the stream, but after tearing down the process context).

htuch: Can we just do a zero timeout with https://en.cppreference.com/w/cpp/thread/future/wait_for to check future state?

Thinking about future semantics though, are we sure that the future even begins execution until we do a wait on it? An efficient runtime should do this, but I wonder if we're making some non-standard guaranteed assumption here.

oschaaf: Done, In the change I pushed we:
- wait with a zero timeout upon observing an active future. when the timeout fails we check the new busy flag.
- we start explicitly passing `std::launch::async`, to avoid running into lazy execution, good point

source/client/service_impl.cc
```diff
@@ -67,20 +67,22 @@ ::grpc::Status ServiceImpl::ExecutionStream(
...
   while (stream->Read(&request)) {
     ENVOY_LOG(debug, "Read ExecutionRequest data: {}", request.DebugString());
     if (request.has_start_request()) {
       // It is possible to receive a back-to-back request here, while the future that is associated
       // to our previous response is still active. We check the running_ flag to see if the previous
       // future has progressed in a state where we can do another one. This avoids the odd flake in
       // ServiceTest.BackToBackExecution.
-      if (busy_) {
+      if (future_.valid() &&
+          future_.wait_for(std::chrono::seconds(0)) != std::future_status::ready && busy_) {
```
htuch: Looks good, thanks. One last question; why do we still need `busy_`? Isn't knowing that the future isn't ready sufficient to know that it's still in-flight?

oschaaf: I think the thing is that the future may have send the reply and still not be ready when the next request comes in. So we may even receive the next request before the future has torn down.
To illustrate: the print I added in the following diff will be seen in the out when reproducing the failure in parallell/repeated tests (note that the busy flag is no longer checked):

```diff
diff --git a/source/client/service_impl.cc b/source/client/service_impl.cc
index fca8290..853cb77 100644
--- a/source/client/service_impl.cc
+++ b/source/client/service_impl.cc
@@ -75,7 +75,8 @@ void ServiceImpl::writeResponse(const nighthawk::client::ExecutionResponse& resp
       // future has progressed in a state where we can do another one. This avoids the odd flake in
       // ServiceTest.BackToBackExecution.
       if (future_.valid() &&
-          future_.wait_for(std::chrono::seconds(0)) != std::future_status::ready && busy_) {
+          future_.wait_for(std::chrono::seconds(0)) != std::future_status::ready) {
+        std::cerr << "expected to not be busy" << std::endl;
         return finishGrpcStream(false, "Only a single benchmark session is allowed at a time.");
       } else {
         busy_ = true;
```

htuch: Yes, that seems entirely possible. I think `busy_` is OK to unblock the import, but it's also super icky IMHO, since we're doing manual concurrency control via atomics. Ideally we would have something more like a giant select on the message loop, rather than a gRPC `Read`. Combining futures and gRPC APIs is kind of messy.Did we explore maybe using something like https://en.cppreference.com/w/cpp/thread/mutex/try_lock and a mutex?

oschaaf:
To tackle this in a clean way, I investigated the async api's for the grpc server, but unfortunately I wan't able to find any examples for that when it comes to bidirectional streaming. 

So instead, I updated the approach here to rely on `CondVar` & `BasicLockableMutex` (with a few comments to describe workings and gotcha's). 

Also added handling and a test for concurrent clients. This is ready for another look!

/assign htuch

source/client/service_impl.cc
```diff
@@ -8,46 +8,58 @@
...
+Envoy::Thread::MutexBasicLockable ServiceImpl::global_lock_;
+
 void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionRequest& request) {
+  // Lock accepted_lock, in case we get here before accepted_event_.wait() is entered.
+  auto accepted_lock = std::make_unique<Envoy::Thread::LockGuard>(accepted_lock_);
+  // Acquire busy_lock_, and signal that we did so, allowing the service to continue
+  // processing inbound requests on the stream.
+  auto busy_lock = std::make_unique<Envoy::Thread::LockGuard>(busy_lock_);
+  accepted_event_.notifyOne();
+  accepted_lock.reset();
```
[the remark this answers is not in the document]

source/client/service_impl.cc
```diff
@@ -64,16 +76,33 @@ ::grpc::Status ServiceImpl::ExecutionStream(
     ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
                                ::nighthawk::client::ExecutionRequest>* stream) {
   nighthawk::client::ExecutionRequest request;
+  // We are able to service a single client at a time, because we can only run a single
```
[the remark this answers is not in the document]

source/client/service_impl.h
```diff
@@ -31,14 +31,24 @@ class ServiceImpl final : public nighthawk::client::NighthawkService::Service,
...
   ::grpc::Status finishGrpcStream(const bool success, absl::string_view description = "");
 
   Envoy::Event::RealTimeSystem time_system_; // NO_CHECK_FORMAT(real_time)
   Envoy::Thread::MutexBasicLockable log_lock_;
   ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
-                             ::nighthawk::client::ExecutionRequest>* stream_;
+                             ::nighthawk::client::ExecutionRequest>*
+      stream_; // guarded by global_lock_
   std::future<void> future_;
+  static Envoy::Thread::MutexBasicLockable global_lock_;
```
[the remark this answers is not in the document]
oschaaf: It's not longer around, but for context,  as I understand it we may see multiple instances of this class when multiple clients are involved.

oschaaf:
@htuch re: complexity
On idea: maybe someday we could use Envoy's grpc client implementation and run it on NH's main dispatcher to connect back to a new internal grpc service method. 
That client would then be running on our own event loop, and the grpc service would lean up and just be passing messages back and forth.

oschaaf:
Oh, also, this is ready for another look :)

source/client/service_impl.h
```diff
@@ -31,14 +31,22 @@ class ServiceImpl final : public nighthawk::client::NighthawkService::Service,
...
   void handleExecutionRequest(const nighthawk::client::ExecutionRequest& request);
-  void writeResponseAndFinish(const nighthawk::client::ExecutionResponse& response);
+  void writeResponse(const nighthawk::client::ExecutionResponse& response);
   ::grpc::Status finishGrpcStream(const bool success, absl::string_view description = "");
 
   Envoy::Event::RealTimeSystem time_system_; // NO_CHECK_FORMAT(real_time)
   Envoy::Thread::MutexBasicLockable log_lock_;
-  ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
-                             ::nighthawk::client::ExecutionRequest>* stream_;
+  static ::grpc::ServerReaderWriter<::nighthawk::client::ExecutionResponse,
```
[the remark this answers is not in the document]
oschaaf: It's static so we can assert when multiple service instances are created because of multiple clients connecting; we assert when we do so.. any suggestions?

PR #123, opened 2019-08-20
Fix a couple typos.

jplevyak:
Signed-off-by: John Plevyak <jplevyak@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ bazel build -c opt //:nighthawk
 
 ```bash
 ➜ bazel-bin/nighthawk_client --help
+```
 
+```
 USAGE:
 
    bazel-bin/nighthawk_client  [--sequencer-idle-strategy <spin|poll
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -162,7 +164,9 @@ The gRPC service can be used to start a server which is able to perform back-to-
 
 ```bash
 ➜ bazel-bin/nighthawk_service --help
```
jplevyak: I noticed that the -> is used here but not above in the "Test it" and "Build it" sections.   Is that intentional?

jplevyak: @oschaaf

README.md
```diff
@@ -38,7 +38,9 @@ bazel build -c opt //:nighthawk
 
 ```bash
 ➜ bazel-bin/nighthawk_client --help
+```
 
+```
```
jplevyak: Note: when this section was ```bash it resulted in the bash keywords being colorized which seemed odd.  @oschaaf

README.md
oschaaf: No the inconsistency is not intentional. I have been manually updating these sections, but actually I consider the source of truth for this to live over at  https://github.com/envoyproxy/nighthawk/blob/master/source/client/options_impl.cc
I wonder if it's possible to somehow automate updating these docs from that.

jplevyak:
I can't merge, so I'll have to let you do the honors.

Issue #119, opened 2019-08-19
Google import 


htuch:
There are a number of things we need to ensure we can cleanly import Nighthawk into Google:
* [x] service_main.cc needs to be able to handle IPv6 addresses. Its test, `service_main_test.cc`, should work with `TestEnvironment::getIpVersionsForTest`.
* [x] service_test.cc and process_test.cc also need to consume from `TestEnvironment::getIpVersionsForTest` when forming loopback addresses.
* [x] We need to switch Envoy import paths from overlay to `external/envoy` based and structurally enforce this.
* [x] We need to eliminate the race being fixed in https://github.com/envoyproxy/nighthawk/pull/118

oschaaf:
I have been looking getting the include path overlay fixed.
There now is a candidate script to test for this: https://github.com/envoyproxy/nighthawk/pull/122, but unfortunately updating the include paths as planned causes the build to fail.

oschaaf:
status update: 
- the fix for the observed race in the service is in review, pending discussion.
- maybe we should brainstorm on slack to see if we can come up with a plan B for disambiguating the Envoy include paths; the script to check for what we do not want to see is implemented, but actually updating the include paths doesn't seem to work for me (see the comment above).
- I'll look into getting the ipv6 handling sorted tomorrow.

PR #126, opened 2019-08-23
Improve service ipv6 arg handling & parameterize tests

oschaaf:
Addresses some of the things needed for https://github.com/envoyproxy/nighthawk/issues/119

### DIFF
--- a/source/client/service_main.cc
+++ b/source/client/service_main.cc
@@ -18,39 +18,49 @@ ServiceMain::ServiceMain(int argc, const char** argv) {
   TCLAP::ValueArg<std::string> internet_address_and_port(
       "", "listen",
       "The address:port on which the Nighthawk grpc service should listen. Default: "
-      "0.0.0.0:8443.",
+      "0.0.0.0:8443",
       false, "0.0.0.0:8443", "address:port", cmd);
-
   Utility::parseCommand(cmd, argc, argv);
 
   try {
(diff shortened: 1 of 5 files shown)


source/client/service_main.cc
```diff
@@ -20,37 +20,44 @@ ServiceMain::ServiceMain(int argc, const char** argv) {
...
-  } catch (Envoy::EnvoyException e) {
-    throw NighthawkException(e.what());
+  parseIpAndMaybePort(internet_address_and_port.getValue());
+  ENVOY_LOG(info, "Nighthawk grpc service listener binding to: {}", getListenerAddress());
+  builder_.AddListeningPort(getListenerAddress(), grpc::InsecureServerCredentials(), &port_);
+  builder_.RegisterService(&service_);
+}
+
+// TODO(oschaaf): move to utility class.
+void ServiceMain::parseIpAndMaybePort(absl::string_view ip_and_maybe_port) {
```
htuch: I would much prefer to have these utilities in Envoy base (and ideally built on existing things in `Envoy::Network::Utility` and `Address::Instance`. It's pretty fidgetty doing all of this, and ideally we do it once in the Envoy world.

oschaaf: I updated this to restrict the parsing to a minimum

oschaaf:
This is ready for another look
/assign htuch

PR #129, opened 2019-08-27
Address the last few ambiguous Envoy includes

oschaaf:
This adresses the last ambiguoys includes paths, plus has some cleanup.

Needed for https://github.com/envoyproxy/nighthawk/issues/119

This needs https://github.com/envoyproxy/envoy/pull/8045


oschaaf:
Updated our Envoy dependency, with that CI is passing

PR #110, opened 2019-08-11
Zipkin tracing

oschaaf:
Opening an early PR for those who would like to have a look-see at progress around adding OpenTracing capabilities in Nighthawk. The current status is that this carves a crude but working path towards end-to-end tracing. 

Tasks left:
- Inventorize what information is missing in the NH part of the trace and add it
- Generalize set up of the tracing in NH so we can have other implementations other then zipkin

To keep the PR small, maybe it's a good idea to land a minimal version first.

Example record in the zipkin UI:

image

ps - At the moment the diff for this PR is rather large, but that will become a lot smaller once we land the outstanding PR's and merge master after that.

/cc @htuch

Issue #28, opened 2019-04-23
Integrate with a tracing API (Like OpenTrace, Lightstep, or Zipkin)


oschaaf:

oschaaf:
Some early groundwork for this is in review, and some exploration of a possible implementation is in progress over at https://github.com/envoyproxy/nighthawk/compare/master...oschaaf:tracing, which is able to send a foo zipkin spans to Envoy set up for tracing. That shows up as a combined records in zipkin's UI.

It looks like we will be able to re-use the full OpenTracing capabilities that come with Envoy, so next to zipkin, I anticipate we can tackle the other implementations in one go with relatively little extra work.

oschaaf:
Basic Zipkin support was landed via https://github.com/envoyproxy/nighthawk/pull/110

PR #130, opened 2019-08-29
Tracing prerequisites: enhanced shutdown handling

oschaaf:
Carves off a non-trivial piece from
https://github.com/envoyproxy/nighthawk/pull/110
This should help reduce noise in review when we get to landing the actual
tracing related stuff.

Most of the changes in here boil down to that when tracing is enabled, we need to be more careful in our shutdown procedures.

Part of https://github.com/envoyproxy/nighthawk/issues/28


source/common/worker_impl.cc
```diff
@@ -8,27 +8,57 @@ namespace Nighthawk {
...
   tls.registerThread(*dispatcher_, false);
 }
 
-WorkerImpl::~WorkerImpl() { tls_.shutdownThread(); }
+WorkerImpl::~WorkerImpl() { shutDown(); }
+
+void WorkerImpl::notifyExit() {
+  // code analysis isn't able to understand what we are doing here.
+  // suppress the errors from that locally, so we can continue to use
+#ifdef __clang__
```
oschaaf: clang-tidy doesn't like this, and also seems to ignore the suppression here :-(

htuch:
@jplevyak could you take a first pass on this?

source/client/client_worker_impl.cc
```diff
@@ -22,8 +22,11 @@ void ClientWorkerImpl::simpleWarmup() {
   if (prefetch_connections_) {
     benchmark_client_->prefetchPoolConnections();
   }
-  benchmark_client_->tryStartOne([this] { dispatcher_->exit(); });
-  dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
+  if (benchmark_client_->tryStartOne([this] { dispatcher_->exit(); })) {
```
jplevyak: So I am new to this codebase so bare with me.  The name "tryStartOne" is a bit confusing.  The class is BenchmarkClient, but "One" here would seem to imply that you are starting a Client, but given that the function is not static or a factory that doesn't make sense.  How about tryStartOneRequest ?

The class BenchmarkClientHttpImpl has a method onComplete() which is part of StreamDecoderCompletionCallback.   AFAICT this is a per-request callback.   Therefore this class seems to be mixing "Client" which handles multiple requests and "Request" methods.  This only make sense if this client can only have a single outstanding request, but then the method would be tryStart rather than tryStartOneRequest and there would have to be many clients.

I would expect that there would be an object to manage the outstanding request and that object would have the onComplete handler which would then call back to the BenchmarkClientHttpImpl handler with a pointer to itself so that the BenchmarkClientHttpImpl could determine which request was failing/succeeding.  Possibly Nighthawk doesn't care specifically which request completes?

source/client/process_impl.cc
```diff
@@ -258,6 +254,7 @@ bool ProcessImpl::run(OutputCollector& collector) {
   try {
     uri.resolve(*dispatcher_, Utility::translateFamilyOptionString(options_.addressFamily()));
   } catch (UriException) {
+    tls_.shutdownGlobalThreading();
```
jplevyak: Why is this handled here instead of at the higher level?  Perhaps you want to add some termination function instead and call it here and below.  A function which checks what is running and shuts it all down that can be called in different states.  That might simplify future error handling.

source/client/stream_decoder.cc
```diff
@@ -37,9 +37,8 @@ void StreamDecoder::onComplete(bool success) {
   }
   ASSERT(!success || complete_);
   decoder_completion_callback_.onComplete(success, *response_headers_);
-  if (success) {
-    caller_completion_callback_();
-  }
+  // XXX(oschaaf): forward success state
```
jplevyak: It this hard to do now?

source/common/worker_impl.cc
```diff
@@ -8,27 +8,57 @@ namespace Nighthawk {
...
+    RELEASE_ASSERT(Envoy::Runtime::LoaderSingleton::getExisting() != nullptr,
+                   "Couldn't get runtime");
+    {
+      Envoy::Thread::LockGuard completion_lock(completion_lock_);
+      wait_event.notifyOne();
+      // Run the dispatcher to let the callbacks posted by registerThread() execute.
+      dispatcher_->run(Envoy::Event::Dispatcher::RunType::NonBlock);
+      work();
+    }
+    // This will block untill shutdown is called.
```
jplevyak: s/untill/until

source/common/worker_impl.cc
```diff
@@ -8,27 +8,57 @@ namespace Nighthawk {
...
-  ASSERT(!started_ && !completed_);
+  RELEASE_ASSERT(!started_, "WorkerImpl::start() expected started_ to be false");
   started_ = true;
-  thread_ = thread_factory_.createThread([this]() {
-    ASSERT(Envoy::Runtime::LoaderSingleton::getExisting() != nullptr);
-    // Run the dispatcher to let the callbacks posted by registerThread() execute.
-    dispatcher_->run(Envoy::Event::Dispatcher::RunType::NonBlock);
-    work();
-  });
+  Envoy::Thread::LockGuard completion_lock(completion_lock_);
```
jplevyak: This is a bit complicated because you are mixing condition variables with futures.  Why not stick with futures.

std::promise<void> launched;
future_ = std::future<void>(std::async(std::launch::async, [this, &launched](
 ...
 launched.set_value();
  ....
}));
launched.get_future().wait().

Also, std::async semantics are somewhat complex in that the implementation has a lot of leeway to do things synchronously.   Perhaps just use a std::thread?

source/client/process_impl.cc
oschaaf: I moved the shutdown code to the destructor. 
(Note: Some of the calls we need to do are forgiving when called unpaired; so it's not being done conditionally right now).

source/client/stream_decoder.cc
oschaaf: Not hard, but it did bloat this PR for a bit.
Added in https://github.com/envoyproxy/nighthawk/pull/130/commits/bd4704f4252471f30cc37e3e9316de47b6a085b2
Let me know if you prefer that backed out of this PR; I can create a separate PR for that too.
(I didn't do it earlier, because I was trying to keep PR size down, so I drew an arbitrary line)

source/common/worker_impl.cc
oschaaf: Done

source/client/client_worker_impl.cc
jplevyak: Do the requests by necessity complete in order?  Does a Client manage a single connection or multiple connections?  HTTP/2 requests need not complete in-order so then the question remains how one can determine latency without knowing which request completed?

source/client/process_impl.cc
jplevyak: I have just been debugging code which did complex work in destructors.  Would it be possible to have a shutdown() method  instead?  Sorry, to cause you trouble about this but complex code (e.g. locking, threading) in constructors and destructors have caused me too many sleepless night :)

source/client/client_worker_impl.cc
oschaaf: Latency is captured at the stream level here:
https://github.com/envoyproxy/nighthawk/blob/master/source/client/stream_decoder.cc#L36

And at the sequencer level here (which contains queue/connect time): https://github.com/envoyproxy/nighthawk/blob/master/source/common/sequencer_impl.cc#L113

include/nighthawk/common/operation_callback.h
```diff
@@ -0,0 +1,6 @@
+#include <functional>
+
+namespace Nighthawk {
+
+using OperationCallback = std::function<void(bool, bool)>;
```
jplevyak: Could you document what these two bools are?  You can add descriptive variable names for example e.g.

 std::function<void(bool done, bool ok)>;

source/client/process_impl.cc
jplevyak: I have had numerous problems with code doing complex things like locking and threads in constructors and destructors... sorry to be a pain, but could you use a shutdown() function instead  and call it before the destructor?

source/client/stream_decoder.cc
jplevyak: I am find with doing it here.  Thanx!

source/client/process_impl.cc
jplevyak: Sorry about the dup, github was slow to recognize the comment :(

source/common/worker_impl.cc
jplevyak: I am a bit confused as to why this seems so complicated.  There must be something that I am not understanding about this code.  I would like to understand it, so I grabbed your branch and hacked a very simple version which just does a join in waitForCompletion:

https://github.com/jplevyak/nighthawk/commit/c56a645673d3a413abb2711961b4fc5fedb64429

It passes all the tests.

What am I not understanding?

oschaaf:
@jplevyak explicit shutdown added: https://github.com/envoyproxy/nighthawk/pull/130/commits/5f5e064e546852a1ddee76f1698cf4f612bb524c

source/common/worker_impl.cc
oschaaf: Later on we need the threads to be running after waitforcompletion() exits. (The tracing branch is a litle ahead of what we are reviewing here). If I remember correctly we need this to be able to properly shut down things owned on the main thread, but I need to revisit to refresh my memory and dig up specifics.

oschaaf: So I patched in  jplevyak@c56a645 in over at the `tracing` branch to see if problems still arise.
Testing with https we would run into an issue upon shutdown.. For example:

```bash
$ gdb --args bazel-bin/nighthawk_client --duration 1 https://127.0.0.
```

Yields the following ASSERT being triggered:

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c6c535 in __GI_abort () at abort.c:79
#2  0x0000000001b4f1a2 in Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl::~ContextManagerImpl (this=0x433e4b0)
    at external/envoy/source/extensions/transport_sockets/tls/context_manager_impl.cc:18
#3  0x0000000001b4f269 in Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl::~ContextManagerImpl (this=0x433e4b0)
    at external/envoy/source/extensions/transport_sockets/tls/context_manager_impl.cc:16
#4  0x000000000049c31f in std::default_delete<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl>::operator() (this=0x7fffffffd860, __ptr=0x433e4b0)
    at /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/unique_ptr.h:81
#5  0x000000000049ab40 in std::unique_ptr<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl, std::default_delete<Envoy::Extensions::TransportSockets::Tls::ContextManagerImpl> >::~unique_ptr (
    this=0x7fffffffd860) at /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/unique_ptr.h:274
#6  0x0000000000494ce1 in Nighthawk::Client::ProcessImpl::~ProcessImpl (this=0x7fffffffc148) at bazel-out/k8-dbg/bin/source/client/_virtual_includes/nighthawk_client_lib/client/process_impl.h:46
#7  0x0000000000491fcc in Nighthawk::Client::Main::run (this=0x429a1b0) at source/client/client.cc:60
#8  0x0000000000459809 in main (argc=6, argv=0x7fffffffdae8) at source/exe/client_main_entry.cc:26
```

Link to the `ASSERT` being triggered:
https://github.com/envoyproxy/envoy/blob/master/source/extensions/transport_sockets/tls/context_manager_impl.cc#L18

When initiating this change earlier, I think we'd actually triggered a different issue (with an earlier Envoy version). 
One upside to the approach with keeping the workers running when `WorkerImpl::waitForCompletion()` exits is that we can accurately mimic Envoy's shutdown procedure, which does some things upon shutdown of its main thread while its workers are still up and running too:
```
  tls_.shutdownGlobalThreading();
  store_root_.shutdownThreading();
```

Mimicking Envoy's shutdown procedure with respect to the above helps us avoid running into issues that will only affect NH and not Envoy, at the expense of some additional complexity in our shutdown handling. 
Given that, what do you think?

include/nighthawk/common/operation_callback.h
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/130/commits/8abb45eebb97f58d202b0963b3b683ce8161c076

source/client/client_worker_impl.cc
jplevyak: Got it.  Thanx for the pointers!

source/common/worker_impl.cc
jplevyak: That makes sense to me.  I tried to move the shutdown order around in that patch and ran into issues. I believe Envoy is simplifying the shutdown procedure by not having the workers take into account any running async operations and instead shutting down those operations by shutting down threading.   So keeping the worker alive make sense.  It isn't clear to me why the std::thread needs to be alive however given that it isn't going to do anything but join.  What resource is it holding which is visible?  I'd like to understand that bit.  Could you clarify it for me?

oschaaf: Good question; I'd have to dig in some more for a bit, but on a hunch I suspect that the thread local memory being destroyed when the worker-thread exits is still needed by the shutdown operations on the main thread?

jplevyak: Ah, yes the std::thread is being used by a Dispatcher which is managing thread local data.

Oh a different note, what envoy.yaml file are you using?  Is there an example setup to benchmark envoy?  What did you use to get the ASSERT above?

oschaaf: I teased it out manually this time to catch in with gdb, but I know that at least //test/integration:* on the tracer branch would catch these issues (when backing out the updated shutdown handling so they show up).
These tests wrap the binaries of NH’s client and test server (which actually is a  Envoy with the fault filter preconfigured), the test server has a configuration sub directory.

The relevant NH client config computation can be inspected over at https://github.com/oschaaf/nighthawk/blob/tracing/source/client/process_impl.cc#L311

jplevyak: I am thinking that https://github.com/jplevyak/nighthawk/commit/041ecea9b641cf6e39a2a6a9cf7b2236948b6dbd would work and if not I would like to understand why not.

Also, is there some trick to running the integration_test?  Perhaps it is a IPv6 thing with my dev box.

oschaaf: I will try that and let you know. Regarding integration_test: sorry, you need to set an environment variable, I run as follows with vscode [1]

bazel test --test_env=ENVOY_IP_TEST_VERSIONS=all --cache_test_results=no  -c dbg //test/integration:integration_test

[1] https://github.com/oschaaf/envoy-perf-vscode/blob/master/.vscode/tasks.json#L288

oschaaf: So, I applied https://github.com/jplevyak/nighthawk/commit/041ecea9b641cf6e39a2a6a9cf7b2236948b6dbd to https://github.com/oschaaf/nighthawk/tree/tracing. I did run into issues, see https://gist.github.com/oschaaf/c621024eca8ef965affa5f1ff6472ed6

jplevyak: OK, so I finally understand what is going on.  Take a look at /test/common/thread_local/thread_local_impl_test.cc ThreadLocalInstanceImplDispatcherTest.Dispatcher.  This seems to imply that what I was suggesting was correct however I was deceived.

Envoy wants shutdownGlobalThreading to be called first then shutdownThread() to be called  *on each* thread *before it exist*.  This is because the "thread_local" contains smart pointers and the thread_atexit code is going to call the destructors.  However Envoy doesn't want them called.  Instead it wants to do the destruction manually in threadShutdown after shutdownGlobalThreading().   Specifically, threads are shutdown in Envoy in InstanceImpl::terminate by calling shutdownGlobalThreading then listener_manager->stopWorkers which then calls stop() on all the workers which causes the dispatcher_->run(Event::Dispatcher::RunType::Block) in WorkerImpl::threadRoutine to exit which causes tls_.shutdownThread to be called *on the running thread* before it exists.  After which threadShutdown() is called on the main thread.

Unfortunately the tests do not make this clear and only work because they do not allocate thread local memory on the new threads.


Furthermore this isn't necessary it is simply an artifact of the Envoy implementation where threads and workers are largely static.  It would be possible to have ephemeral threads with some small modifications of the Envoy code.

However, again for compatibility, it is probably best to stick with the Envoy model.

That means that the implementation here is incorrect because it is not calling shutdownThread on the thread itself before it exists, but instead just on the main thread.

I have modified my solution yet again to follow the Envoy model.  PTAL

Sorry to be so pedantic about this but threading issues have a tendency to cause very hard to debug issues and I want to make sure I understand what is going on as best I can.

BTW: this fixes my integration_test issue.

PTAL at https://github.com/jplevyak/nighthawk/commit/f3976f9fb1593fb2e807589f6dc18ec87f17e3fa
and the branch https://github.com/jplevyak/nighthawk/tree/tracing-prelude-1

oschaaf: Awesome, I just confirmed your proposed changes to work on the tracer branch, and with your modifications the NH shutdown code gets easier to understand. The missing call to `threadShutdown()` on the main thread is a great catch. Thanks for persisting!! 

Perhaps it is worth considering capturing some of the value of the debugging/analysis here by adding asserts in Envoy to guide those who come after us? (For example, looking at `ThreadLocal::InstanceImpl`, at destruction time there is an assert which ensures that `shutdownGlobalThreading()` was called, maybe something similar for `shutdownThread()` would be helpful to those who come after us).

jplevyak: Yes adding an assert to the destructor for TSL would help, as well as documentation, and adding/modifying the threading tests so that they cover this case would help as well.

PR #133, opened 2019-09-05
Python-test: switch to pytest / x-dist

oschaaf:
- Currently the integration tests take 120 seconds or more to complete, and we cannot shard.
This PR switches us to use `pytest` with the `x-dist` plugin. This brings down test time to approximately 20 seconds total with a modest hard-coded concurrency level specified (10).
- Also, we couldn't parameterize tests so we had an awkward custom solution for that. We can now
rely on pytest for that.

As a bonus we get *much* better failure outputs, as well as edge a step closer towards running distributed tests.


oschaaf:
@jplevyak it may make sense to hold off reviewing this for a bit

(I think it might make things easier to queue this up and review when we have merged the other outstanding changes changes; I wish we had recursive pull requests)

test/integration/BUILD
```diff
@@ -33,5 +33,21 @@ py_test(
         requirement("chardet"),
         requirement("certifi"),
         requirement("idna"),
+        requirement("pytest"),
```
jplevyak: If more than one tests might use pytest, consider adding a macro to include all these dependencies.

test/integration/integration_test.py
```diff
@@ -1,45 +1,13 @@
-#!/usr/bin/env python3
+#! /home/venv/bin/python
```
jplevyak: Just for my edification: where is the virtual environment configured?

oschaaf: Backed out the venv shebang: this is cruft from an experiment which I accidentally left in

Issue #102, opened 2019-08-01
Allow configuration of server-side request latency distributions


tonya11en:
To test various behaviors for [certain Envoy filters](https://github.com/envoyproxy/envoy/issues/7789), it would be very handy to be able to specify request latencies with a weighted probability. This would allow us to observe the effect of tail latencies on filter behavior.

Something like this should work:
`--latency_distribution=zoned:3/1ms:1/4ms:1/20ms`

That would translate into:
- 3/5 of latencies should be 1ms
- 1/5 of latencies should be 4ms
- 1/5 of latencies should be 20ms

PR #137, opened 2019-09-11
RemoteHeaderSource & stub grpc service HeaderStreamRequest

oschaaf:
Adds a `--request-source` option, which directs Nighhtawk's workers to connect to a remote grpc service to pull to-be-replayed request headers from. These headers will be merged on the client side with any uri configuration specified on the client side.

```
--request-source <uri format>
Remote gRPC source that will deliver to-be-replayed traffic. Each
worker will separately connect to this source. For example
grpc://127.0.0.1:8443/.
```

Summary of introduced components:
- Introduces a grpc client that NH's workers rely on to talk to a service to get to-be-replayed contents
- Introduces a new `RequestSource` implementation which uses the above to source remote request data.
- Introduces a new dummy grpc service method which implements a stub header stream for testing
- Introduces a python wrapper for our grpc service for use in integration testing
- Makes our grpc service listen to certain signals so we can shut it down nicely from the integration tests.

Related issues:
- https://github.com/envoyproxy/nighthawk/issues/34
- https://github.com/envoyproxy/nighthawk/issues/102

List of things to do before we can proceed with this PR:

-------------

- [x] Land #212 
- [x] Land #213 
- [x] Land #214 
- [x] Land #220 
- [x] Land #216
- [x] gRPC python service test wrapper seems to occasionally timeout while we wait for it to start in sanitizer runs. Possibly we just need to increase the amount of time we allow for that.
- [x] Fix [tsan failure](https://github.com/envoyproxy/nighthawk/pull/137#issuecomment-558235691) on signal handling 
- [x] Finalize and land the api changes in #229


oschaaf:
While this PR may be a bit larger then ideal size for a review, splitting it up isn't trivial as comonents in there rely on each other for testing. Please let me know if this is workable.

api/client/options.proto
```diff
@@ -91,6 +95,7 @@ message CommandLineOptions {
   oneof oneof_request_options {
     // See :option:`--request_options` for details.
     RequestOptions request_options = 12;
+    HeaderSource header_source = 19;
```
htuch: I think we should bike shed the data model a bit before diving into implementation. My thoughts were that we would ideally structure the record-replay along trace lines, i.e. build up a conceptual trace and then replay it, including various attributes of the request, not just header. This could be streamed, in the way that you plan for headers here, but it would be good to describe the roadmap for the data model here.

oschaaf: So right now the `Content-Length:` request header dictates how many bytes NH will upstream for a request in the entity body, so it's not explicitly modeled in the proto.. I pinged you on Slack to bikeshed about that :-)

htuch: Yeah, I reckon it makes sense to create the trace format that we want to be heading for, even if we just replicate constants in it today.

api/client/service.proto
```diff
@@ -37,10 +38,46 @@ message ExecutionResponse {
...
+
+message Predicate {
+  /*
+  TODO(oschaaf): define this, ideally we can express:
+  >, >=, ==, <, <=, in
+  */
+  string name = 1;
+}
+
+message HeaderStreamResponse {
```
oschaaf: @htuch update `HeaderStreamResponse`, does this look better?

oschaaf: (edit: typo)

api/client/service.proto
```diff
@@ -37,10 +39,57 @@ message ExecutionResponse {
...
+  // so technically they are redundant. In case of ambiguity headers may end up being duplicated.
+  // Note that as of today, the target upstream/cluster that will be used to replay traffic to
+  // fixes the destination address/port
+  google.protobuf.StringValue method = 1;
+  google.protobuf.StringValue uri = 2; // XXX(oschaaf): path?
+  google.protobuf.StringValue authority = 3;
+  envoy.api.v2.core.HeaderMap headers = 4;
+  google.protobuf.UInt32Value content_length = 5;
+
+  repeated Expectation expectations = 6;
```
htuch: Why do we have these complicated expectation conditions? Why not just exact match criteria?

api/client/service.proto
```diff
@@ -37,10 +39,57 @@ message ExecutionResponse {
...
+
+  // TODO(oschaaf): nice to have
+  // google.protobuf.StringValue sni_hostname = 10;
+}
```
htuch: How far is this from the original replay format we discussed when we started this exercise? if you can share an abstract (i.e. non Google specific) idea of this, it would be good to compare.

api/client/service.proto
```diff
@@ -37,10 +39,57 @@ message ExecutionResponse {
...
+  DataType input_header_data_type = 3;
+  Condition condition = 4;
+}
+
+message RequestStreamResponse {
+  // method/path/uri/authority/content_length can all be expressed in request headers.
+  // so technically they are redundant. In case of ambiguity headers may end up being duplicated.
+  // Note that as of today, the target upstream/cluster that will be used to replay traffic to
+  // fixes the destination address/port
+  google.protobuf.StringValue method = 1;
```
htuch: So, this is actually a `Request` specifier? If so, it should probably live in a message called something like `RequestSpecifier`, as it seems like it's somehow response related as is.

api/client/service.proto
oschaaf: The draft replay format we came up with looked like:
 
``` proto
message KeyValue {
  string key = 1;
  string value = 2;
}

string request_method = 1;
string request_url = 2;
string authority = 3;
repeated KeyValue request_headers = 4;
string request_body = 5; (for POSTs)
 
optional int expected_response_code;
optional int expected_response_size; // (possibly also some to bound response size being at least N bytes or at most M bytes)
optional string expected_response_body;
optional KeyValue expected_response_headers; // (I assume subset of response headers that must be present)
 
// Possibly:
string sni_hostname = 6;
```

api/client/service.proto
oschaaf: Well, I thought it might be useful to be able to express more then just exact matches, so we can set up expectations to match ranges of response-codes, and specify upper and/or lower bounds on expected response sizes. Also, see the original replay format draft, which suggests that being able to express this might be desirable.

htuch: I think we should add this as needed. I think in general, we should have a framework that is relatively deterministic. It will send some request and get back the same thing each time.

oschaaf: As per Slack discussion, +1 on simplifying request-level expecations. 
I'll simplify and modify this PR to prepare it for code-level review.

oschaaf:
Feedback addressed in https://github.com/envoyproxy/nighthawk/pull/137/commits/7a39bc9c149f69b4cabbb4ea9039913f3aea8095

Note that this leaves some code commented for future use, with that we pass coverage.

/assign htuch

api/client/service.proto
```diff
@@ -37,10 +39,40 @@ message ExecutionResponse {
   }
 }
 
+message RequestStreamRequest {
+  uint64 amount = 1;
```
htuch: I'm curious why the client needs to specify this at all? Is this some form of basic flow control so that the client isn't overwhelmed?

api/client/service.proto
```diff
@@ -37,10 +39,40 @@ message ExecutionResponse {
...
+  google.protobuf.StringValue authority = 3;
+  envoy.api.v2.core.HeaderMap headers = 4;
+  google.protobuf.UInt32Value content_length = 5;
+  // TODO(oschaaf): nice to have
+  // google.protobuf.StringValue sni_hostname = 10;
+}
+
+message Expectations {
+  google.protobuf.UInt32Value response_code = 1;
+  google.protobuf.UInt32Value content_length = 2;
```
htuch: Please comment all proto fields here and elsewhere.

api/client/service.proto
oschaaf: Yes, added a comment about that. 
I couldn't find a way to cleanly add this capability otherwise (without that becoming a significant effort in itself).

api/client/service.proto
oschaaf: Done

htuch:
@aghost527 @codeblooded could you folks weigh in on this from a Google EngProd perspective? Specifically, we're after feedback on whether the data model in the proto and the RPC service architecture matches what you would expect for a load generator consuming a replay trace. Thanks!

codeblooded:
cc: @paulosjca, since he can provide critical insight here.

htuch:
Friendly ping to @envoyproxy/nighthawk-reviewers, would be great to get the take of some of the interested parties who will be integrating with Nighthawk.

eric846:
> Friendly ping to @envoyproxy/nighthawk-reviewers, would be great to get the take of some of the interested parties who will be integrating with Nighthawk.

I think this could be important for some of our upcoming tests. 

I definitely agree with the principle. If we want to do something as simple as make interleaved requests to 2 different URLs, I don't think we can do that with a basic Nighthawk benchmark, which (correct me if I'm wrong) hits a single URL over and over. And we have talked about randomized requests with some custom statistical model.

gRPC would certainly be a convenient interface for test writers who want to implement custom logic. I think we could conceptually implement all our current plans using what's in this PR.

The only question is whether having gRPC in the loop for every request could completely disrupt the performance of Nighthawk. I don't know much about it -- if gRPC to localhost is a million times faster than the network with nanosecond latency, no problem. If we would need to worry or do a lot of work to measure and validate it, we should consider a way to move request generation in-process.

What if instead we plan for the test writer to compile extra C++ plugins into Nighthawk? They could pass a command line arg that selects and configures a request source plugin.

 We could ship Nighthawk with 2 plugins:
- the gRPC client in this PR (taking a host:port arg)
- a file reader (taking a pathname arg) (lower priority)

Anything else would be the test writer's custom C++ plugin.

All they would have to implement would be some plugin interface derived from `RequestSource` or `RequestGenerator`, and the base class could take care of registering the plugin. I would actually find it easier to implement that than a gRPC server.

oschaaf:
@eric846 Thanks for the feedback; I like the idea of having a plugin model here, and I can see that would make sense. Determining the performance impact of the gRPC service might be doable without a lot of effort.. Are you on Slack or hangouts? Maybe by chatting up over there we can figure this out fast.

mum4k:
@oschaaf we had an offline discussion about this and concluded that while the performance is an important factor, we are happy to proceed with the current PR to get the functionality in, as it brings something our testing framework depends on (as per @eric846's response).

We are still interested in performance testing this and refactoring it into a plugin framework, but we are happy to do that outside of this PR assuming you agree on that as the path forward. Please let us know.

With that I will start reviewing this PR, I expect this will take me a couple of days due to its size. I will send partial review comments or questions once I have any.

README.md
```diff
@@ -68,6 +69,9 @@ format>
 
 Where:
 
+--request-source <string>
+replay source description
```
mum4k: Can we expand this description? As it stands I don't really understand what it would be used for. We could also consider including an example.

api/client/options.proto
```diff
@@ -93,6 +97,7 @@ message CommandLineOptions {
   oneof oneof_request_options {
     // See :option:`--request_options` for details.
     RequestOptions request_options = 12;
+    RequestSource request_source = 25;
```
mum4k: Can we add comments for the new messages and fields?

api/client/options.proto
```diff
@@ -14,6 +14,10 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+message RequestSource {
+  google.protobuf.StringValue uri = 19; // [(validate.rules).string.uri = true];
```
mum4k: Curious, why did we choose to wrap the uri in a message? Do we expect more qualifiers for a request source?

api/client/service.proto
```diff
@@ -2,8 +2,10 @@ syntax = "proto3";
 
 package nighthawk.client;
```
mum4k: Before I review the changes here I have a few questions that can help me to understand this better. What is the definition of a "client" and a "server" when it comes to this RPC service. Do I understand it correctly that NH acts as the client and this proto file defines a server that should be implemented by the users who want to use NH in performance tests?

I.e. is it NH that will call the defined RPC methods? Would you mind adding a top level file comment with some background and few notes about the architecture so that this is clear?

oschaaf:
Note: Just split out https://github.com/envoyproxy/nighthawk/pull/212 from this PR

oschaaf:
Note: also split out https://github.com/envoyproxy/nighthawk/pull/214

Hopefully we can get this down to a more fun-to-review size by iterating on splitting out things like this.

I'm [working on document](https://docs.google.com/document/d/1k21xIVCxQX51OORw_WQYjRLzi0m7pM1MN0859XXqWxI/edit?usp=sharing) with a high-level description of Nighthawk. While not specific to this PR, I'm hoping it will address some of the high-level requests to document here when I am done.

api/client/service.proto
oschaaf: @mum4k Please see "Executables / nighthawk_service" over at https://docs.google.com/document/d/1k21xIVCxQX51OORw_WQYjRLzi0m7pM1MN0859XXqWxI/edit?usp=sharing

(note: doc is a work in progress)

oschaaf:
tsan fails with:

```
WARNING: ThreadSanitizer: signal-unsafe call inside of a signal (pid=28514)
    #0 free /local/mnt/workspace/bcain_clang_bcain-ubuntu_23113/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/tsan/rtl/tsan_interceptors.cc:706:3 (nighthawk_service+0x1d5d518)
    #1 gpr_free /proc/self/cwd/external/com_github_grpc_grpc/src/core/lib/gpr/alloc.cc:77:3 (nighthawk_service+0x4536f6d)
    #2 grpc_cq_internal_unref(grpc_completion_queue*, char const*, char const*, int) /proc/self/cwd/external/com_github_grpc_grpc/src/core/lib/surface/completion_queue.cc:605:5 (nighthawk_service+0x450124c)
    #3 grpc_completion_queue_destroy /proc/self/cwd/external/com_github_grpc_grpc/src/core/lib/surface/completion_queue.cc:1409:3 (nighthawk_service+0x45021e5)
    #4 grpc::CoreCodegen::grpc_completion_queue_destroy(grpc_completion_queue*) /proc/self/cwd/external/com_github_grpc_grpc/src/cpp/common/core_codegen.cc:67:3 (nighthawk_service+0x4193694)
    #5 grpc_impl::CompletionQueue::~CompletionQueue() /proc/self/cwd/external/com_github_grpc_grpc/include/grpcpp/impl/codegen/completion_queue_impl.h:116:39 (nighthawk_service+0x239e243)
    #6 grpc_impl::Server::ShutdownInternal(gpr_timespec) /proc/self/cwd/external/com_github_grpc_grpc/src/cpp/server/server_cc.cc:1326:1 (nighthawk_service+0x41bd99e)
    #7 grpc::ServerInterface::Shutdown() /proc/self/cwd/external/com_github_grpc_grpc/include/grpcpp/impl/codegen/server_interface.h:106:5 (nighthawk_service+0x1e40a90)
    #8 Nighthawk::Client::ServiceMain::shutdown() /proc/self/cwd/source/client/service_main.cc:90:12 (nighthawk_service+0x1e3ecb2)
    #9 Nighthawk::Client::ServiceMain::wait()::$_0::operator()(int) const /proc/self/cwd/source/client/service_main.cc:82:40 (nighthawk_service+0x1e3f434)
    #10 std::_Function_handler<void (int), Nighthawk::Client::ServiceMain::wait()::$_0>::_M_invoke(std::_Any_data const&, int&&) /usr/lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/bits/std_function.h:316:2 (nighthawk_service+0x1e3f1e1)
    #11 std::function<void (int)>::operator()(int) const /usr/lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/bits/std_function.h:706:14 (nighthawk_service+0x1e46620)
    #12 Nighthawk::Client::(anonymous namespace)::signal_handler(int) /proc/self/cwd/source/client/service_main.cc:20:35 (nighthawk_service+0x1e3ebeb)
    #13 __tsan::CallUserSignalHandler(__tsan::ThreadState*, bool, bool, bool, int, __sanitizer::__sanitizer_siginfo*, void*) /local/mnt/workspace/bcain_clang_bcain-ubuntu_23113/llvm/utils/release/final/llvm.src/projects/compiler-rt/lib/tsan/rtl/tsan_interceptors.cc:1914:7 (nighthawk_service+0x1d6552e)
    #14 gpr_cv_wait /proc/self/cwd/external/com_github_grpc_grpc/src/core/lib/gpr/sync_posix.cc:128:11 (nighthawk_service+0x453bed5)
    #15 grpc::CoreCodegen::gpr_cv_wait(pthread_cond_t*, pthread_mutex_t*, gpr_timespec) /proc/self/cwd/external/com_github_grpc_grpc/src/cpp/common/core_codegen.cc:92:10 (nighthawk_service+0x4193c33)
    #16 grpc::internal::CondVar::Wait(grpc::internal::Mutex*, gpr_timespec const&) /proc/self/cwd/external/com_github_grpc_grpc/include/grpcpp/impl/codegen/sync.h:134:38 (nighthawk_service+0x41d4e88)
    #17 grpc::internal::CondVar::Wait(grpc::internal::Mutex*) /proc/self/cwd/external/com_github_grpc_grpc/include/grpcpp/impl/codegen/sync.h:130:12 (nighthawk_service+0x41c3347)
```

oschaaf:
List of things to do before we can proceed with this PR:

- [] Land #212 
- [] Land #214 
- [] Fix tsan failure on signal handling (see above)

oschaaf:
We landed all the prerequisites, @envoyproxy/nighthawk-reviewers is this PR now downsized enough to comfortably take it forward here?

mum4k:
Thank you for downsizing this PR @oschaaf, I think it is at a good size code-wise. How would you feel about taking the public API (proto changes) into its own PR so that we can agree on the interaction and what can and cannot be done before we dive into the implementation? We have moved forward with the design of our testing framework internally and are in a better position to comment and explain what we need. I would add more reviewers (all of us) to the API PR if we split it.

Alternatively (if you prefer) we can start here by reviewing just the protos first and then heading down to code. Please let us know.

oschaaf:
@mum4k I will split out the proto changes into a separate PR

oschaaf:
Api changes split out to https://github.com/envoyproxy/nighthawk/pull/229

oschaaf:
Speculatively merged #229 into this. 1108afada3628f419835023f189a35a3c2b1a20c has the nesecary adjustments. Notably, that still has a single binary, but allows service selection on the cli via `--service controller|request-source`

mum4k:
@eric846, this has been thinned significantly and is now ready for a review. Please assign back to me once ready.

source/client/service_impl.cc
```diff
@@ -107,5 +109,65 @@ ::grpc::Status ServiceImpl::ExecutionStream(
...
+::grpc::Status RequestSourceServiceImpl::RequestStream(
+    ::grpc::ServerContext* /*context*/,
+    ::grpc::ServerReaderWriter<::nighthawk::request_source::RequestStreamResponse,
+                               ::nighthawk::request_source::RequestStreamRequest>* stream) {
+  nighthawk::request_source::RequestStreamRequest request;
+  bool ok = true;
+  while (stream->Read(&request)) {
+    ENVOY_LOG(trace, "Inbound RequestStreamRequest {}", request.DebugString());
+
+    // TODO(oschaaf): this is useful for integration testing purposes, but sending
```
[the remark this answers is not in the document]
eric846: I finally understand what's going on here. So this is the implementation of a trivial request source gRPC service, and internally it instantiates yet another request source, but a static in-process one. And this request source server code ends up performing the same request_generator() calls that the nighthawk_client does.

I think we should assume that someone else could also be confused by this situation, even if they figure it out 10x faster. We might want to add a comment calling out the fact that the usage here is a clever one, and the one benchmark_client_impl.cc is a normal one.

And now that I think about it more, I don't see why future enhancements to Nighthawk like reading from a file should be placed in the RequestSource service, behind a layer of indirection through gRPC -- especially if the file reading will be implemented as a RequestSource anyway, and the RequestSource factory could just instantiate the file-reading RequestSource in the nighthawk_client process directly.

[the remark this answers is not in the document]
oschaaf: yeah, exactly. So the current state should be taken as a first step, and when pulling traffic in over gRPC via the remote request source it is just a (very elaborate) no-op when comparing what nighthawk_client does today. 

But it has some use, in proving that it all works, landing test coverage, and also that the new abstractions we introduced while working on this are on par for the final destination we have in mind. 

When I wrote this, what I had in mind is that we would have a request source abstraction, and users can plug implementations of that into either nighthawk client or nighthawk service. The idea behind that is that we build generic tools and let users decide how and where use them. I think that when we get to factorizing `RequestSource` based on options, this will automatically propagate to both the `nighthawk_client` and `nighthawk_service`. Is that what you are suggesting as well?

source/common/request_source_impl.h
```diff
@@ -16,10 +18,31 @@ class StaticRequestSourceImpl : public BaseRequestSourceImpl {
...
+  RemoteRequestSourceImpl(Envoy::Upstream::ClusterManagerPtr& cluster_manager,
+                          Envoy::Event::Dispatcher& dispatcher, Envoy::Stats::Scope& scope,
+                          absl::string_view service_cluster_name,
+                          Envoy::Http::HeaderMapPtr&& base_header, uint32_t header_buffer_length);
+  RequestGenerator get() override;
+  void initOnThread() override;
+
+private:
+  void connectToRequestStreamGrpcService();
+  Envoy::Upstream::ClusterManagerPtr& cluster_manager_;
```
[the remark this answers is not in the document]
oschaaf: See my earlier comment; but perhaps we can const this, I'll try.

source/client/factories_impl.cc
```diff
@@ -138,7 +141,16 @@ RequestSourcePtr RequestSourceFactoryImpl::create() const {
     setRequestHeader(*header, option_header.header().key(), option_header.header().value());
   }
 
-  return std::make_unique<StaticRequestSourceImpl>(std::move(header));
+  if (options_.requestSource() == "") {
```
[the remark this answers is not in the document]
oschaaf: that's a nice solution, +1

include/nighthawk/client/factories.h
```diff
@@ -66,7 +66,9 @@ class OutputFormatterFactory {
 class RequestSourceFactory {
 public:
   virtual ~RequestSourceFactory() = default;
-  virtual RequestSourcePtr create() const PURE;
+  virtual RequestSourcePtr create(Envoy::Upstream::ClusterManagerPtr& cluster_manager,
```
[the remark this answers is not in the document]
eric846: I see, sounds good.

If the convention for this project is:

_`unique_ptr` parameter takes ownership_
vs.
_`unique_ptr&` parameter explicitly doesn't take ownership,_

that seems just as effective as 

 _`unique_ptr` parameter takes ownership_
vs.
_ raw pointer parameter explicitly doesn't take ownership_

source/client/service_impl.cc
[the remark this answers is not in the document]
eric846: Yes. I think we're good with the code in this PR since it leaves it open to handle each of those potential RequestSource implementations in its own way.

README.md
```diff
@@ -65,6 +65,9 @@ bazel-bin/nighthawk_client  [--trace <uri format>]
...
```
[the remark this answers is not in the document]
oschaaf: I think this comment was on stale code, this looked different on the latest revision

include/nighthawk/client/factories.h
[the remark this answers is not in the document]
oschaaf: In terms of ownership yes, and I would be happy either way; but in this case we expect the the unique ptr to get updated at a later point in time. I did const this in 1536e837535d2b242049ef19595393e9a3c18629 to make it clear we won't modify the Ptr ourselves.

source/client/options_impl.cc
```diff
@@ -174,6 +174,9 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
                                             "but in case of https no certificates are validated.",
                                             true, "", "uri format", cmd);
 
+  TCLAP::ValueArg<std::string> request_source(
+      "", "request-source", fmt::format("replay source description"), false, "", "string", cmd);
```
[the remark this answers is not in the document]
oschaaf: stale comment; I did make some updates, so it might be worth it to check the latest revision to see if that looks better.

source/common/request_source_impl.h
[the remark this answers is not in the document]
oschaaf: The const'ing is done in 1536e837535d2b242049ef19595393e9a3c18629

source/client/factories_impl.cc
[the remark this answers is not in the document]
oschaaf: Tightened up scheme constraining / defaulting for this in 1536e837535d2b242049ef19595393e9a3c18629

source/common/request_source_impl.cc
```diff
@@ -21,4 +21,35 @@ RequestGenerator StaticRequestSourceImpl::get() {
...
+                 "Source cluster not found");
+  envoy::api::v2::core::GrpcService grpc_service;
+  grpc_service.mutable_envoy_grpc()->set_cluster_name(service_cluster_name_);
+  auto cm =
+      cluster_manager_->grpcAsyncClientManager().factoryForGrpcService(grpc_service, scope_, true);
+  grpc_client_ = std::make_unique<RequestStreamGrpcClientImpl>(
+      cm->create(), dispatcher_, *base_header_, header_buffer_length_);
+  grpc_client_->start();
+  // Wait for the client's initial stream setup to complete.
+  while (!grpc_client_->streamStatusKnown()) {
```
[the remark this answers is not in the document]
oschaaf: Bounded in 1536e837535d2b242049ef19595393e9a3c18629

README.md
```diff
@@ -212,12 +218,16 @@ The gRPC service can be used to start a server which is able to perform back-to-
 ```
 USAGE:
 
-bazel-bin/nighthawk_service  [--listener-address-file <>] [--listen
+bazel-bin/nighthawk_service  [--service <controller|request-source>]
```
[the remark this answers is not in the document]
oschaaf: Amended as suggested in 1536e837535d2b242049ef19595393e9a3c18629

api/client/service.proto
```diff
@@ -34,10 +36,58 @@ message ExecutionResponse {
...
+  // google.protobuf.StringValue sni_hostname = 10;
+}
+
+// Response-level expectations. If set, any unmet expectations will result in stats counter
+// increases in the future.
+message Expectations {
+  // Expected response status code
+  google.protobuf.UInt32Value response_code = 1;
+  // Expected response body content length
+  google.protobuf.UInt32Value content_length = 2;
```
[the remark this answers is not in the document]
eric846: Haha, for some reason I was thinking we had to commit to the ultimate interface up front, not realizing we would immediately have to implement everything that was declared

source/client/options_impl.cc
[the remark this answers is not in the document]
eric846: Looks good

README.md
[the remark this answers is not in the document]
eric846: Yep, looks good

source/client/factories_impl.cc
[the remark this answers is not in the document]
eric846: This is great

README.md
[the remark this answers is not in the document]
eric846: Thanks, looks good

eric846:
LGTM

api/client/service.proto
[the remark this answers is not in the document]
oschaaf: > Haha, for some reason I was thinking we had to commit to the ultimate interface up front, not realizing we would immediately have to implement everything that was declared

That is an option. As far as I am concerned, we can iterate here in this PR or go and try to land the next steps in follow ups. Or we can go hybrid and try to shape this PR to the final state we want, and then split out sub PRs for in depth review when we are ready for that. But I stopped expanding this PR  earlier when I realized this change was getting to a complexity and size where it would be challenging to assess/land (even when already familiar with the code base).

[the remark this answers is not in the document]
eric846: Filed #260 to propose adding expected headers and body later. We definitely want to keep this initial PR manageable to be able to release something sooner.

mum4k:
@oschaaf, I am now ready to start the review here, can you help by resolving the conflicts?

mum4k:
@oschaaf also please take a look at the PR description in case any updates need to go in.

oschaaf:
> @oschaaf, I am now ready to start the review here, can you help by resolving the conflicts?

Done in 0958491e912233bc84381f1d84550cf885010350

oschaaf:
> @oschaaf also please take a look at the PR description in case any updates need to go in.

Updated it. (I left the todo list in the description for history tracking purposes, probably there's no value for having that in the commit message).

source/common/request_stream_grpc_client_impl.h
```diff
@@ -0,0 +1,78 @@
...
+namespace Nighthawk {
+
+class ProtoRequestHelper {
+public:
+  static RequestPtr
+  messageToRequest(const Envoy::Http::HeaderMap& base_header,
+                   const nighthawk::request_source::RequestStreamResponse& message);
+};
+
+class RequestStreamGrpcClientImpl
```
mum4k: Similarly here, can we add a short comment explaining what the class is for and document its constructor?

source/common/utility.cc
```diff
@@ -22,10 +22,14 @@ Utility::mapCountersFromStore(const Envoy::Stats::Store& store,
       std::string stat_name = stat->name();
       // Strip off cluster.[x]. & worker.[x]. prefixes.
       std::vector<std::string> v = absl::StrSplit(stat_name, '.');
-      if ((v[0] == "cluster" || v[0] == "worker") && v.size() > 1) {
-        v.erase(v.begin(), v.begin() + 2);
-        stat_name = absl::StrJoin(v, ".");
+      if (v[0] == "cluster" || v[0] == "worker") {
```
mum4k: (optional) Can we use this opportunity and add a unit test for this utility?

oschaaf:
@mum4k Thanks. I think I addressed all comments in the last two commits.

PR #146, opened 2019-09-19
[WIP] Conditional Zipkin tracing support

oschaaf:
Fails the integration test when zipkin is ripped out, hence still a work in progress.

This PR aims to restore the Zipkin tracing feature, but now with a build-time opt-out. 

For example:

```
bazel build --define zipkin_disabled=1 //...
```

The change compared to what was landed earlier is currently contained
in https://github.com/envoyproxy/nighthawk/commit/ce5979549c9d62ef5acfc2da399f828ed117e2a7


bazel/BUILD
```diff
@@ -1,3 +1,8 @@
 licenses(["notice"])  # Apache 2
 
 package(default_visibility = ["//visibility:public"])
+
+config_setting(
+    name = "zipkin_disabled",
+    values = {"define": "zipkin_disabled=1"},
```
htuch: Nit: it would be clearer in the code to make this `ZIPKIN_ENABLED` and invert the logic.

source/client/process_impl.cc
```diff
@@ -23,6 +24,11 @@
 #include "external/envoy/source/common/runtime/runtime_impl.h"
 #include "external/envoy/source/common/singleton/manager_impl.h"
 #include "external/envoy/source/common/thread_local/thread_local_impl.h"
+#include "external/envoy/source/extensions/tracers/well_known_names.h"
+
+#ifndef zipkin_disabled
```
htuch: In general, we try and avoid `#ifdef` in Envoy, maybe take a `TODO` to use some kind of static module registration (as done by Envoy) when we have more than null/Zipkin tracers?

htuch:
CC @stevenzzzz WDYT?

stevenzzzz:
LGTM.
Need to learn how to disable this feature in our codebase tho. :)

bazel/BUILD
oschaaf: I inverted this with respect to the define we add later on and consume in the c++ code. So that part is better.. But at this place I left it as-is pending a question I left on Slack; I'm a bit hesitant to make this opt-in instead of opt-out.

source/client/process_impl.cc
oschaaf: Added a TODO for that.

PR #148, opened 2019-09-24
Integration tests: retry server start

oschaaf:
It's possible that ports(s) we came up with for tests servers to use get stolen in the time it takes
to get to the test server to boot up.

This moves us to using a combination of generating port 0 bindings in configuration and querying the admin interface for the listeners that we are interested in.


oschaaf:
Some context on the practice of first figuring out a free port for the test server to use, 
and then pushing it through configuration and actually starting the server:
With that we can use configuration generation as a generic concept for this, while otherwise,
for supporting multiple server types, we'd have to do per-server implementations to get to the
port(s) it used.

test/integration/nighthawk_test_server.py
```diff
@@ -60,8 +60,14 @@ def waitUntilServerListening(self):
       while tries > 0:
         time.sleep(0.5)
         tries -= 1
-        if sock.connect_ex((self.server_ip, self.server_port)) == 0:
-          return True
+        result = subprocess.run(
+            ["lsof", "-p",
```
htuch: Why can't the integration test server fail directly vs. us having to try indirectly to measure this? I'm worried about potential races here.

test/integration/integration_test_fixtures.py
```diff
@@ -70,13 +70,19 @@ def setUp(self):
...
-    assert (self.test_server.start())
+    attempts = 10
+    while attempts > 0:
+      self.server_port = self.getFreeListenerPortForAddress(self.server_ip)
+      self.admin_port = self.getFreeListenerPortForAddress(self.server_ip)
+      self.parameters["admin_port"] = self.admin_port
+      self.test_server = NighthawkTestServer(self.nighthawk_test_server_path,
+                                             self.nighthawk_test_config_path, self.server_ip,
+                                             self.server_port, self.ip_version, self.parameters)
+      if self.test_server.start():
```
htuch: This is OK... but it would be much nicer if we had end-to-end port zero bind and assignment. If you can add a TODO and/or point to a tracking issue that would be fine for now.

oschaaf: I ended up reverting this, several tries at getting to the pid/ports in the Docker environment failed. The most successful attempt was with `fuser` but that required docker to run privileged. So I just pushed a change to move to port 0 binding and querying the admin interface for the listener port that we are interested in instead.

test/integration/nighthawk_test_server.py
```diff
@@ -27,45 +30,63 @@ def __init__(self, server_binary_path, config_template_path, server_ip, server_p
...
-      sock.settimeout(1)
-      tries = 10
-      while tries > 0:
-        time.sleep(0.5)
-        tries -= 1
-        if sock.connect_ex((self.server_ip, self.server_port)) == 0:
-          return True
-      logging.error("Timeout while waiting for server listener at %s:%s to accept connections.",
-                    self.server_ip, self.server_port)
+  def tryUpdateFromAdminInterface(self):
```
htuch: This is great.. my only question is would it make sense to have a common admin access method that can do both stats scraping and listener fetching etc?

oschaaf: Yes; Done in 5109fd408c2bbf62fd2e188c12c7c04bf4030fe2

PR #149, opened 2019-09-27
Envoy dependency update

oschaaf:
Update to ef9139a991fb9f7df676430f626b32db93a06932


oschaaf:
Proposed https://github.com/envoyproxy/envoy/pull/8435, which would fix the failing integration tests reported in CI here.

oschaaf:
While the Envoy-change is under discussion, I pushed a change here to unblock this by moving to testing with another cipher (one of the few that is currently registered)

test/integration/test_integration_basics.py
```diff
@@ -192,11 +192,11 @@ def test_https_h1_tls_context_configuration(https_test_server_fixture):
 
   parsed_json, _ = https_test_server_fixture.runNighthawkClient([
       "--h2", "--duration 1",
-      "--tls-context {common_tls_context:{tls_params:{cipher_suites:[\"-ALL:ECDHE-RSA-AES256-GCM-SHA384\"]}}}",
+      "--tls-context {common_tls_context:{tls_params:{cipher_suites:[\"-ALL:ECDHE-RSA-CHACHA20-POLY1305\"]}}}",
       https_test_server_fixture.getTestServerRootUri()
   ])
   counters = https_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
-  assertEqual(counters["ssl.ciphers.ECDHE-RSA-AES256-GCM-SHA384"], 1)
+  assertEqual(counters["ssl.ciphers.ECDHE-RSA-CHACHA20-POLY1305"], 1)
```
htuch: Arguably changing ciphers will create differences in performance. I think for now this is fine, but please do take on the long term fix.

oschaaf: Added https://github.com/envoyproxy/nighthawk/issues/156 for tracking

Issue #136, opened 2019-09-07
Option to output a flat key-value list


oschaaf:
An idea was floated during review [1], it could be useful to have a way to output a flat key-value list on top of the output options that we have today. 

Context: https://github.com/envoyproxy/nighthawk/pull/116#pullrequestreview-282273357

PR #150, opened 2019-09-28
Add `dotted` output format

oschaaf:
This adds `--output-format dotted` as an option. Doing so will make
the client emit a flattened list of dotted strings with `key: value`s.
This simple format is useful for processing output with minimal
dependencies on other libraries like json/proto/etc, and prometheus
also supports it.

Fixes https://github.com/envoyproxy/nighthawk/issues/136


oschaaf:
@jplevyak this is ready for review

source/client/output_collector_impl.cc
```diff
@@ -122,5 +122,51 @@ std::string YamlOutputCollectorImpl::toString() const {
   return Envoy::MessageUtil::getYamlStringFromMessage(toProto(), true, true);
 }
 
+DottedStringOutputCollectorImpl::DottedStringOutputCollectorImpl(Envoy::TimeSource& time_source,
+                                                                 const Options& options)
+    : OutputCollectorImpl(time_source, options) {}
+
+std::string DottedStringOutputCollectorImpl::toString() const {
```
htuch: As per my comment on the other review, can we make this a pure function that takes a Nighthawk proto output and performs the format? That way we can avoid putting this on the wire and allow clients to decide how they want to do the output formatting.

oschaaf: So as we are planning to have [a separate CLI](https://github.com/envoyproxy/nighthawk/pull/159) instead, I think we can pursue landing this as-is, correct?

source/client/output_collector_impl.cc
```diff
@@ -44,7 +44,7 @@ std::string ConsoleOutputCollectorImpl::toString() const {
 
         // The proto percentiles are ordered ascending. We write the first match to the stream.
         double last_percentile = -1.;
-        for (const double p : {.0, .5, .75, .8, .9, .95, .99, .999, 1.}) {
+        for (const double p : {.5, .75, .8, .9, .95, .99, .999, 1.}) {
```
htuch: Is this a bug fix?

test/integration/test_integration_basics.py
```diff
@@ -271,3 +271,16 @@ def test_https_log_verbosity(https_test_server_fixture):
...
+def test_dotted_output_format(http_test_server_fixture):
+  """
+  Ensure we get the dotted string output format when requested.
+  and ensure we get latency percentiles.
+  """
+  output, _ = http_test_server_fixture.runNighthawkClient([
+      "--duration 1", "--rps 10", "--output-format dotted",
+      http_test_server_fixture.getTestServerRootUri()
+  ],
+                                                          as_json=False)
```
htuch: This seems very strange formatting, are we running pyformat on the NH repo? See what Envoy does.

source/client/output_collector_impl.h
```diff
@@ -50,5 +50,11 @@ class YamlOutputCollectorImpl : public OutputCollectorImpl {
   std::string toString() const override;
 };
 
+class DottedStringOutputCollectorImpl : public OutputCollectorImpl {
```
htuch: Can you link to where this format is specified? i.e. is there an RFC or something?

source/client/output_collector_impl.cc
```diff
@@ -122,5 +122,51 @@ std::string YamlOutputCollectorImpl::toString() const {
...
+                        Envoy::Protobuf::util::TimeUtil::DurationToMicroseconds(statistic.mean()))
+         << std::endl;
+      ss << fmt::format("{}.pstdev: {}", prefix,
+                        Envoy::Protobuf::util::TimeUtil::DurationToMicroseconds(statistic.pstdev()))
+         << std::endl;
+
+      // The proto percentiles are ordered ascending. We write the first match to the stream.
+      // Revisit to see if we can factor out a common denominator with the CLI output formatter.
+      double last_percentile = -1.;
+      for (const double p : {.5, .75, .8, .9, .95, .99, .999, 1.}) {
```
htuch: This iteration over the histogram seems like a common pattern shared by the different output formatters. Would it be possible to create a higher-order function or class that implements this, with the inner block supplied by a lambda?

source/client/output_collector_impl.cc
oschaaf: Yes, this is a bug-fix, forgot to mention it in the PR description. 
Should I file a separate issue on this?

test/integration/test_integration_basics.py
oschaaf: This is what `tools/check_format fix` outputs. I double checked it, because I also thought it looked rather strange. Maybe we're hitting some weird edge case?
And yup, we are running formatting checks in CI (and they are content with this).

source/client/output_collector_impl.h
oschaaf: I'm actually not sure if there's a formal spec. @jplevyak would you know?

source/client/output_collector_impl.cc
oschaaf: Sure; can I take a TODO on that?

source/client/output_collector_impl.cc
htuch: It's fine, just best in future to split these out :)

test/integration/test_integration_basics.py
htuch: Fair enough, kind of strange though.

source/client/output_collector_impl.cc
htuch: I think it could be good to do in this PR, it should actually shrink the size of the PR.

test/test_data/output_collector.dotted.gold
```diff
@@ -0,0 +1,24 @@
...
+worker_1.stat_id.pstdev: 816
+worker_1..samples: 0
+worker_1..mean: 0
+worker_1..pstdev: 0
+worker_1.bar:2
+worker_1.foo:1
+global.stat_id.samples: 3
+global.stat_id.mean: 2000
+global.stat_id.pstdev: 816
+global..samples: 0
```
jplevyak: What do the double periods mean? ".." ?

source/client/output_collector_impl.cc
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/150/commits/3836ba1cc1e39f0265e7ec02714a6de417a4fb30

source/client/output_collector_impl.cc
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/150/commits/3836ba1cc1e39f0265e7ec02714a6de417a4fb30

test/test_data/output_collector.dotted.gold
oschaaf: re: `What do the double periods mean? ".." ?`
Sorry, forgot to get back to this. We're outputting a stat with an empty name.
We won't be seeing that in practice.

source/client/output_collector_impl.cc
oschaaf: So actually, not having it introduced a bug, we would miss the minimal value. Added it back.

Issue #151, opened 2019-09-28
Add CLI for transforming output


oschaaf:
For convenience, it would be useful to emit transformed output when an output format is specified.
Today, we only acknowledge the option in the CLI.

oschaaf:
After discussing on Slack, consensus is that a separate CLI for transforming output is the better approach.

oschaaf:
Done in https://github.com/envoyproxy/nighthawk/pull/159

PR #152, opened 2019-09-28
Acknowledge requested output format in service

oschaaf:
Add `formatted` string field to output which will be set
when an output format is requested in a service call.

Fixes https://github.com/envoyproxy/nighthawk/issues/151


oschaaf:
> What motivates this? Can't the client just convert the result proto to JSON using the protobuf lib at its end?

Well @nareddyt and me had a chat on Slack, and he pointed out we can't access any transformed output when using the grpc service. While with the current status I agree this PR doesn't add much value, it becomes more useful as we land [the dotted string formatter](https://github.com/envoyproxy/nighthawk/pull/150), and [we might also see a fortio output formatter](https://github.com/envoyproxy/nighthawk/issues/155)

htuch:
@oschaaf can we make this library transform something that happens in the client library? I'd prefer not to have redundancy and UX on the wire.

oschaaf:
So after discussing on Slack with @htuch @nareddyt, the optimal approach seems to be a separate CLI. Therefore I am closing this PR in favor of https://github.com/envoyproxy/nighthawk/pull/159

Issue #140, opened 2019-09-16
Get the python integration tests to count for coverage testing


oschaaf:
Right now we have some c++ unit tests that would be better of written as python integration tests. Unfortunately, the python tests do not count for coverage measurement. 
Look into ways to make that happen.

PR #153, opened 2019-09-28
Coverage: C++ test wrapper for the python tests

oschaaf:
Get coverage test runs to consider the code hit by our python-based testing.

Fixes https://github.com/envoyproxy/nighthawk/issues/140

### DIFF
--- a/test/integration/BUILD
+++ b/test/integration/BUILD
@@ -61,7 +61,7 @@ py_library(
     deps = [":integration_test_base"],
 )
 
-py_test(
+py_binary(
     name = "integration_test",
     srcs = ["integration_test.py"],
     args = select({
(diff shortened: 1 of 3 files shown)


test/integration/BUILD
```diff
@@ -61,7 +61,7 @@ py_library(
     deps = [":integration_test_base"],
 )
 
-py_test(
+py_binary(
```
jplevyak: So py_test doesn't count for coverage but coverage works over forking.  OK.

oschaaf: That was the plan, make python run the with instrumented binaries.. 
I thought that worked, however, double checking again this actually didn't work as intended.
So I just pushed another commit which declares some of the top-level libs as dependencies.
With that I could verify this to work.

jplevyak: Did it work?

oschaaf: Yes, I could verify that as of the last commit, it worked

Issue #154, opened 2019-10-02
Add a way to terminate test runs based on observed conditions


oschaaf:
In open-loop testing mode, the client should be able to detect and act upon certain thresholds being exceeded. For example, when bursts of bad status codes or connection failures are observed.

PR #158, opened 2019-10-04
Conditions for terminating a test run

oschaaf:
Conceptualize a simple system based on linked predicates that
represent when and how we would like a test to terminate.

Doesn't wire anything up yet, just ground work.

Part of https://github.com/envoyproxy/nighthawk/issues/154


oschaaf:
> Cool, nice to see this coming together. One question; what's the motivation for the recursive linking etc? Can't we just pass in a predicate and use normal function composition to combine?

My idea was to convert all the exit conditions to this. So, for example:
- We convert the current way that we track duration to a predicate that tracks time and indicates successful completion
- A user can configure  a boundary on, say, 5XX. We translate that to a predicate that indicates  failure when exceeded, and link it to the first predicate.
- Repeat with more user-specified thresholds.

So we build a chain of those predicates and pass the first link to the sequencer, which will repeatedly query that to get status. This call to get the status is a single, well-defined call, which we can extend meaningfully later on if we want to.

include/nighthawk/common/termination_predicate.h
```diff
@@ -0,0 +1,49 @@
...
+#include "envoy/common/pure.h"
+
+namespace Nighthawk {
+
+class TerminationPredicate;
+
+using TerminationPredicatePtr = std::unique_ptr<TerminationPredicate>;
+
+/**
+ * Determines if and how a test is terminated.
```
htuch: Do we have consistent terminology for what constitutes a "test"? https://github.com/envoyproxy/nighthawk/blob/master/api/client/service.proto talks about executions. Would be good to have a standard lexicon.

oschaaf: Right, good point. Shall I rename `test` here to execution, and create a TODO to put up a `TERMINOLOGY.md` in the root of the repo?

jplevyak: Sounds good to me.

include/nighthawk/common/termination_predicate.h
```diff
@@ -0,0 +1,49 @@
...
+
+using TerminationPredicatePtr = std::unique_ptr<TerminationPredicate>;
+
+/**
+ * Determines if and how execution is terminated.
+ */
+class TerminationPredicate {
+public:
+  enum class Status {
+    PROCEED,   // indicates a test should proceed
```
htuch: "test" terminology used in these comments as well.

oschaaf: Amended!

oschaaf:
Renames done, filed https://github.com/envoyproxy/nighthawk/issues/161 for the clang-tidy flakes in CI.

PR #160, opened 2019-10-06
Integration test: enableCpuProfiler

oschaaf:
Enable profiling for a single stress test in the integration tests.
Also, adds `PROFILING.md` with steps for manual profiling of Envoy.

Not intended to merged as-is, but meant to open discussion on if this would be deemed useful in general. 

I was thinking that we could add a series of benchmarks, to be run via the existing python integration test facilities offered here, and dump collected profile data for those. 
Visualizing the profile via `pprof` is as simple as:

```
pprof -http=localhost:8888 /path/to/envoy.prof
```

This approach could facilitate consolidation of perf scenario's we would like to test, and store them in a way so that they can be easily repeated, reproduced locally, discussed, etc. as well as verified for correctness.


oschaaf:
`PROFILING.md` can be previewed with markup applied [here](https://github.com/oschaaf/nighthawk/blob/test-profiling/PROFILING.md).

PROFILING.md
```diff
@@ -0,0 +1,78 @@
...
+The important part is that the admin interface needs to be set up to allow one to enable / disable
+Profiling via http, as well as specify a location to dump the profiling data.
+
+``` yaml
+admin:
+  access_log_path: /tmp/admin_access.log
+  profile_path: /tmp/envoy.prof
+  address:
+    socket_address: { address: $server_ip, port_value: 0 }
+Static_resources:
```
htuch: Type: lower case `static_resources`

PROFILING.md
```diff
@@ -0,0 +1,78 @@
...
+.. your configuration ..
+```
+
+## Run Envoy
+
+```bash
+taskset /path/to/envoy-repo/bazel-bin/envoy-static --config-path /path/to/envoy-config.yaml
```
htuch: Do we want to be prescriptive about running under `taskset`?

PROFILING.md
```diff
@@ -0,0 +1,78 @@
+# Profiling Envoy
+
+## Install pprof
+
+From https://github.com/google/pprof#building-pprof
+go get -u github.com/google/pprof
+
+## Envoy build
+
+See [building envoy with bazel](https://github.com/envoyproxy/envoy/tree/master/bazel#building-envoy-with-bazel).
```
htuch: Nit: "building Envoy with Bazel"

PROFILING.md
```diff
@@ -0,0 +1,78 @@
...
+Profiling via http, as well as specify a location to dump the profiling data.
+
+``` yaml
+admin:
+  access_log_path: /tmp/admin_access.log
+  profile_path: /tmp/envoy.prof
+  address:
+    socket_address: { address: $server_ip, port_value: 0 }
+Static_resources:
+.. your configuration ..
```
jplevyak: Could you include a full working configuration example?

PROFILING.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+taskset /path/to/nighthawk-repo/bazel-bin/nighthawk_client --concurrency 5 --rps 10000 --duration 30 http://envoy-cluster-host:envoy-cluster-port
+```
+
+Run pprof web UI
+
+```bash
+pprof -http=localhost:8888 /tmp/envoy.prof
+```
+
```
jplevyak: Can you include a example script (e.g. nighthawk/examples ?

oschaaf:
This is ready for another round of feedback

PROFILING.md
```diff
@@ -0,0 +1,100 @@
...
+Nighthawk comes with a [small framework and experimental benchmark suite](/benchmarks/) that
+will write `.prof` files to `/tmp/`:
+
+```bash
+bazel test --cache_test_results=no --compilation_mode=opt --cxxopt=-g --cxxopt=-ggdb3 //benchmarks:*
+```
+
+Note that it is possible to override Nighthawk's Envoy dependency
+to link against a local version, by adding a line to `.bazelrc`:
```
htuch: It's cleaner and more minimally correct just to recommend adding this as a CLI option. Any CLI option can be put in `.bazelrc` at a user's discretion.

PROFILING.md
```diff
@@ -0,0 +1,100 @@
...
+See [building Nighthawk](https://github.com/envoyproxy/nighthawk#nighthawk).
+
+```
+bazel build -c opt //:nighthawk
+```
+
+### Envoy configuration
+
+The important part is that the admin interface needs to be set up to allow one to enable / disable
+Profiling via http, as well as specify a location to dump the profiling data.
```
htuch: Nit: s/Profiling/profiling/

PR #163, opened 2019-10-12
[WIP] Experimental latency low pass filter

oschaaf:
Based on some experimental older code I found, I thought I'd update it to
work with the current state to see if there is interest.

Allows one to filter latency samples before they enter statistics
processing, hopefully improving resolution of the signal by cutting
out noise.

Example usage, dropping samples below 500us:

```
./nighthawk_client -latency-sampling-global-low-pass 500000 ...
```

The filtering is global, it would apply to all histograms.


oschaaf:
Closing this, as proper visualisation tooling will address this better, so the added weight of the code introduced here isn't worth it.

PR #164, opened 2019-10-13
Update Envoy dependency

oschaaf:
Updates Envoy to 94ac19f5e16ad012aa3a77258c0868c2fea2a7f0

Leaves a TODO regarding tools/check_format in `api/client/BUILD`.
Will not pass the format check in CI, but passes unit/integration tests.


oschaaf:
@htuch there's a TODO over at https://github.com/envoyproxy/nighthawk/pull/164/files#diff-04dcaeb167b260b75a61e7f95b910e72R5, somehow `tools/check_format fix` insists on making a breaking change. I suspect that to address this, an upstream change is needed.

api/client/BUILD
```diff
@@ -1,16 +1,14 @@
...
-envoy_package()
-
-load("@envoy_api//bazel:api_build_system.bzl", "api_proto_library")
 load("@com_github_grpc_grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")
 
-api_proto_library(
+# TODO(oschaaf): `tools/check_format fix` insists on adding/changing
+# load("bazel:envoy_build_system.bzl", "envoy_package")
+# That breaks the build.
+# load("@envoy//bazel:envoy_build_system.bzl", "envoy_package")
```
htuch: I think you can configure exclusions, see https://github.com/envoyproxy/envoy/blob/4825facf2451b234af737b5af8c792c31bca5d41/tools/check_format.py#L800. I think `check_format` should ideally have a finer grained notion, of saying "we want to do buildifier, but not the `envoy_package` jazz outside of the Envoy repo.

oschaaf: Excluded at the file-level, added https://github.com/envoyproxy/nighthawk/issues/165 to track optimizing this

oschaaf:
Ready for another look!
/assign htuch

PR #167, opened 2019-10-16
Wire up termination predicates

oschaaf:
Wires up termination predicates as a generic way of stopping execution.

High level change description:
- In the sequencer we no longer wait for outstanding requests to come in after the test duration
  expires. The only way the sequencer will stop is if one of the termination predicates indicates so.
- The benchmark client shutdown has been moved in time after we collect the counters we want to
  output. This is avoids polluting the stats with increments that occur after the execution duration has
  transpired.
- If we stop because of a termination predicate indicating failure, we will reflect that in the exit code of `ProcessImpl`.
- Makes a small change to `output.proto`: we allow both output and `error_detail` to be emitted at the same time. This allows us to output counters when a termination predicate indicated a failure. 
- Adds default termination predicates for failure status codes and connection failures, so executions fail fast by default upon misconfiguration.
This PR is part of https://github.com/envoyproxy/nighthawk/issues/154

Planned followups:
- In a follow-up configuration of the predicates will be added, which -iff set- will replace the stock ones we introduced in this PR (see above).
- Tangentially related, when https://github.com/envoyproxy/nighthawk/pull/137 lands, a follow up to that (and this) will be tracking violations of request-level expectations via counters. These counters can then (optionally) be used in termination predicates to specify the desired thresholds for successful execution. 


test/integration/test_integration_basics.py
```diff
@@ -269,3 +261,16 @@ def test_dotted_output_format(http_test_server_fixture):
   ],
                                                           as_json=False)
   assertIn("global.benchmark_http_client.request_to_response.permilles-500.microseconds", output)
+
+
+# TODO(oschaaf): add percentiles to the gold testing in the C++ output formatter
+# once the fortio formatter has landed (https://github.com/envoyproxy/nighthawk/pull/168)
+def test_cli_output_format(http_test_server_fixture):
```
oschaaf: Note: this test is here to recover some lost coverage.

test/integration/test_integration_basics.py
```diff
@@ -33,7 +31,7 @@ def test_http_h1(http_test_server_fixture):
   assertCounterEqual(counters, "upstream_rq_pending_total", 1)
   assertCounterEqual(counters, "upstream_rq_total", 25)
   assertCounterEqual(counters, "default.total_match_count", 1)
-  assertEqual(len(counters), 14)
+  assertEqual(len(counters), 12)
```
oschaaf: Note: we expect less counters, because we grab the statistics before we shutdown the pools.
This implies that we won't observe some connections being closed, which I think makes sense.

README.md
```diff
@@ -130,9 +130,7 @@ Where:
      Use HTTP/2
 
    --timeout <uint32_t>
-     Timeout period in seconds used for both connection timeout and grace
-     period waiting for lagging responses to come in after the test run is
-     done. Default: 30.
+     Connection timeout period in seconds. Default: 30.
```
oschaaf: Note: We had this concept of a timeout for the test waiting for outstanding requests to wrap up post-execution-duration.
This gets dropped here (the timeout will now be used solely as a connection timeout). This is done in this PR because it doesn't fit well with the termination predicate concept, and I think it also makes for easier reasoning about results. If we want something like this, then maybe we should think about introducing "phases" as a first-class citizen (e.g. warmup, cooldown, etc).

htuch: Sure. But, as a courtesy to reviewers, really prefer that these unrelated things get cleaned up in a prelude PR.

api/client/service.proto
```diff
@@ -30,11 +30,8 @@ message ExecutionRequest {
 }
 
 message ExecutionResponse {
-  oneof status_specifier {
-    option (validate.required) = true;
-    Output output = 1;
-    google.rpc.Status error_detail = 7;
-  }
+  Output output = 1;
```
htuch: FWIW, these breaking changes are going to have to not be allowed as soon as Google or some external is making use of Nighthawk in any CI etc.

source/client/benchmark_client_impl.cc
```diff
@@ -82,8 +82,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
   // exactly as possible here.
   // NOTE: We can't consistently rely on resourceManager()::requests()
   // because that isn't used for h/1 (it is used in tcp and h2 though).
-  if ((max_pending_requests_ == 1 &&
-       (requests_initiated_ - requests_completed_) >= connection_limit_)) {
+  if (max_pending_requests_ == 1 &&
+      ((requests_initiated_ - requests_completed_) >= connection_limit_)) {
```
htuch: Nit: these parens are redundant.

source/client/client.cc
```diff
@@ -50,16 +50,16 @@ bool Main::run() {
   ProcessImpl process(*options_, time_system);
   OutputFormatterFactoryImpl output_formatter_factory;
   OutputCollectorImpl output_collector(time_system, *options_);
-  if (process.run(output_collector)) {
-    auto formatter = output_formatter_factory.create(options_->outputFormat());
-    std::cout << formatter->formatProto(output_collector.toProto());
-    process.shutdown();
+  bool res = process.run(output_collector);
```
htuch: Nit: `const`

source/client/client_worker_impl.cc
```diff
@@ -46,23 +50,28 @@ void ClientWorkerImpl::work() {
...
     if (stat->value() && absl::StartsWith(stat_name, worker_prefix)) {
       thread_local_counter_values_[std::string(absl::StripPrefix(stat_name, worker_prefix))] =
           stat->value();
     }
   }
 }
 
+void ClientWorkerImpl::shutdownThread() {
+  // Terminate will shut down the pool and run the dispatcher, which may increment certain stats
+  // like the number of connections destroyed. We do that here, so we don't pollute the reported
```
htuch: Here vs. where? This is not clear from the comment.

source/client/factories_impl.cc
```diff
@@ -126,5 +129,32 @@ HeaderSourcePtr HeaderSourceFactoryImpl::create() const {
   return std::make_unique<StaticHeaderSourceImpl>(std::move(header));
 }
 
+TerminationPredicateFactoryImpl::TerminationPredicateFactoryImpl(const Options& options)
+    : OptionBasedFactoryImpl(options) {}
+
+TerminationPredicatePtr
+TerminationPredicateFactoryImpl::create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
+                                        const Envoy::MonotonicTime start) const {
+  // By default, we don't tolerate status code errors & connection-level errors.
```
htuch: Is this new behavior? There already are a bunch of folks who are getting used to Nighthawk. If we are going to start changing things like this, we probably need a version history file that tracks this kind of breaking change.

source/common/sequencer_impl.cc
```diff
@@ -84,36 +82,29 @@ void SequencerImpl::run(bool from_periodic_timer) {
...
     while (rate_limiter_->tryAcquireOne()) {
       // The rate limiter says it's OK to proceed and call the target. Let's see if the target is OK
       // with that as well.
       const bool target_could_start = target_([this, now](bool, bool) {
         const auto dur = time_source_.monotonicTime() - now;
         latency_statistic_->addValue(dur.count());
         targets_completed_++;
-        // Immediately schedule us to check again, as chances are we can get on with the next task.
-        spin_timer_->enableTimer(0ms);
+        if (spin_timer_ != nullptr) {
```
jplevyak: comment why might this be null and why is it safe to not schedule?

test/client_test.cc
```diff
@@ -39,7 +39,7 @@ TEST_F(ClientTest, AutoConcurrencyRun) {
   argv.push_back("error");
   argv.push_back("http://localhost:63657/");
   Main program(argv.size(), argv.data());
-  EXPECT_TRUE(program.run());
+  EXPECT_FALSE(program.run());
```
jplevyak: Why false?

test/process_test.cc
```diff
@@ -48,10 +48,10 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProcessTest,
                          Envoy::TestUtility::ipTestParamsToString);
 
 TEST_P(ProcessTest, TwoProcessInSequence) {
-  runProcess(RunExpectation::EXPECT_SUCCESS);
+  runProcess(RunExpectation::EXPECT_FAILURE);
```
jplevyak: Why failure?

test/service_test.cc
```diff
@@ -120,7 +121,7 @@ TEST_P(ServiceTest, Basic) {
   r->Write(request_, {});
   r->WritesDone();
   EXPECT_TRUE(r->Read(&response_));
-  ASSERT_FALSE(response_.has_error_detail());
+  ASSERT_TRUE(response_.has_error_detail());
```
jplevyak: Check the error detail?

README.md
oschaaf: It is actually forwarded to `mutable_connect_timeout()` on the cluster. Updated.

api/client/service.proto
oschaaf: Ack; AFAICT this didn't break compilation, but it's a behavioral change nonetheless.

include/nighthawk/common/worker.h
```diff
@@ -30,6 +30,11 @@ class Worker {
    * and mandatory.
    */
   virtual void shutdown() PURE;
+
+  /*
+  Called just before exitting the thread, allows for the thread to clean up.
```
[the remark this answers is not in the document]
oschaaf: Added some comments, let me know if that clarifies it sufficiently.

source/client/benchmark_client_impl.cc
oschaaf: Backed out this change, it was included accidentally.

source/client/client.cc
oschaaf: done

source/client/client_worker_impl.cc
oschaaf: Modified comment and moved it above, let me know if that looks better.

source/client/factories_impl.cc
oschaaf: Yes this is new behavior. We take responsibility for determining success/failure based on the termination predicates.
I think these defaults represent an improvement, but backing it out is easy if it causes more trouble then it is worth.

source/client/factories_impl.cc
```diff
@@ -126,5 +129,32 @@ HeaderSourcePtr HeaderSourceFactoryImpl::create() const {
...
+      std::make_unique<StatsCounterAbsoluteThresholdTerminationPredicateImpl>(
+          scope.counter("benchmark.http_4xx"), 0, TerminationPredicate::Status::FAIL);
+  TerminationPredicatePtr code_5xx_predicate =
+      std::make_unique<StatsCounterAbsoluteThresholdTerminationPredicateImpl>(
+          scope.counter("benchmark.http_5xx"), 0, TerminationPredicate::Status::FAIL);
+  TerminationPredicatePtr pool_failure_predicate =
+      std::make_unique<StatsCounterAbsoluteThresholdTerminationPredicateImpl>(
+          scope.counter("benchmark.pool_connection_failure"), 0,
+          TerminationPredicate::Status::FAIL);
+  code_5xx_predicate->link(std::move(pool_failure_predicate));
```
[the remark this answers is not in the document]
oschaaf: Done, looks way nicer, thanks.

source/common/sequencer_impl.cc
oschaaf: Done

test/client_test.cc
oschaaf: Ah, yes, this surely deserves some explanation. 

So the c++ tests talk to a non-existing server. We've gone through some trouble with the c++ integration test server because of issues with singleton runtimes in the past.

Before this change, NH didn't care about that there was no server listening, but now that we have these default termination predicates it marks the execution as a failure, because connection failures are observed. That's why this line has `EXPECT_FALSE`.

In https://github.com/envoyproxy/nighthawk/pull/137 we wrap the grpc service in the python integration tests.
That takes us one step closer towards porting these tests to that, which would be helpful because then we can easily spin up test servers again. 

(this also applies to some of your comments below)

test/service_test.cc
oschaaf: Sure; I will address this in a follow up

oschaaf:
@htuch @jplevyak I'll split out some of the changes here into separate PR's. 
The first one is https://github.com/envoyproxy/nighthawk/pull/171

I propose we freeze review here, and I'll update this PR to include the prerequisites as they land.
Once that is done, we can finalize here. WDYT?

oschaaf:
@htuch @jplevyak the following PR's are preludes to this one:
https://github.com/envoyproxy/nighthawk/pull/175
https://github.com/envoyproxy/nighthawk/pull/174
https://github.com/envoyproxy/nighthawk/pull/173

test/client_test.cc
jplevyak: OK.  We probably want to get c++ integration tests working in general, but I know that difficult.  Do we have a core Envoy issue for this?  Also could we add a comment?

test/run_nighthawk_bazel_coverage.sh
```diff
@@ -57,7 +57,7 @@ then
   COVERAGE_VALUE=$(llvm-cov export "${COVERAGE_BINARY}" -instr-profile="${COVERAGE_DATA}" \
     -ignore-filename-regex="${COVERAGE_IGNORE_REGEX}" -summary-only | \
     python3 -c "import sys, json; print(json.load(sys.stdin)['data'][0]['totals']['lines']['percent'])")
-  COVERAGE_THRESHOLD=98.5
+  COVERAGE_THRESHOLD=98.4
```
jplevyak: That's one way to fix the coverage :)

oschaaf: Heh :-) Yeah, so actually I think I did this because the dip coverage % is caused by a reduction in the number of total code lines. But let me double check to make sure, and follow up here.

test/client_test.cc
oschaaf: We played whack-a-mole for a while as the issues would jump around as dependencies on the singleton where moved/introduced in the Envoy code-base. So we went through various hoops to work around (including using fork() to dish out a new process to run the actual tests), each attempt introduced new kinds of trouble, and finally we gave up. There is no core Envoy issue for this that I know of.

I'll add a comment here to make note of why we expect these failures, makes sense

test/service_test.cc
oschaaf: Done in https://github.com/envoyproxy/nighthawk/pull/167/commits/b7b40efce024e4f19248617e8c7dac897713b1fb, plus filed https://github.com/envoyproxy/nighthawk/issues/181 for tracking an improvement over that.

oschaaf:
The prerequisites we split out  from this [1]  have all been merged, and I followed up on comments here. If everyone is happy with moving this forward based on the current changes: this is ready for another look. (I think the remaining changes are pretty much on target).

/assign htuch
/assign jplevyak

[1] https://github.com/envoyproxy/nighthawk/pull/167#issuecomment-545312450

include/nighthawk/common/termination_predicate.h
```diff
@@ -29,7 +29,7 @@ class TerminationPredicate {
    *
    * @param child the child predicate to link.
    */
-  virtual void link(TerminationPredicatePtr&& child) PURE;
+  virtual TerminationPredicate& link(TerminationPredicatePtr&& child) PURE;
```
htuch: Please document `@return` in the comment above, not sure why this was needed, but the comment can explain :)

source/client/factories_impl.cc
htuch: I think it's fine to do this, but can you add a file similar to https://github.com/envoyproxy/envoy/blob/master/docs/root/intro/version_history.rst (it can just be a simple Markdown for now) that tracks changes like this? Ideally we update it as we go, and then we have version release notes.

source/client/service_impl.cc
```diff
@@ -37,12 +37,14 @@ void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionReque
           nighthawk::client::Verbosity::VerbosityOptions_Name(options->verbosity())),
       "[%T.%f][%t][%L] %v", log_lock_);
   OutputCollectorImpl output_collector(time_system_, *options);
-  if (process.run(output_collector)) {
-    response.clear_error_detail();
-    *(response.mutable_output()) = output_collector.toProto();
-  } else {
+  bool ok = process.run(output_collector);
```
htuch: Nit: const

test/client_worker_test.cc
```diff
@@ -34,17 +34,23 @@ class ClientWorkerTest : public Test {
...
+    EXPECT_CALL(sequencer_factory_, create(_, _, _, _, _, _))
         .Times(1)
         .WillOnce(Return(ByMove(std::unique_ptr<Sequencer>(sequencer_))));
 
     EXPECT_CALL(header_generator_factory_, create())
         .Times(1)
         .WillOnce(Return(ByMove(std::unique_ptr<HeaderSource>(header_generator_))));
+
+    EXPECT_CALL(termination_predicate_factory_, create(_, _, _))
+        .Times(1)
```
htuch: Nit: `Times(1)` is redundant, it's implied by just the basic `EXPECT`.

test/benchmark_http_client_test.cc
```diff
@@ -31,7 +31,7 @@ class BenchmarkClientHttpTest : public Test {
   BenchmarkClientHttpTest()
       : api_(Envoy::Api::createApiForTest()), dispatcher_(api_->allocateDispatcher()),
         cluster_manager_(std::make_unique<Envoy::Upstream::MockClusterManager>()),
-        cluster_info_(std::make_unique<Envoy::Upstream::MockClusterInfo>()),
+        cluster_info_(std::make_unique<NiceMock<Envoy::Upstream::MockClusterInfo>>()),
```
htuch: I'm surprised this change is related.

test/client_test.cc
```diff
@@ -60,7 +63,7 @@ TEST_F(ClientTest, TracingRun) {
   argv.push_back("--trace");
   argv.push_back("zipkin://localhost:9411/api/v1/spans");
   Main program(argv.size(), argv.data());
-  EXPECT_TRUE(program.run());
+  EXPECT_FALSE(program.run());
```
htuch: That seems a dramatic change :)

include/nighthawk/common/termination_predicate.h
oschaaf: Done

source/client/factories_impl.cc
oschaaf: Sure, updated that file for things related to this PR, I'll update it some more in a follow up to include earlier work.

source/client/service_impl.cc

test/benchmark_http_client_test.cc
oschaaf: And rightly so, nice catch. Backed this out, the `NiceMock` wrapper is no longer needed.

test/client_test.cc
oschaaf: Yes .. also see https://github.com/envoyproxy/nighthawk/pull/167/files/bcc8e4fee772dbbc967d976145e2a9fbdd3465d0#r337765347
I have added comments about this (also in this file), fortunately the python integration tests cover this with a happier flow (a running server to talk to).

test/client_worker_test.cc
oschaaf: Removed

docs/root/version_history.md
```diff
@@ -1,6 +1,19 @@
...
+
+### Notable (breaking) changes
+
+- In `service.proto` a change was made to allow both `Output` and `error_detail` to co-exist at the same time.
+- Both `nighthawk_client` and `nighthawk_service` will indicate execution failure (respectively through exit code or grpc reply) when connection errors and/or status code errors are observed by default.
+
+
+### Changelist
+
+- Introducing termination predicates (https://github.com/envoyproxy/nighthawk/pull/167)
```
htuch: Also would be good to describe the behavioral change.

oschaaf: Well, I did create a section with breaking changes which does so, see https://github.com/envoyproxy/nighthawk/pull/167/files#diff-381db20629d045ee9d5e7567c36a451cR10
Given that, WDYT?

PR #168, opened 2019-10-18
Add support for Fortio output format

nareddyt:
Description: Add support for Fortio output format

Testing: Added a unit test the compare the output to a golden file

Docs: Added sample data/reports, added docs to README.md


nareddyt:
Current output:

```
{
 "RetCodes": {
  "200": "60"
 },
 "Labels": "A random label",
 "StartTime": "2019-10-16T00:30:12Z",
 "RequestedQPS": 1,
 "RequestedDuration": "5s",
 "ActualQPS": 0,
 "ActualDuration": 0,
 "NumThreads": 0,
 "Version": "",
 "Exactly": 0,
 "URL": "http://127.0.0.1:10000/",
 "SocketCount": 1
}
```

Just need to figure out the distribution part and understand the remaining fields that are zero-valued...

nareddyt:
This mapping seems to work. Here is an example output:

image

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,119 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+
+  // This displays the endpoint URL in the UI
+  fortio_output.mutable_url()->set_value(output.options().uri().value());
+
+  // Get the result that represents all workers (global)
+  const auto& nh_global_result = this->getGlobalResult(output);
+
+  // Fill in the number of successful responses.
+  // Fortio-ui only reads the 200 OK field, other fields are only informational.
+  const auto& nh_2xx_counter = this->getCounterByName(nh_global_result, "benchmark.http_2xx");
```
oschaaf: I think it is possible that this counter can not be found in NH's output. For example, when all responses returned 404. (0 statistics are not emitted). I'm not sure how to translate that into the domain of Fortio's json, and perhaps the exception raised in that case is the right thing to do (though mentioning the input being untranslatable might be helpful).

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,119 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+  nighthawk::client::FortioResult fortio_output;
+
+  // TODO(nareddyt): Not needed but nice to have, displays in the UI
+  fortio_output.mutable_labels()->set_value("A random label");
+
+  fortio_output.mutable_starttime()->set_seconds(output.timestamp().seconds());
+  fortio_output.mutable_requestedqps()->set_value(output.options().requests_per_second().value());
+  fortio_output.mutable_requestedduration()->set_seconds(output.options().duration().seconds());
+
+  // FIXME: Set the actual to the same as the requested since NH does not expose these fields
```
oschaaf: You may be able to compute the actual qps by dividing the `upstream_rq_total` counter by the specified duration.

source/client/output_formatter_impl.cc
nareddyt: I think we can handle this by just setting it to 0.

source/client/output_formatter_impl.cc
nareddyt: Thanks!

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,123 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+  fortio_output.mutable_requestedqps()->set_value(output.options().requests_per_second().value());
+  fortio_output.mutable_url()->set_value(output.options().uri().value());
+
+  // Actual and requested durations are the same
+  const auto& nh_duration = output.options().duration().seconds();
+  fortio_output.mutable_requestedduration()->set_seconds(nh_duration);
+  fortio_output.mutable_actualduration()->set_value(nh_duration);
+
+  // This displays as "connections" in the UI, not threads
+  fortio_output.mutable_numthreads()->set_value(output.options().connections().value());
```
oschaaf: This is the requested setting, and not the effective setting, right? If I remember correctly, in Fortio this corresponds to the number of goroutines, right? Thinking about it, what we want here might be a bit more tricky; we must take into account Nighthawk's `--concurrency` flag, which acts as a multiplier. Furthermore, we have `--connections` when using h1, but with h2 that is ignored today as Envoy's h2 pool mostly uses a single connection. So we'd probably want the number of concurrent streams (governed by `--max-active-requests`) here, multiplied by the number specified in concurrency.

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,123 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+  // TODO(nareddyt): Not needed but nice to have, displays in the UI
+  fortio_output.mutable_labels()->set_value("");
+  fortio_output.mutable_starttime()->set_seconds(output.timestamp().seconds());
+  fortio_output.mutable_requestedqps()->set_value(output.options().requests_per_second().value());
+  fortio_output.mutable_url()->set_value(output.options().uri().value());
+
+  // Actual and requested durations are the same
+  const auto& nh_duration = output.options().duration().seconds();
+  fortio_output.mutable_requestedduration()->set_seconds(nh_duration);
+  fortio_output.mutable_actualduration()->set_value(nh_duration);
```
oschaaf: Could you file an issue for this? I feel we need this one, I can take care of it.

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,123 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
 
+const nighthawk::client::Result&
+FortioOutputFormatterImpl::getGlobalResult(const nighthawk::client::Output& output) const {
+  for (const auto& nh_result : output.results()) {
+    if (nh_result.name() == "global") {
+      return nh_result;
+    }
+  }
+
+  throw NighthawkException("Nighthawk output was malformed, contains no 'global' results.");
```
oschaaf: Perhaps consider covering this throw and the ones below in tests?
As a suggestion, there are the python integration tests which might be useful for that, you can perform an end-to-end test with bad inputs, this would also ensure the throw are handled as intended:
https://github.com/envoyproxy/nighthawk/blob/master/test/integration/test_integration_basics.py

nareddyt:
Thanks for the comments, will address soon. In the mean time, could you help me determine why the coverage CI job failed?

oschaaf:
> Thanks for the comments, will address soon. In the mean time, could you help me determine why the coverage CI job failed?

Coverage dipped below the threshold currently in place. You can view the report via
the “artifacts“ tab in the ci report. e.g there are a couple of uncovered lines here that might be actionable:

https://5049-180498819-gh.circle-artifacts.com/0/coverage/coverage/proc/self/cwd/source/client/output_formatter_impl.cc.html

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,123 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+    const auto& nh_percentile = nh_stat.percentiles().at(i);
+
+    // fortio_percent = 100 * nh_percentile
+    fortio_data_entry.mutable_percent()->set_value(nh_percentile.percentile() * 100);
+
+    // fortio_count = nh_count - prev_fortio_count
+    fortio_data_entry.mutable_count()->set_value(nh_percentile.count() - prev_fortio_count);
+
+    // fortio_end = nh_duration (need to convert formats)
+    const double nh_duration_in_double = nh_percentile.duration().nanos() / 1e9;
```
oschaaf: I'm not sure this will work with larger durations, you might also need to consider the `.seconds()` part.

oschaaf: Yeah, see Envoy::Protobuf::util::TimeUtil::DurationToNanoseconds. which may be useful here.

nareddyt:
Comments make sense, I need to spend some more time understanding how the tests work.

Sorry for the delay, got caught up in some envoy PRs. I'm travelling for the remainder of the week, so hopefully I can work on this next Monday!

source/client/output_formatter_impl.cc
nareddyt: Ok, I see why we shouldn't be using `--connections`.

However, I don't know if using `--concurrency` will work, since it's a string value that can be set to "auto". How can I determine the concurrency for auto values?

nareddyt: I could get the number of CPU cores like the source code does, but its possible the output transform might run on a different machine with a different number of CPU cores, so it wouldn't be accurate.

Can we file a bug for this and address it in another PR, since this isn't a core value that is needed?

source/client/output_formatter_impl.cc
nareddyt: Done, #182

source/client/output_formatter_impl.cc
nareddyt: I think I see, could you confirm I did it correctly now?

source/client/output_formatter_impl.cc
nareddyt: I think it's pretty hard to simulate all 3 errors with invalid inputs to the fortio client. I think most invalid inputs will just be caught by the first `getGlobalResult` function (?).

I'll look into adding unit tests instead.

nareddyt: Done!

oschaaf: Thanks, looks great. 
Double checking; one gotcha is that the some consumers don't currently expect the output formatters to throw [1]. I feel that should be addressed outside of this PR, I created https://github.com/envoyproxy/nighthawk/issues/184 for tracking.

source/client/output_formatter_impl.cc
oschaaf: Sounds good to me. Maybe also mention that a prerequisite to that is that we offer the effective concurrency in NH's proto, it sounds like we need something like that.

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,124 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+    // fortio_percent = 100 * nh_percentile
+    fortio_data_entry.mutable_percent()->set_value(nh_percentile.percentile() * 100);
+
+    // fortio_count = nh_count - prev_fortio_count
+    fortio_data_entry.mutable_count()->set_value(nh_percentile.count() - prev_fortio_count);
+
+    // fortio_end = nh_duration (need to convert formats)
+    const double nh_percentile_duration_sec =
+        Envoy::Protobuf::util::TimeUtil::DurationToNanoseconds(nh_percentile.duration()) / 1e9;
+    fortio_data_entry.mutable_end()->set_value(nh_percentile_duration_sec);
```
oschaaf: Sorry; one more nit: looking at https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.time_util I suspect that `DurationToSeconds()` may be more convenient here for the precision we need

samples/fortio_data/large.json
```diff
@@ -0,0 +1,456 @@
+{
```
oschaaf: Is this file used anywhere?

oschaaf:
Oh, just noticed coverage is unhappy; there may be an easy fix over at https://github.com/envoyproxy/nighthawk/blob/master/test/factories_test.cc#L127 by adding nighthawk::client::OutputFormat::FORTIO to the list of formats being tested.

source/client/output_formatter_impl.cc
nareddyt: #186

source/client/output_formatter_impl.cc
nareddyt: Looking at the documentation, it seems like `DurationToSeconds()` only returns an `int64`. So I think it would truncate all the decimals, which is definitely what we want. I can test it.

samples/fortio_data/large.json
nareddyt: This is the data that generates the sample pngs in `samples/fortio_reports/`. I kept this data here so that any users of NH could compare the PNG with the underlying data. It also allows them to test running Fortio-UI on the data to generate interactive graphs, as documented in the README.md

oschaaf: Ah, makes sense, sgtm!

source/client/output_formatter_impl.cc
nareddyt: Yeah, it truncates to 0.

```
-    \"Start\": 0.055568383,
-    \"End\": 0.055568383,
+    \"Start\": 0,
+    \"End\": 0,
```

oschaaf: Sorry about the noise, looks great as-is then.

nareddyt:
Huh, it looks like `clang_tidy` failed. I can't figure out the error, any ideas?

oschaaf:
> Huh, it looks like `clang_tidy` failed. I can't figure out the error, any ideas?

I think that's a flake. I've seen it earlier, I think it is related to the CI machine running out of memory.
Just restarted CI. Just filed https://github.com/envoyproxy/nighthawk/issues/187 to track fixing the clang-tidy flakes.

api/client/transform/fortio.proto
```diff
@@ -0,0 +1,54 @@
...
+
+  google.protobuf.Timestamp StartTime = 2;
+
+  google.protobuf.UInt32Value RequestedQPS = 3;
+
+  google.protobuf.Duration RequestedDuration = 4;
+
+  google.protobuf.DoubleValue ActualQPS = 5;
+
+  google.protobuf.DoubleValue ActualDuration = 6;
```
htuch: Only wrap types that are optional.

api/client/transform/fortio.proto
```diff
@@ -0,0 +1,54 @@
...
+package nighthawk.client;
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+
+message FortioResult {
+  // Disable validate because it doesn't work well with fields that are not all lowercase.
+  // The field cases must match the final output JSON that fortio-ui expects.
```
htuch: Please put a top-level comment explaining the proto3-JSON trick that is being carried out here.

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,125 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+
+const nighthawk::client::Statistic& FortioOutputFormatterImpl::getRequestResponseStatistic(
+    const nighthawk::client::Result& result) const {
+  for (auto const& nh_stat : result.statistics()) {
+    if (nh_stat.id() == "benchmark_http_client.request_to_response") {
+      return nh_stat;
+    }
+  }
+
+  throw NighthawkException("Nighthawk result was malformed, contains no "
```
htuch: This seems an anti-pattern, passing around expected output results as string map values.. should we have more structure here?

source/client/output_formatter_impl.cc
```diff
@@ -138,5 +144,125 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
...
+    }
+  }
+
+  throw NighthawkException("Nighthawk output was malformed, contains no 'global' results.");
+}
+
+const nighthawk::client::Counter&
+FortioOutputFormatterImpl::getCounterByName(const nighthawk::client::Result& result,
+                                            absl::string_view counter_name) const {
+  for (const auto& nh_counter : result.counters()) {
```
htuch: This seems potential O(n^2), why don't we have a map of name -> counter provided?

nareddyt: I filed #169 a few days ago for this

htuch: Fair enough, resolved.

source/client/output_formatter_impl.cc
nareddyt: @oschaaf perhaps you could look into this? I think this would require changes outside of my PR.

api/client/transform/fortio.proto
nareddyt: Done.

api/client/transform/fortio.proto
nareddyt: Done

nareddyt:
Ready for review.

Fairly certain that all the code is covered based on @oschaaf's comments. I can't see the coverage artifact on CI for some reason now, it keeps giving me a 404.

oschaaf:
A fix/format for the CLI docs was recently introduced for `README.md`.
Can you run this command? Hopefully that fixes the CI failure:
```bash
tools/update_cli_readme_documentation.sh --mode fix
```

source/client/output_formatter_impl.cc
oschaaf: I filed https://github.com/envoyproxy/nighthawk/issues/195 to discuss.

oschaaf: (and track)

oschaaf:
The current CI failure is a flake, and my bad. Apparently `test_https_h2` isn't stable with `asan`

PR #170, opened 2019-10-22
Update Envoy dependency

oschaaf:
- Switch to Envoy 94ac19f5e16ad012aa3a77258c0868c2fea2a7f0
- Sync CI build image & .bazelrc
- Tweak setup_clang_toolchain() in do_ci.sh
- Updates for making `clang-tidy` & `clang-format` work
- Switches CI to use clang-9 (previously clang-8)


oschaaf:
This needed some more tweaking because the `clang-tidy` & `clang-format` CI jobs broke. That is fixed now, so this is ready for another look.

/assign htuch

tools/gen_compilation_database.py
```diff
@@ -1,18 +1,47 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
htuch: I'm kind of unhappy that we have to manually sync these files. Do you think we could move the common CI and dev tooling support files to an additional repository, e.g. https://github.com/envoyproxy/envoy-build-tools?

This is non-actionable for this review, but I'm hoping between you, me and @lizan we can figure out a good story here, as I would like CI and dev tools to not create undue burden on adding new envoyproxy org repos.

htuch:
@oschaaf LGTM, thanks! Can you merge master to resolve conflict?

PR #174, opened 2019-10-23
Postpone and enhance connection pool shutdown

oschaaf:
Delays termination of the benchmark client, so that the associated
connection pool shutdown will not influence reported counters.
Also makes the pool shut down cleanly.

Split out from https://github.com/envoyproxy/nighthawk/pull/167


include/nighthawk/common/worker.h
```diff
@@ -30,6 +30,12 @@ class Worker {
    * and mandatory.
    */
   virtual void shutdown() PURE;
+
+  /*
+  Called on-thread after its designated task finishes. Last chance to clean up while the associated
```
htuch: Nit: `* ` formatting at the start of each line.

source/client/client_worker_impl.cc
```diff
@@ -61,8 +60,13 @@ void ClientWorkerImpl::work() {
           stat->value();
     }
   }
+  // Note that benchmark_client_ is not terminated here, but in shutdownThread() below. This is to
+  // prevent it from influencing counters: the main thread still needs to be able to read the
```
htuch: Nit: "to prevent the shutdown artifacts from influencing the test result counters."

include/nighthawk/common/worker.h
oschaaf: Done

source/client/client_worker_impl.cc
oschaaf: Done

oschaaf:
Feedback addressed, thanks!

/assign htuch

PR #175, opened 2019-10-23
Sequencer enhancements & cleanup

oschaaf:
- Guard against null dereference in the completion callback.
  In the current state we won't hit this case in practice,
  but in the future we will once the termination predicate
  feature lands.
- Remove cancelled flag. I think this was accidentally left
  in because of an incomplete revert in the past.

Split out from https://github.com/envoyproxy/nighthawk/pull/167


source/common/sequencer_impl.cc
```diff
@@ -112,8 +112,14 @@ void SequencerImpl::run(bool from_periodic_timer) {
         const auto dur = time_source_.monotonicTime() - now;
         latency_statistic_->addValue(dur.count());
         targets_completed_++;
-        // Immediately schedule us to check again, as chances are we can get on with the next task.
-        spin_timer_->enableTimer(0ms);
+        // Callbacks may fire after stop(), as during teardown the dispatcher may run. (Concretely
+        // when BenchmarkClient::terminate() is called, and the connection pool shuts down). Hence
```
htuch: Should we have an additional drain phase before termination, i.e. to empty out the dispatcher? It seems it would be easier to reason about this than arbitrary destructed state during the callback.

oschaaf: I amended this to check for `this->running_` here instead? We only need to worry about things like this if `stop()` has been called, and that flag explicitly tracks that. Also attempted to improve the comment. WDYT?

Regarding a drain phase: imho ideally we'd model phases explicitly, and then define a (default) pre-termination phase via termination predicates that ensure we're bounded in time and exit with a failure code when we don't see the relevant gauges drop to 0 within that duration.

PR #176, opened 2019-10-23
Optionize termination predicates

oschaaf:
Allows configuration of termination & failure predicates.

Sample CLI usage:

```
# Stop after 10 requests, indicating success
--termination-predicate "benchmark.http_2xx: 10"
# Fail after a single connection failure
--failure-predicate "upstream_cx_connect_fail: 0"
```

Adds tracking and publishing of the effective execution duration. As we are now able
to terminate before the specified execution duration.

Note: renames the `worker` tag to `cluster` so we share the same counter namespace. This allows the predicates to query the involved counters without going through hoops. As this is change does not propagate to the output, this is for internal handling only, and as such it is not a breaking change.

Together with https://github.com/envoyproxy/nighthawk/pull/167, fixes https://github.com/envoyproxy/nighthawk/issues/154


oschaaf:
@envoyproxy/nighthawk-reviewers I merged master here, and made a few final tweaks, now that the preludes have landed. 
Ready for review!

README.md
```diff
@@ -65,6 +67,15 @@ bazel-bin/nighthawk_client  [--trace <uri format>]
 
 Where:
 
+--failure-predicate <<string, uint32_t>>  (accepted multiple times)
+Failure predicate. Allows specifying a counter name plus threshold
```
htuch: Do thresholds cover all cases? I'm curious if we might want to consider using something more powerful, like CEL (already in the Envoy dependency tree due to its use in RBAC), or some other custom expression format.

oschaaf: I think you are right in that it is not hard to come up with exit conditions that would be useful, but can't be expressed with just the chaining of thresholds for absolute counter values we accomplish here. So I was hoping this current PR state embodies a reasonable start, by capturing some necessities with a simple approach. Next to being able to express more complex logical conditions, arguably more input to the evaluating those could be useful: latency, delta's of counters / gauges, maybe even environment variables for simple signalling from tested application code.

Glossing over the CEL / RBAC implementation you point out, I think that looks like a great opportunity to enhance this. If we'd want move towards that, we can probably do so without breaking existing setups, by implementing (a) new type(s) of `TerminationPredicate`?

README.md
```diff
@@ -43,7 +43,9 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--trace <uri format>]
+bazel-bin/nighthawk_client  [--failure-predicate <<string, uint32_t>>]
+... [--termination-predicate <<string,
+uint32_t>>] ... [--trace <uri format>]
```
eric846: If we do intend to support latency, would it be better to establish the proto with uint64 since 5 billion nanoseconds would overflow uint32?

README.md
```diff
@@ -65,6 +67,15 @@ bazel-bin/nighthawk_client  [--trace <uri format>]
...
 Where:
 
+--failure-predicate <<string, uint32_t>>  (accepted multiple times)
+Failure predicate. Allows specifying a counter name plus threshold
+value for failing execution. Defaults to not tolerating error status
+codes and connection errors.
+
+--termination-predicate <<string, uint32_t>>  (accepted multiple times)
+Termination predicate. Allows specifying a counter name plus threshold
+value for terminating execution.
```
eric846: I wasn't familiar with the `<< >>` notation. What about `string:uint32_t`? 

Also could you add an example (`benchmark.http_4xx:50`) for both termination and failure predicate? I would have guessed `client.benchmark.http_4xx=50`.

README.md
oschaaf: Yes, I think this makes a lot of sense, will follow up on this

README.md
oschaaf: This notation originates from TCLAP .. (so we generate this fragment pre-commit via running `nighthawk_client --help`). Given that, wdyt? It would be possible to add a step in the format/fix step for this to massage it the way we want it to be, but I'm not sure it's worth it? (And then we'd still have the actual `--help` output using this notation).

As for an example, sure, I will add that. Its also easy to replace `:` with `=` at this stage, I agree that would be more intuitive.

eric846: I see, if a tool is interpreting `<<>>` and the `:` isn't built in to `<<>>`, we should definitely keep `<<>>`.

README.md
oschaaf: Redefined as `uint64` in https://github.com/envoyproxy/nighthawk/pull/176/commits/fadc55a539127813bf1532d87aef2d8973c252ba

thanks for catching this!

oschaaf:
Feedback addressed!
/assign @eric846

eric846:
@htuch @jplevyak this PR is ready for review

README.md
htuch: Let's ship with what you have. The user community is small enough that I think we can cope with evolution of this.

Issue #25, opened 2019-04-23
Add support for open-loop latency measurement


oschaaf:
Currently Nighthawk is capable of doing closed-loop testing, which means that when configured resource limits are met (e.g. max connections, max streams), no new requests will be issued, even when that means not reaching the requested request pacing configuration (this wil show up in the output as a time-spend-blocked histogram).
In real life, clients will not wait like that, and supporting open-loop testing will help measuring latencies under these circumstances.

An important part of this feature is that load generation should auto-terminate upon detecting a certain amount of in-flight requests, and maybe when it detects certain resource shortages (like running out of file descriptors).

oschaaf:
A lot of the prerequisites to this will land in #88 but to close this out it would be great to to add time-spend tracking closed-loop mode. In terms of implementation [that tracking should probably go here](https://github.com/envoyproxy/nighthawk/pull/88/files#diff-1d2696707301c1f4e20ca097e73e5891R199).

mum4k:
@oschaaf can you help me by expanding on what is missing here? I was under the impression that we have all we need for the open loop mode. I see the comment about "time-spend tracking closed-loop mode", but am not clear on what it means.

oschaaf:
This was fixed via #177. The line you refer to misses the word ‘blocking’: We track time spend waiting on the server while sending one or more new requests is due. We named that blocking. 

Closing this as there’s nothing remaining to be done here.

PR #177, opened 2019-10-24
Open loop testing mode

oschaaf:
Adds `--open-loop`, which will make the benchmark client refrain from applying backpressure. Instead, stream creation is unconditional and will be handled by the connection pool.
In practice this means we'll not be measuring time spend blocking,
instead we'll see overflow counters increases.

Part of https://github.com/envoyproxy/nighthawk/issues/25
Prerequisite: https://github.com/envoyproxy/nighthawk/pull/178


source/client/benchmark_client_impl.cc
```diff
@@ -69,7 +71,9 @@ StatisticPtrMap BenchmarkClientHttpImpl::statistics() const {
 
 bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completion_callback) {
   const uint64_t max_in_flight =
-      max_pending_requests_ + (use_h2_ ? max_active_requests_ : connection_limit_);
+      provide_resource_backpressure_
+          ? max_pending_requests_ + (use_h2_ ? max_active_requests_ : connection_limit_)
+          : UINT64_MAX;
```
eric846: What about:

```
if (provide_resource_backpressure_) {
  const uint64_t max_in_flight =
      max_pending_requests_ + (use_h2_ ? max_active_requests_ : connection_limit_);

  if (requests_initiated_ - requests_completed_ >= max_in_flight) {
    // When we allow client-side queueing, we want to have a sense of time spend waiting on that
    // queue. So we return false here to indicate we couldn't initiate a new request.
    return false;
  }
}
```

test/integration/test_integration_basics.py
```diff
@@ -109,6 +114,30 @@ def test_http_h2_mini_stress_test_without_client_side_queueing(http_test_server_
   assertNotIn("upstream_rq_pending_overflow", counters)
 
 
+def test_http_h1_mini_stress_test_open_loop(http_test_server_fixture):
+  """
+  H1 open loop stress test. We expect higher pending and overflow counts 
+  """
+  counters = mini_stress_test(http_test_server_fixture, [
+      http_test_server_fixture.getTestServerRootUri(), "--rps", "2500", "--max-pending-requests",
+      "1", "--duration 10", "--open-loop", "--max-active-requests", "1", "--connections", "1"
```
eric846: Nit: I think you split all the others yesterday to `"--duration", "10"`

source/client/benchmark_client_impl.cc
oschaaf: Much nicer indeed -- I'm sold :-) Done in 0fcba50e4bbd0a645251ec81a9e377562aa433e6

test/integration/test_integration_basics.py
oschaaf: Fixed in 0fcba50e4bbd0a645251ec81a9e377562aa433e6

oschaaf:
comments addressed!
/assign @eric846

README.md
```diff
@@ -66,6 +66,10 @@ format>
 
 Where:
 
+--open-loop
+Enable open loop mode. When enabled, the benchmark client will not
+provide backpressurewhen resource limits are hit.
```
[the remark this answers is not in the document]
oschaaf: Amended

PR #178, opened 2019-10-24
Closed-loop fixes & max-active-request default

oschaaf:
- A fix for http/2: consider `--max-active-requests` instead of `--connections`.
- Change the default for `--max-active-requests` to 1 to be consistent with `--connections`
- Test coverage for connection management
- Don't specialize on `--max-pending-requests == 1` as that makes the pending/overflow counts harder to interpret, as it's unclear who's gating requests. Instead, explicitly add a separate `--open-loop` option (see below).
- Allow `0` as an option value for `--max-pending-requests`.

A prelude to https://github.com/envoyproxy/nighthawk/pull/177

(That adds an `--open-loop` option, which will make the benchmark client ask the connection pool to issue requests without checking configured resource limits itself, trading measurement of time spend being blocked for counter increments upon being resource-bound).


README.md
```diff
@@ -77,7 +77,8 @@ Where:
      Max requests per connection (default: 4294937295).
 
    --max-active-requests <uint32_t>
-     Max active requests (default: 4294937295).
+     Max active requests (default: 1). H2-only, the number of concurrent
```
htuch: Do you think this might distort common case results? I think the principle that should govern defaults is "would someone using NH out-of-the-box get reasonable results if performing common case benchmarking?". From that perspective, I'd be interested in hearing what you reckon.

oschaaf: (pushed a change to implement defaulting to `100` for `max-active-requests` )

README.md
```diff
@@ -77,7 +77,8 @@ Where:
      Max requests per connection (default: 4294937295).
 
    --max-active-requests <uint32_t>
-     Max active requests (default: 4294937295).
+     Max active requests (default: 100). H2-only, the number of concurrent
```
jplevyak: The description says that the default is 1.

source/client/benchmark_client_impl.cc
```diff
@@ -68,22 +68,12 @@ StatisticPtrMap BenchmarkClientHttpImpl::statistics() const {
...
-    return false;
-  }
-  // When no client side queueing is disabled (max_pending equals 1) we control the pacing as
-  // exactly as possible here.
-  // NOTE: We can't consistently rely on resourceManager()::requests()
-  // because that isn't used for h/1 (it is used in tcp and h2 though).
-  if ((max_pending_requests_ == 1 &&
-       (requests_initiated_ - requests_completed_) >= connection_limit_)) {
+  const uint64_t max_in_flight =
+      (max_pending_requests_ - 1) + (use_h2_ ? max_active_requests_ : connection_limit_);
```
jplevyak: Why the -1 ?  Comment.

source/client/options_impl.cc
```diff
@@ -140,8 +140,8 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
 
   TCLAP::ValueArg<uint32_t> max_active_requests(
       "", "max-active-requests",
-      fmt::format("Max active requests (default: {}).", max_active_requests_), false, 0, "uint32_t",
-      cmd);
+      fmt::format("Max active requests (default: {}). HTTP/2 only.", max_active_requests_), false,
+      0, "uint32_t", cmd);
```
jplevyak: This seems to say that the default is zero.

source/client/options_impl.h
```diff
@@ -75,7 +75,9 @@ class OptionsImpl : public Options {
   uint32_t request_body_size_{0};
   envoy::api::v2::auth::UpstreamTlsContext tls_context_;
   uint32_t max_pending_requests_{1};
-  uint32_t max_active_requests_{largest_acceptable_uint32_option_value};
+  // This default is based the minimum recommendation for SETTINGS_MAX_CONCURRENT_STREAMS over at
+  // https://tools.ietf.org/html/rfc7540#section-6.5.2
+  uint32_t max_active_requests_{100};
```
jplevyak: 100?

source/client/options_impl.cc
oschaaf: We ignore TCLAP's capability to set defaults. Instead, we define the (trivial) defaults in the option_impl header. Also added a comment above to point that out to future readers.

source/client/options_impl.h
oschaaf: This header provides the canonical source for default values.
(We also need them in the grpc service, and we don't use TCLAP there)

source/client/benchmark_client_impl.cc
oschaaf: So after thinking about this, and refreshing my memory for a bit, I changed this.
We do our own gating here in front of Envoy's own mechanism for that. However, that doesn't support
a zero-length max pending requests size. I made some changes in this PR to add support for `0` in NH's `--max-pending-requests`, and we now specialize on that value in some places, but no longer here. I think this is much cleaner. I also added a couple of integration tests to ensure the connection counts make sense.

oschaaf:
I addressed comments, and made some additional changes to this PR. Please take another look!

/assign jplevyak
/assign htuch

README.md
htuch: In this case, we probably want to shoot for "what would be reasonable in a performance benchmarking scenario".

README.md
```diff
@@ -138,8 +139,8 @@ Where:
      The number of seconds that the test should run. Default: 5.
 
    --connections <uint32_t>
-     The number of connections per event loop that the test should
-     maximally use. HTTP/1 only. Default: 1.
+     The maximum allowed number of concurrent connections per event loop.
+     HTTP/1 only. Default: 1.
```
htuch: Thinking about it, I think we should to set the defaults to "what would give the same result that some really simple `while` loop QPS generator would on a redline` test. Many folks will just want to measure max QPS (even though this is not the meaningful IMHO). To make this comparable to other tools, it should work for this out-of-the-box. The other settings can be used for more sensitive experiments, like latency vs. QPS characterizaiton.

oschaaf: So one the one hand, if someone downloads the tool and goes `./nighthawk_client arbitrarydomain.com` to try it out, it would be nice if that didn't result in a dead server or blacklisting :-) That is what motivated lowish defaults. To minimize friction for common cases, to float an idea, how about adding flags like `--run-XXX-test` (where XXX could be max-rps, which could then perhaps load a packaged `max-rps-test-options.proto` from disk) to reduce friction?

oschaaf: (but, arguably, `--connections` could be defaulted higher with the current create-as-needed- strategy, as the low rps and duration will probably prevent problems there)

oschaaf: I bumped the default for `--connections` to `100` as that will allow putting a decent amount of pressure on most servers by just tweaking `--rps`. @htuch @jplevyak does that sound good to you?

htuch:
I'm a little busy right now, I'd also like to get the other Google folks who have a stake in NH involved in these reviews. @envoyproxy/nighthawk-reviewers could one of you folks give this a pass?

mum4k:
I will be happy to look at this after my vacation (coming back on the 12 Nov). I realize that is an unwanted delay, maybe someone else has the cycles in the meantime?

eric846:
I can take this up on 4 Nov.

source/client/process_impl.cc
```diff
@@ -246,7 +246,10 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
     // We do not support any retrying.
     thresholds->mutable_max_retries()->set_value(0);
     thresholds->mutable_max_connections()->set_value(options_.connections());
-    thresholds->mutable_max_pending_requests()->set_value(options_.maxPendingRequests());
+    // We specialize on 0 below, as that is not supported natively. The benchmark client will track
+    // in flight work and avoid creating pending requests in this case.
+    thresholds->mutable_max_pending_requests()->set_value(
+        options_.maxPendingRequests() == 0 ? 1 : options_.maxPendingRequests());
```
eric846: I didn't understand this. The circuit breaker doesn't allow a value of 0?

If the user passed --max_pending_requests 0 on the command line, meaning disable client-side queueing, then we set maxPendingRequests to 1 here. Does that mean we actually allow client-side queueing with a queue size of 1?

oschaaf: Correct, the circuit breaker does not allow a value of 0. 

When 0 is specified for max-pending-requests in NH, we configure the circuit breaker to its minimum, 1. 
But: in the current mode (closed-loop) we actually attempt to avoid relying on the circuit breaker, because the idea is that we perform that gating ourselves so we can get a sense of  time spend being blocked on waiting for resources.
 
In the upcoming open-loop mode [1], Nighthawk won't try to avoid hitting the circuit breaker limits. A configured value of `0` should result in an error or warning there, because we'll solely rely on the circuit breaker.  

[1] https://github.com/envoyproxy/nighthawk/pull/177

test/integration/test_integration_basics.py
```diff
@@ -52,37 +55,60 @@ def mini_stress_test_h1(fixture, args):
...
+  assertCounterEqual(counters, "upstream_rq_pending_overflow", 10)
+
+
+def test_http_h2_mini_stress_test_without_client_side_queueing(http_test_server_fixture):
+  """
+  Run a max rps test with the h2 pool against our test server, with no client-side
+  queueing. 
+  """
+  counters = mini_stress_test(http_test_server_fixture, [
+      http_test_server_fixture.getTestServerRootUri(), "--rps", "999999", "--duration 2", "--h2",
```
eric846: Should all these duration args be
`"--duration", "2", `
?

test/integration/test_integration_basics.py
```diff
@@ -52,37 +55,60 @@ def mini_stress_test_h1(fixture, args):
...
   assertNotIn("upstream_cx_overflow", counters)
 
 
+def test_http_h2_mini_stress_test_with_client_side_queueing(http_test_server_fixture):
+  """
+  Run a max rps test with the h2 pool against our test server, using a small client-side
+  queue. 
+  """
+  counters = mini_stress_test(http_test_server_fixture, [
+      http_test_server_fixture.getTestServerRootUri(), "--rps", "999999", "--max-pending-requests",
```
eric846: Any chance 999999 is too small to truly stress it with this test harness? Is there any huge value like 999999999 that causes it to fail or behave differently? I'm remembering the time I got burned by using 9999 for infinity, and also the time Dr. Evil asked for one MILLION dollars :)

test/integration/test_integration_basics.py
oschaaf: yes, nice catch. It won't break us here, but sticking to this convention allows the `mini_stress_test()` to easily query which flags are set.

test/integration/test_integration_basics.py
oschaaf: Maybe we should consider adding a special value to indicate "as-fast-as-possible"? E.g. `MAX` or something like that. We would then internally translate to that to a large value.
For now I'd like to file an issue to track that, as this value is used in other places as well, and we update all relevant places in one go when Nighthawk understands that special value. WDYT?
(I think it is wishful thinking that mainstream processors will break this test anytime soon :-) )

eric846: Yep, sounds good. There's also a value `requests_per_connection=1024 * 1024` in test/integration/test_connection_management.py that might fall under the same scope.

test/integration/test_integration_basics.py
oschaaf: Fixed in 467ff20ed1387c83d044f8931bee454c4a172a39

test/integration/test_integration_basics.py
oschaaf: Filed https://github.com/envoyproxy/nighthawk/issues/190

oschaaf:
Merged master in here & resolved conflicts.

test/integration/test_connection_management.py
```diff
@@ -0,0 +1,96 @@
...
+  requests = counters["upstream_rq_total"]
+  assertCounterBetweenInclusive(counters, "upstream_cx_total", requests, requests + 1)
+
+
+# Test h2 with a single request_per_connection
+@pytest.mark.skipif(isSanitizerRun(), reason="Unstable in sanitizer runs")
+def test_http_h2_connection_management_single_request_per_conn_1(http_test_server_fixture):
+  counters = run_with_number_of_connections(
+      http_test_server_fixture,
+      1,
```
eric846: Would it make sense to copy this to an additional test with a value other than 1, as a sanity check?

source/client/benchmark_client_impl.cc
```diff
@@ -68,22 +68,12 @@ StatisticPtrMap BenchmarkClientHttpImpl::statistics() const {
...
-  }
-  // When no client side queueing is disabled (max_pending equals 1) we control the pacing as
-  // exactly as possible here.
-  // NOTE: We can't consistently rely on resourceManager()::requests()
-  // because that isn't used for h/1 (it is used in tcp and h2 though).
-  if ((max_pending_requests_ == 1 &&
-       (requests_initiated_ - requests_completed_) >= connection_limit_)) {
+  const uint64_t max_in_flight =
+      max_pending_requests_ + (use_h2_ ? max_active_requests_ : connection_limit_);
+  if ((requests_initiated_ - requests_completed_ >= max_in_flight) || pool_ptr == nullptr) {
```
eric846: Nit: The pool_ptr == nullptr check might be clearer in a separate `if`, directly after pool_ptr was defined.

test/integration/test_connection_management.py
oschaaf: yes, that makes sense to me, done in 0fcaf9185769ad68e07c51c9e3fee55e0035e6f3

source/client/benchmark_client_impl.cc
oschaaf: Also done in 0fcaf9185769ad68e07c51c9e3fee55e0035e6f3

test/integration/test_connection_management.py
oschaaf: ugh, sorry, plus https://github.com/envoyproxy/nighthawk/pull/178/commits/f3ae31052b07d8108c385b6808bc85b720525daf (which makes us test h2 again)

oschaaf:
@eric846 thanks for the review! one request, could you re-approve this? I pushed out a change that fixes the formatting issue I accidentally introduced in the last commit :/

PR #180, opened 2019-10-27
Generate the command line docs in README.md

oschaaf:
- Add tooling to help updating the CLI documentation to the latest, to avoid inconsistencies from creeping in. 
- While testing this I found two enum options (`--verbosity` and `--sequencer-idle-strategy`) displaying their defaults as int values. This PR corrects that.
- Updates documentation fragments that have CLI `--help` output to the versions generated by this change.


oschaaf:
Note to self: the command to run this and update `README.md` is:

```
bazel run -c dbg //tools:update_cli_readme_documentation
```

tools/BUILD
```diff
@@ -3,3 +3,23 @@ licenses(["notice"])  # Apache 2
 exports_files([
     "check_envoy_includes.py",
 ])
+
+py_library(
+    name = "tools_lib",
+    srcs = [
```
htuch: Prefer to just inline this in the target below, these intermediate targets aren't that helpful..

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,43 @@
...
+      '--readme', help='Relative path to the target documentation file, for example: "README.md"')
+  args = parser.parse_args()
+
+  project_root = os.path.join(os.path.dirname(os.path.join(os.path.realpath(__file__))), "../")
+  # Change directory to avoid TCLAP outputting a full path specification to the binary
+  # in its help output.
+  os.chdir(os.path.realpath(project_root))
+  readme_md_path = os.path.join(project_root, args.readme)
+  process = Popen([args.binary, "--help"], stdout=PIPE)
+  (output, err) = process.communicate()
```
htuch: Why not just `subprocess.run`?

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,43 @@
...
+  exit_code = process.wait()
+  if exit_code != 0:
+    sys.stderr.writelines(["Failure running --help on with target command"])
+    sys.exit(exit_code)
+
+  # Avoid trailing spaces, as they introduce markdown linting errors.
+  cli_help = [s.strip() for s in output.decode().splitlines()]
+
+  with open(readme_md_path, "r") as f:
```
htuch: Consider using `pathlib` as we are Python 3.

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,43 @@
...
+    replaced = re.sub("\nUSAGE\:[^.]*.*%s[^```]*" % args.binary, str.join("\n", cli_help), f.read())
```
htuch: Can markdown documents inline other docs? That way we could avoid having to do a substitution.

tools/update_cli_readme_documentation.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
```
htuch: Should this be run or verified by CI?

oschaaf: Ideally, yes we would verify in CI, but this requires a full build of the cli tools, which is why I hesitated to do it. If you feel it's worth the added cost there I'm +1, and happy to add it.

tools/update_cli_readme_documentation.py
oschaaf: I couldn't find a way to accomplish that. We could, however, create links to separate `.md` files, but I'm not sure if that's ideal?

tools/update_cli_readme_documentation.py
oschaaf: will amend

tools/update_cli_readme_documentation.py
oschaaf: Well, there's no good reason, I'll change this to use `subprocess.run`

tools/BUILD
oschaaf: Sure, will do

oschaaf: Done in https://github.com/envoyproxy/nighthawk/commit/b1e3ad8e779f4ad0cac5498ead7e4d2e30778fb9

tools/update_cli_readme_documentation.py
oschaaf: Done: https://github.com/envoyproxy/nighthawk/commit/b1e3ad8e779f4ad0cac5498ead7e4d2e30778fb9

tools/update_cli_readme_documentation.py
oschaaf: Done: https://github.com/envoyproxy/nighthawk/commit/b1e3ad8e779f4ad0cac5498ead7e4d2e30778fb9

oschaaf:
/assign htuch

thanks for the feedback, I addressed comments, modulo the ones where I had questions

tools/update_cli_readme_documentation.sh
htuch: I think we could just tack this onto the end of the basic build/test run, have it do an extra step to regenerate the README.md and ensure there is no delta. Doing it in the format CI job would be much more overhead as you point out.

tools/update_cli_readme_documentation.py
htuch: The alternative is to make the entire README.md written in the Python `strings.Template` language, load this file and do the substitution with the generate content. WDYT?

oschaaf:
Ok so this now works to fail the `build` task in CI, log tail:

```
INFO: Build completed successfully, 12 total actions
2019-10-30 10:17:58,858 - INFO - CLI documentation in /README.md needs to be updated for bazel-bin/nighthawk_client
Exited with code 255
CircleCI received exit
```

I'll now push the change that ought to make this pass.

tools/update_cli_readme_documentation.sh
oschaaf: Done, this is now integrated into the build step.

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,68 @@
...
+  os.chdir(os.path.realpath(project_root))
+  readme_md_path = os.path.join(project_root, args.readme)
+  process = subprocess.run([args.binary, "--help"], stdout=subprocess.PIPE, check=True)
+
+  # Avoid trailing spaces, as they introduce markdown linting errors.
+  cli_help = [s.strip() for s in process.stdout.decode().splitlines()]
+
+  target_path = pathlib.Path(readme_md_path)
+  with target_path.open("r", encoding="utf-8") as f:
+    original_contents = f.read()
```
htuch: See pathlib's `read_text()`

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,68 @@
...
+  parser.add_argument(
+      '--readme',
+      required=True,
+      help='Relative path to the target documentation file, for example: "README.md"')
+  parser.add_argument(
+      '--mode',
+      default="check",
+      required=True,
+      choices={"check", "fix"},
+      help='Either "check" or "fix"')
```
htuch: Nit: don't mix quoting styles (see line above and elsewhere) for strings.

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,68 @@
...
+  process = subprocess.run([args.binary, "--help"], stdout=subprocess.PIPE, check=True)
+
+  # Avoid trailing spaces, as they introduce markdown linting errors.
+  cli_help = [s.strip() for s in process.stdout.decode().splitlines()]
+
+  target_path = pathlib.Path(readme_md_path)
+  with target_path.open("r", encoding="utf-8") as f:
+    original_contents = f.read()
+    replaced = re.sub("\nUSAGE\:[^.]*.*%s[^```]*" % args.binary, str.join("\n", cli_help),
+                      original_contents)
```
htuch: Did you have any thoughts on string library templates?

tools/update_cli_readme_documentation.py
```diff
@@ -0,0 +1,68 @@
...
+      choices={"check", "fix"},
+      help='Either "check" or "fix"')
+
+  args = parser.parse_args()
+  root = logging.getLogger()
+  root.setLevel(logging.INFO)
+
+  handler = logging.StreamHandler(sys.stdout)
+  handler.setLevel(logging.INFO)
+  formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
```
htuch: Why do we have this complex logging for a simple script like this?

tools/update_cli_readme_documentation.py
oschaaf: Done

tools/update_cli_readme_documentation.py
oschaaf: Simplified

tools/update_cli_readme_documentation.py
oschaaf: amended

tools/update_cli_readme_documentation.py
oschaaf: +1, but as the current approach works and is better then nothing, I created https://github.com/envoyproxy/nighthawk/pull/180#discussion_r340425493 for postponing this to a follow up. I do agree it would be cleaner and less fragile.

tools/update_cli_readme_documentation.py
oschaaf: (yes, added to the earlier discussion so I could link to that for context in the issue I filed)

PR #183, opened 2019-10-29
Test request body transmission

oschaaf:
Adds a small integration test for ensuring `--request-body-size`
does what it says on the tin

(split out from https://github.com/envoyproxy/nighthawk/pull/167, which didn't end up needing it.
this also gets us coverage for a line or two)

### DIFF
--- a/test/integration/test_integration_basics.py
+++ b/test/integration/test_integration_basics.py
@@ -262,6 +262,7 @@ def test_dotted_output_format(http_test_server_fixture):
                                                           as_json=False)
   assertIn("global.benchmark_http_client.request_to_response.permilles-500.microseconds", output)
 
+
 def check_upload_expectations(fixture, parsed_json, expected_transmitted_bytes,
                               expected_received_bytes):
   counters = fixture.getNighthawkCounterMapFromJson(parsed_json)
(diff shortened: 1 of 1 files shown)


test/integration/test_integration_basics.py
```diff
@@ -261,3 +261,39 @@ def test_dotted_output_format(http_test_server_fixture):
...
+  """
+  upload_bytes = 10000
+
+  # test h1
+  parsed_json, _ = http_test_server_fixture.runNighthawkClient([
+      http_test_server_fixture.getTestServerRootUri(), "--duration", "1", "--rps", "2",
+      "--request-body-size",
+      str(upload_bytes)
+  ])
+  check_upload_expectations(http_test_server_fixture, parsed_json, upload_bytes * 2,
```
htuch: Why *2?

test/integration/test_integration_basics.py
```diff
@@ -261,3 +261,39 @@ def test_dotted_output_format(http_test_server_fixture):
   ],
                                                           as_json=False)
   assertIn("global.benchmark_http_client.request_to_response.permilles-500.microseconds", output)
+
+
+def check_upload_expectations(fixture, parsed_json, expected_transmitted_bytes,
```
htuch: Either inline or add more comment on parameters etc.

oschaaf: inlined

test/integration/test_integration_basics.py
oschaaf: comments added

PR #194, opened 2019-11-07
Change parsing of header arguments to client, to support header values with ':'.

jaaju:
This is needed to send server configuration through the `x-nighthawk-test-server-config` request header.
Besides, ':' and whitespace are allowed in header values.

Tested using `bazel test test:options_test` and ivoking `nighthawk_client` with argument `--request-header 'x-nighthawk-test-server-config: { "response_body_size": 1 }'`.


source/client/options_impl.cc
```diff
@@ -371,13 +371,12 @@ CommandLineOptionsPtr OptionsImpl::toCommandLineOptions() const {
...
-    request_header->set_key(split_header[0]);
-    if (split_header.size() == 2) {
-      request_header->set_value(split_header[1]);
-    }
+    auto pos = header.find(':');
+    if (pos != std::string::npos) {
+      request_header->set_key(header.substr(0, pos));
+      // Any visible char, including ':', is allowed in header values.
+      request_header->set_value(header.substr(pos + 1));
+    } // else throw? a ':' is required in a header.
```
oschaaf: +1, e.g. `throw MalformedArgvException("A ':' is required in a header");` ?
It would be great to also nail the case down in a test in `options_test.cc`. This would also help with keeping coverage up. Some pseudo-code for that:

```
TEST_F(OptionsImplTest, RequestHeaderWithoutColon) {
  EXPECT_THROW_WITH_REGEX(TestUtility::createOptionsImpl("client --request-header bar",
                          MalformedArgvException, "is required in a header");
}
```

oschaaf:
Oh one last remark: the CI `DCO` check is failing, fixing that might require a force push.

source/client/options_impl.cc
jaaju: Thanks for the review, I've added the exception and test.

jaaju:
@oschaaf ,
> Oh one last remark: the CI `DCO` check is failing, fixing that might require a force push.
I tried the `git rebase --signoff ...` suggested by the DCO check, but it seems to be failing from commits prior to the amends! When good to merge, let me know how I should do the "force push". Thanks!

oschaaf:
Also @envoyproxy/nighthawk-reviewers -- this LGTM pending one clang-tidy nit, anyone else who likes to take a pass at this?

source/client/options_impl.cc
```diff
@@ -371,12 +371,13 @@ CommandLineOptionsPtr OptionsImpl::toCommandLineOptions() const {
...
     auto request_header = header_value_option->mutable_header();
-    std::vector<std::string> split_header = absl::StrSplit(
-        header, ':',
-        absl::SkipWhitespace()); // TODO(oschaaf): maybe throw when we find > 2 elements.
-    request_header->set_key(split_header[0]);
-    if (split_header.size() == 2) {
-      request_header->set_value(split_header[1]);
+    auto pos = header.find(':');
+    if (pos != std::string::npos) {
+      request_header->set_key(header.substr(0, pos));
```
mum4k: (optional) Seems that we are changing a behavior related to whitespace here, the previous code used the absl::SkipWhitespace() predicate. I am unsure whether that was required or if the change is intentional.

Would it be useful to add a test case with whitespace to capture the "correct" behavior that we want to keep? Please let me know if there already is one that I missed.

jaaju: @mum4k , this change will retain whitespaces, which are allowed in HTTP header values.

The `RequestHeaderValueWithColonsAndSpaces` test uses a header value that has an embedded space and expects it to be preserved.

mum4k: Thank you for the clarification.

jaaju:
> Not sure what I'm looking at here, seems to be a bunch of upstream Envoy changes..

I missed the DCO signoff in my initial commit. I followed the `git rebase --signoff ...` and `git push --force-with-lease` steps recommended by the DCO check (https://github.com/envoyproxy/nighthawk/pull/194/checks?check_run_id=303864449). But that seems to include all commits merged in the meantime to this PR! Only the last 4 commits are mine.

Should I just discard this PR and create a new one (will remember to include DCO signoff)?

htuch:
@jaaju sure; this is a small PR. Please tag me on that.

jaaju:
Dropping this PR, will create another with only header parsing related commits.

PR #196, opened 2019-11-10
H1 pool behavior tweaks

oschaaf:
- Re-implement a cleaner and enhanced version of `--prefetch-connections` via our recent H1 pool derivation, that will continuously prefetch instead of only once and up front.
- Add `--h1-connection-reuse-strategy <HOT|FAIR>`. `HOT` represents stock Envoy behavior.

Status: needs tests


oschaaf:
Closing this in favor of #197 and #198

PR #197, opened 2019-11-11
Continuous h1-pool prefetching

oschaaf:
Now that we derive our own H1 pool implementation, clean up the way we implement connection prefetching, as well as make it a continuous process instead of only doing it up front.


oschaaf:
@envoyproxy/nighthawk-reviewers this PR is ready for review

test/benchmark_http_client_test.cc
```diff
@@ -220,33 +220,6 @@ TEST_F(BenchmarkClientHttpTest, StatusTrackingInOnComplete) {
   client_.reset();
 }
 
-TEST_F(BenchmarkClientHttpTest, ConnectionPrefetching) {
```
mum4k: Do we have any new behaviors that we should test here? Say the difference between having the setPrefetchConnections enabled and disabled?

oschaaf: Sorry, should have probably mentioned this in the PR description, but the test at https://github.com/envoyproxy/nighthawk/blob/master/test/integration/test_integration_basics.py#L246 is also in place.

That python-based integration testing is often convenient for end-to-end tests, but was not around initially. With the changes made here, this C++ test needed maintenance, but I figured it didn't really add any value on top of the python-based test. So I dropped it instead.

mum4k: Thank you @oschaaf.

source/client/benchmark_client_impl.cc
```diff
@@ -21,11 +21,18 @@ using namespace std::chrono_literals;
...
-void Http1PoolImpl::createConnections(const uint32_t connection_limit) {
-  ENVOY_LOG(error, "Prefetching {} connections.", connection_limit);
-  for (uint32_t i = 0; i < connection_limit; i++) {
-    createNewConnection();
+Envoy::Http::ConnectionPool::Cancellable*
+Http1PoolImpl::newStream(Envoy::Http::StreamDecoder& response_decoder,
+                         Envoy::Http::ConnectionPool::Callbacks& callbacks) {
+  // In prefetch mode we try to keep the amount of connections at the configured limit.
+  if (prefetch_connections_) {
+    while (host_->cluster().resourceManager(priority_).connections().canCreate()) {
```
htuch: What guarantee do we have that this doesn't live lock?

oschaaf: So.. how would you see this lock? 
Note: each worker in NH has a unique cluster assigned to it / they are not shared across workers. As far as I can tell, the vanilla Envoy pool implementations also use this in vanilla Envoy.

htuch: I'm wondering what guarntees we have that `canCreate()` returns false at some point?

oschaaf: The call to `ConnPoolImpl::createNewConnection()` unconditionally constructs a new `ActiveClient`, which in turn unconditionally and directly calls `resourceManager(parent_.priority_).connections().inc();`.

The resource manager is configured to cap the number of connections to our `--connection` limit option, and at some point we're going to hit that at which point `canCreate()` will return false.

htuch:
@oschaaf needs master merge.

oschaaf:
> @oschaaf needs master merge.

@htuch done!

PR #198, opened 2019-11-11
Add "fair" HTTP/1 connection reuse strategy

oschaaf:
When selecting a connection from the available ones for re-use, stock Envoy behavior is to pick the one that was most recently added (back) to the pool.

This adds an opt-in alternative to pick the least recently used available connection instead.
Effectively the will cycle through all the available connections, instead of efficiently restricting to the hot set.

CLI usage:
```
--h1-connection-reuse-strategy <HOT|FAIR> (default: "HOT")
```

Status: blocked  - we need a discuss and decide how to best move forward with and connection-reuse strategies in NH.

- [x] Needs https://github.com/envoyproxy/nighthawk/pull/197 to go in first.
- [x] Needs a decision around https://github.com/envoyproxy/nighthawk/pull/200 -- if we land that we probably want this option to not be h1-specific: not worth it as the option will be added as --experimental. We'll wait for Envoy to piggy back on unified h1/h2 pool capabilities regarding this.
- [x] Address review feedback - make the connection reuse strategy an explicit (mandatory) choice: not done as we will be landing this as experimental
- [x] Address review feedback - change option names to lru/mru, and lowercase them
- [ ] File issues on enhancements based on @antoniovicente 's [feedback](https://github.com/envoyproxy/nighthawk/pull/198#pullrequestreview-329123438)


source/client/options_impl.cc
```diff
@@ -169,6 +169,18 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
+      h1_connection_reuse_strategies);
+  TCLAP::ValueArg<std::string> h1_connection_reuse_strategy(
+      "", "h1-connection-reuse-strategy",
+      fmt::format(
+          "Choose picking the most recently used, or least-recently-used connections for re-use."
+          "(default: {}).",
+          absl::AsciiStrToLower(
+              nighthawk::client::H1ConnectionReuseStrategy_H1ConnectionReuseStrategyOptions_Name(
+                  h1_connection_reuse_strategy_))),
+      false, "", &h1_connection_reuse_strategies_allowed, cmd);
```
mum4k: Shouldn't we set and document the default as HOT, as if I understand this correctly, that is what happens when the flag isn't specified?

api/client/options.proto
```diff
@@ -107,4 +116,6 @@ message CommandLineOptions {
   google.protobuf.StringValue uri = 18; // [(validate.rules).string.uri = true];
   // See :option:`--trace` for details.
   google.protobuf.StringValue trace = 19; // [(validate.rules).string.uri = true];
+  // See :option:`--h1-connection-reuse-strategy` for details.
+  H1ConnectionReuseStrategy h1_connection_reuse_strategy = 23;
```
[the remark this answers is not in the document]
oschaaf: This is on purpose, some of the other id's are taken in other PR's :-)

source/client/options_impl.cc
oschaaf: It will be documented here, as that will replace {} with the default: https://github.com/envoyproxy/nighthawk/pull/198/files/59cefde2ee9417889f98a5af69a5e968e9cc3bb5#diff-54af86e901ab4e3f97d4dcf19e326cc3R178

The default is defined here in the options header (all trivially typed options have their defaults set over there): https://github.com/envoyproxy/nighthawk/pull/198/files/59cefde2ee9417889f98a5af69a5e968e9cc3bb5#diff-139219ef3d84f04ddb20aef0d38419dfR89

oschaaf:
re: review #197 first
Yes, that makes sense; this PR builds on top of it. Unfortunately we can't do recursive pull requests here.

oschaaf:
@eric846 So #197 was merged, and I merged master here. This is ready for review!

oschaaf:
Sorry -- I tagged @eric846 where I should have tagged @mum4k

README.md
```diff
@@ -72,6 +74,10 @@ Where:
 Enable open loop mode. When enabled, the benchmark client will not
 provide backpressure when resource limits are hit.
 
+--h1-connection-reuse-strategy <HOT|FAIR>
+Choose picking the most recently used, or least-recently-used
+connections for re-use.(default: hot).
```
htuch: Do you think this is the best "out of box" experience? Would fair have some advantage in making it harder to do the wrong thing?

api/client/options.proto
```diff
@@ -62,6 +62,15 @@ message SequencerIdleStrategy {
   SequencerIdleStrategyOptions value = 1;
 }
 
+message H1ConnectionReuseStrategy {
+  enum H1ConnectionReuseStrategyOptions {
+    DEFAULT = 0;
```
htuch: Best to make that `NOT_SET` and reject IMHO, to avoid surprising behavior if defaults change.

source/client/options_impl.cc
```diff
@@ -179,6 +179,18 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "failing execution. Defaults to not tolerating error status codes and connection errors.",
       false, "<string, uint64_t>", cmd);
 
+  std::vector<std::string> h1_connection_reuse_strategies = {"HOT", "FAIR"};
+  TCLAP::ValuesConstraint<std::string> h1_connection_reuse_strategies_allowed(
```
htuch: This is not particular to this PR or Nighthawk, but it seems kind of crazy there is so much boiler plate involved in adding new commands. I would find it ideal if both Envoy and NH were to take their set of options from a proto and then generate the CLI parser. This could be done with proto annotations etc. It's a pretty big project, so definitely not this PR, but it's a wishlist item.

test/integration/test_connection_management.py
```diff
@@ -105,3 +105,23 @@ def test_http_h2_connection_management_single_request_per_conn_1(http_test_serve
 @pytest.mark.skipif(isSanitizerRun(), reason="Unstable in sanitizer runs")
 def test_http_h2_connection_management_single_request_per_conn_1(http_test_server_fixture):
   connection_management_test_request_per_connection(http_test_server_fixture, 5, True)
+
+
+def test_h1_pool_strategy(http_test_server_fixture):
+  """
+  Test that with the "HOT" strategy only the first created connection gets to send requests.
+  Then, with the "FAIR" strategy, we expect the other connection to be used as well.
```
htuch: If we're adverting fairness, I think we want to show that there is roughly equal number of C0/C1 across a run of a fair few RPS. Do you think that could be tested easily?

README.md
oschaaf: Initially I wanted to propose `FAIR` as a new default, because in certain scenario's it may help the distribute load better across available computing capacity (threads/processes/machines).
But on the other hand, I guess `HOT` is the more realistic approach, as most client side (pool) implementations follow that pattern I guess.. 
So I was kind of torn and backed out the update to the default behavior.  (I also mulled over adding `RANDOM` here as an more challenging pattern, which may be representative for certain use cases as well.. so I ended up thinking that what is best here this kind of might just depend on the kind of traffic that you want to simulate).
Given all that, do you have any thoughts here?

source/client/options_impl.cc
oschaaf: I totally agree. Definitely sounds doable if you ask me.

oschaaf: Filed https://github.com/envoyproxy/nighthawk/issues/209

test/integration/test_connection_management.py
oschaaf: Actually, yes, so done in https://github.com/envoyproxy/nighthawk/pull/198/commits/cf86b00ae5648d81afad563b975611999a0ee602

api/client/options.proto
oschaaf: Rephrasing to make sure I understand, are you proposing we make this a mandatory option?

oschaaf:
@htuch feedback mostly addressed, added a question for clarification on a change you suggested to the proto spec. Also merged master here. In short, ready for another round!

api/client/options.proto
htuch: Yep, that way you don't get any surprises when defaults change in different endpoints.

README.md
htuch: I think "fair" is a better out-of-the-box experience for characterizing performance; when a user wants to get to the nuances of connection reuse and modeling the behavior of other clients, we can advise them to switch to "hot".

source/client/options_impl.cc
```diff
@@ -240,6 +252,16 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
                        upper_cased, &sequencer_idle_strategy_),
                    "Failed to parse sequencer idle strategy");
   }
+
+  if (h1_connection_reuse_strategy.isSet()) {
+    std::string upper_cased = h1_connection_reuse_strategy.getValue();
+    absl::AsciiStrToUpper(&upper_cased);
+    RELEASE_ASSERT(
```
htuch: Should these be exceptions and we catch them at the top? I think this makes testing easier.

test/integration/test_connection_management.py
```diff
@@ -105,3 +105,36 @@ def test_http_h2_connection_management_single_request_per_conn_1(http_test_serve
...
+
+  _, logs = http_test_server_fixture.runNighthawkClient([
+      "--rps", "20", "-v trace", "--connections",
+      str(connections), "--prefetch-connections", "--h1-connection-reuse-strategy", "FAIR",
+      "--termination-predicate",
+      "benchmark.http_2xx:%d" % requests,
+      http_test_server_fixture.getTestServerRootUri()
+  ])
+  for i in range(1, connections):
+    assertGreaterEqual(
```
htuch: Is an upper bound also possible to specify?

source/client/options_impl.cc
oschaaf: This should be invariant; we have a problem when we would raise that exception as that means we didn't set up TCLAP's constraints at all. In the regular user-flow, entering a value not listed as valid will result in TCLAP throwing an exception, which we will handle and show as an error to the user.

Added some testing around this in 49e4d2dfbf3e3c3a0a7a9d7bd5befab0c1192c5e

test/integration/test_connection_management.py
oschaaf: Yes, it was possible to tighten this up, done in bf86ebee79eccd3195bc22c701adb8977516d23f

source/client/options_impl.cc
htuch: If it's an invariant then this would ideally be ASSERT.

test/integration/test_connection_management.py
htuch: Looks good, thanks.

README.md
```diff
@@ -72,6 +74,10 @@ Where:
 Enable open loop mode. When enabled, the benchmark client will not
 provide backpressure when resource limits are hit.
 
+--h1-connection-reuse-strategy <HOT|FAIR>
```
antoniovicente: Drive by comment: FAIR doesn't map to least-recently-used in my mind.  Consider other names like:
most-recently-used | least-recently-used
HOT | COLD

Also should this be lowercase or uppercase?  The default below appears in lowercase, but the values here are uppercase.

README.md
```diff
@@ -43,11 +43,13 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--open-loop] [--failure-predicate <<string,
-uint64_t>>] ... [--termination-predicate
-<<string, uint64_t>>] ... [--trace <uri
-format>] [--sequencer-idle-strategy <spin
-|poll|sleep>] [--max-requests-per-connection
+bazel-bin/nighthawk_client  [--open-loop]
+[--h1-connection-reuse-strategy <HOT|FAIR>]
```
antoniovicente: Are we missing some other parameters like:
- Keep at least N open connections.
- Reuse connections that have been idle for at least N msec.

I think that the least-recently-used strategy results in connections not idling out after a burst.  It also helps spread load across envoy worker threads, but only if you manage to create enough connections in the first place.  Hence the parameter to create at least N connections to get better steady state behavior.

README.md
oschaaf: I like lru/mru, and in fact that is what these options where called initially. I can't remember why I changed it to this. Also, lowercasing makes sense, I'll amend that.

README.md
oschaaf: - We have `--prefetch-connections` which is an all or nothing proposition. It will switch to proactive connection prefetching mode and keep open the number of connections specified. An enhancement could then be to let that take an int to specify the minimum amount?
- Yeah, those are nice, I also mulled over adding "random/bursty" as a nastier pattern. 

Shall I file issues for these?

antoniovicente: sgtm.

README.md
antoniovicente: lru/mru sgtm.

mum4k:
@htuch, can you help to move this forward?

oschaaf:
Linking https://github.com/envoyproxy/nighthawk/pull/200#issuecomment-566718517 for context on why this is marked as blocked

oschaaf:
@htuch comments addressed, and marked the option as `--experimental`. Notes:

- As we will piggy-back on Envoy for this feature later on (and probably deprecate this), I feel it doesn't make sense to spend time on unifying h1/h2 pool capabilities regarding this here. So I left the option h1-specific.
- Another TODO was to make the connection reuse strategy an explicit (mandatory) choice: not done as we will be landing this as experimental, and this would force people to depend on this. Also, this will default to `mru` (current behavior / stock Envoy).

api/client/options.proto
htuch: Ping on this one.

test/integration/test_connection_management.py
```diff
@@ -95,3 +95,39 @@ def test_http_h2_connection_management_single_request_per_conn_1(http_test_serve
 @pytest.mark.skipif(isSanitizerRun(), reason="Unstable in sanitizer runs")
 def test_http_h2_connection_management_single_request_per_conn_1(http_test_server_fixture):
   connection_management_test_request_per_connection(http_test_server_fixture, 5, True)
+
+
+@pytest.mark.skipif(isSanitizerRun(), reason="Unstable in sanitizer runs")
```
htuch: Why are these unstable? Ideally we would have all inputs hermetic and not have any environmental deps.

test/integration/test_connection_management.py
```diff
@@ -95,3 +95,39 @@ def test_http_h2_connection_management_single_request_per_conn_1(http_test_serve
...
+  _, logs = http_test_server_fixture.runNighthawkClient([
+      "--rps 20", "-v", "trace", "--connections", "2", "--prefetch-connections",
+      "--experimental-h1-connection-reuse-strategy", "mru", "--termination-predicate",
+      "benchmark.http_2xx:10",
+      http_test_server_fixture.getTestServerRootUri()
+  ])
+
+  requests = 60
+  connections = 3
+  assertNotIn("[C1] message complete", logs)
```
htuch: It would be nice if Envoy has or could be taught to have better direct stats here. Relying on log messages is somewhat fragile.

api/client/options.proto
oschaaf: See https://github.com/envoyproxy/nighthawk/pull/198#issuecomment-569518828; I was thinking that maybe it is better to we postpone this until we land the non-experimental (Envoy-based) approach?

test/integration/test_connection_management.py
oschaaf: I backed out the unstable marking of this specific test, let's see how that goes. 
So far the "unstable in sanitizer runs" marking has been limited to the ones related to connection management. 

In general, these end-to-end integration tests are using the binaries we produce and verifying they work as intended. They aren't using simulated time... now the sanitizers sometimes induce excessive processing slowdowns, which can make some of these tests become unstable when they rely on "reasonable" timings.
For example: if we expect a connection to be re-used, excessive reply latency can cause test failures, because the connection doesn't free up in time (before we issue a new request).

I opted for marking them as unstable, but a different solution might be to tune parameters to allow more timing slack; however I suspect that doing so is going to increase test duration considerably overall (unless we start introducing specializations in tests for sanitizer awareness).

test/integration/test_connection_management.py
oschaaf: I agree, what's here today is best-effort based on the means currently available, but going forward perhaps a histogram (or just an average) which tracks replies/connection could be useful to get a sense of connection re-use?

htuch: Something like that, sure. Just wanted to make sure it was on your radar. Thanks.

PR #199, opened 2019-11-11
RateLimiter: Add UniformDistributingRateLimiter

oschaaf:
Introduces `RandomDistributingRateLimiter` which takes a delegate which
can will be used to generate timing offsets for the underlying rate
limiter. `UniformDistributingRateLimiter` builds on top of that and plugs
in a random uniform distribution.

Next steps:
- optionize enabling this & wire it through
- add support for more distributions next to `uniform_int_distribution`


oschaaf:
@envoyproxy/nighthawk-reviewers this PR is ready for review

source/common/rate_limiter_impl.cc
```diff
@@ -81,4 +83,41 @@ void LinearRateLimiter::releaseOne() {
   acquired_count_--;
 }
 
+RandomDistributingRateLimiter::RandomDistributingRateLimiter(
+    Envoy::TimeSource& time_source, RateLimiterPtr&& rate_limiter,
+    const std::function<const std::chrono::nanoseconds()>& random_distribution_generator)
```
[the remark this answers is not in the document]
oschaaf: Correct, nice catch. Ideally we would somehow constrain this

source/common/rate_limiter_impl.cc
```diff
@@ -6,6 +6,8 @@
 
 namespace Nighthawk {
 
+using namespace std::chrono_literals;
```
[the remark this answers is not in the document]
dubious90: Hi Otto, Jakub just explained to me what this does in this case, and I drop this comment, as the 1ns is definitely a clearer way to express time notation.  Sorry for my confusion!

source/common/rate_limiter_impl.cc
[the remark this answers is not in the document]
oschaaf: Added https://github.com/envoyproxy/nighthawk/pull/199/commits/92554697a4e533e03136c5af9b2e91f79a046cdc, wdyt?

source/common/rate_limiter_impl.cc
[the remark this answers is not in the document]
oschaaf: ok; so yeah Jakub and me discussed this on Slack and both ended up leaning towards that the convenience outweigh the contra arguments in this specific case as well. It's much nicer to look at :-)

test/rate_limiter_test.cc
```diff
@@ -104,4 +104,54 @@ TEST_F(BurstingRateLimiterIntegrationTest, BurstingLinearRateLimiterTest) {
...
+    if (rate_limiter->tryAcquireOne()) {
+      acquisitions++;
+    }
+    // We test the release gets propagated to the mock rate limiter.
+    // also, the release will force RandomDistributingRateLimiter to propagate tryAcquireOne.
+    rate_limiter->releaseOne();
+  }
+  // Surely we can expect that we didn't acquire every time we tried, considering we're using a
+  // uniform distributions.
+  EXPECT_LT(acquisitions, tries);
```
jplevyak: You can tighten this up, say (tries/2) + 30 // 1 in a billion chance of failure.

source/common/rate_limiter_impl.cc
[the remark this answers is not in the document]
dubious90: Looks good.

dubious90:
LGTM

test/rate_limiter_test.cc
oschaaf: Nice, done: e4f44f7495f72e3892ecb4c2f41bd79d906afc33

htuch:
@oschaaf @envoyproxy/nighthawk-reviewers one thing I think would be interesting to see in the future, based on NH's predecessor, is Zipfian distributions (https://en.wikipedia.org/wiki/Zipf%27s_law) for request release. This is a simple mathematical model for capturing the behaviors that occur in reality, e.g. a bunch of clustered requests and then many sparser requests. It looks like this is a good framework to build on here.

source/common/rate_limiter_impl.cc
```diff
@@ -81,4 +83,43 @@ void LinearRateLimiter::releaseOne() {
...
+    Envoy::TimeSource& time_source, RateLimiterPtr&& rate_limiter,
+    RandomDistributionGenerator random_distribution_generator)
+    : random_distribution_generator_(std::move(random_distribution_generator)),
+      time_source_(time_source), rate_limiter_(std::move(rate_limiter)) {}
+
+bool RandomDistributingRateLimiter::tryAcquireOne() {
+  if (!distributed_start_set_) {
+    if (rate_limiter_->tryAcquireOne()) {
+      distributed_start_ = time_source_.monotonicTime() + random_distribution_generator_();
+      distributed_start_set_ = true;
```
htuch: Can you just use `absl::optional` here?

source/common/rate_limiter_impl.h
```diff
@@ -54,4 +56,38 @@ class LinearRateLimiter : public RateLimiter,
   Envoy::MonotonicTime started_at_;
 };
 
+using RandomDistributionGenerator =
```
htuch: Can you comment on the return value here?

source/common/rate_limiter_impl.h
```diff
@@ -54,4 +56,38 @@ class LinearRateLimiter : public RateLimiter,
...
+
+// Allows adding uniformly distributed random timing offsets to an underlying rate limiter.
+class UniformDistributingRateLimiter : public RandomDistributingRateLimiter {
+public:
+  UniformDistributingRateLimiter(Envoy::TimeSource& time_source, RateLimiterPtr&& rate_limiter,
+                                 const std::chrono::nanoseconds upper_bound);
+
+private:
+  std::default_random_engine generator_;
+  std::uniform_int_distribution<uint64_t> distribution_;
```
htuch: Is there a way to do dependency injection of this or control the seed for deterministic tests?

oschaaf: Added dc748584f1d4a27739e33dfd2b3ee5e8b6935ce5 which refactors a bit. And rewrote a test to leverage the result.
Let me know if this looks better

PR #200, opened 2019-11-14
Experimental multi-connection H2 pool

### LINKED ISSUE #46: Add support for multiple concurrent H2 connections
Currently Nighthawk mostly uses a single http/2 connection to issue requests. This may lead to hotspotting processes on the benchmark target. The expectation is that this will be fixed upstream in Envoy.

oschaaf:
Adds a new pool, which experimentally creates multiple vanilla H2 pools
under the hood. Subsequently, the workload will be distributed in round-robin fashion
accross these pools.

The `--connections` option controls the number of h2 pools that will be created.

Fixes #46


oschaaf:
@htuch wdyt, is this a viable (NH-specific) approach?
Manual testing shows this helps saturating multi-threaded backends in h2 tests, especially when combining --max-requests-per-connection

source/client/benchmark_client_impl.h
```diff
@@ -50,6 +51,73 @@ class Http1PoolImpl : public Envoy::Http::Http1::ProdConnPoolImpl {
   void createConnections(const uint32_t connection_limit);
 };
 
+// Vanilla Envoy's HTTP/2 pool is single connection only (or actually sometimes dual in connection
+// drainage scenarios). Http2PoolImpl is an experimental pool, which uses multiple vanilla Envoy
+// HTTP/2 pools under the hood. Using multiple connections is useful when testing backends that need
+// multiple connections to distribute load internally. Combining multiple connections with
+// --max-requests-per-connection may help as well, as doing periodically initiating new connections
+// may help the benchmark target by giving it an opportunity to rebalance.
+class Http2PoolImpl : public Envoy::Http::ConnectionPool::Instance,
```
htuch: This is remarkably simple; I suggest you sync with @antoniovicente about how this might be delivered in Envoy.

oschaaf: @antoniovicente do you have any thoughts / comments on this experimental approach to adding support for multiple h2 connections? Maybe we can take this forward as a Nighthawk-specific approach?

antoniovicente: Sorry, I missed this review until now.  I'll try to look later today.

I wonder what's the status of the proposal at: https://docs.google.com/document/d/1gzY1y8CdbvZZ_82y9Fb6kNXAlzOBC0VIeiuMdn_xCRs/edit

oschaaf: @antoniovicente 
That draft certainly looks pretty relevant to this, and https://github.com/envoyproxy/nighthawk/pull/198 also :-)

antoniovicente: Commented on #198

antoniovicente: I think that the general Envoy case would benefit from minimizing the number of H2 connections in use.  This implementation looks good for the use case of us having a fairly steady known QPS that would result in good utilization of the backend connection when operating in LRU mode.  One potential worry is connections idling out before they are reused via RR, but that may be a desirable thing in a loadtest scenario.

I pinged in the document, I'ld like multi H2 connection support to be implemented in Envoy core.  Several projects depend on that fix.

mum4k:
@oschaaf the description indicates that this is still experimental/draft. How can we help at this point to get this moving? Do you need advice on direction from @htuch or @antoniovicente or is this ready for general review?

oschaaf:
> @oschaaf the description indicates that this is still experimental/draft. How can we help at this point to get this moving? Do you need advice on direction from @htuch or @antoniovicente or is this ready for general review?

To get this moving, we need to decide how to best take it forward. There are a couple of things:
- #198 introduces a keep-alive connection re-use strategy. do we want that to be h1 specific? We may want to unify
- There's a [proposal](https://github.com/envoyproxy/nighthawk/pull/200#discussion_r353858694) for Envoy in flight, which upon landing could replace this. That could provide an alternative but it's not clear if what that will end up looking like, and long that will take.
- I think there's also an effort to unify some things for the h1 and h2 pools in Envoy, possibly also involving connection re-use strategy.

I can come up with four options for both H2-pool multi connection support and connection re-use:
1) DIY, then we have to decide if we want to unify H1 / H2 behavior with respect to selecting a connection
  to put a request on.
2) Wait for this to arrive in Envoy and then piggyback
3) Get involved in the Envoy effort
4) Shoot for 1) and 3) in parallel. We land this soon-ish, and make a push in the relevant proposal(s) to get our requirements in so we can convert later and piggy back on Envoy

htuch:
@oschaaf I'm strongly inclined towards (2) or (3). Long term this is the most sustainable.

oschaaf:
> @oschaaf I'm strongly inclined towards (2) or (3). Long term this is the most sustainable.

how about landing these features with `--experimental` prefixed flags, and a big fat disclaimer in the docs that they are volatile and will be substituted with possibly non-identical Envoy-based replacements in the future?

htuch:
@oschaaf that's a reasonable option. I'd want to get a feeling for who would be consuming this and the value of having it out there. If one of our known NH users finds this useful, we can do this.

oschaaf:
This feature is now hidden under an `--experimental` flag, and plus some light test coverage

mum4k:
@dubious90 please review and assign back to me when ready.

source/client/benchmark_client_impl.h
```diff
@@ -66,6 +67,73 @@ class Http1PoolImpl : public Envoy::Http::Http1::ProdConnPoolImpl {
...
+                      public Envoy::Http::ConnPoolImplBase {
+public:
+  Http2PoolImpl(
+      Envoy::Event::Dispatcher& dispatcher, Envoy::Upstream::HostConstSharedPtr host,
+      Envoy::Upstream::ResourcePriority priority,
+      const Envoy::Network::ConnectionSocket::OptionsSharedPtr& options,               // NOLINT
+      const Envoy::Network::TransportSocketOptionsSharedPtr& transport_socket_options) // NOLINT
+      : Envoy::Http::ConnPoolImplBase(std::move(host), priority), dispatcher_(dispatcher),
+        socket_options_(options), transport_socket_options_(transport_socket_options) {
+    for (uint32_t i = 0; i < host_->cluster().resourceManager(priority_).connections().max(); i++) {
```
dubious90: is there a preference here for having this implementation in the header file, rather than the cc file?

test/run_nighthawk_bazel_coverage.sh
```diff
@@ -57,7 +57,7 @@ then
   COVERAGE_VALUE=$(llvm-cov export "${COVERAGE_BINARY}" -instr-profile="${COVERAGE_DATA}" \
     -ignore-filename-regex="${COVERAGE_IGNORE_REGEX}" -summary-only | \
     python3 -c "import sys, json; print(json.load(sys.stdin)['data'][0]['totals']['lines']['percent'])")
-  COVERAGE_THRESHOLD=98.0
+  COVERAGE_THRESHOLD=97.7
```
dubious90: This is a pretty high coverage threshold, and is likely to result in more edits in the future. Is the idea that whenever the threshold is reduced, we want to make the developer actively note it by lowering this value? Or could this value be reduced to a more reasonable longterm threshold?

source/client/benchmark_client_impl.h
dubious90: My personal thinking is that even though this and the other functions in this file are all fairly small, it makes it harder to get a clean look at the api in general, but I'm open to other opinions.

oschaaf: No good reason, so module one liners (getters/setters) which I feel don't really interfere with readablility of the api, I moved implementations to the `.cc`. Wdyt?

test/run_nighthawk_bazel_coverage.sh
oschaaf: So far the strategy was to keep this borderline high indeed, to queue authors/reviewers when something is up with coverage.  Personally I like that: there's some upkeep to this, but figuring out what changed and what needs to be done with respect to coverage is really easy. And we get auto-queued when something is off. So I bump in decrease coverage as-needed. E.g. the in flight grpc stuff will give a boost to coverage (that PR has `98.9%`), I had to lower it while splitting that up into pieces to not make master fail.

But as far as I am concerned, this isn't written in stone. I can imagine that with more people working on this might get uncomfortable. I believe @htuch and me have been aiming for `97.5%` initially.
It's worth noting that the code base is relatively small, so small changes in covered lines and even code-base size make a relatively big difference.

source/client/benchmark_client_impl.h
dubious90: that makes sense to me. Happy to follow your instincts on this in general. I do think the api is clearer for me now though, so thank you for doing that!

dubious90:
LGTM, assigning to @mum4k for approval

source/client/benchmark_client_impl.h
```diff
@@ -66,6 +67,48 @@ class Http1PoolImpl : public Envoy::Http::Http1::ProdConnPoolImpl {
...
+// Vanilla Envoy's HTTP/2 pool is single connection only (or actually sometimes dual in connection
+// drainage scenarios). Http2PoolImpl is an experimental pool, which uses multiple vanilla Envoy
+// HTTP/2 pools under the hood. Using multiple connections is useful when testing backends that need
+// multiple connections to distribute load internally. Combining multiple connections with
+// --max-requests-per-connection may help as well, as doing periodically initiating new connections
+// may help the benchmark target by giving it an opportunity to rebalance.
+class Http2PoolImpl : public Envoy::Http::ConnectionPool::Instance,
+                      public Envoy::Http::ConnPoolImplBase {
+public:
+  Http2PoolImpl(
```
mum4k: Can we document the constructor and its arguments? 

Applies to all other public portions of this API, unless they have been documented in a parent class already.

source/client/benchmark_client_impl.h
```diff
@@ -66,6 +67,48 @@ class Http1PoolImpl : public Envoy::Http::Http1::ProdConnPoolImpl {
...
+      Envoy::Event::Dispatcher& dispatcher, Envoy::Upstream::HostConstSharedPtr host,
+      Envoy::Upstream::ResourcePriority priority,
+      const Envoy::Network::ConnectionSocket::OptionsSharedPtr& options,                // NOLINT
```
mum4k: Why do we need the "NOLINT"s here?

oschaaf: clang-tidy doesn't like this [1]. But as the argument list is implied by a base class that originates from Envoy, it is hard to change this according to what it wants.

[1] https://app.circleci.com/jobs/github/envoyproxy/nighthawk/9194

```
clang-tidy -extra-arg-before=-xc++ -p=/root/project -quiet /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/source/client/benchmark_client_impl.cc
/build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/source/client/benchmark_client_impl.cc:50:5: error: pass by value and use std::move [modernize-pass-by-value,-warnings-as-errors]
    const Envoy::Network::ConnectionSocket::OptionsSharedPtr& options,
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Envoy::Network::ConnectionSocket::OptionsSharedPtr 
/build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/source/client/benchmark_client_impl.cc:51:5: error: pass by value and use std::move [modernize-pass-by-value,-warnings-as-errors]
    const Envoy::Network::TransportSocketOptionsSharedPtr& transport_socket_options)
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Envoy::Network::TransportSocketOptionsSharedPtr 
14185 warnings generated.
```

source/client/benchmark_client_impl.h
oschaaf: Added a single doc comment pointing those looking for doc comments to the interface/base class

PR #202, opened 2019-11-15
Change header arg parsing to support ':' in values.

jaaju:
(This is a redo of pull#194).

This is needed to send server configuration through the
`x-nighthawk-test-server-config` request header. Besides, ':' and
whitespace are allowed in header values.

A header argument must contain a ':' that delimits the name and value.

Tested using `bazel test test:options_test` and invoking
`nighthawk_client` with argument `--request-header
'x-nighthawk-test-server-config: { "response_body_size": 1 }'`.

Signed-off-by: Jay Ramamurthi <jramamurthi@google.com>


jaaju:
@oschaaf , @mum4k ,  @htuch - This is the same as the earlier PR#194, with only the changes that I intended.

Thanks!

oschaaf:
LGTM
/assign @htuch

Issue #203, opened 2019-11-16
Add an option to enable sequencer loop latency tracking


oschaaf:
Tracking time delta's between calls to `SequencerImpl::run(_)` would be useful to get a sense
how noisy an execution was as well in tests.

PR #204, opened 2019-11-16
Sequencer:  improve initial loop resolution

oschaaf:
When it is time to start releasing requests, don't initiate
work via the deferred path (e.g. timers, sleep), but enter the tight
loop right away. 

Filed https://github.com/envoyproxy/nighthawk/issues/203 for tracking a good follow up to this.


oschaaf:
@mum4k  Sure; so this change is driven by some manual tests and subsequent observations; effectively it proposes we wait with a spin loop until it is time to release the first request instead of polling at a coarse 1ms resolution for that through a periodic timer. This targets to improve accuracy of the first request-release timings.

PR #206, opened 2019-11-22
Update Envoy dependency

oschaaf:
Updates Envoy to fb8235971792d924a429577f87cd86ea6cbc1a75

### DIFF
--- a/source/server/README.md
+++ b/source/server/README.md
@@ -85,7 +85,6 @@ admin:
 USAGE:
 
 bazel-bin/nighthawk_test_server  [--use-fake-symbol-table <bool>]
-[--use-libevent-buffers <bool>]
 [--cpuset-threads]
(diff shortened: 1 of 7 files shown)


source/server/README.md
```diff
@@ -143,7 +140,7 @@ normally) or 'validate' (validate configs and exit).
 Hot restart parent shutdown time in seconds
 
 --drain-time-s <uint32_t>
-Hot restart drain time in seconds
+Hot restart and LDS removal drain time in seconds
```
htuch: Nit: prefer non-update changes to be distinct PRs.

oschaaf: Our fix format flagged this, and this belongs to the associated envoy update as our test server CLI changed.. so I thought I'd include it here

PR #208, opened 2019-11-22
Test-server: request entity body handling

### LINKED ISSUE #207: nighthawk_test_server filter vs request entity bodies
The test server filter does not handle incoming request entity bodies well. 
It will respond right after decoding the request headers, but before reading all of the data.
This should be fixed, and subsequently the test code that was disabled in https://github.com/envoyproxy/nighthawk/pull/205 should be re-enabled.

oschaaf:
When sending request data, the test server may put a reply on the wire
before finishing up reading all request data. Address that.

Preludes that should go in first:

- [x] #206
- [x] #205

Fixes #207


mum4k:
Eric is on vacation this week, I will take this one.

source/server/http_test_server_filter.h
```diff
@@ -68,8 +68,11 @@ class HttpTestServerDecoderFilter : public Envoy::Http::StreamDecoderFilter {
                                     nighthawk::server::ResponseOptions& response_options);
 
 private:
+  void sendReply();
   const HttpTestServerDecoderFilterConfigSharedPtr config_;
   Envoy::Http::StreamDecoderFilterCallbacks* decoder_callbacks_;
+  nighthawk::server::ResponseOptions base_config_;
```
mum4k: Can you help me understand the lifetime of a HttpTestServerDecoderFilterConfig instance? Is it created just for one pair of request and reply?

I am trying to understand why we are storing the error message and the base config as instance attributes rather than local variables in a function.

oschaaf: So I think `HttpTestServerDecoderFilterConfig` is what Envoy hands our filter instance, based on static (disk based) configuration, and possibly more sources. It's a shared pointer, so I assume these are shared across different instances of our filter. I guess that during configuration updates, two (or more) versions of `HttpTestServerDecoderFilterConfigPtr` are in existence, until all users of the old configuration(s) have dropped their references. Fortunately, this is probably transparant to our filter, and it seems we don't have to worry about configuration reloading as a result.

Our filter is instantiated per request, as far as I can tell, and therefore I added tracking of per-request state as instance attributes in there -- as we need to carry some information across various member calls.  

@htuch ^^ can you please verify this for correctness?

oschaaf:
FWIW, the prerequisites to this have landed, and I merged master into this, so this now is in a state that we can take it forward.

source/server/http_test_server_filter.h
htuch: Yeah, filter instance is per-request, config is shared.

mum4k: Thank you for the explanation.

source/server/http_test_server_filter.cc
```diff
@@ -53,33 +53,42 @@ void HttpTestServerDecoderFilter::applyConfigToResponseHeaders(
...
     decoder_callbacks_->sendLocalReply(
         static_cast<Envoy::Http::Code>(500),
-        fmt::format("test-server didn't understand the request: {}", *error_message), nullptr,
+        fmt::format("test-server didn't understand the request: {}", *error_message_), nullptr,
         absl::nullopt, "");
   }
+}
+
+Envoy::Http::FilterHeadersStatus
+HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::HeaderMap& headers, bool done) {
```
htuch: Nit: rename done to end_stream here and below.

oschaaf: done in 8ec6e11fbac09daab718fef309f3fc122491c930 + merged master

Issue #125, opened 2019-08-22
Publish release to docker


howardjohn:
It would be useful to have releases published to docker. This would allow easy testing in environments like Kubernetes, or where I don't have bazel, etc

Issue #210, opened 2019-11-22
Stabilize pytest under tsan/asan


oschaaf:
Some of the tests are flaking frequently in sanitizer runs.
Now that we landed the termination predicate feature, they can be stabilized by relying on that to complete a fixed amount of requests before checking expectations, instead capping on a fixed duration.

PR #211, opened 2019-11-24
Add docker build

howardjohn:
For https://github.com/envoyproxy/nighthawk/issues/125

This adds a basic docker file for the client and server

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,13 @@ jobs:
     steps:
       - checkout
       - run: ci/do_ci.sh check_format
+  docker:
+    docker:
+      - image: *envoy-build-image
+    resource_class: xlarge
+    steps:
+      - checkout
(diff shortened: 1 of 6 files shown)


oschaaf:
Cool! FYI tsan/asan runs are flaky in CI at the moment because they are too slow to reliably satisfy some of the expectations in the python-based integration tests. 
Tracking that in https://github.com/envoyproxy/nighthawk/issues/210, I'm working a PR to stabilize that.

howardjohn:
Thanks! I'll still need to fix this up, the docker stage depends on the binaries being built already so I'll either need to make it copy them over from the previous steps or rebuild maybe. But I did build these locally so it's at least mostly working

oschaaf:
As discussed on Slack, taking this forward in https://github.com/envoyproxy/nighthawk/pull/239

PR #214, opened 2019-11-25
Integration tests: add gRPC service test infrastructure

oschaaf:
Split out from https://github.com/envoyproxy/nighthawk/pull/137

- Add python gRPC service wrapper
- Add signal handling to the c++ gRPC service so we can make it shut down gracefully
- Temporarily lower the coverage threshold to 98% to avoid continuously being broken while we upstream the request-source feature.

We don't use this yet, but we'll need it later on. This helps leaning up the above PR,
and make review more manageable.


source/client/service_main.h
```diff
@@ -19,18 +23,40 @@ class ServiceMain : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
 public:
   ServiceMain(int argc, const char** argv);
   void start();
+  /**
+   * Can be used to wait for the server to exit.
```
mum4k: Can we explicitly say that this call is blocking?

Can we also indicate that it registers for certain signals that will terminate the server?

source/client/service_main.h
```diff
@@ -19,18 +23,40 @@ class ServiceMain : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
...
+
   grpc::ServerBuilder builder_;
   ServiceImpl service_;
   std::unique_ptr<grpc::Server> server_;
   std::shared_ptr<grpc::Channel> channel_;
   std::unique_ptr<nighthawk::client::NighthawkService::Stub> stub_;
   int listener_port_{-1};
   std::string listener_bound_address_;
+  std::string listener_output_path_;
+  std::vector<int> pipe_fds_;
```
mum4k: Can we add a comment here indicating what the pipe is for?

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,71 @@
...
+import threading
+import time
+from string import Template
+
+from common import IpVersion
+
+
+class NighthawkGrpcService(object):
+  """
+    Base class for running a server in a separate process.
```
mum4k: (nit) Doc comment seems misaligned.

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+  def __init__(self, server_binary_path, server_ip, ip_version):
```
mum4k: Can we add a comment for the initializer documenting its arguments? Something like:

def __init__(self, server_binary_path, server_ip, ip_version):
  """Initializes Nighthawk gRPC service.

  Args:
    server_binary_path: A string, path to the server binary.
    ...
  """

test/integration/nighthawk_grpc_service.py
mum4k: (optional) Can we also include an example of how this class would be used?

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,71 @@
...
+    assert ip_version != IpVersion.UNKNOWN
+    self.ip_version = ip_version
```
mum4k: Not sure if enforced elsewhere, but since this is a new file and a new class - can we distinguish between public and private attributes? We can prefix all private attributes with a single underscore. E.g. self._ip_version.

Additionally we could document any public attributes in the docstring of the class by adding an "Attributes" section.

class NighthawkGrpcService(object):
  """ Base class for running a server in a separate process.

  Attributes:
    server_binary_path: A string, path to the server binary.
    ...
  """

Ideally we would mark as many of these private as possible.

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,71 @@
...
+    self.server_binary_path = server_binary_path
+    self.server_thread = threading.Thread(target=self.serverThreadRunner)
+    self.server_process = None
+    self.server_ip = server_ip
+    self.socket_type = socket.AF_INET6 if ip_version == IpVersion.IPV6 else socket.AF_INET
+    self.server_port = 0
+    self.address_file = ""
+
+  def serverThreadRunner(self):
```
mum4k: Similarly for methods - can we indicate what the public API is by marking anything private with a leading underscore?

def _serverThreadRunner(self):

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,71 @@
...
+        self.server_port = int(tmp[len(tmp) - 1])
+        return True
+      time.sleep(0.5)
+      tries -= 1
+
+    logging.error("Timeout while waiting for server listener at %s:%s to accept connections.",
+                  self.server_ip, self.server_port)
+    return False
+
+  def start(self):
```
mum4k: These two methods seem to be the public API of this class, can we add comments on them even if just single sentences?

oschaaf:
@mum4k Thanks for the review; I added a round of enhancements based on that in 162ca7af0060c99e3f753a72ffbde1e9b95ed6ca

/assign mum4k

test/integration/nighthawk_grpc_service.py
```diff
@@ -60,12 +73,21 @@ def waitUntilServerListening(self):
...
+  """
+  Starts the Nighthawk gRPC service. Returns True upon success, after which the server_port attribute
+  can be queried to get the listening port.
+  """
+
   def start(self):
-    self.server_thread.daemon = True
-    self.server_thread.start()
-    return self.waitUntilServerListening()
+    self._server_thread.daemon = True
```
mum4k: (nit) In Python we typically add doc comments after the opening of the scope we are commenting on. To comment on a method we would do:

```python
def start(self):
  """
  Starts the Nighthawk gRPC service. Returns True upon success, after which the server_port
  attribute can be queried to get the listening port.
  """
```

Applies here and below.

test/integration/nighthawk_grpc_service.py
```diff
@@ -1,50 +1,63 @@
...
-    """
+  Class for running the Nighthawk gRPC service in a separate process.
+  Usage:
+    grpc_service = NighthawkGrpcService("/path/to/nighthawk_service"), "127.0.0.1", IpVersion.IPV4)
+    if grpc_service.start():
+      .... You can talk to the Nighthawk gRPC service at the 127.0.0.1:grpc_service.server_port ...
+
+  Attributes:
+  server_ip: IP address used by the gRPC service to listen. 
+  server_port: An integer, indicates the port used by the gRPC service to listen.
```
mum4k: Looks like in the current design this would be a value of zero before start is called. That indicates that the object here can be in an "invalid" state. If we really wanted to fix it, we would make the  NighthawkGrpcService class private and add a public factory function that would both create the instance and call start on it. The factory function would then return the class instance in a valid state.

The other option is to at least document the behavior here where we are talking about the attribute. Indicating that it will be zero if the class wasn't used properly.

test/integration/nighthawk_grpc_service.py
oschaaf: Whoops, fixed in 75ed6acc89f5937e90bc9ef617c7e1b535d1e0ac

test/integration/nighthawk_grpc_service.py
oschaaf: Added a comment about that, and made sure to reset the port to 0 after stopping the server to be consistent: 75ed6acc89f5937e90bc9ef617c7e1b535d1e0ac.

oschaaf:
@mum4k thanks, I addressed your comments, so this is ready for another look

source/client/service_main.cc
```diff
@@ -49,13 +67,53 @@ void ServiceMain::start() {
...
+    RELEASE_ASSERT(close(pipe_fds_[1]) == 0, "write side close failed");
+    pipe_fds_.clear();
+    server_->Shutdown();
+  });
 }
 
-void ServiceMain::wait() { server_->Wait(); }
+void ServiceMain::wait() {
+  signal_handler_delegate = [this](int) { onSignal(); };
+  signal(SIGTERM, signal_handler);
```
htuch: Do we want to mix raw signals and libevent like this? I thought libevent has dedicated APIs for signal handling.

oschaaf: Right; looking into that I think ideally we'd use `DispatcherImpl::listenForSignal()`..

But we'd need a `DispatcherImpl` instance - or a means to allocate one - to leverage that, and I can't see a convenient way to accomplish either. This file just wraps/manages the Nighthawk gRPC service, and has no ties to NH/Envoy concepts (yet) otherwise.. 
If I'm correct, pulling this of would require wiring a way to request signal handling up accross the gRPC service implementation in `ServiceImpl` and then `ProcessImpl` where the main dispatcher lives.

when I scanned Envoy's codebase, I couldn't find any other signal hooking going on except in Envoy's server.cc - which will not run concurrently to this, so based on that this seemed viable to me (as an intermediate step at least).

htuch: Ah, right, that's fair.

source/client/service_main.cc
```diff
@@ -49,13 +67,53 @@ void ServiceMain::start() {
     absl::StrAppend(&listener_bound_address_, listener_port_);
   }
   ENVOY_LOG(info, "Nighthawk grpc service listening on: {}", listener_bound_address_);
+  if (listener_output_path_ != "") {
+    std::ofstream myfile(listener_output_path_);
+    if (myfile.is_open()) {
+      myfile << listener_bound_address_;
+      myfile.close();
```
htuch: This is probably redundant due to RAII, but doesn't hurt either.

source/client/service_main.cc
```diff
@@ -49,13 +67,53 @@ void ServiceMain::start() {
...
+    }
+  }
   channel_ = grpc::CreateChannel(listener_bound_address_, grpc::InsecureChannelCredentials());
   stub_ = std::make_unique<nighthawk::client::NighthawkService::Stub>(channel_);
+  pipe_fds_.resize(2);
```
htuch: I realize that signal handlers are a horrible execution environment, but is a pipe the best that is practical? We can't lock etc., and we dont' want to write to shared memory. Maybe we should have a dispatcher and just post to it?

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,94 @@
...
+    self.server_ip = server_ip
+    self._server_process = None
+    self._ip_version = ip_version
+    self._server_binary_path = server_binary_path
+    self._socket_type = socket.AF_INET6 if ip_version == IpVersion.IPV6 else socket.AF_INET
+    self._server_thread = threading.Thread(target=self._serverThreadRunner)
+    self._address_file = ""
+
+  def _serverThreadRunner(self):
+    with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".tmp") as tmp:
```
htuch: Isn't the benefit of having a context manager here that it will delete/clean up for us when we're done? Why not run the code below under it?

test/integration/nighthawk_grpc_service.py
```diff
@@ -0,0 +1,94 @@
...
+    self._server_process.communicate()
+
+  def _waitUntilServerListening(self):
+    tries = 30
+    while tries > 0:
+      contents = ""
+      try:
+        with open(self._address_file) as f:
+          contents = f.read().strip()
+      except (IOError):
```
htuch: Nit: no need for parens.

source/client/service_main.cc
oschaaf: Some arguments I weighed with respect to the pipe approach:

Pro:
- easy to verify that it is signal safe

Con:
- Need to verify the dispatcher / post implementation to make sure it restricts to the list of safe calls over at http://man7.org/linux/man-pages/man7/signal-safety.7.html
- Generally `foreign` code may change; is the dispatcher post routine guaranteeing signal safety? Will it remain so if a different event system is swapped in? 
- Lastly we don't have a dispatcher around here today (see my earlier [comment](https://github.com/envoyproxy/nighthawk/pull/214#discussion_r356139568))

oschaaf:
@htuch addressed feedback in f34669c as well as left a comment detailing the reasoning behind the pipe-based approach.

source/client/service_main.cc
htuch: Fair enough, let's go with this, we can clean up later as needed.

PR #217, opened 2019-11-27
Switch to enableHRTimer -- improve request-release timing accuracy

oschaaf:
Improves accuracy of request-release timings when the sequencer cannot use spinning.

Requirement on the Envoy side needs to go in first:
- [x] https://github.com/envoyproxy/envoy/pull/9155
- [x] https://github.com/envoyproxy/envoy/pull/9229
- [x] update the Envoy dep to a revision that has ^^: update the revision over at https://github.com/envoyproxy/nighthawk/pull/222
- [x] merge master after ^^, which hopefully makes CI happy regarding this PR


oschaaf:
For best results, libevent needs to be made aware too that we'd like to use precise timers.
One way to do that is by setting an environment variable called EVENT_PRECISE_TIMER.

```
EVENT_PRECISE_TIMER=1 bazel-bin/nighthawk_client ...
```

oschaaf:
readying this up for review, as we are closing in on landing all the prerequisites to this.
Tests should green-light once that happens.

source/client/process_impl.cc
```diff
@@ -408,5 +409,14 @@ bool ProcessImpl::run(OutputCollector& collector) {
   return counters.find("sequencer.failed_terminations") == counters.end();
 }
 
+void ProcessImpl::setupForHRTimers() {
+  // We override the local environment to indicate to libevent that we favor precision over
+  // efficiency. Note that it is also possible to do this at setup time via libevent's api's.
+  // The upside of the approach below is that we are very loosely coupled and have a one-liner.
+  // Getting to libevent for the other approach is going to introduce more code as we would need to
+  // derive our own customized versions of certain Envoy concepts.
+  putenv(const_cast<char*>("EVENT_PRECISE_TIMER=1"));
```
htuch: Is it safe to change this at runtime? It seems to be relying on the fact that libevent isn't caching this value or using it for its own setup. Should we at least push it to prior to wherever we initialize libevent?

source/common/sequencer_impl.cc
```diff
@@ -82,8 +82,9 @@ void SequencerImpl::updateStartBlockingTimeIfNeeded() {
 
 void SequencerImpl::run(bool from_periodic_timer) {
   ASSERT(running_);
-  const auto now = last_event_time_ = time_source_.monotonicTime();
+  const auto now = time_source_.monotonicTime();
   const auto running_duration = now - start_time_;
+  last_event_time_ = now;
```
htuch: What's the reason for this change?

source/client/process_impl.cc
oschaaf: So libevent applies this [here](https://github.com/libevent/libevent/blob/13b8fc39fd4000de2c836b0acdc52ae5b4c757c7/event.c#L633).

So I think this is safe / in time for the workers.  But I do agree earlier is less risky going forward. Let me see what I can do there.

oschaaf: (doing this early makes this a little more resilient to changes, and would also make this apply to the dispatcher we use above)

oschaaf:
@htuch backed out the no-op change, and moved the `setupForHRTimers()` call + added a comment about the timing of the call.

source/client/process_impl.cc
htuch: Yeah, this is better. I'm still a bit sketched by using env vars here, but let's see how it goes.

Issue #31, opened 2019-04-23
Support ramping of load and connections


oschaaf:

oschaaf:
Work in progress associated to this:

- [ ] ramping rate limiters - https://github.com/envoyproxy/nighthawk/pull/218
- [ ] phases concept - https://github.com/envoyproxy/nighthawk/pull/219

PR #218, opened 2019-11-27
Linear ramping and probabilistic ramping

oschaaf:
Contains new RateLimiter features:

- A linear rate limiter, which ramps up to a frequency over a configurable period. 
  Precise and relatively easy to reason about
- A probabilistic rate limiter filter, which gradually opens up over a configurable period.
  Useful for ramping up when wrapping an underlying rate limiter which is non-deterministic, 
  and possibly in distributed scenarios as well to add some entropy / jitter. 

This needs a bit of polish, but ideally we discuss this early so we can cherry pick what we like
from this and possibly discard what we don't.

TODO:

- [x] Land https://github.com/envoyproxy/nighthawk/pull/223
- [x] Make `RampingLinearRateLimiter` take a ramp duration argument & final frequency to unify the interface with the probabilistic filtering one

In a follow up we can wire up options for these new rate limiters. 
(May also be done in a follow-up PR instead).

In conjunction with the upcoming [Phases](https://github.com/envoyproxy/nighthawk/pull/219) this would allow implementing a better warmup phase, which ramps up to the targeted frequency.

Groundwork for #31


oschaaf:
@htuch finishing up Zipf support needs some of your attention here.
It can be split out into a separate change too, not sure how much work it that needs to ready up?

htuch:
@oschaaf which aspects of Zipf support are you after my feedback on? One thing I'd suggest is exposing out the parameters in the constructor https://github.com/abseil/abseil-cpp/blob/master/absl/random/zipf_distribution.h#L105 to the end user.

oschaaf:
> @oschaaf which aspects of Zipf support are you after my feedback on? One thing I'd suggest is exposing out the parameters in the constructor https://github.com/abseil/abseil-cpp/blob/master/absl/random/zipf_distribution.h#L105 to the end user.

Thanks, that helps. I'll expose `double q` and `double v` (after figuring out if they also have less cryptic names) :-)

oschaaf:
> @oschaaf which aspects of Zipf support are you after my feedback on? One thing I'd suggest is exposing out the parameters in the constructor https://github.com/abseil/abseil-cpp/blob/master/absl/random/zipf_distribution.h#L105 to the end user.

done, exposed the args and a way to wire in determinism plus added a test and some doc comments

mum4k:

mum4k:
@eric846 I will take this one and will be happy to give you another one after your vacation, please do enjoy it.

source/common/rate_limiter_impl.cc
```diff
@@ -78,6 +78,43 @@ void LinearRateLimiter::releaseOne() {
...
+    throw NighthawkException("duration must be positive");
+  }
+}
+
+bool LinearRampingRateLimiterImpl::tryAcquireOne() {
+  if (acquireable_count_) {
+    acquired_count_++;
+    return acquireable_count_--;
+  }
+  const auto elapsed_time = elapsed() + 1ns;
```
mum4k: (optional) This may be a good place to spell out the type instead of using auto to help the reader:

```cc
const std::chrono::nanoseconds elapsed_time = ...
```

source/common/rate_limiter_impl.cc
```diff
@@ -78,6 +78,43 @@ void LinearRateLimiter::releaseOne() {
...
+  double elapsed_fraction = 1.0;
+  if (elapsed_time < ramp_time_) {
+    elapsed_fraction -= static_cast<double>(ramp_time_.count() - elapsed_time.count()) /
+                        static_cast<double>(ramp_time_.count());
+  }
+
+  const double current_frequency = elapsed_fraction * (frequency_.value() * 1.0);
+  const double chrono_seconds =
+      std::chrono::duration_cast<std::chrono::duration<double>>(elapsed_time).count();
+  const double total = chrono_seconds * current_frequency / 2.0;
```
mum4k: Sorry I may be tired, can you help me understand the division by 2.0?

test/rate_limiter_test.cc
```diff
@@ -172,4 +172,123 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplSchedulingTest) {
...
+class LinearRampingRateLimiterImplTest : public Test {
+public:
+  std::vector<int64_t> getAcquisitionTimings(const Frequency frequency,
+                                             const std::chrono::seconds duration) {
+    Envoy::Event::SimulatedTimeSystem time_system;
+    std::vector<int64_t> aquisition_timings;
+    LinearRampingRateLimiterImpl rate_limiter(time_system, duration, frequency);
+    auto total_ms_elapsed = 0ms;
+    const auto clock_tick = 1ms;
+    auto last_acquisition_timestamp = 0ms;
```
mum4k: What is last_acquisition_timestamp used for?

source/common/rate_limiter_impl.h
```diff
@@ -157,4 +177,44 @@ class FilteringRateLimiterImpl : public ForwardingRateLimiterImpl,
...
+public:
+  /**
+   * From the absl header associated to the zipf distribution:
+   * Preconditions: v > 0, q > 1
+   * The precondidtions are validated when NDEBUG is not defined via
+   * a pair of assert() directives.
+   * If NDEBUG is defined and either or both of these parameters take invalid
+   * values, the behavior of the class is undefined.
+   */
+  ZipfRateLimiterImpl(RateLimiterPtr&& rate_limiter, bool deterministic, double q = 2.0,
```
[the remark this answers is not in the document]
oschaaf: I like that; +1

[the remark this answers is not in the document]
oschaaf: Maybe we should have a place to put code-level guidance like this?

source/common/rate_limiter_impl.h
```diff
@@ -157,4 +177,44 @@ class FilteringRateLimiterImpl : public ForwardingRateLimiterImpl,
...
+  const std::chrono::nanoseconds ramp_time_;
+};
+
+/**
+ * Thin wrapper around absl::zipf_distribution that will pull zeroes and ones from the distribution
+ * with the intent to probabilistically suppress the wrapped rate limiter.
+ * This may need further consideration, because it will shoot holes in the pacing, lowering the
+ * actual achieved frequency.
+ */
+class ZipfRateLimiterImpl : public FilteringRateLimiterImpl {
```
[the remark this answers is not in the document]
oschaaf: Yes, let's move this into its own PR. I'm exploring an initial take on this here to figure out how we  this could fit zipf distributions into our current model / get a sense of how we can land this. @htuch came up with the idea for this feature and I think it would be a really cool to have.

Having said that, I feel the zipf stuff is kind of on a different trajectory compared to the other rate limiters in here. 

Initially I planned to only do the ramping ones as that can be used directly to get us a better warm-up phase, but maybe I got a little (over-)enthusiastic ;-)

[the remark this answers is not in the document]
oschaaf: Moved the zipf stuff over to #251

source/common/rate_limiter_impl.h
```diff
@@ -157,4 +177,44 @@ class FilteringRateLimiterImpl : public ForwardingRateLimiterImpl,
   const RateLimiterFilter filter_;
 };
 
+/**
+ * Takes a probabilistic approach to suppressing an arbitrary wrapper rate limiter.
```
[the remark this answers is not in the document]
oschaaf: Done in 6329377a31b0b35828d81c2e2a22cfd8f3943e8a

source/common/rate_limiter_impl.cc
```diff
@@ -78,6 +78,43 @@ void LinearRateLimiter::releaseOne() {
   acquired_count_--;
 }
 
+LinearRampingRateLimiterImpl::LinearRampingRateLimiterImpl(Envoy::TimeSource& time_source,
+                                                           const std::chrono::nanoseconds ramp_time,
+                                                           const Frequency frequency)
+    : RateLimiterBaseImpl(time_source), ramp_time_(ramp_time), frequency_(frequency) {
+  if (frequency_.value() <= 0) {
+    throw NighthawkException("frequency must be > 0");
```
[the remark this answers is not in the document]
oschaaf: Done in 6329377a31b0b35828d81c2e2a22cfd8f3943e8a

source/common/rate_limiter_impl.cc
oschaaf: Done in 6329377a31b0b35828d81c2e2a22cfd8f3943e8a

source/common/rate_limiter_impl.cc
oschaaf: Added comment in 6329377a31b0b35828d81c2e2a22cfd8f3943e8a

test/rate_limiter_test.cc
oschaaf: Removed in 6329377a31b0b35828d81c2e2a22cfd8f3943e8a

source/common/rate_limiter_impl.cc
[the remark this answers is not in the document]
oschaaf: Ugh, I have to refresh my memory. Backing it out it seems to make the first release timing
in tests go off, e.g. we start at `1001` instead of `1000`, I need to dive in to figure out why that was again. The test used to have a 1 ms resolution, possibly that has something to do with it.
I'll figure this out and get back to this.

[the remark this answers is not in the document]
oschaaf: Please see 5cfa171ba9bcc920fe36ad220b81d0e092ecd24b
Summary:

- Backed out the 1 ns adjustment
- Add comment with explanation & amend test expectations
- Also, fix a TODO and make the test use microsecond step resolution now that we can (as of https://github.com/envoyproxy/envoy/pull/9229)

source/common/rate_limiter_impl.h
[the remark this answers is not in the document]
mum4k: Thank you for moving it over to another PR. Let's continue the discussion over there, I am interested to hear more about the Zipf use case and motivation.

source/common/rate_limiter_impl.h
```diff
@@ -157,4 +177,44 @@ class FilteringRateLimiterImpl : public ForwardingRateLimiterImpl,
...
+ */
+class GraduallyOpeningRateLimiterFilter : public FilteringRateLimiterImpl {
+public:
+  GraduallyOpeningRateLimiterFilter(const std::chrono::nanoseconds ramp_time,
```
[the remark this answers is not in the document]
mum4k: I believe this comment still applies after the latest changes. Maybe it got lost?

source/common/rate_limiter_impl.cc
mum4k: Thank you, that makes perfect sense.

source/common/rate_limiter_impl.cc
```diff
@@ -118,4 +160,21 @@ bool FilteringRateLimiterImpl::tryAcquireOne() {
...
+    RateLimiterPtr&& rate_limiter)
+    : FilteringRateLimiterImpl(
+          std::move(rate_limiter),
+          [this]() {
+            if (elapsed() < ramp_time_) {
+              const double chance_percentage =
+                  100.0 - (static_cast<double>(ramp_time_.count() - elapsed().count()) /
+                           (ramp_time_.count() * 1.0)) *
+                              100.0;
+              return std::round(provider_->getValue() / 10000.0) <= chance_percentage;
```
mum4k: I am trying to understand the fixed division by 10k. It seems to imply some expectations about the discrete numeric distribution provider that don't seem to be specified on the API.

This may be related to my comment that we should add more documentation next to the constructor for the GraduallyOpeningRateLimiterFilter that will explain the arguments and their roles.

source/common/rate_limiter_impl.cc
```diff
@@ -118,4 +160,21 @@ bool FilteringRateLimiterImpl::tryAcquireOne() {
   return rate_limiter_->tryAcquireOne() ? filter_() : false;
 }
 
+GraduallyOpeningRateLimiterFilter::GraduallyOpeningRateLimiterFilter(
+    const std::chrono::nanoseconds ramp_time, DiscreteNumericDistributionSamplerPtr&& provider,
```
mum4k: Should we validate some of the arguments?

test/rate_limiter_test.cc
```diff
@@ -172,4 +172,102 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplSchedulingTest) {
...
+}
+
+TEST_F(LinearRampingRateLimiterImplTest, TimingVerificationTest) {
+  // The first acquisition timing (1000050) seems slightly off because of floating point
+  // rounding causing the first timing to fall into the next clock tick. This is
+  // a test only issue: in practice we don't have fixed microsecond step sizes,
+  // and the rate limiter computes at nanosecond precision internally. As we
+  // want to have microsecond level precision, this should be more then sufficient.
+  EXPECT_EQ(getAcquisitionTimings(5_Hz, 5s),
+            std::vector<int64_t>({1000050, 1732100, 2236100, 2645800, 3000000, 3316650, 3605600,
```
mum4k: Tests apart from testing are often times used by the readers to understand what the code does. (i.e. as additional documentation). Feels like the values that we chose for these tests don't really help the reader. Additionally the unit of these values isn't obvious since the last change to micro-seconds.

I am wondering if we could choose different parameters than 5Hz over 5sec or maybe add test cases with other parameters that will be easier / more obvious for the reader to comprehend. We don't want to decrease test coverage, but we should aim towards self-documenting and readable test cases. Can we try to find a balance?

source/common/rate_limiter_impl.cc
```diff
@@ -118,4 +160,21 @@ bool FilteringRateLimiterImpl::tryAcquireOne() {
   return rate_limiter_->tryAcquireOne() ? filter_() : false;
 }
 
+GraduallyOpeningRateLimiterFilter::GraduallyOpeningRateLimiterFilter(
```
mum4k: (fyi) Now that I read the code and understand the implementation here, I have to say that this is truly beautiful, including your use of class inheritance for the ForwardingRateLimiter.

I have learned something today and I appreciate it, thanks.

test/rate_limiter_test.cc
oschaaf: @mum4k I couldn't really obtain easy-to-grok numbers by picking different parameter values here; but how about this: 9d248f2a914e6cf667216ebac2a1ba46d5e5e618 ?

Instead of matching expectations against hard-coded numbers, that uses the second law of motion to figure out control acquisition timings for comparison, and adds a little slack to when verifying expectations with a comment explaining the need for that. I'm hoping that will help future readers of this code. wdyt, will this be better?

(As a side-effect, this approach also allowed adding tests with higher frequencies / duration arguments)

source/common/rate_limiter_impl.cc
oschaaf: That is nice to hear, thanks!

oschaaf:
@mum4k  I think I addressed all comments, so marking this as ready-for-review. Thanks for the review!

include/nighthawk/common/rate_limiter.h
```diff
@@ -49,6 +49,8 @@ class DiscreteNumericDistributionSampler {
 public:
   virtual ~DiscreteNumericDistributionSampler() = default;
   virtual uint64_t getValue() PURE;
+  virtual uint64_t min() const PURE;
```
mum4k: Can we document the new methods?

(optional / unrelated) Since we are here, can we document the entire interface?

source/common/rate_limiter_impl.cc
```diff
@@ -166,15 +163,32 @@ GraduallyOpeningRateLimiterFilter::GraduallyOpeningRateLimiterFilter(
...
-              return std::round(provider_->getValue() / 10000.0) <= chance_percentage;
+            const auto elapsed_time = elapsed();
+            if (elapsed_time < ramp_time_) {
+              // We want to linearly increase the probability of returning true
+              // below. We can derive that from the elapsed fraction of ramp_time.
+              const double probability =
+                  1.0 - static_cast<double>(ramp_time_.count() - elapsed_time.count()) /
+                            (ramp_time_.count() * 1.0);
+              // Get a random number r, where 0 < r ≤ 1.
+              const double random_between_0_and_1 = 1.0 * provider_->getValue() / provider_->max();
```
mum4k: Thank you, this is easier to follow with the new comments and the min() and max() methods.

test/rate_limiter_test.cc
```diff
@@ -174,23 +174,47 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplSchedulingTest) {
 
 class LinearRampingRateLimiterImplTest : public Test {
 public:
-  std::vector<int64_t> getAcquisitionTimings(const Frequency frequency,
-                                             const std::chrono::seconds duration) {
+  void checkAcquisitionTimings(const Frequency frequency, const std::chrono::seconds duration) {
```
mum4k: Can we document this helper? Specifically its arguments.

test/rate_limiter_test.cc
```diff
@@ -207,31 +231,65 @@ TEST_F(RateLimiterTest, LinearRampingRateLimiterImplInvalidArgumentTest) {
...
-  // rounding causing the first timing to fall into the next clock tick. This is
-  // a test only issue: in practice we don't have fixed microsecond step sizes,
-  // and the rate limiter computes at nanosecond precision internally. As we
-  // want to have microsecond level precision, this should be more then sufficient.
-  EXPECT_EQ(getAcquisitionTimings(5_Hz, 5s),
-            std::vector<int64_t>({1000050, 1732100, 2236100, 2645800, 3000000, 3316650, 3605600,
-                                  3873000, 4123150, 4358900, 4582600, 4795850, 5000000}));
-  EXPECT_EQ(getAcquisitionTimings(4_Hz, 2s),
-            std::vector<int64_t>({707150, 1224750, 1581150, 1870850}));
+  checkAcquisitionTimings(1_Hz, 3s);
```
mum4k: This is indeed much better for readability, but i am debating with myself whether it doesn't decrease our coverage. Could there later be a bug that would go undetected? Should we add at least one test case that verifies the exact values as before? WDYT?

oschaaf:
@mum4k added f480248669da8e5a839d23fd000b1c3113731406 based on your feedback, thanks!

test/rate_limiter_test.cc
oschaaf: yeah this makes sense, done (f480248669da8e5a839d23fd000b1c3113731406)

PR #219, opened 2019-11-27
Introduce Phases

oschaaf:
Introduces a concept of `Phases` into the system.

Phases represent named pieces of the overall benchmark execution.
For example, a warm-up phase could perform a ramp-up to a certain qps, and then a main phase could enable latency measurement and sustain the requested qps.

Currently this adds a warmup phase as an example, which is functionally on par with what we had before. Logical follow ups to this would:
- be per-phase reporting
- allowing phase configuration (generic or just adding a warmup/cooldown)
- possibly dynamically injecting phases on the fly (in response to a gRPC call and/or CLI input and/or special test-server responses)

Groundwork for #31

-----------

### Prerequisites

- [x] Land #281 and update this to leverage it


oschaaf:
This is ready for review -- but it's probably easier to do so when we've merged https://github.com/envoyproxy/nighthawk/pull/222 and merged master afterwards in here.

mum4k:

Some background about the vision of phases is documented in #246.

source/client/process_impl.cc
```diff
@@ -385,7 +385,7 @@ bool ProcessImpl::run(OutputCollector& collector) {
   int i = 0;
   std::chrono::nanoseconds total_execution_duration = 0ns;
   for (auto& worker : workers_) {
-    auto sequencer_execution_duration = worker->sequencer().executionDuration();
+    auto sequencer_execution_duration = worker->phases().back()->sequencer().executionDuration();
```
dubious90: Can you explain why it's safe to pick just the last phase for this?  Why don't we need the duration of all of the phases?

source/client/factories_impl.cc
```diff
@@ -47,16 +47,15 @@ SequencerFactoryImpl::SequencerFactoryImpl(const Options& options)
 
 SequencerPtr SequencerFactoryImpl::create(Envoy::TimeSource& time_source,
                                           Envoy::Event::Dispatcher& dispatcher,
-                                          Envoy::MonotonicTime start_time,
                                           BenchmarkClient& benchmark_client,
-                                          TerminationPredicate& termination_predicate,
-                                          Envoy::Stats::Scope& scope) const {
+                                          TerminationPredicatePtr&& termination_predicate,
+                                          Envoy::Stats::Scope& scope, const bool warmup) const {
```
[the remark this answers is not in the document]
oschaaf: My goal is to wire in the phase concept as a no-op, with test coverage; my idea is that in a followup we can optionize this as generic as we want to, eg allow arbitrary phase names and configs.. so one step at a time

source/client/process_impl.cc
oschaaf: See above; this is safe now because of some implicit knowledge abo this first step to land phases internally; in a follow up this has to change

include/nighthawk/common/phase.h
```diff
@@ -0,0 +1,50 @@
...
+ * High level, a worker statically configure a vector of phases, and will transfer the hot
+ * connection pool when transitioning between them. At this time, nothing is stopping us from
+ * dynamically injecting phases later, be it via grpc calls and/or live CLI input.
+ */
+class Phase {
+public:
+  virtual ~Phase() = default;
+
+  /**
+   * @return absl::string_view Contains the id of the pase. Should be unique but that is not
```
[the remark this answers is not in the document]
oschaaf: Fixed!

include/nighthawk/client/factories.h
```diff
@@ -37,9 +37,9 @@ class SequencerFactory {
 public:
   virtual ~SequencerFactory() = default;
   virtual SequencerPtr create(Envoy::TimeSource& time_source, Envoy::Event::Dispatcher& dispatcher,
-                              Envoy::MonotonicTime start_time, BenchmarkClient& benchmark_client,
-                              TerminationPredicate& termination_predicate,
-                              Envoy::Stats::Scope& scope) const PURE;
+                              BenchmarkClient& benchmark_client,
+                              TerminationPredicatePtr&& termination_predicate,
+                              Envoy::Stats::Scope& scope, const bool warmup) const PURE;
```
mum4k: Similar to our discussion in one of the previous PRs, how about we use an enum instead of the boolean flag argument? Looks like this boolean is used in a couple of interfaces that plumb the phases through. We might be able to reuse the enum in multiple locations.

This however makes me wonder how the end result is going to look like. If we go down the path of dynamically configured phases, what are we going to replace the boolean here with? Do we need a class carrying information about phases?

With that said - is this something we should leave to the next PR? Or would it be more beneficial to come up with a good structure now?

include/nighthawk/client/factories.h
```diff
@@ -72,8 +72,8 @@ class RequestSourceFactory {
 class TerminationPredicateFactory {
 public:
   virtual ~TerminationPredicateFactory() = default;
-  virtual TerminationPredicatePtr create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
-                                         Envoy::MonotonicTime time) const PURE;
+  virtual TerminationPredicatePtr create(Envoy::TimeSource& time_source,
```
mum4k: (optional) If you don't mind, can we document this interface since we are touching this? Optional, since this is unrelated to this PR.

include/nighthawk/common/phase.h
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * @return Sequencer& Sequencer associated to this phase.
+   */
+  virtual Sequencer& sequencer() const PURE;
+
+  /**
+   * @return bool Indicates if latencies should be tracked for this phase.
+   */
+  virtual bool measureLatencies() const PURE;
```
mum4k: (nit) The current name sounds like a verb (a command) and could be misread as a request to start measuring latencies. It helps to include the word "should" in boolean functions that are technically asking a question. How about: `shouldMeasureLatencies()` or similar?

include/nighthawk/common/phase.h
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Runs the sequencer associated to this phase and blocks until completion, which means this phase
+   * has ended as well.
+   */
+  virtual void run() const PURE;
```
mum4k: Doesn't the phase need to communicate any result status / error code, etc?

If there already is a way for us to examine the result, can we document it?

source/client/client_worker_impl.cc
```diff
@@ -17,34 +22,55 @@ ClientWorkerImpl::ClientWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Ins
...
-    dispatcher_->run(Envoy::Event::Dispatcher::RunType::RunUntilExit);
-  } else {
-    ENVOY_LOG(warn, "> worker {}: failed to initiate warmup request.", worker_number_);
-  }
-  ENVOY_LOG(debug, "> worker {}: warmup done.", worker_number_);
+  // We add a short warmup phase, which ends when either the first successful response
+  // is observed or two seconds have passed, whichever comes first. These warmup conditions are
+  // registered on top of the original configured predicates.
+  // TODO(oschaaf): allow configuration of this phase once the ramping rate limiters land,
+  // or a generic configuration of multiple phases (id, duration, termination predicates, rate
```
mum4k: (for the future) Instinctively I lean towards generic configuration of multiple phases, rather than built-in phase types, but I haven't given it as much thought as you did. Generic configuration for phases will allow users to utilize phases in ways that we didn't account for.

On the other hand the warmup phase is very likely going to be fairly common. Maybe we could come up with providing some sort of templating for command line / service arguments that we could ship with the final version of phases. I.e. instead of building the warmup phase (or others) into the code we could bake it into config templates. WDYT?

source/client/client_worker_impl.cc
```diff
@@ -69,7 +95,8 @@ void ClientWorkerImpl::shutdownThread() { benchmark_client_->terminate(); }
 StatisticPtrMap ClientWorkerImpl::statistics() const {
   StatisticPtrMap statistics;
   StatisticPtrMap s1 = benchmark_client_->statistics();
-  StatisticPtrMap s2 = sequencer_->statistics();
+  auto& sequencer = phases_.back()->sequencer();
```
mum4k: Can we prefer to mention the Sequencer type explicitly (instead of auto) unless it is too unwieldy?

source/client/client_worker_impl.h
```diff
@@ -34,23 +37,26 @@ class ClientWorkerImpl : public WorkerImpl, virtual public ClientWorker {
   const std::map<std::string, uint64_t>& thread_local_counter_values() override {
     return thread_local_counter_values_;
   }
-  const Sequencer& sequencer() const override { return *sequencer_; }
+
+  const std::vector<PhasePtr>& phases() const override { return phases_; }
```
mum4k: Can we document the new public method?

source/common/phase_impl.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include "nighthawk/common/phase.h"
+
+#include "external/envoy/source/common/common/logger.h"
+
+namespace Nighthawk {
+
+class PhaseImpl : public Phase, public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  PhaseImpl(absl::string_view id, SequencerPtr&& sequencer, bool measure_latencies)
```
mum4k: Can we document the constructor?

mum4k: The same comment about boolean flags applies here to `measure_latencies`. Would it be worth coming up with something more readable to replace it with?

source/common/sequencer_impl.cc
```diff
@@ -33,9 +34,7 @@ SequencerImpl::SequencerImpl(
 void SequencerImpl::start() {
   ASSERT(!running_);
   running_ = true;
-  if (start_time_ < time_source_.monotonicTime()) {
```
mum4k: Can you help me understand the history behind the evaluation that was deleted from here and lead to the error message "Sequencer start called too late"?

I am trying to figure out why it is no longer needed and how does it relate to phases.

source/common/sequencer_impl.cc
```diff
@@ -83,48 +82,43 @@ void SequencerImpl::updateStartBlockingTimeIfNeeded() {
 void SequencerImpl::run(bool from_periodic_timer) {
   ASSERT(running_);
   const auto now = last_event_time_ = time_source_.monotonicTime();
-  const auto running_duration = now - start_time_;
-
-  // The running_duration we compute will be negative until it is time to start.
-  if (running_duration >= 0ns) {
```
mum4k: Similarly here, can you help me understand the history behind this evaluation and why it is no longer necessary?

source/common/termination_predicate_impl.h
```diff
@@ -15,22 +17,34 @@ class TerminationPredicateBaseImpl : public TerminationPredicate {
     linked_child_ = std::move(child);
     return *linked_child_;
   }
+  TerminationPredicate& appendToChain(TerminationPredicatePtr&& child) final {
+    RELEASE_ASSERT(child != nullptr, "child == nullptr");
```
mum4k: I am wondering if this RELEASE_ASSERT follows the advice of the Envoy style guide:

https://github.com/envoyproxy/envoy/blob/master/STYLE.md

Wouldn't this crash in an obvious way if left without the assertion?

test/client_test.cc
```diff
@@ -37,7 +37,7 @@ TEST_F(ClientTest, AutoConcurrencyRun) {
   argv.push_back("--duration");
   argv.push_back("1");
   argv.push_back("--rps");
-  argv.push_back("1");
+  argv.push_back("10");
```
mum4k: For my understanding, why was this change necessary?

test/sequencer_test.cc
```diff
@@ -162,29 +165,14 @@ TEST_F(SequencerTestWithTimerEmulation, RateLimiterInteraction) {
   sequencer.waitForCompletion();
 }
 
-TEST_F(SequencerTestWithTimerEmulation, StartingLate) {
```
mum4k: This is probably related to one of my previous questions - it is not clear to me why / how was this behavior removed. Can you help me understand?

include/nighthawk/client/factories.h
oschaaf: I'm happy to iterate here. My goal with this is to land a decent chunk of the changes required to implement phases as a no-op, as an intermediate step to make review more fun. 
Later on we'll use them generically, we can also iterate on this in here in this PR here if you prefer. Or we can go for shipping this, and iterate on a second step in another one.

include/nighthawk/client/factories.h
oschaaf: Maybe I should just go over all the interfaces, and document everything that isn't in one go in a separate PR? I'd be happy to.

include/nighthawk/common/phase.h
oschaaf: I think the convention is that `field_name` translates into `(set)fieldName()` for the getter/setter. Possibly the naming is unfortunate here, for the `field` it is pretty clear, but for the associated getter/setter it turns out unfortunate. I'll amend.

include/nighthawk/common/phase.h
oschaaf: I'm mulling this over; we can query the stats and have this decoupled, but we could probably also return the last termination predicate result here.

source/client/client_worker_impl.cc
oschaaf: My instinct is the same; however because of the intermediate step performed here we're applying the generic phases concept in a very specific way to ensure we're doing a no-op. Hopefully we won't have hard coded build-in phases when we are done; I love the config templates idea.

source/common/sequencer_impl.cc
oschaaf: Ah I forgot about this; I'll ping you on Slack, this needs some explaining, and also some polish, as right now we're not doing a no-op. Let me try to share some context here: 

Today, when preparing the worker threads, the main thread computes the times at which load generation should start for each worker. When doing so it tries to carefully offset these timings so that requests that are released by each worker will be evenly spaced in time from a global perspective (we want to avoid that workers all release requests with exactly the same timings). 
This needs some thought now that we introduce phases as well as have the new rate limiters ready for wiring up as features.

include/nighthawk/client/factories.h
mum4k: Assuming you don't mind doing that, that would be well worth the time and would help us when reading the code base.

test/client_test.cc
oschaaf: It wasn't needed, accidentally left it in. Backed it out.

include/nighthawk/client/factories.h
oschaaf: I eliminated the warmup bool arg, to minimize changes that won't have future value as per earlier discussion.

include/nighthawk/client/factories.h
oschaaf: Filed https://github.com/envoyproxy/nighthawk/issues/268

include/nighthawk/common/phase.h
oschaaf: Renamed to `shouldMeasureLatencies`

include/nighthawk/common/phase.h
oschaaf: Added a doc comment here

source/client/client_worker_impl.cc
oschaaf: Backed some of this out: we retain the old warmup call, and have a single `phase` for now to minimize changes that won't have future value.

source/client/client_worker_impl.cc
oschaaf: Done

source/client/client_worker_impl.h
oschaaf: This one is now renamed. Doc comments related to this reside in the interface class, does that suffice?

source/common/phase_impl.h
oschaaf: Done

oschaaf: Done

source/common/sequencer_impl.cc
oschaaf: I made the `phase` execute the wait for the start offset of a worker. So we are doing a no-op again. Thanks for catching this!

mum4k:
@dubious90 agreed to take another look after the recent changes. Thank you Nate.

source/client/client_worker_impl.cc
```diff
@@ -17,18 +22,21 @@ ClientWorkerImpl::ClientWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Ins
...
       benchmark_client_(benchmark_client_factory.create(
           api, *dispatcher_, *worker_number_scope_, cluster_manager, http_tracer_,
           fmt::format("{}", worker_number), *request_generator_)),
-      termination_predicate_(
-          termination_predicate_factory.create(time_source_, *worker_number_scope_, starting_time)),
-      sequencer_(sequencer_factory.create(time_source_, *dispatcher_, starting_time,
-                                          *benchmark_client_, *termination_predicate_,
-                                          *worker_number_scope_)) {}
+      phase_(std::make_unique<PhaseImpl>(
+          "main",
```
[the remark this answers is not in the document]
oschaaf: No this shouldn’t matter for now, it will only show up in logs but otherwise is not visible externally.

source/common/phase_impl.h
```diff
@@ -0,0 +1,42 @@
...
+
+class PhaseImpl : public Phase, public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  /**
+   * @param id Unique identifier of the pase (uniqueness not enforced).
+   * @param sequencer Sequencer that will be used to execute this phase.
+   * @param should_measure_latencies Indicates if latencies should be tracked for requests issued
+   * during execution of this phase.
+   * @param time_source Time source that will be used to query the clock.
+   * @param start_time Optional starting time of the phase. Can be used to schedule phases ahead.
```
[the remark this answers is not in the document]
oschaaf: This is a re-implementation of something used today as a means to offset load test starts, to ensure request timings are evenly spaced from a global perspective. See https://github.com/envoyproxy/nighthawk/pull/219/files/c927a6ed9c006a0778813cf495c920408d1785bd#discussion_r363545616 for more context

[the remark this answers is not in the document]
dubious90: I'll leave this up to you and @mum4k for if that makes sense architecturally.  My concern is that we're defining an ordering mechanism directly into phases that doesn't seem to fit for phases, and we're repurposing it in ClientWorkerImpl for replacing a currently existing functionality that does not seem to have much to do with phases.

For whatever it's worth, that discussion link doesn't work, so I don't know what context you're referring to.

source/client/client_worker_impl.cc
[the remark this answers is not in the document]
dubious90: Just to clarify, does that mean this is temporary code that will be replaced later?  Or that the use of main here is a secret phase that outside users won't need to know about?  If it's the latter, I'm concerned about name collision, but if it's the first, then agree there's no concern here.

dubious90:
I've left a couple questions, but as they likely do not require changes, assigning back to @mum4k for approval

source/client/client_worker_impl.cc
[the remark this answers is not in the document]
oschaaf: It's temporary code. This will be replaced with a generic mechanism that translates phases into configuration. So the id chosen here can be an arbitrary value, perhaps I should have better called it `foo` or some such to avoid confusion.

source/common/phase_impl.h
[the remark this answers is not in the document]
oschaaf: The right link to the earlier discussion ought to have been https://github.com/envoyproxy/nighthawk/pull/219#discussion_r363545616. Somehow the previous one went bad, sorry. 
But point taken, somehow this doesn't fit well with me either, but so far I wasn't able to come up with something better. But reading back the linked discussion:

Maybe this should move into a different concept. I also had a concern with that this concept won't hold well with the new rate limiters we recently landed, and with non-probabilistic timing it doesn't make sense either. So maybe we should contain this functionality in a `DelayStartingRateLimiter` instead of here in `Phase`. How would you feel about that?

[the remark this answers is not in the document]
mum4k: I support your last suggestion @oschaaf, it makes the concept of phase more focused.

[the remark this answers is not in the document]
oschaaf: Great; I added #281 as a prerequisite to this PR

oschaaf:
This now leverages `ScheduledStartingRateLimiter`, and `PhaseImpl` has been cleaned up.

PR #220, opened 2019-11-28
Introduce Request / prepare dynamic request generation

oschaaf:
This adds infrastructure in preparation of
sending dynamic requests / validating request-level expectations.

No functional changes.

Prelude to #137, where add a feature which
streams  to-be replayed requests from a gRPC remote service method.

### DIFF
--- a/include/nighthawk/common/request.h
+++ b/include/nighthawk/common/request.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <functional>
+
+#include "envoy/http/header_map.h"
+
+namespace Nighthawk {
+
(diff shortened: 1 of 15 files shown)


include/nighthawk/common/request.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
htuch: Do you think we should have more directory structure than `common`?

oschaaf: Yes I agree; I opened an issue to reorganize source file layout a while ago. Just have not got around do proposing a new structure yet

PR #222, opened 2019-12-03
Update Envoy dep for access to enableHRTimer()

oschaaf:
Prerequisite to https://github.com/envoyproxy/nighthawk/pull/217


oschaaf:
Note: the prerequisite in Envoy changed to https://github.com/envoyproxy/envoy/pull/9229 as we have had to back out the first attempt. It probably makes sense to update this PR to use the forthcoming Envoy revision that has that, after it lands.. Sorry, should have remarked about that earlier here.

oschaaf:
This is ready for review (again).
I updated to the Envoy revision that has `enableHRTimer()` and also amended what broke while doing so.

PR #223, opened 2019-12-06
rate-limiting: test enhancement + refactoring 

oschaaf:
Modulo an improved test for the batching rate limiter,
this pulls off some refactoring for upcoming new rate
limiters.

This has no functional changes, and is an intermediate step for draft PR #218,
which contains new rate limiters (linear & probabilistic ramping, first stab at a zipf distribution support).


source/common/rate_limiter_impl.cc
```diff
@@ -81,38 +78,45 @@ void LinearRateLimiter::releaseOne() {
...
-    : random_distribution_generator_(std::move(random_distribution_generator)),
-      time_source_(time_source), rate_limiter_(std::move(rate_limiter)) {}
+DelegatingRateLimiterImpl::DelegatingRateLimiterImpl(
+    RateLimiterPtr&& rate_limiter, RateLimiterDelegate random_distribution_generator)
+    : ForwardingRateLimiterImpl(std::move(rate_limiter)),
+      random_distribution_generator_(std::move(random_distribution_generator)) {}
 
-bool DelegatingRateLimiter::tryAcquireOne() {
+bool DelegatingRateLimiterImpl::tryAcquireOne() {
+  const auto now = timeSource().monotonicTime();
```
dubious90: Is there a slight functional change here?  I'm not 100% sure I understand how monotonicTime() works here, but because you're pulling this value before tryAcquireOne here, if tryAcquireOne were to ever have a delay for any reason (even if only in certain impls), what `now` is capturing here is not equivalent to what `now` was capturing before.

If you don't think this is a legitimate concern, I will completely defer to you on this.

source/common/rate_limiter_impl.h
```diff
@@ -14,6 +14,60 @@
...
+  int64_t acquireable_count_{0};
+  uint64_t acquired_count_{0};
+  const Frequency frequency_;
+};
+
+/**
+ * Base for a rate limiter which wraps another rate limiter, and forwards
+ * some calls.
+ */
+class ForwardingRateLimiterImpl : public RateLimiter {
```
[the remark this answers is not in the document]
oschaaf: I can see your point, but I have a hard time coming up with a better name; hopefully the comment
I added in 6d1e7471261e67bbf0b083d22135dff210fcfa95 help

source/common/rate_limiter_impl.cc
oschaaf: Nice, catch; I backed this out. 
The behavioral change is quite small but as the PR description promised to not introduce these type of changes, this shouldn't go in here - and probably there is no value in making the change at all.

dubious90: Thank you!

source/common/rate_limiter_impl.h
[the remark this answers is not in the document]
dubious90: I think that your added comment clarifies things nicely.  Thank you

dubious90:
LGTM, htuch to approve

include/nighthawk/common/rate_limiter.h
```diff
@@ -24,6 +28,12 @@ class RateLimiter {
    * Releases a controlled resource.
    */
   virtual void releaseOne() PURE;
+
+  /**
+   * @return Envoy::TimeSource& time_source
+   */
+  virtual Envoy::TimeSource& timeSource() PURE;
+  virtual std::chrono::nanoseconds elapsed() PURE;
```
htuch: Can we add more docs here? Like, what does `elapsed` do and what is it good for?

oschaaf: Done in 4d6448b1e80306c0c20ef3dc1c63ba16fec5f2ee

PR #224, opened 2019-12-08
Optimization: static request body vs memory copying

oschaaf:
We always send the same request body today.
It is possible to optimize that case and reduce copies.
This greatly speeds up achieved transfer rates as well as allows
much higher concurrencies to be used when testing uploads.
(This was initiated based on a field report containing memory
exhaustion symptoms)


source/client/stream_decoder.cc
```diff
@@ -87,11 +87,15 @@ void StreamDecoder::onPoolReady(Envoy::Http::StreamEncoder& encoder,
     // we take the risk of erroneously reporting that we did send all the bytes, instead of always
     // reporting 0 bytes.
     stream_info_.addBytesReceived(request_body_size_);
-    // TODO(oschaaf): We can probably get away without allocating/copying here if we pregenerate
-    // potential request-body candidates up front. Revisit this when we have non-uniform request
-    // distributions and on-the-fly reconfiguration in place.
-    std::string body(request_body_size_, 'a');
-    Envoy::Buffer::OwnedImpl body_buffer(body);
+    // Revisit this when we have non-uniform request distributions and on-the-fly reconfiguration in
```
mum4k: On the fly reconfiguration is something we are after, so this is probably a temporary performance increase. How are we going to deal with the performance regression that will come with switching this back to copying?

Are we going to make this optional and only copy on reconfiguration? Do you have other plans? I am worried about users starting to depend on this newly increased performance if we are going to end up reverting it.

oschaaf: I guess that what is missing is an abstraction to source the actual contents we are sending. We only have a content-length concept, which will work well even when that requested content-length changes dynamically. But today there's no way in place to change the actual content that we are sending. The series of 'a' characters we use today probably is too simple / unfit for certain purposes (for one that is highly compressible), maybe we should take this to slack for a bit and discuss where we want to take this?

source/client/stream_decoder.cc
```diff
@@ -87,11 +87,15 @@ void StreamDecoder::onPoolReady(Envoy::Http::StreamEncoder& encoder,
...
+    // place. The string size below MUST match the cap we put on RequestOptions::request_body_size
+    // in api/client/options.proto!
```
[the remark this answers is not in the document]
oschaaf: sounds like a great idea to me, consider it done :-)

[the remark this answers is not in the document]
oschaaf: Done in 2a5a787368360adc163417257fa884c0c99a4e50

source/client/stream_decoder.cc
mum4k: Discussed offline and the questions were answered, this won't prevent us from implementing any dynamic body functionality.

source/client/stream_decoder.cc
```diff
@@ -87,11 +87,15 @@ void StreamDecoder::onPoolReady(Envoy::Http::StreamEncoder& encoder,
...
+    static const std::string body(4194304, 'a');
```
htuch: Due to static initialization fiasco, I recommend we follow Envoy practice around `CONSTRUCT_ON_FIRST_USE`

oschaaf: > Due to static initialization fiasco, I recommend we follow Envoy practice around `CONSTRUCT_ON_FIRST_USE`

I have a surprisingly hard time getting that to work; the macro forwards the args to an initializer list.
```
    CONSTRUCT_ON_FIRST_USE(std::string, 4194304, 'a');
```

translates into this build error 

```
n file included from source/client/stream_decoder.cc:1:
bazel-out/k8-dbg/bin/source/client/_virtual_includes/nighthawk_client_lib/client/stream_decoder.h:85:41: error: implicit conversion from 'int' to 'const char' changes value from 4194304 to 0 [-Werror,-Wconstant-conversion]
    CONSTRUCT_ON_FIRST_USE(std::string, 4194304, 'a');
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
bazel-out/k8-dbg/bin/external/envoy/source/common/common/_virtual_includes/macros/common/common/macros.h:34:45: note: expanded from macro 'CONSTRUCT_ON_FIRST_USE'
    static const type* objectptr = new type{__VA_ARGS__};                                          \
```

htuch: Not sure why it's not doing the right thing here, probably some constructor override issue interacting with templating etc. You can always implement the `CONSTRUCT_ON_FIRST_USE` idiom by hand in this file.

oschaaf: Done in 1bb39a09507f676f009a2bd62548e9f3daf8db62

source/client/stream_decoder.h
```diff
@@ -81,6 +81,10 @@ class StreamDecoder : public Envoy::Http::StreamDecoder,
   streamResetReasonToResponseFlag(Envoy::Http::StreamResetReason reset_reason);
   void finalizeActiveSpan();
   void setupForTracing(std::string& x_request_id);
+  static const std::string& staticUploadContent() {
+    static const std::string s(4194304, 'a');
```
htuch: This needs to be a `new` and a pointer as per `CONSTRUCT_ON_FIRST_USE`, otherwise you have static destruction fiasco.

oschaaf: I read up over at https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2 and followed the approach for a when a non-trivial destructor is involved -- however, reading more, it's not like the `std::string` is needs to write to a file or some such in its destructor, so I'll amend. Sorry about the noise.

oschaaf: Amended

source/client/stream_decoder.h
```diff
@@ -81,6 +81,10 @@ class StreamDecoder : public Envoy::Http::StreamDecoder,
   streamResetReasonToResponseFlag(Envoy::Http::StreamResetReason reset_reason);
   void finalizeActiveSpan();
   void setupForTracing(std::string& x_request_id);
+  static const std::string& staticUploadContent() {
```
htuch: Nit: should be private

oschaaf: Done: cb36e1fabb0018fc2a747d5c5e40ef047a3ad962

PR #226, opened 2019-12-11
Fortio output: add missing fields & correct some of the output fields

### LINKED ISSUE #186: Fortio numthreads field is inaccurate
In reference to the discussion in #168 

The field should probably be set as the number of concurrent streams (governed by `--max-active-requests`), multiplied by the number specified in concurrency. However, `--concurrency` can be set to `auto` in the CLI, making it difficult for the output transform code to read it.

The output transform code shouldn't use the number of CPUs on the machine when `concurrency == 'auto'` because it may run on a different machine.

As a prerequisite: We need to offer an "effective concurrency" field in the NH output proto. Then the output transform code will not need to infer the concurrency.

oschaaf:
# PR Description

- Define NH version in the code in a canonical place.
- Update the fortio output formatter - add missing version field
- Consume the new version information in various places to make
  `--version`  output of executables sensible.
- Implement a couple of missing fields in the Fortio output
- Amend a couple of fields for correctness

Fixes #186, #235, #231

# Prerequisites to merging this
- [x] Move RELEASE_PROCEDURE.md into its own PR
- [x] Land envoyproxy/envoy#9301
- [x] [WIP] Update the NH Envoy dependency so we can use ^^:  #261
- [x] Update this PR to leverage the structured versioning that now is available
- [x] Fix missing stuff in the Fortio output, punt what cannot be completed yet to TODO/issues and/or task lists here. 
- [X] amend inaccurate things
- [x] ~~Discuss that the label field in Fortio output also contains machine name, we set it to `Nighthawk`. Check which one is preferrable~~.
- [x] Get consensus on what we want to show for the concurrency value in the Fortio visualization, [discussion](https://github.com/envoyproxy/nighthawk/issues/186)
- [x] use experimental pool features (h1 connection re-use strategy, h2 multiple connection support)


RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,7 @@
+# Release procedure
+
+- Update [version_history.md](docs/root/version_history.md). 
+  - Make sure breaking changes are explicitly called out.
+  - Ensure the release notes are complete.
+- Tag the release in git using the current version number
```
mum4k: Would it help if we include example commands for some of these steps?

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,7 @@
+# Release procedure
+
+- Update [version_history.md](docs/root/version_history.md).
```
mum4k: (optional) Not sure if worth it (depends on how often this happens). Would it make sense to add a script that will automate the version update and the version "Bump" noted below?

include/nighthawk/common/version.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "absl/strings/str_cat.h"
+
+namespace Nighthawk {
+
+const int MAJOR_VERSION{0};
+const int MINOR_VERSION{3};
+
+class Globals {
```
mum4k: Curious - why wrap it in a class?

Considering its generic name - won't this attract more possibly unrelated methods?

source/client/output_formatter_impl.cc
```diff
@@ -187,6 +188,7 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
 
   // TODO(#182): Not needed but nice to have, displays in the UI
   fortio_output.set_labels("");
+  fortio_output.set_version("0.0");
```
mum4k: Don't we want to set this to the actual version?

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,11 @@
+# Release procedure
+
+- Update [version_history.md](docs/root/version_history.md). 
+  - Make sure breaking changes are explicitly called out.
```
htuch: Ideally we do these as we go, rather than a major effort at release time. The experience form Envoy has been that this scales better.

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,11 @@
...
+  - Ensure the release notes are complete.
+- Tag the release in git using the current version number:
+```bash
+git tag -a "v0.3" -m "Release v0.3"
+git push origin v0.3
```
htuch: Do we need to do the tagging? If we do a GitHub tagged release, does it generate these for us?

include/nighthawk/common/version.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "absl/strings/str_cat.h"
+
+namespace Nighthawk {
+
+const int MAJOR_VERSION{0};
+const int MINOR_VERSION{3};
```
htuch: Do you want full semantic major.minor.patch style versions?

api/client/transform/fortio.proto
```diff
@@ -34,6 +34,8 @@ message FortioResult {
   map<string, uint64> RetCodes = 11;
 
   string URL = 12;
+
+  string Version = 13;
```
htuch: You could make this a structured type, @yanavlasov is introducing this in https://github.com/envoyproxy/envoy/pull/9301

RELEASE_PROCEDURE.md
oschaaf: Those currently exist, so if that is what you did earlier, then yes :-) I'll amend

RELEASE_PROCEDURE.md
oschaaf: I'll make sure we mention that here - though maybe we could also have a pre-merge reviewer checklist that captures this; when someone about to perform a release read it here it's probably too late?

api/client/transform/fortio.proto
oschaaf: +1

include/nighthawk/common/version.h
oschaaf: Lets go for that, I'll amend

oschaaf:
@mum4k @htuch I addressed feedback and iterated on the release procedure; 
As we can only land this after envoyproxy/envoy#9301 goes in, it's a nice time to
discuss this some more, so please let me know what you think! I tried to capture
some of the things we did in the earlier releases, as well as added some ideas.

RELEASE_PROCEDURE.md
htuch: Sure. Could be useful for new maintainers, +1

oschaaf: > Sure. Could be useful for new maintainers, +1

Filed https://github.com/envoyproxy/nighthawk/issues/230, let me know if you have any more things that should go in there.

source/client/output_formatter_impl.cc
oschaaf: > Don't we want to set this to the actual version?

I'm sorry, I had another look at this, and I somehow misunderstood your comment here.
That is correct, and this is fixed now.

oschaaf:
This is now in flux for a bit, because I noticed a couple of issues as wel as some fields missing in the output. As this is blocked on waiting for Envoy to land the structured versioning proto definition I may as well go for getting the output sorted completely while I'm at it.

RELEASE_PROCEDURE.md
```diff
@@ -0,0 +1,17 @@
+# Release procedure
```
oschaaf: Note to self - let's move this out of this PR and into a new one

oschaaf: Done

oschaaf:
@envoyproxy/nighthawk-reviewers This is now in a state again where we can pursue reviewing it, and has most of the things to get the Fortio output right. Some notes:

- I updated description of this PR with a checklist so we can track status.
- We need to discuss some fields that Fortio emits in its output, but are not used in its visualizations. Do we want Nighthawk to emit those? (details in the PR description).
- Setting the `jitter` field to the correct value cannot be done that as that depends on #227. I punted that to #235

api/client/transform/fortio.proto
```diff
@@ -42,6 +47,18 @@ message DurationHistogram {
...
+
+  double Max = 4;
+
+  double Sum = 5;
+
+  double Avg = 6;
+
+  double StdDev = 7;
+
+  repeated FortioPercentile Percentiles = 8;
```
nareddyt: Does this message have to be prefixed with `Fortio`, since it's already in the `FortioResult` proto? I guess it makes it easier to use in the transform code, where you have Fortio and NH percentiles.

oschaaf: Well, there’s already a ‘Percentile’ message defined in NH’s own output proto and using the same name here broke the build.

include/nighthawk/common/version.h
```diff
@@ -0,0 +1,18 @@
...
+const int PATCH_VERSION{0};
+
+class VersionUtils {
```
mum4k: Sorry I wasn't clear enough. I was wondering if this even needs to be in a class? How about a plain top-level function?

htuch:
@oschaaf should this be "waiting" or are you after another review regardless of blocked status?

oschaaf:
> @oschaaf should this be "waiting" or are you after another review regardless of blocked status?

See https://github.com/envoyproxy/nighthawk/pull/226#issue-351884360 for context. We can still review the current state, and polishing PR state will mean progress, but we cannot finalize this at time.

include/nighthawk/common/version.h
oschaaf: Ah, sure! How about 7de573f15e7d85ee12d8364d88880a7c18cf9e06?

oschaaf:
Note fore reviewers: It's better to land https://github.com/envoyproxy/nighthawk/pull/261 first, as I merged that into this one so I could take support for structured versioning onwards.

source/client/output_formatter_impl.cc
```diff
@@ -225,14 +266,11 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
...
-  // Set the count (number of data points)
   nighthawk::client::DurationHistogram fortio_histogram;
-  fortio_histogram.set_count(nh_stat.count());
-
   uint64_t prev_fortio_count = 0;
   double prev_fortio_end = 0;
-  for (int i = 0; i < nh_stat.percentiles().size(); i++) {
-
+  const int percentiles_size = nh_stat.percentiles().size();
+  for (int i = 0; i < percentiles_size; i++) {
```
jplevyak: You can use a range-based for loop here.

source/common/version_info.cc
```diff
@@ -0,0 +1,46 @@
...
+#include "absl/strings/string_view.h"
+
+namespace Nighthawk {
+
+const std::string& VersionInfo::version() {
+  CONSTRUCT_ON_FIRST_USE(std::string, NIGHTHAWK_BUILD_VERSION_NUMBER);
+}
+
+const envoy::config::core::v3alpha::BuildVersion& VersionInfo::buildVersion() {
+  static const auto* result = new envoy::config::core::v3alpha::BuildVersion(
```
jplevyak: Can you use CONSTRUCT_ON_FRIST_USE here as well?

oschaaf:
@jplevyak feedback addressed in 13b5222fc91fc30cbc6b7eec7d12cc7166344d9f

source/common/version_info.h
```diff
@@ -0,0 +1,25 @@
...
+namespace Nighthawk {
+
+/*
+ * TODO(#267): This class is heavily based on Envoy's source/common/common/version.h
+ * There's some code duplication going on as we cannot easily directly fully
+ * re-use it. It would be great if we can make that happen.
+ */
+class VersionInfo {
+public:
+  static const std::string& version();
```
mum4k: Can we document these public methods?

api/client/transform/fortio.proto
```diff
@@ -41,6 +41,20 @@ message FortioResult {
   map<string, uint64> RetCodes = 11;
 
   string URL = 12;
+
+  string Version = 13;
```
mum4k: (optional) Unsure what role this proto plays. If it is "public" by any means we should improve the documentation of these fields. We can leave this to a follow-up PR.

mum4k:
Discussed offline, we will address the documentation nits in another PR. This one going in will unblock two other PRs.

api/client/transform/fortio.proto
oschaaf: Followed up on the requested doc comments in 77a7eb267423cac7b15cde7a2a2c83e699177756 over at #265

PR #227, opened 2019-12-11
Optionize uniformly distributed request-release timing jitter 

oschaaf:
Adds `--jitter-uniform .001s`, which provides a way to add jitter to request-release timings.

Implemented by wiring up and optionizing the distribution sampling rate limiter that landed earlier.


README.md
```diff
@@ -68,6 +69,11 @@ format>
 
 Where:
 
+--jitter-uniform <duration>
+Add uniformly distributed absolute uniformly distributed
```
mum4k: (nit) Doesn't seem to read well (uniformly distributed repeats twice). Unless I am misunderstanding the sentence.

source/client/options_impl.cc
```diff
@@ -184,7 +184,11 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "Enable open loop mode. When enabled, the benchmark client will not provide backpressure "
       "when resource limits are hit.",
       cmd);
-
+  TCLAP::ValueArg<std::string> jitter_uniform(
+      "", "jitter-uniform",
+      "Add uniformly distributed absolute uniformly distributed request-release timing jitter. For "
```
mum4k: (nit) Looks like "uniformly distributed" is doubled here as well.

README.md
oschaaf: Ah, I messed this up with a bad replacement, let me amend that.

oschaaf: Done!

api/client/options.proto
```diff
@@ -114,4 +114,6 @@ message CommandLineOptions {
   map<string, uint64> failure_predicates = 24;
   // See :option:`--open-loop` for details.
   google.protobuf.BoolValue open_loop = 21;
+  // See :option:`--jitter-uniform`
+  google.protobuf.Duration jitter_uniform = 25 [(validate.rules).duration.lt.seconds = 1];
```
htuch: Why this limit? I don't think we need to be more opinonated than what the implementation dictates in the API.

source/client/options_impl.cc
```diff
@@ -244,6 +248,19 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
   parsePredicates(failure_predicates, failure_predicates_);
   TCLAP_SET_IF_SPECIFIED(open_loop, open_loop_);
+  if (jitter_uniform.isSet()) {
+    Envoy::ProtobufWkt::Duration duration;
+    if (Envoy::Protobuf::util::TimeUtil::FromString(jitter_uniform.getValue(), &duration)) {
+      if (duration.nanos() < 0 || duration.seconds() != 0) {
+        throw MalformedArgvException("--jitter-uniform is out of range");
+      }
+      if (duration.nanos() > 0) {
+        jitter_uniform_ = std::chrono::nanoseconds(duration.nanos());
```
htuch: I think there are utils for this conversion already in Envoy.

api/client/options.proto
oschaaf: My reasoning here is that higher values may be indicative of accidental bad input, I think the usual input here is some fraction of the targetted frequency. Our minimal rps is 1/s..
Do you think we will miss usecases when disallowing >1s uniform jitter? 

ps - The seconds part of the proto can be disregarded when staying below 1 second, which is why below the conversion api isn't used.

oschaaf: I'm open to changing this, otherwise I'll add a comment summarizing the reasoning here, let me know which one you prefer

htuch: It's probably fine to limit to a second. I don't think in general though our philosophy is to prevent the user misconfiguring via PGV limits beyond the implementation preconditions; we let folks pick their own adventure, maybe they have a legitimate unforeseen need for some esoteric config value.

oschaaf: fixed!

PR #229, opened 2019-12-12
gRPC service API changes for sourcing request data

oschaaf:
Split out from https://github.com/envoyproxy/nighthawk/pull/137
for separate review.


mum4k:
Thank you @oschaaf for splitting out the API for a more focused discussion.

@eric846, @dubious90 please take a look and comment in relation to the testing framework we are building. I know that @dubious90 is also putting together a document of use cases that we will need, which will help us navigate the API in the required direction. We will share that to better explain our requirements for this API.

api/client/service.proto
```diff
@@ -34,10 +36,59 @@ message ExecutionResponse {
...
+
 service NighthawkService {
   //  Accepts a stream of ExecutionRequests, which can be used to initiate, update, or cancel a
   //  test. Returns a stream of ExecutionResponse objects which will contain results for all
   //  Nighthawk runs.
   rpc ExecutionStream(stream ExecutionRequest) returns (stream ExecutionResponse) {
   }
+
+  // Obtains a stream request-level specifications plus expectations.
+  rpc RequestStream(stream RequestStreamRequest) returns (stream RequestStreamResponse) {
```
mum4k: To start a discussion - one point we were wondering about is whether this new RPC method would work better under a separate (new) service definition.

If I understand the architecture correctly, Nighthawk is meant to connect to some remote service to request the stream of request specifiers for the test run. Let's call that remote service the "traffic source". If this is true, we are really dealing with two binaries, Nighthawk itself and the traffic source.

Nighthawk should implement the ExecutionStream while the traffic source should implement the RequestStream. The traffic source shouldn't need to implement the ExecutionStream rpc, but any implementations of it will have to implement it to satisfy the service definition.

If I do have correct understanding - would it make sense to start by defining a separate service (e.g. TrafficSourceService) for the traffic source binary to implement? Alternatively please correct my understanding.

oschaaf: A separate service absolutely makes sense to me. A separate binary as well.

api/client/service.proto
```diff
@@ -34,10 +36,59 @@ message ExecutionResponse {
...
+
+message RequestSpecifier {
+  // method/path/authority/content_length can all be expressed in request headers.
+  // so technically they are redundant. In case of ambiguity, the explicit fields will prevail.
+  // Note that as of today, the target upstream/cluster that will be used to replay traffic to
+  // fixes the destination address/port.
+  // Request method
+  google.protobuf.StringValue method = 1;
+  // Request path
+  google.protobuf.StringValue path = 2;
```
dubious90: Just want to make sure I understand - Is the idea behind `path` that it would be appended to the URIs?

api/client/service.proto
```diff
@@ -34,10 +36,59 @@ message ExecutionResponse {
   google.rpc.Status error_detail = 7;
 }
 
+// Used to request a RequestStreamResponse.
+message RequestStreamRequest {
+  // Amount of RequestStreamResponses that should be yielded.
+  // Used to implement basic flow control by the client. Once AsyncClientImpl gets
+  // a way to apply backpressure this could probably be dropped.
+  uint64 amount = 1;
```
dubious90: can you clarify in this comment what this is the amount of?

My assumption is number of requests to specify.  Is the contract on the server side supposed to be that that's how many responses they'll stream back as output for that one request?

api/client/service.proto
```diff
@@ -34,10 +36,59 @@ message ExecutionResponse {
...
+  envoy.api.v2.core.HeaderMap headers = 5;
+  // TODO(oschaaf): nice to have
+  // google.protobuf.StringValue sni_hostname = 10;
+}
+
+// Response-level expectations. If set, any unmet expectations will result in stats counter
+// increases in the future.
+message Expectations {
+  // Expected response status code
+  google.protobuf.UInt32Value response_code = 1;
```
dubious90: Is there already default behavior in NH around response_code validation?  In terms of requiring 200 responses or 2xx responses?  Are there possible scenarios where a consumer might want to specify not a specific response_code but a response_code range or array of possible response_codes?

I'm not positive that this design isn't sufficient, but just want to pose those questions just in case.

api/client/service.proto
```diff
@@ -34,10 +36,59 @@ message ExecutionResponse {
...
+  // Request path
+  google.protobuf.StringValue path = 2;
+  // Request authority
+  google.protobuf.StringValue authority = 3;
+  // Request content length. The client will transfer the number of bytes specified here for the
+  // request body.
+  google.protobuf.UInt32Value content_length = 4;
+  // Request header replacements. Any existing header(s) with the same name will be removed
+  // before setting.
+  envoy.api.v2.core.HeaderMap headers = 5;
```
dubious90: Considering this spec is likely to become a source of truth for people to refer to, would it be possible to link to a reference of what headers are supported by envoy/NH backend?  Does such a list exist somewhere?

oschaaf: Well, I guess this can take arbitrary string/string combinations. But there's quite a few standardized and semi-standardized headers, e.g.: https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

api/client/service.proto
oschaaf: Recently we landed the `termination-predicate` feature, which allows one to configure exit and failure conditions. It's a pretty basic implementation: you can state a counter threshold, and what happens when that gets exceeded: termination or failure. We atomically introduced default failure predicates to quickly stop upon observing connection failures and / or http error status codes. When one explicitly configures these predicates, that will override the default ones.
The idea was that this feature could also serve here; we track failed request-level expectations by bumping one or more counters, and then one could configure termination/failure-predicates to specify the exit condition and status.

api/client/service.proto
oschaaf: The path will override whatever path is configured on the remote side if specified here.

oschaaf:
@dubious90 I changed/added some comments based on your feedback in e38b19d2835f360f47e814b3ce172b92675f11d4

api/client/service.proto
dubious90: My initial comment was really unclear, and I'm sorry.  I did some more digging to explain what I really mean. What I thought would be helpful here would be to callout in a comment some pointer to documentation on x-nighthawk-test-server-config which I think has its options defined here: https://github.com/envoyproxy/nighthawk/blob/master/api/server/response_options.proto, so that users could understand what might be useful to specify here.  However, I now see that content_length is currently the only variable controlled by nighthawk-specific headers.  Is there an idea that we expect to implement more features like that?  If not, probably no need to call it out here, because you've abstracted that behavior away from the headers anyway

api/client/service.proto
dubious90: Thank you for the explanation!

api/client/service.proto
dubious90: Actually, looking again, I think that the content_length specified about is request length, not response length, so I do think providing a pointer to response_options.proto would probably be helpful, if you agree.

oschaaf: Oh I see, that would be when we're configured to send these request headers to our own test server, which may interpret some of the request headers as configuration directives. In that case there's quite a bit of inception going on when we talk about that here :-) To be honest, I'm hesitant about documenting here because it is not written in stone that we'll be talking to the test server, and because of the inception it might take a lot of stack space to grok what we're talking about? 
I think we need some more high level documentation / cookbooks / getting started. There we could document
setups and describe the flow of what happens where, and this could fit more naturally. WDYT?

dubious90: That's a really good point. I wasn't thinking about other consumers who may not be using the NH Test Server.  Having a recipe book of sorts, some high level documentation would be really awesome.  I'm fine with omitting it here in that case.

dubious90:
LGTM

eric846:
LGTM

api/client/service.proto
```diff
@@ -34,10 +36,62 @@ message ExecutionResponse {
...
+
+  // Request method
+  google.protobuf.StringValue method = 1;
+  // Request path
+  google.protobuf.StringValue path = 2;
+  // Request authority
+  google.protobuf.StringValue authority = 3;
+  // Request content length. The client will transfer the number of bytes specified here for the
+  // request body.
+  google.protobuf.UInt32Value content_length = 4;
```
htuch: Why is this wrapped? Doesn't i have a default value of 0?

api/client/service.proto
```diff
@@ -34,10 +36,62 @@ message ExecutionResponse {
...
+  // google.protobuf.StringValue sni_hostname = 10;
+}
+
+// Response-level expectations. If set, any unmet expectations will result in stats counter
+// increases in the future.
+message Expectations {
+  // Expected response status code
+  google.protobuf.UInt32Value response_code = 1;
+  // Expected response body content length
+  google.protobuf.UInt32Value content_length = 2;
```
htuch: Ditto here. We should only wrap for non-zero default values (and comment what they are).

api/client/service.proto
```diff
@@ -34,10 +36,62 @@ message ExecutionResponse {
   google.rpc.Status error_detail = 7;
 }
 
+// Used to request a RequestStreamResponse.
+message RequestStreamRequest {
+  // Indicates the amount of RequestStreamResponses that should be yielded by the server
+  // in response to this request.
+  // Used to implement basic flow control by the client. Once AsyncClientImpl gets a way
+  // to apply backpressure this could probably be dropped.
+  uint64 amount = 1;
```
htuch: Nit: I would probably find `quantity` a bit clearer.

api/client/service.proto
```diff
@@ -34,10 +36,62 @@ message ExecutionResponse {
   google.rpc.Status error_detail = 7;
 }
 
+// Used to request a RequestStreamResponse.
+message RequestStreamRequest {
+  // Indicates the amount of RequestStreamResponses that should be yielded by the server
+  // in response to this request.
+  // Used to implement basic flow control by the client. Once AsyncClientImpl gets a way
+  // to apply backpressure this could probably be dropped.
```
htuch: Non-actionable, but one thing you'll need to watch out for is potential coordination between Nighthawk when it's under load and backed up and the gRPC request stream. There may be situations in which we're applying backpressure that somehow related to the system-under-test. Not sure how likely this is, probably the best practice is to always ensure that Nighthawk itself is not max loaded, but is kept to something like 50% or 80% of CPU.

oschaaf: Good point; putting some do's and dont's out there in documentation will surely be helpful. Cross linking #241 so we'll remember this particular comment. It might also be a good practice to use different network cards for sourcing data and replaying.  Also leaving a tangentially related thought here so I don't forget: hypothetically the replay service might not keep up. In the WIP PR associated to this that will be tracked as blocking. This might be either valid (the gRPC data source implements a poor-mans version of request-release timing, or problematic when it's not intentional). This should be addressed at the `RequestSource` level as it's not specific to gRPC. Possibly a special `--rps` value to enable users to explicitly express intent for `as fast as possible` would help here, as we can then discern between problematic and (probably) intentional scenarios.

api/client/service.proto
oschaaf: Done in d43efd24a7d9991aca07b559dccd606c033f3322

api/client/service.proto
oschaaf: The other side overrides its configured values with inbound values originating from this. 
While doing so it needs to be able to discern between not-set and `0`. 
Concretely, if we omit `content_length` here, the idea is that the other side will use its configured value and
not have it overridden to `0`.

api/client/service.proto
mum4k: Can we make the corresponding change (splitting this out into a standalone service definition) before we move forward with this PR?

oschaaf:
> Looks good, except the comment about splitting the RequestStream RPC method into its own service definition.
> 
> Once addressed, I think we can proceed by merging this PR in without the implementation so that we can start building on top of this API in parallel.

Thanks for reminding me, I forgot about that, sorry, let me fix that right away. Also, shall I move `service.proto` out of `api/client/` and into `api/service`?

api/client/service.proto
oschaaf: Never mind moving the service definition into `api/service/`, that's probably a breaking change. 
Split into two separate service definitions in 8065182ef35c1eda851107750d7a9f77941251da

mum4k: Moving this new service out of the client subdirectory actually isn't a bad idea, because it doesn't logically fit into what is understood as the "client".

Can you help me understand why would it be a breaking change? The new service comes with a set of exclusively new messages that don't seem to have any dependencies on other types defined in the client protos. Am I missing something?

If we do end up moving it, I would recommend `api/request_source` since `api/service` feels bit too generic.

oschaaf: To be honest, I'm not sure what I was thinking, this wouldn't be a breaking change. Done in 1dd144a99ea5af9025af0d705d81e660659d68e7

api/client/service.proto
```diff
@@ -2,8 +2,10 @@ syntax = "proto3";
 
 package nighthawk.client;
 
+import "google/protobuf/wrappers.proto";
```
mum4k: Are these new imports still necessary now that we are adding a new proto file?

(wrappers.proto and base.proto)

api/request_source/service.proto
```diff
@@ -0,0 +1,60 @@
...
+// increases in the future.
+message Expectations {
+  // Expected response status code
+  google.protobuf.UInt32Value response_code = 1;
+  // Expected response body content length
+  google.protobuf.UInt32Value content_length = 2;
+}
+
+service NighthawkRequestSourceService {
+  // Obtains a stream request-level specifications plus expectations.
```
mum4k: Possible typo: Obtains a stream of ...

api/client/service.proto
oschaaf: they can be removed indeed, done in 944259382bb86f75316099c256b964de492b48b5

api/request_source/service.proto
oschaaf: fixed in 944259382bb86f75316099c256b964de492b48b5

oschaaf:
> Sorry, missed this before. Do we want to define a BUILD target for this so it gets built by the CI and remains fresh?

Currently https://github.com/envoyproxy/nighthawk/pull/137 is in sync with what we have here,and succeeds in CI, does that suffice? (If we introduce the associated BUILD file here we also need a temporary way to make sure it gets build in CI, and possibly back that out later?). If you foresee concurrent changes to it while we work on further getting #137 ready I think it probably is worth it?

mum4k:
> > Sorry, missed this before. Do we want to define a BUILD target for this so it gets built by the CI and remains fresh?
> 
> Currently #137 is in sync with what we have here,and succeeds in CI, does that suffice? (If we introduce the associated BUILD file here we also need a temporary way to make sure it gets build in CI, and possibly back that out later?). If you foresee concurrent changes to it while we work on further getting #137 ready I think it probably is worth it?

Good point, I don't think there will be concurrent changes with #137, since they pretty much work towards the same target, we just took this out to make #137 smaller.

Out of curiosity though - how does our CI setup work. Do we need to list all build targets explicitly? I would expect we have something like "bazel build ./..." somewhere so that even targets that aren't explicitly linked into one of our mains get built. Do we need to revisit this at some point?

api/client/service.proto
mum4k: I believe this was answered, I do see a value in giving the server the ability to see which of these fields were set. Planning to proceed and merge this in. @htuch, please let me know if I misread this and you would like us to make further changes to this API. We can still iterate on it in #137.

PR #240, opened 2019-12-17
Add --transport-socket command line option

eric846:
This PR adds the --transport-socket command line parameter, taking a JSON blob in the same way as --tls-context. This enables Nighthawk to run with customized transport_socket settings on the output cluster.

Since tls_context in Envoy config is deprecated in favor of transport_socket, I assume anyone currently using --tls-context will eventually need migrate to --transport-socket, and after this change everything should be ready for them.


source/client/options_impl.cc
```diff
@@ -284,6 +294,16 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       throw MalformedArgvException(e.what());
     }
   }
+  if (!transport_socket.getValue().empty()) {
+    try {
+      envoy::api::v2::core::TransportSocket ts;
+      Envoy::MessageUtil::loadFromJson(transport_socket.getValue(), ts,
```
dubious90: Does loadFromJson also work on yaml?

source/client/options_impl.cc
```diff
@@ -129,10 +130,19 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
 
   TCLAP::ValueArg<std::string> tls_context(
       "", "tls-context",
-      "Tls context configuration in yaml or json. Example (json):"
+      "DEPRECATED, use --transport-socket instead. "
+      "Tls context configuration in yaml or json. Example (json): "
```
dubious90: Not sure of the proper place for this in the pull request, but is there any bizarre behavior if somebody provides both transport_socket and tls_context in the same request?  If so, should we fail up front if a user does that? Or alternatively, explicitly ignore tls_context?

oschaaf:
Cool! I'll do a review pass tomorrow morning, but:

- you can use `tools/format_python_tools.sh fix` to make the format check happy.
- you can use `BAZEL_BUILD_OPTIONS=\"-c dbg\" tools/update_cli_readme_documentation.sh --mode fix` to update the cli docs in `README.md` and make the build CI task happy

README.md
```diff
@@ -1,13 +1,13 @@
...
-*A L7 (HTTP/HTTPS/HTTP2) performance characterization tool*
+_A L7 (HTTP/HTTPS/HTTP2) performance characterization tool_
 
 ## Current state
 
 Nighthawk currently offers:
 
 - A load testing client which supports HTTP/1.1 and HTTP/2 over HTTP and HTTPS.
-(HTTPS certificates are not yet validated).
+  (HTTPS certificates are not yet validated).
```
oschaaf: It seems we use different linting tools for markdown files, and both ways satisfy the formatting checks. It would be nice of we could somehow force-sync that up across contributors. Probably out of scope here, just an observation.

test/integration/test_integration_basics.py
```diff
@@ -269,6 +269,32 @@ def test_https_h1_tls_context_configuration(https_test_server_fixture):
   assertCounterEqual(counters, "ssl.ciphers.ECDHE-RSA-CHACHA20-POLY1305", 1)
 
 
+def test_https_h1_transport_socket_configuration(https_test_server_fixture):
```
oschaaf: - possibly we can refactor this a little bit to share code with the existing `test_XXX_tls_context_configuration` tests, as only the arguments to `runNighthawkClient()` change between those?
- also a tiny nit, it's better to make separate elements for arg / value pairs, e.g. `""--transport-socket ", "{..}"`. This makes it easier to inspect/massage the arguments in `runNighthawkClient()`. (There's may be some existing inconsistencies with respect to this).

source/client/options_impl.cc
eric846: Good point: 

Two things are going on:
- this particular json parser tolerates not putting quotes around dictionary keys
- yaml's compact syntax seems to be the same as json except without quotes around dictionary keys

So I think we may actually be supporting both json and yaml with loadFromJson. Technically the documentation should say "json or compact yaml."

There's also generic loadFromFile that loads the whole file into memory to detect the format, but it takes an extra Api parameter that depends on a thread pool, and I didn't want to get involved with all that if loadFromJson happens to work. loadFromFile would be the correct thing to call if users needed to pass a multiline yaml string on the command line. (Or try loadFromJson, see if it throws an exception, and if so, try loadFromYaml...)

README.md
eric846: Agreed, I'll revert these changes. I thought they had come from upstream but they were from my own editor all along! 

I would be interested to match any tools you're using.

source/client/options_impl.cc
eric846: I'm adding a warning rather than an error to reduce effort. 

The command line test would become more complicated if we made them mutually exclusive. The cost of refactoring and reviewing the test could be too high considering:
- the low chance that someone specifies both and also misses the warning during the time --tls-context still exists
- the fact that the test would need to be refactored back to being simple when --tls-context is removed in the near future

Also, added a warning that --tls-context is deprecated.

test/integration/test_integration_basics.py
eric846: Done and done!

oschaaf:
(Modulo formatting fixes to satisfy CI)

oschaaf:
Inspired by the discussion on conflicting options and warnings, I filed https://github.com/envoyproxy/nighthawk/issues/242

source/client/options_impl.cc
```diff
@@ -373,6 +401,16 @@ OptionsImpl::OptionsImpl(const nighthawk::client::CommandLineOptions& options) {
   trace_ = PROTOBUF_GET_WRAPPED_OR_DEFAULT(options, trace, trace_);
   open_loop_ = PROTOBUF_GET_WRAPPED_OR_DEFAULT(options, open_loop, open_loop_);
   tls_context_.MergeFrom(options.tls_context());
+
+  if (options.has_transport_socket()) {
+    envoy::api::v2::core::TransportSocket ts;
+    if (transport_socket_.has_value()) {
+      ts = transport_socket_.value();
```
oschaaf: Sorry - one tiny optional nit, this line isn't covered in tests. If it's not a lot of effort, it would be nice to cover it,
because that keeps coverage up to 100% for this file.

source/client/options_impl.cc
dubious90: Perfect.  This works for me

README.md
oschaaf: > Agreed, I'll revert these changes. I thought they had come from upstream but they were from my own editor all along! (or the script that generates README.md via some linter on my machine?)
> 
> I would be interested to match any tools you're using.

I'm using vscode with this extension: https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint. But imho ideally we'd have a check/fix format that would be more helpful here, apparently the current check is satisfied with both your and mine version.

source/client/options_impl.cc
eric846: I realized this is the constructor, so there's never going to be an existing value in transport_socket_. Deleting the impossible branch brings us back to 100%.

source/client/options_impl.cc
```diff
@@ -373,6 +402,13 @@ OptionsImpl::OptionsImpl(const nighthawk::client::CommandLineOptions& options) {
   trace_ = PROTOBUF_GET_WRAPPED_OR_DEFAULT(options, trace, trace_);
   open_loop_ = PROTOBUF_GET_WRAPPED_OR_DEFAULT(options, open_loop, open_loop_);
   tls_context_.MergeFrom(options.tls_context());
+
+  if (options.has_transport_socket()) {
+    envoy::api::v2::core::TransportSocket ts;
+    ts.MergeFrom(options.transport_socket());
+    transport_socket_.emplace(ts);
```
mum4k: (optional) What is the reasoning behind using .emplace() here? "ts" is an lvalue so it might not necessary, we could as well write something like this:

transport_socket_ = ts;

Which might paint a cleaner picture for the reader.

source/client/process_impl.cc
```diff
@@ -244,6 +244,11 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
       }
     }
 
+    if (options_.transportSocket().has_value()) {
+      auto* transport_socket = cluster->mutable_transport_socket();
```
mum4k: This may not be a good use of auto, since it hides useful type information from the reader. Can we expand it into the type name?

test/integration/test_integration_basics.py
```diff
@@ -247,47 +247,58 @@ def test_https_h2(https_test_server_fixture):
   assertEqual(len(counters), 17)
 
 
+def _do_tls_configuration_test(https_test_server_fixture, cli_parameter, use_h2):
```
mum4k: Can we document the arguments as per our Python style guide?

```python
def _do_tls_configuration_test(https_test_server_fixture, cli_parameter, use_h2):
  """Runs tests for different ciphers.

  For a given choice of (--tls-context, --transport-socket) x (H1, H2),
  run a series of traffic tests with different ciphers.

  Args:
    https_test_server_fixture: type, what it is ...
    ...
  """
```

source/client/options_impl.cc
```diff
@@ -293,6 +303,15 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
   }
 
+  if (!tls_context.getValue().empty()) {
+    ENVOY_LOG(warn, "--tls-context is deprecated. "
+                    "It can be replaced by an equivalent --transport-socket. "
+                    "See --help for an example.");
+  }
+  if (!tls_context.getValue().empty() && !transport_socket.getValue().empty()) {
+    ENVOY_LOG(warn, "Setting both --tls-context and --transport-socket "
+                    "may produce undefined behavior.");
```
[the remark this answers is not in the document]
oschaaf: Drive-by comment: I wonder if it would make sense for us to wire up a way to change the execution status if we observe errors and/or warnings being send to the logs.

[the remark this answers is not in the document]
oschaaf: +1 for a hard error here, sorry about the noise.

source/client/process_impl.cc
eric846: Simplified to a single line now.

source/client/options_impl.cc
eric846: Replaced with a clearer usage of emplace.

test/integration/test_integration_basics.py
eric846: Done

source/client/options_impl.cc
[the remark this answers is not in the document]
eric846: Done

test/options_test.cc
```diff
@@ -369,6 +419,21 @@ TEST_F(OptionsImplTest, BadTlsContextSpecification) {
       MalformedArgvException, "envoy.api.v2.auth.UpstreamTlsContext reason INVALID_ARGUMENT");
 }
 
+TEST_F(OptionsImplTest, BadTransportSocketSpecification) {
```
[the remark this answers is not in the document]
eric846: Done

source/client/options_impl.cc
eric846: Done

PR #246, opened 2019-12-20
Add overview.md

oschaaf:
First stab at adding more documentation.


docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### Streaming parameterizations and output stats
+
+One we have phases, the gRPC service, and perhaps the CLI, would be natural candidates to follow up with to allow dynamic phase injection, as well as send back reports per phase.
+
+## Key concept descriptions
+
+### Process
+
+**Process** represents the primary entry point to a Nighthawk execution run. Only one Process is allowed to exist at the same point in time within an OS Process. **Process **is responsible for performing process-wide initialization and termination, as well as handle input configuration, deliver output, and co-ordinate Workers. **ProcessImpl **is re-used accross the CLI and the gRPC service.
```
mum4k: Would this section benefit from links to header files where the individual concepts are defined?

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
+# Nighthawk: architecture and key concepts
```
mum4k: Can we line-wrap to some reasonable length, say similar to what we use in code?

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+## Key concept descriptions
+
+### Process
+
+**Process** represents the primary entry point to a Nighthawk execution run. Only one Process is allowed to exist at the same point in time within an OS Process. **Process **is responsible for performing process-wide initialization and termination, as well as handle input configuration, deliver output, and co-ordinate Workers. **ProcessImpl **is re-used accross the CLI and the gRPC service.
+
+### Worker
+
+**Worker **is responsible for performing correct initialization and termination of its thread, as well as execution of it’s designated task and offering a way for consumers to wait for that task to complete.
```
mum4k: (nit) its instead of it's in the second occasion. Or at least I think so ;)

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+**TerminationPredicate **is responsible for if and how to terminate. As of today, there are two types: one that will indicate that it is time to terminate based on a pre-configured duration, and one that will do so based on absolute counter thresholds. 
+
+### Sequencer
+
+**SequencerImpl** resides on a worker-thread, and drives itself via timers that run on the dispatcher, and coordinates interaction between **RateLimiter,** **BenchmarkClient**, and **TerminationPredicate** to drive execution to completion.
+
+### RateLimiter
+
+**RateLimiter **is responsible for indicating when it is time to release a request. **RateLimiter **offers a semaphore-like interaction model, as in closed-loop mode it may be that **BenchmarkClient **is not able to satisfy request-release timings, in which case acquisitions from **RateLimiter** need to be cancelled. Concretely, as of today there is **LinearRateLimiterImpl** which offers a straight-paced plain frequency, as well as work in progress on **DistributionSamplingRateLimiterImpl **(adding uniformly distributed random timing offsets to an underlying **RateLimiter**) and **RampingRateLimiter**.
```
mum4k: We should add a section on the modes before this one and define what closed and open loop modes mean in the context of Nighthawk.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### BenchmarkClient
+
+As of today, there’s a single implementation called** BenchmarkClientImpl,** which wraps Envoy’s **Upstream** concept and (slightly) customized H1/H2 **Pool** concepts. For executing requests, the pool will be requested to create a **StreamEncoder**, and Nighthawk will pass its own **StreamDecoderImpl** into that as an argument. The integration surface between **BenchmarkClient **is defined via **BenchmarkClient::tryStartRequest()** and a callback specification which will be fired upon completion of a successfully started request.
+
+For H3, it is anticipated that it will fit into this model, but if all else fails, it will be entirely possible to wire in a new type of **BenchmarkClient.**
+
+### RequestSource
+
+**RequestSource **is an abstraction that allows us to implement different ways for **BenchmarkClient** to get information on what the request that it is about to fire off should look like. Today, two implementations exist: a static one, which will repeat the same request over and over, as well as one that pulls dynamic request data from a grpc service. The latter can, for example, be used to implement log-replay.
```
mum4k: This is another place where we are referring to a "grpc service", can we replace it with a name of the service, since this is actually referring to a different service?

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### StreamDecoder
+
+StreamDecoder is a Nighthawk-specific implementation of an Envoy concept. It is responsible for coordinating lifetime events of a request to upper abstraction layers (**BenchmarkClient**, **Sequencer) **as well as latency recording.
```
mum4k: Without having the pre-existing knowledge on this Envoy concept, this paragraph doesn't help me to understand what StreamDecoder does. Any way how we can expand it, maybe link to the Envoy concept or provide more context?

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+OutputCollector is a container that facilitates building up the native output format of Nighthawk (proto3, nighthawk::client::Output). It is the basis for all output formats offered by Nighthawk, including CLI human output.
+
+### OutputFormatter
+
+OutputFormatter is responsible for transformations of nighthawk::client::Output to requested formats (e.g. human, json, fortio, etc)
+
+### Statistic
+
+Nighthawk’s statistic is responsible for administrating latencies. The most notable implemenation that exists today wraps HdrHistogram, but Nighthawk also has a couple of other implementations which mostly exist to ensure that floating point math is correct in tests. For various reasons, HdrHistogram might get replaced by [libcirclhist](https://github.com/envoyproxy/nighthawk/issues/115) in the near future.
```
mum4k: Typo: Implementation

mum4k: Can we define / link explaining what HdrHistogram is?

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+## Executables
+
+### nighthawk_client
+
+This is the CLI interface
+
+### nighthawk_service
+
+Nighthawk’s gRPC service is able to execute load tests, and also offers a stub request source gRPC service method implementation.
```
mum4k: I think we have moved the request source service method onto its own service and binary.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### H1Pool & H2Pool
+
+Nighthawk derives its own version of these from the vanilla Envoy ones. It does that to implement things like pro-active connection prefetching and H2 multi-connection support, as well as offer more connection management strategies. 
+
+## Executables
+
+### nighthawk_client
+
+This is the CLI interface
```
mum4k: Can we add something like: ... it is used to control the nighthawk_test_server (or whichever name we choose for it, e.g. Nighthawk Traffic API). If we do change the name, we should apply that throughout this document.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### nighthawk_service
+
+Nighthawk’s gRPC service is able to execute load tests, and also offers a stub request source gRPC service method implementation. 
+
+The stub request source gRPC method can be used for testing remote request source configuration (either with nighthawk_client, or by configuring request-source to loop back when requesting nighthawk_service to perform a load test).
```
mum4k: I am having a hard time parsing this paragraph and can't propose an edit, because I am not sure what we were trying to say.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+This is the CLI interface
+
+### nighthawk_service
+
+Nighthawk’s gRPC service is able to execute load tests, and also offers a stub request source gRPC service method implementation. 
+
+The stub request source gRPC method can be used for testing remote request source configuration (either with nighthawk_client, or by configuring request-source to loop back when requesting nighthawk_service to perform a load test). 
+
+While the stub gRPC method has no production use today, perhaps in the future this service method may get configurable so it can read traffic streams from disk, or from yet another remote source (which would allow the gRPC service to serve as a workload distributor for its Workers in a benchmark execution, as well as be convenient for those implementing a gRPC request source as they only have a single inbound gRPC call instead of one per worker).
```
mum4k: Why are we calling it a "stub gRPC method"? Or maybe I am unclear which method we are referring to. Is this the new method added in #229 ?

Secondly the language sounds fairly uncertain about its use. If this is about #229, we probably have a better understanding today as to how this will be used. It was already mentioned earlier in this doc as a reference to log-replay (non-static requests). Can we reword this accordingly? We can still keep the uncertain language about the fact that we may refactor this into a plugin-like architecture where we will have other request sources. Like loading from a file.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+
+### nighthawk_test_server
+
+Nighthawk’s test server, based on Envoy. It is able to synthesize delays and responses based on configuration via request headers (next to on-disk configuration).
+
+### nighthawk_output_transform
+
+Utility for transfoming the nighthawk-native "lossless" json output format into other formats (e.g. whuman, fortio). It can be very useful to always store the json output format, yet be able to easily get to one of the other output formats. It’s like having the cake and eating it too!
```
mum4k: +1 on the cake note ;)

docs/root/overview.rst
oschaaf: Sure; however, I added a single one under "#Key concept descriptions ..." which points to the directory where they live to lower the maintenance burden on this. Are you ok with that?

docs/root/overview.rst
oschaaf: Linked to the new `terminology.md` section on that.

docs/root/overview.rst
oschaaf: So at this time we're mentioning something that has not landed yet; but when that does we could link to the service definition that one should implement to make use of this.. maybe we should consider updating this part of #137?

docs/root/overview.rst
oschaaf: Linked to Envoy's codebase & added some content to (hopefully) clarify

docs/root/overview.rst
oschaaf: Fixed

oschaaf: Added a link

docs/root/overview.rst
oschaaf: Added more text; for the rename I refer to my earlier comment on the subject.

docs/root/overview.rst
oschaaf: I leaned this up; I propose we iterate on this more in #137

docs/root/overview.rst
oschaaf: Eliminated most of this.

docs/root/overview.rst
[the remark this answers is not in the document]
mum4k: Sounds good to me.

docs/root/overview.rst
```diff
@@ -0,0 +1,97 @@
+# Nighthawk: architecture and key concepts
+
+## High level interaction model
+
+**Process** creates one or more **Workers**. **Worker** will run **Sequencer**, which in turn queries **RateLimiter** for request-release timings. When it is time to release a request, **BenchmarkClient** will be requested to do so by **Sequencer**. **BenchMarkClient** will then ask its underlying **Pool** to create a **StreamDecoder** release the actual request. **StreamDecoder** will query the request data it needs to send from the configured **RequestSource**, and send it off. **StreamDecoder** will emit events as it progresses (pool ready, completion, etc), and timings will subsequently be recorded into **Statistic** as well as get bubbled up to **Sequencer** for tracking in-flight work and **Statistic** bookkeeping.
```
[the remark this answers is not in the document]
mum4k: Do we want to add a link pointing to terminology.md from the overview.md?

docs/root/overview.rst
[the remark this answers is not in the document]
mum4k: I have had similar experience with automated conversion, when you find one that works, please let me know. ;)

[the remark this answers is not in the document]
mum4k: Sounds good, do we want to add an issue so that we don't forget? We can use the same issue to propose and discuss possible names.

mum4k: Sounds good, less links means less maintenance.

docs/root/overview.rst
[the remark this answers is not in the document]
mum4k: Just noticed the link.

docs/root/overview.rst
mum4k: Sounds good.

docs/root/terminology.md
```diff
@@ -0,0 +1,40 @@
...
+In closed-loop mode, there is a feedback loop between the test subject and the
+test system. Concretely, Nighthawk will track and report [blocking](#blocking).
+When this happens, the current linear rate limiter will attempt to compensate
+pacing.
+
+When significant blocking is observed, measurements are skewed. Therefore,
+open-loop testing should be considered the preferred mode when measuring latency.
+
+Closed-loop testing can be useful to get a sense of the maximum requests per
+second that can be squeezed out of a system, as well as for comparing numbers
```
mum4k: I would suggest we add: "...that can be squeezed out of a system irrespective of the high latency incurred."

Or something similar to point out that the resulting measurement might not be indicative of the maximum load a system can reasonably handle in production without customer visible impact.

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
+# Nighthawk: architecture and key concepts
+
+## High level interaction model
+
+**Process** creates one or more **Workers**. **Worker** will run **Sequencer**,
+which in turn queries **RateLimiter** for [request-release
+timings](terminology.md#request-release-timings). When it is time to release a
+request, **BenchmarkClient** will be requested to do so by **Sequencer**.
+**BenchMarkClient** will then ask its underlying **Pool** to create a
```
mum4k: (nit) This part doesn't read well:

"to create a **StreamDecoder** release the actual request"

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+**Statistic** bookkeeping.
+
+**Sequencer** will query the configured **TerminationPredicates** to terminate
+when and how to terminate execution. When all **Workers** have finished,
+**Process** will collect the results from all **Workers** via
+**OutputCollector**, and run **OutputFormatter** to transform to the requested
+output format.
+
+However, it is said that pictures say more then words, so here is a block
+diagram: (TODO(#249): create a real diagram)
```
mum4k: (optional) Consider making the #249 an actual link to the issue.

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+## Notable upcoming changes
+
+Calling out two new concepts that may get proposed in the future, and cause some
+churn in the code base as we inject them.
+
+### Phases
+
+One notable addition / change that may get proposed in the near future is the
+introduction of **Phase**. **Phase** would represent a distinct stage of an
+execution, like, for example, warm-up. It would then be useful to have
```
mum4k: (nit) Feels like this is bit too many commas in a row, how about:

...a distinct stage of an execution, for example a warm-up.

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+
+**RateLimiter** is responsible for indicating when it is time to release a
+request. **RateLimiter** offers a semaphore-like interaction model, as in
+[closed-loop](terminology.md#closed-loop) mode it may be that
+**BenchmarkClient** is not able to satisfy request-release timings, in which
+case acquisitions from **RateLimiter** need to be cancelled. Concretely, as of
+today there is **LinearRateLimiterImpl** which offers a straight-paced plain
+frequency, as well as work in progress on
+**DistributionSamplingRateLimiterImpl** (adding uniformly distributed random
+timing offsets to an underlying **RateLimiter**) and **RampingRateLimiter**.
```
mum4k: I think we ended up calling it LinearRampingRateLimiter instead of RampingRateLimiter.

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+
+As of today, there’s a single implementation called **BenchmarkClientImpl**,
+which wraps Envoy’s **Upstream** concept and (slightly) customized H1/H2
+**Pool** concepts. For executing requests, the pool will be requested to create
+a **StreamEncoder**, and Nighthawk will pass its own **StreamDecoderImpl** into
+that as an argument. The integration surface between **BenchmarkClient** is
+defined via `BenchmarkClient::tryStartRequest()` and a callback specification
+which will be fired upon completion of a successfully started request.
+
+For H/3, it is anticipated that it will fit into this model, but if all else
```
mum4k: (nit) Is the slash here intentional or do we want to say "H3"?

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+
+### OutputCollector
+
+**OutputCollector** is a container that facilitates building up the native output
+format of Nighthawk (`proto3`, `nighthawk::client::Output`). It is the basis for all
+output formats offered by Nighthawk, including CLI human output.
+
+### OutputFormatter
+
+**OutputFormatter** is responsible for transformations of nighthawk::client::Output
```
mum4k: Do we want back-ticks around `nighthawk::client::Output` here as well?

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+Nighthawk derives its own version of these from the vanilla Envoy ones. It does
+that to implement things like pro-active connection pre-fetching and H2
+multi-connection support, as well as offer more connection management
+strategies.
+
+## Nighthawk binaries
+
+### nighthawk_client
+
+The CLI interface of the Nighthawk client. It will synthesize traffic according
```
mum4k: (nit) This seems to be the only place where we used the future tense, maybe:

The CLI interface of the Nighthawk client. It synthesizes...

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+to the requested configuration and report results in the requested output format.
+
+### nighthawk_service
+
+Nighthawk’s gRPC service is able to execute load tests, and report results.
+Under the hood it shared much of the code of nighthawk_client, and effectively
```
mum4k: typo: ...it shares

docs/root/overview.md
```diff
@@ -0,0 +1,189 @@
...
+it allows you to efficiently perform remote back-to-back executions of that.
```
mum4k: (nit) The "you" here feels weirdly personal. How about:

...and allows to efficiently perform...

oschaaf:
@mum4k thanks! Comments addressed in 859551cac9ff53a04b2910409de043055b7a63c9

PR #250, opened 2019-12-22
nighthawk_client remote execution option

oschaaf:
Adds an option to make it  easy to perform a remote execution via a running nighthawk_service instance and get the results out:

```
--nighthawk-service <uri format>
Nighthawk service uri. Example: grpc://localhost:8843/. Default is empty.
```

Irrespective of local or remote execution, output and behavior of nighthawk_client will be the same (except for logging which isn't transferred).


source/client/process_impl.cc
```diff
@@ -433,5 +436,53 @@ void ProcessImpl::setupForHRTimers() {
...
+  UriPtr uri;
+
+  try {
+    uri = std::make_unique<UriImpl>(options_.nighthawkService());
+  } catch (const UriException&) {
+    ENVOY_LOG(error, "Bad service uri: {}", options_.nighthawkService());
+    return false;
+  }
+
+  auto channel = grpc::CreateChannel(fmt::format("{}:{}", uri->hostWithoutPort(), uri->port()),
```
mum4k: I am wondering if it would be better to adjust the API of RemoteProcessImpl so that it takes the stub in by pointer (raw or otherwise). That way it will be agnostic to how the gRPC stub is created, with what credentials, etc. WDYT?

source/client/process_impl.h
```diff
@@ -107,5 +107,18 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
   bool shutdown_{true};
 };
 
+/**
+ * Will delegate execution to a remote nighthawk_service using gRPC.
+ */
+class RemoteProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
```
mum4k: (optional) Would this work better in a new file called say remote_process.[h|cc]?

source/client/process_impl.cc
oschaaf: That would be better, but it I presume it would also force handling of options outside of this class; to propose an alternative, maybe a `GrpcChannelFactory` concept would be useful here we'd pass that into the API, and we can use that to create the right channel for us?
Do you know if there a existing proto definition around somewhere which generalizes grpc channel configuration? (If so, maybe we can re-use that).

source/client/client.cc
```diff
@@ -41,19 +41,23 @@ Main::Main(Client::OptionsPtr&& options) : options_(std::move(options)) {}
 
 bool Main::run() {
   Envoy::Thread::MutexBasicLockable log_lock;
-
+  std::string lower = absl::AsciiStrToLower(
```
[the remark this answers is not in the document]
oschaaf: Done: b5c266df0c6c3dccbeb3605a2a60c701c58e6450

source/client/process_impl.h
oschaaf: Done: 8b32683c8b89cf8060237eaa28100e7eb7bb7123

source/client/process_impl.h
```diff
@@ -107,5 +107,18 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
...
+public:
+  RemoteProcessImpl(const Options& options);
+  bool run(OutputCollector& collector) override;
```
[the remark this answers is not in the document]
oschaaf: You are right that we should do better here.
Currently we just have this in the implementation:

```
    if (response.has_error_detail()) {
      ENVOY_LOG(error, "have error detail: {}", response.error_detail().DebugString());
    }
```

But this isn't specific to this service IMHO. We also need this in the current flow on the principle that we define our output in the proto. Currently we (ab)use logging to visualize the error details as the sole means.
So .. to resolve, possibly define an error description in the output?

source/client/process_impl.h
```diff
@@ -107,5 +107,18 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
   bool shutdown_{true};
 };
 
+/**
+ * Will delegate execution to a remote nighthawk_service using gRPC.
+ */
+class RemoteProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  RemoteProcessImpl(const Options& options);
```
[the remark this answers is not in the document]
oschaaf: Done: 0985dc07124832687eb4b656965bdd92e124f86e

source/client/remote_process_impl.cc
```diff
@@ -0,0 +1,69 @@
...
+  UriPtr uri;
+
+  try {
+    uri = std::make_unique<UriImpl>(options_.nighthawkService());
+  } catch (const UriException&) {
+    ENVOY_LOG(error, "Bad service uri: {}", options_.nighthawkService());
+    return false;
+  }
+
+  auto channel = grpc::CreateChannel(fmt::format("{}:{}", uri->hostWithoutPort(), uri->port()),
```
dubious90: When this moved files, conversation got lost.  Reinserting it here, and replying below:

mumak:
> I am wondering if it would be better to adjust the API of RemoteProcessImpl so that it takes the stub in by pointer (raw or otherwise). That way it will be agnostic to how the gRPC stub is created, with what credentials, etc. WDYT?

oschaaf
> That would be better, but it I presume it would also force handling of options outside of this class; to propose an alternative, maybe a GrpcChannelFactory concept would be useful here we'd pass that into the API, and we can use that to create the right channel for us?
> Do you know if there a existing proto definition around somewhere which generalizes grpc channel configuration? (If so, maybe we can re-use that).

dubious90: The problem for me with using insecure credentials here is that consumers in cloud environments (aws, google cloud, etc) will probably have protected their services using oauth, tls, etc.  And it would be very difficult to support all of that natively here.  It's also a common paradigm in grpc that ChannelCredentials/ServerCredentials are implemented in project repositories, some of which are not open source.

If you introduce dependency injection here and allow passing in of the credentials from the main file, as Jakub is suggesting, then you allow people to make shallow wrappers around this that implement specifics.  For whatever it's worth, this is already how the `nighthawk-service` works.  It was very easy to pull just the service implementation out of nighthawk, and apply our own credentials to it.  Without dependency injection here, though, that'd be very difficult.

(In terms of preventing additional config or command-line arguments, you could keep it as just using InsecureCredentials in this case, but pass that in from source/client/client.cc)

oschaaf: Ah ok I see, I'll amend. It's a tiny change -- especially if we can persist the hard coded insecure credentials.

oschaaf: @dubious90 does 896f14b1f29bfd880f764caf18a73b010ef59ffd look like what you had in mind?

dubious90: I think this helps a fair amount.  We can always come back to expanding to natively supporting other potential credentials in a separate issue.  What do you think @mum4k ?

mum4k: This seems to be aligned with our intentions, thank you for making the changes.

source/client/process_impl.h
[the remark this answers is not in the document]
mum4k: Ideally we would at least document how to retrieve the error detail for the user.

source/client/remote_process_impl.cc
```diff
@@ -0,0 +1,58 @@
...
+
+RemoteProcessImpl::RemoteProcessImpl(const Options& options,
+                                     nighthawk::client::NighthawkService::Stub& stub)
+    : options_(options), stub_(stub) {}
+
+bool RemoteProcessImpl::run(OutputCollector& collector) {
+  nighthawk::client::ExecutionRequest request;
+  nighthawk::client::ExecutionResponse response;
+  grpc::ClientContext context;
+  auto r = stub_.ExecutionStream(&context);
```
mum4k: (nit) Can we give this variable a slightly longer; more readable name?

source/client/remote_process_impl.cc
```diff
@@ -0,0 +1,58 @@
...
+  // We don't forward the option that requests remote execution. Today,
+  // nighthawk_service will ignore the option, but if someone ever changes that this
+  // is probably desireable.
+  request.mutable_start_request()->mutable_options()->mutable_nighthawk_service()->Clear();
+
+  if (r->Write(request, {}) && r->Read(&response)) {
+    if (response.has_output()) {
+      collector.setOutput(response.output());
+    } else {
+      ENVOY_LOG(error, "Remote execution failed");
```
mum4k: Should this message be closer to "remote execution provided no output"? Or is the lack of output a clear enough signal of a failure?

source/client/remote_process_impl.cc
oschaaf: Done in 5f7fa78

source/client/remote_process_impl.cc
oschaaf: This is a clear signal of a failure; the service will always provide output (with at least some counters indicating connection pool failures).

source/client/process_impl.h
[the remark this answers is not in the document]
oschaaf: I think this comment refers to outdated code; this file was renamed. The current version has some doc comments.. does that look good?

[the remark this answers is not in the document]
mum4k: Indeed, sorry for missing that. The current version looks good.

PR #251, opened 2019-12-23
Add zipf distributing ratelimiter

oschaaf:
Adds a probabilistic filtering rate limiter, which will suppress an underlying rate limiter's pacing using a zipf distribution.


oschaaf:
Note to self: address first round of comment from #218 here before proceeding with review

source/common/rate_limiter_impl.h
```diff
@@ -203,4 +204,30 @@ class GraduallyOpeningRateLimiterFilter : public FilteringRateLimiterImpl {
...
+public:
+  /**
+   * From the absl header associated to the zipf distribution:
+   * Preconditions: v > 0, q > 1
+   * The precondidtions are validated when NDEBUG is not defined via
+   * a pair of assert() directives.
+   * If NDEBUG is defined and either or both of these parameters take invalid
+   * values, the behavior of the class is undefined.
+   */
+  ZipfRateLimiterImpl(RateLimiterPtr&& rate_limiter, bool deterministic, double q = 2.0,
```
mum4k: I believe we had some open comments for the API and its documentation here in the original PR where this was introduced before we split it out. Would you mind resolving those comments before we proceed?

oschaaf: Done, addressed the [prior feedback round](https://github.com/envoyproxy/nighthawk/pull/218#pullrequestreview-335035033) in 10dbc29f4affec32379b40a4e0d9a3f117a36474

mum4k:
Could you update the description of this PR? Unless there is still something blocking this?

oschaaf:
> Could you update the description of this PR? Unless there is still something blocking this?

Done!

PR #252, opened 2020-01-02
Spread generated traffic across multiple backends

eric846:
Introduces the idea of a MultiTarget, multiple backends that all receive the same HTTP request with round robin distribution.

Instead of a URI, specify:
--multi-target-path /x/y/z
--multi-target-endpoint 1.1.1.1:1
--multi-target-endpoint 2.2.2.2:2
--multi-target-use-https

Also fixes a clang-tidy error where a virtual function was detected as being called from the constructor (fixed by referring to fields directly instead of using the virtual getters). We don't know why the problem only started to fail clang-tidy in this PR.


api/client/options.proto
```diff
@@ -128,4 +128,6 @@ message CommandLineOptions {
   google.protobuf.BoolValue open_loop = 21;
   // See :option:`--jitter-uniform`
   google.protobuf.Duration jitter_uniform = 25 [(validate.rules).duration.gte.nanos = 1];
+  // See :option:`--backend-endpoint`
+  repeated google.protobuf.StringValue backend_endpoints = 27;
```
dubious90: We talked about this in person, but officially putting it here. Would it be possible to change the API here slightly, so this is in a `oneof` with `uri`, and maybe have a different way of pushing path through?  If that is unduly difficult, feel free to ignore.

dubious90: Separately, I'm not sure that you're gaining any value here from using the protobuf.StringValue wrapper, rather than just using a string.  Maybe @oschaaf would have a better idea of whether that's true or not.

source/client/process_impl.cc
```diff
@@ -260,11 +262,31 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
         options_.maxPendingRequests() == 0 ? 1 : options_.maxPendingRequests());
     thresholds->mutable_max_requests()->set_value(options_.maxActiveRequests());
 
-    cluster->set_type(envoy::api::v2::Cluster::DiscoveryType::Cluster_DiscoveryType_STATIC);
-    auto* host = cluster->add_hosts();
-    auto* socket_address = host->mutable_socket_address();
-    socket_address->set_address(uri.address()->ip()->addressAsString());
-    socket_address->set_port_value(uri.port());
+    if (options_.backendEndpoints().empty()) {
+      cluster->set_type(envoy::api::v2::Cluster::DiscoveryType::Cluster_DiscoveryType_STATIC);
```
dubious90: nit: can you add a shorthand for this?  DiscoveryType isn't a namespace, but you could do either
```namespace EnvoyCluster = ::envoy::api::v2::Cluster;
EnvoyCluster::DiscoveryType::Cluster_DiscoveryType_STATIC```
or:
```const discovery_type = envoy::api::v2::Cluster::DiscoveryType;
discovery_type::Cluster_DiscoveryType_STATIC
```
I realize this part isn't new, but since you're now calling this long path at least 3 times, I think it'd help readability

source/client/process_impl.cc
```diff
@@ -260,11 +262,31 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
...
+      auto* host = cluster->add_hosts();
+      auto* socket_address = host->mutable_socket_address();
+      socket_address->set_address(uri.address()->ip()->addressAsString());
+      socket_address->set_port_value(uri.port());
+    } else {
+      if (Utility::hostAddressTypeFromHostPort(options_.backendEndpoints()[0]) ==
+          HostAddressType::DNS) {
+        cluster->set_type(envoy::api::v2::Cluster::DiscoveryType::Cluster_DiscoveryType_STRICT_DNS);
```
dubious90: There's probably something I'm missing here, but why is this possible now, but wasn't before?  If this is a new feature (allowing Strict DNS), should we allow that for uri above as well?

source/common/utility.h
```diff
@@ -54,6 +57,14 @@ class Utility {
...
    */
   static void parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* const* argv);
+
+  /**
+   * @param host_port host:port as a string, where host can be IPv4, [IPv6], or a DNS
+   * name.
+   * @return HostAddressType classification of the host address as IPV4, IPV6,
+   * or DNS, or INVALID for basic parse errors
+   */
+  static HostAddressType hostAddressTypeFromHostPort(const std::string& host_port);
```
dubious90: nit: would absl::string_view be more readable here than the string reference?

source/client/options_impl.cc
```diff
@@ -189,6 +189,17 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
...
       "of jitter, specify .00001s. Default is empty / no uniform jitter.",
       false, "", "duration", cmd);
+  TCLAP::MultiArg<std::string> backend_endpoints(
+      "", "backend-endpoint",
+      "Backend endpoint overrides. "
+      "This argument is intended to be specified multiple times. "
+      "Nighthawk will generate the same traffic as usual, but "
+      "will spread it across all backend endpoints with "
+      "round robin distribution. "
+      "Endpoint formats: IPv4:port, [IPv6]:port, DNS:port. "
```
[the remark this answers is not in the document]
oschaaf: I haven't really thought this through yet, but one fleeting thought I just had is if maybe it would make sense to make this interact with the rate limiters. Consider ramping vs round robin over multiple backends.. do we want to the ramp specific per backend? In that case, this could act like a multiplier of sorts, Nighthawk strifes to keep release timings close-to-equal across backends, e.g. one rate limiter aquisition causes a request to be released for each backend. Possibly this is the opposite of what's desirable, just a braindump.

mum4k:
@dubious90, @oschaaf please take a look and assign to me once you are done with the review.

api/client/options.proto
mum4k: To answer this, wrapping a scalar (like a string) in a message in proto3 adds a pointer. Whether the pointer is set or not can then be used as a signal of whether the field was explicitly set or not.

With a plain string in proto3, we can't recognize unset string from a string explicitly set to an empty string. Both would just read as an empty string. With the wrapper we can. For the case of an unset string, the outer message (StringValue) would have a nil pointer.

Practically we should avoid wrappers unless we need to distinguish between a set and an unset default value.

oschaaf: @mum4k First -by no means am I a proto expert.

but this came up before, so let me share some rationale about why there is consistent wrapping used in this file. That is indeed done for the purpose of being able to discern between set and not set; but as of today for the options definition it is not concretely used. 

Having them wrapped is going to make things easy if we want to support hierarchical merging of options. Examples:
- have a base config on disk for nighthawk_client, and override from the CLI
- configure nighthawk_service with a base config, and override with incoming option values
This principle is used in the nighthawk-specific `nighthawk_test_server` configuration when performing overrides based on configuration delivered via incoming request headers. I think we can do this hierarchical merge with a single c++ call.

If we want to support this elsewhere, it will be easier with the wrapped values as we'll be able to discern between set and not set as you say. (Or actually, I think we close that door and make it impossible if we do not wrap?). I am curious what the penalty is that we're paying for the wrapping; if it's just being less efficient at the wire level / a performance penalty, I would be still inclined to go for wrapping as the consumers of this probably won't need to be able to do many transactions/second.. while we keep the door open for hierarchical configuration merging.

Also, perhaps less important, but I found the the consistent wrapping makes me have to think less as handling in code is uniform.

mum4k: Thanks for the added background Otto, I think we are on the same page in regards to why wrapping is required. Yes for determining if the value was set and also for the reason you provided - merging options. I think both of those are valid reasons to use the wrappers and while I am no proto expert myself, from what I know this won't add any significant penalty.

As a matter of fact, we are still using proto2 internally which uses pointers everywhere, so it follows the same pattern that allows to determine if a value was set. I don't see a reason to break away from this pattern and now we have good documented trail of why it is needed.

eric846: I updated it so that now the user sets either a URI or these options:
--multi-target-path
one or more --multi-target-endpoint host:port
optionally --multi-target-use-https 

This removes the strangeness of overriding part of the URI, part of the behavior being determined by the URI, etc. In the proto it's clean, and on the CLI it's fully exposed in hopefully a logical way.

source/client/process_impl.cc
eric846: Removed this code in the updated version

source/client/process_impl.cc
eric846: Good point, in URI mode it was resolving the DNS to an IP address ahead of time, according to an auto/v4/v6 option. I now updated --multi-target-* mode to uniformly reuse the same resolution code, so this DNS-related code no longer exists.

source/common/utility.h
eric846: I tried string_view but the RE2::FullMatch() in the implementation didn't understand it.

Also note that in the updated version, we don't attempt to classify the address anymore, only split host:port into host and port.

source/client/process_impl.cc
```diff
@@ -260,11 +262,31 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
...
+      } else {
+        cluster->set_type(envoy::api::v2::Cluster::DiscoveryType::Cluster_DiscoveryType_STATIC);
+      }
+      for (const std::string& backend_endpoint : options_.backendEndpoints()) {
+        auto* host = cluster->add_hosts();
+        auto* socket_address = host->mutable_socket_address();
+        std::vector<absl::string_view> parts = absl::StrSplit(backend_endpoint, ':');
+        int port = -1;
+        (void)absl::SimpleAtoi(parts.back(), &port);
```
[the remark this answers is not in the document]
eric846: No longer exists in the rewritten version

source/client/options_impl.cc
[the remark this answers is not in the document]
eric846: I rewrote it all so that we don't do anything hacky with the URI. Instead for multi-target mode the URI is omitted and  the path and endpoints are directly spelled out. I left the port mandatory in each endpoint since it would take extra logic to determine a default.

source/common/utility.cc
```diff
@@ -72,4 +72,17 @@ void Utility::parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* cons
   }
 }
 
-} // namespace Nighthawk
\ No newline at end of file
+HostAddressType Utility::hostAddressTypeFromHostPort(const std::string& host_port) {
```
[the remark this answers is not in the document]
eric846: Note: This function no longer classifies the address, it just splits out the :port. I still kept this code separate because it's very short. I started relying on the resolver, with its existing unit tests, to catch malformed addresses later.

source/common/utility.cc
```diff
@@ -72,4 +72,17 @@ void Utility::parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* cons
...
+  if (RE2::FullMatch(host_port, R"(\d+\.\d+\.\d+\.\d+:\d+)")) {
+    return HostAddressType::IPV4;
+  }
+  if (RE2::FullMatch(host_port, R"(\[[.:0-9a-fA-F]+\]:\d+)")) {
+    return HostAddressType::IPV6;
+  }
+  if (RE2::FullMatch(host_port, R"([-.0-9a-zA-Z]+:\d+)")) {
+    return HostAddressType::DNS;
+  }
+  return HostAddressType::INVALID;
```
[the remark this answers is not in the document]
eric846: I did a fairly big rewrite that simplifies the implementation and removes the URI overriding hackiness.

I'm now constructing multiple URIs and calling the existing resolver mechanism as-is.

If we switch to another URI parser, the multi-backend code will also get the improvement with little extra effort.

source/client/options_impl.cc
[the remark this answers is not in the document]
eric846: About the rate limiter, I think I see what you mean. I think we might be all right as-is. If someone activates the multi-backend feature with 3 backends, they have signed up for a 3x reduction in the steady state rps seen at each backend. It sounds like Nighthawk today will also naturally produce a 3x reduction in the ramp rate seen at each backend. 

We don't have any ramping needs at the moment, and it would certainly be safe for us to accept any slow ramping.

Could the user who sets up 3 backends deliberately configure a 3x higher ramp rate to compensate for the 3x reduction that will occur?

Someone who needed to fully control the ramp rate on each backend could also run multiple Nighthawks as today.

By the way, we don't have that much confidence that this multiple backends mode will work better than running multiple Nighthawk processes. The sole motivation is to simplify our testbed by reducing the number of processes. The simplification could come into play more when we implement statistical models in the request source.

test/client/utility.cc
```diff
@@ -16,6 +16,7 @@ std::unique_ptr<OptionsImpl> TestUtility::createOptionsImpl(absl::string_view ar
 }
 
 std::unique_ptr<OptionsImpl> TestUtility::createOptionsImpl(const std::vector<const char*>& argv) {
+  TCLAP::OptionalUnlabeledTracker::alreadyOptional() = false;
```
eric846: This works around an error thrown by TCLAP about multiple unlabeled optional args not being allowed. It appeared in many tests as soon as I made URI optional. TCLAP has a global flag that detects multiple unlabeled optional args. It assumes there will be only one command line in the lifetime of the process. In unit tests we parse multiple TCLAP command lines, so we need to reset TCLAP's flag to simulate a fresh process.

test/options_test.cc
```diff
@@ -35,21 +35,21 @@ TEST_F(OptionsImplTest, BogusInput) {
...
       "--burst-size 13 --address-family v6 --request-method POST --request-body-size 1234 "
       "--transport-socket {} "
-      "--request-header f1:b1 --request-header f2:b2 --request-header f3:b3:b4 {} "
+      "--request-header f1:b1 --request-header f2:b2 --request-header f3:b3:b4 "
       "--max-pending-requests 10 "
       "--max-active-requests 11 --max-requests-per-connection 12 --sequencer-idle-strategy sleep "
       "--termination-predicate t1:1 --termination-predicate t2:2 --failure-predicate f1:1 "
       "--failure-predicate f2:2 --jitter-uniform .00001s "
-      "--experimental-h1-connection-reuse-strategy lru ",
+      "--experimental-h1-connection-reuse-strategy lru {}",
```
eric846: The unlabeled URI arg now has to be at the end because it's now optional.

source/client/options_impl.cc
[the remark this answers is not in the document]
oschaaf: sounds great; this makes sense to me.

source/common/utility.cc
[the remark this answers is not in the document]
oschaaf: great!

source/client/options_impl.cc
```diff
@@ -476,39 +531,67 @@ void OptionsImpl::validate() const {
...
+    if (multi_target_endpoints_.empty()) {
+      throw MalformedArgvException("A URI or --multi-target-* options must be specified.");
+    }
+    if (multi_target_path_.empty()) {
+      throw MalformedArgvException("--multi-target-path must be specified.");
+    }
   }
   try {
-    Envoy::MessageUtil::validate(*toCommandLineOptions(),
+    Envoy::MessageUtil::validate(*toCommandLineOptionsInternal(),
```
eric846: I couldn't figure out how this indirect call to a virtual method (constructor -> validate -> toCommandLineOptions) is escaping clang-tidy on the master branch but not this branch. In any case, it does seem to be a real issue, and this does seem to be a solution.

oschaaf: Oh; that is a good catch. It is pretty weird that this wasn't caught earlier. This does trigger a memory, earlier
we also ran into some odd and inconsistent clang-tidy complaints mentioning TCLAP. I don't think they pointed out the same problem -- they pointed to something in TCLAP instead (and Envoy had similar pre-existing suppressions in its codebase).

I'm not sure if what is resolved here is related to that, but I thought it might be worth bringing it up. At the time we couldn't figure out the underlying cause, and thought it was a benign issue related to TCLAP code. There are a couple of odd `NOLINT`s [1](https://github.com/envoyproxy/nighthawk/blob/67ef1094c5708d55b203b44a203b197d132c4540/source/client/options_impl.cc#L92) [2](https://github.com/envoyproxy/nighthawk/blob/67ef1094c5708d55b203b44a203b197d132c4540/source/client/options_impl.cc#L162) in this file to suppress those.

source/common/utility.h
```diff
@@ -54,6 +55,15 @@ class Utility {
...
   static void parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* const* argv);
+
+  /**
+   * @param host_port host:port as a string, where host can be IPv4, [IPv6], or a DNS
+   * name.
+   * @param host string* to receive the host if the parse succeeds
+   * @param port int* to receive the port if the parse succeeds
+   * @return bool true if the input could be parsed as host:port
+   */
+  static bool parseHostPort(const std::string& host_port, std::string* host, int* port);
```
oschaaf: Optional -- One potential alternative internal implementation is to construct a mock temp `tcp://host:port` uri based on the host/port, and rely on `Envoy::Network::Utility::portFromTcpUrl()` / `Envoy::Network::Utility::hostFromTcpUrl()`
(I think Envoy also does that in some places): https://github.com/envoyproxy/envoy/blob/21aa1a72ed976288ccd5b1612cc9d20f72e5a25d/source/common/network/utility.h#L111

source/client/options_impl.cc
eric846: I see, I will be on the lookout for such things involving TCLAP. I have no idea how master passes but this branch doesn't, but it might have to do with the order.

Is it better to keep all these `virtualAccessor() -> raw_field_name_` changes instead of add NOLINT to fix this particular warning?

oschaaf: +1 as is. I think NOLINT here is no good as this actually looks like a real issue that got caught?

source/common/utility.h
eric846: I tried those out, but unfortunately they don't seem to work with the colons in IPv6 addresses -- malformed url: tcp://[::1]:5" thrown in the test body.

I don't see any existing utility that supports a mixture of DNS, IPv4, IPv6, other than the UriImpl which comes into play later.

oschaaf: Ah, I see, LGTM as is then

ci/do_ci.sh
```diff
@@ -197,7 +197,7 @@ case "$1" in
         exit 0
     ;;
     *)
-        echo "must be one of [build,test,clang_tidy,test_with_valgrind,coverage,asan,tsan,docker]"
+        echo "must be one of [build,test,clang_tidy,test_with_valgrind,coverage,asan,tsan,docker,check_format,fix_format]"
```
oschaaf: Consider moving this into a separate PR? (It's good to have this, but it's not related to the PR topic?)

source/client/options_impl.cc
```diff
@@ -453,6 +508,7 @@ OptionsImpl::OptionsImpl(const nighthawk::client::CommandLineOptions& options) {
     jitter_uniform_ = std::chrono::nanoseconds(
         Envoy::Protobuf::util::TimeUtil::DurationToNanoseconds(options.jitter_uniform()));
   }
+
```
oschaaf: nit: accidental whitespace?

test/client_test.cc
```diff
@@ -30,6 +30,12 @@ TEST_F(ClientTest, NormalRun) {
 }
 
 TEST_F(ClientTest, AutoConcurrencyRun) {
+  // This works around an error thrown by TCLAP about multiple unlabeled optional args not being
```
oschaaf: Maybe just add a pointer to the other comment and move this higher up in the file, e.g.  
```c++
// See the comment in TestUtility::createOptionsImpl()
````

File: test/client/utility.cc [hunk 16]
```diff
@@ -16,6 +16,12 @@ std::unique_ptr<OptionsImpl> TestUtility::createOptionsImpl(absl::string_view ar
 }
 
 std::unique_ptr<OptionsImpl> TestUtility::createOptionsImpl(const std::vector<const char*>& argv) {
+  // This works around an error thrown by TCLAP about multiple unlabeled optional args not being
+  // allowed. TCLAP has a global flag that detects multiple unlabeled optional args. It assumes
+  // there will be only one command line in the lifetime of the process. In unit tests we parse
+  // multiple TCLAP command lines, so we need to reset TCLAP's flag to simulate a fresh process.
+  TCLAP::OptionalUnlabeledTracker::alreadyOptional() = false;
```
oschaaf: I wonder if there would be anything against doing this in the after we're done parsing TCLAP args in the related `OptionImpl` constructor (plus maybe in the destructor of `OptionImpl` for good measure)..

test/integration/integration_test_fixtures.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, ip_version):
     self.server_ip = "::1" if ip_version == IpVersion.IPV6 else "127.0.0.1"
     self.socket_type = socket.AF_INET6 if ip_version == IpVersion.IPV6 else socket.AF_INET
     self.test_server = None
+    self.test_servers = []
```
oschaaf: Looking at `test_server`/`test_servers`, somehow I wonder if we should just rewrite/design this class as if multiple backends is the default case,
And then possibly write a derived convenience class which provides easy access for the single-server case.
Probably out of scope in this PR :-)

test/integration/test_integration_basics.py
```diff
@@ -441,3 +442,67 @@ def test_bad_arg_error_messages(http_test_server_fixture):
       expect_failure=True,
       as_json=False)
   assert "Bad argument: Termination predicate 'a:a' has an out of range threshold." in err
+
+
+def test_multiple_backends_http_h1(multi_http_test_server_fixture):
+  """
+  Runs the CLI configured to use plain HTTP/1 against multiple test servers, and sanity
```
oschaaf: Could we also verify we spread requests as intended across individual back-ends? (So the avg number of replies observed from a backend / received by a backend looks as expected?)

ci/do_ci.sh
eric846: Moving to #257

test/integration/integration_test_fixtures.py
eric846: Agreed, filed #258

test/client/utility.cc
eric846: I was torn about this -- putting it in OptionImpl would eliminate the problem for all future tests, but I don't know if we have any rule against modifying production code to fix tests.

test/integration/test_integration_basics.py
eric846: Done

source/client/options_impl.cc
eric846: Removed

test/client_test.cc
eric846: Commented both occurrences here with the pointer. What did you mean we should move?

test/integration/test_integration_basics.py
eric846: Done

test/client_test.cc
oschaaf: Sorry that remark about moving didn't make sense, thanks for updating this, marking as resolved.

test/client/utility.cc
oschaaf: Test only changes should be avoided... but thinking about it .. arguably this isn't test only. It may be possible to accidentally misuse this? 
 
So maybe, to guard against incidents / foolproof it, a helper class that facilitates enforcing non-concurrent usage as well as taking care of resetting the flag in a RAII-style manner is worth considering here? If feel this would bloat this PR though, it's probably something that should have been done earlier if TCLAP assumes there will be only one command line per process.
Shall we file an issue for this one?

eric846: Sounds good, filed #259

api/client/options.proto
```diff
@@ -71,6 +71,20 @@ message SequencerIdleStrategy {
...
+  message Endpoint {
+    google.protobuf.StringValue address = 1;
+    google.protobuf.UInt32Value port = 2 [(validate.rules).uint32 = {gte: 1, lte: 65535}];
+  }
+  // Whether to use HTTPS in requests to all backends; otherwise HTTP.
+  google.protobuf.BoolValue use_https = 1;
+  // One or more address-port pairs to receive traffic distributed with round robin.
+  repeated Endpoint endpoints = 2;
+  // The absolute HTTP request path (the part of the URL after host:port).
+  // A single path is requested from all backends. Ignored when using a RequestSource.
```
mum4k: (nit) Might not be clear whether the path should or shouldn't start with a forward slash, can we include at least one example?

include/nighthawk/client/options.h
```diff
@@ -29,7 +29,7 @@ class Options {
   virtual uint32_t connections() const PURE;
   virtual std::chrono::seconds duration() const PURE;
   virtual std::chrono::seconds timeout() const PURE;
-  virtual std::string uri() const PURE;
+  virtual absl::optional<std::string> uri() const PURE;
```
mum4k: This might be a good place to document why it is optional.

source/client/factories_impl.cc
```diff
@@ -118,16 +118,24 @@ void RequestSourceFactoryImpl::setRequestHeader(Envoy::Http::HeaderMap& header,
 }
 
 RequestSourcePtr RequestSourceFactoryImpl::create() const {
-  // Note: we assume a valid uri.
-  // Also, we can't resolve, but we do not need that.
-  UriImpl uri(options_.uri());
   Envoy::Http::HeaderMapPtr header = std::make_unique<Envoy::Http::HeaderMapImpl>();
+  if (options_.uri().has_value()) {
+    // Note: we assume a valid uri.
+    // Also, we can't resolve, but we do not need that.
```
mum4k: The comment as it stands begs a question - why can't we resolve? This is probably a result of the review discussion that happened here, but can we improve (or remove) the comment?

source/client/factories_impl.cc
```diff
@@ -118,16 +118,24 @@ void RequestSourceFactoryImpl::setRequestHeader(Envoy::Http::HeaderMap& header,
...
+    UriImpl uri(options_.uri().value());
+    header->setPath(uri.path());
+    header->setHost(uri.hostAndPort());
+    header->setScheme(uri.scheme() == "https" ? Envoy::Http::Headers::get().SchemeValues.Https
+                                              : Envoy::Http::Headers::get().SchemeValues.Http);
+  } else {
+    header->setPath(options_.multiTargetPath());
+    header->setHost("dummy");
```
mum4k: Can we improve the string literal to make this a bit more self-documenting? E.g. "host is not used with multi target" or similar.

source/client/options_impl.cc
```diff
@@ -482,39 +537,67 @@ void OptionsImpl::validate() const {
...
   }
-  try {
-    UriImpl uri(uri_);
-  } catch (const UriException&) {
-    throw MalformedArgvException("Invalid URI");
+  if (uri_.has_value()) {
+    try {
+      UriImpl uri(uri_.value());
+    } catch (const UriException&) {
+      throw MalformedArgvException("Invalid URI");
```
mum4k: (pre-existing, optional) Can we include the invalid uri in the error message?

source/client/process_impl.cc
```diff
@@ -238,10 +240,12 @@ ProcessImpl::mergeWorkerStatistics(const StatisticFactory& statistic_factory,
...
 
 void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Bootstrap& bootstrap,
-                                               const Uri& uri, int number_of_clusters) const {
+                                               const std::vector<UriPtr>& uris,
+                                               int number_of_clusters) const {
   for (int i = 0; i < number_of_clusters; i++) {
     auto* cluster = bootstrap.mutable_static_resources()->add_clusters();
-    if (uri.scheme() == "https") {
+    RELEASE_ASSERT(!uris.empty(), "illegal configuration with zero endpoints");
+    if (uris[0]->scheme() == "https") {
```
mum4k: A comment explaining why all URIs are guaranteed to have the same scheme might be good here.

source/client/process_impl.cc
```diff
@@ -271,10 +275,12 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v2::Boo
     thresholds->mutable_max_requests()->set_value(options_.maxActiveRequests());
 
     cluster->set_type(envoy::api::v2::Cluster::DiscoveryType::Cluster_DiscoveryType_STATIC);
-    auto* host = cluster->add_hosts();
-    auto* socket_address = host->mutable_socket_address();
-    socket_address->set_address(uri.address()->ip()->addressAsString());
-    socket_address->set_port_value(uri.port());
+    for (const UriPtr& uri : uris) {
+      auto* host = cluster->add_hosts();
```
mum4k: (nit, optional) Are the types too unwieldy to mention here?

source/common/utility.cc
```diff
@@ -72,4 +72,10 @@ void Utility::parseCommand(TCLAP::CmdLine& cmd, const int argc, const char* cons
   }
 }
 
-} // namespace Nighthawk
\ No newline at end of file
+bool Utility::parseHostPort(const std::string& host_port, std::string* address, int* port) {
+  return RE2::FullMatch(host_port, R"((\d+\.\d+\.\d+\.\d+):(\d+))", address, port) ||
```
mum4k: Did we get these REs from somewhere or are they created by us? If we got them from somewhere, can we include a comment linking there? If they are crafted by us, can we add a comment above each one including an example string they are intending to match? That tends to help the posterity a lot when reading REs.

source/common/utility.h
mum4k: Can we add a unit test for this function exercising the various behaviors? It looks like a good candidate (well isolated) and the unit test would help in documenting what it is and isn't supposed to do.

test/client/utility.cc
mum4k: Do we also want to include a "TODO(#259)" here?

test/integration/integration_test_fixtures.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, ip_version):
     self.server_ip = "::1" if ip_version == IpVersion.IPV6 else "127.0.0.1"
     self.socket_type = socket.AF_INET6 if ip_version == IpVersion.IPV6 else socket.AF_INET
     self.test_server = None
```
mum4k: Keeping both self.test_server and self.test_servers can be confusing to new readers. Would it be too much work to switch all existing code to using self.test_servers ? (with index of zero as appropriate)

Alternatively, can we add a comment explaining when is self.test_server expected to remain None?

test/integration/integration_test_fixtures.py
```diff
@@ -113,12 +124,35 @@ def getTestServerRootUri(self, https=False):
     uri = "%s://%s:%s/" % ("https" if https else "http", uri_host, self.test_server.server_port)
     return uri
 
+  def getAllTestServerRootUris(self, https=False):
+    """
+    Utility for getting the list of http://host:port/ that can be used to query the servers we started
+    in setUp()
+    """
+    uri_host = self.server_ip
```
mum4k: (optional) It tends to be better when if<->else branches are explicit. If we set a variable to value "A" and then conditionally change it to value "B", the reader might get confused, not noticing the condition.

How about:

```python
if self.ip_version == IpVersion.IPV6:
  uri_host = ...
else:
  uri_host = ...
```

test/integration/integration_test_fixtures.py
```diff
@@ -178,6 +212,23 @@ def getTestServerRootUri(self):
     return super(HttpIntegrationTestBase, self).getTestServerRootUri(False)
 
 
+class MultiServerHttpIntegrationTestBase(IntegrationTestBase):
+  """
+  Base for running plain http tests against multiple Nighthawk test servers
+  """
+
+  def __init__(self, ip_version, backend_count):
```
mum4k: Can we add documentation for the constructor, including an Args section describing the arguments?

test/integration/integration_test_fixtures.py
```diff
@@ -178,6 +212,23 @@ def getTestServerRootUri(self):
...
+    super(MultiServerHttpIntegrationTestBase, self).__init__(ip_version, backend_count)
+    self.nighthawk_test_config_path = os.path.join(
+        self.test_rundir, "test/integration/configurations/nighthawk_http_origin.yaml")
+
+  def getTestServerRootUri(self):
```
mum4k: Can we add doc comments for these? if there are just overrides for methods on the base class, we can just do:

```python
def getTestServerRootUri(self):
  """See base class."""
```

Of course we should ensure that the base class has documentation.

test/integration/integration_test_fixtures.py
```diff
@@ -196,6 +247,27 @@ def getTestServerRootUri(self):
     return super(HttpsIntegrationTestBase, self).getTestServerRootUri(True)
 
 
+class MultiServerHttpsIntegrationTestBase(IntegrationTestBase):
+  """
+  Base for https tests against multiple Nighthawk test servers
+  """
+
+  def __init__(self, ip_version, backend_count):
```
mum4k: Same comments about documentation apply to this class.

test/integration/test_integration_basics.py
```diff
@@ -441,3 +442,71 @@ def test_bad_arg_error_messages(http_test_server_fixture):
...
+      "benchmark.http_2xx:24"
+  ]
+  for uri in multi_http_test_server_fixture.getAllTestServerRootUris():
+    nighthawk_client_args.append("--multi-target-endpoint")
+    nighthawk_client_args.append(uri.replace("http://", "").replace("/", ""))
+
+  parsed_json, stderr = multi_http_test_server_fixture.runNighthawkClient(nighthawk_client_args)
+
+  counters = multi_http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
+  assertCounterEqual(counters, "benchmark.http_2xx", 25)
```
mum4k: This is a fairly large amount of assertions. Are all of them related to verifying the multi-backend functionality or are we combining multiple behaviors in this test?

Can you help me understand which of these assertions are actually verifying the multi-backend behavior?

test/options_test.cc
```diff
@@ -203,6 +204,67 @@ TEST_F(OptionsImplTest, TlsContext) {
...
+  EXPECT_EQ(cmd->multi_target().endpoints(3).port().value(), 6);
+
+  // Now we construct a new options from the proto we created above. This should result in an
+  // OptionsImpl instance equivalent to options. We test that by converting both to yaml strings,
+  // expecting them to be equal. This should provide helpful output when the test fails by showing
+  // the unexpected (yaml) diff.
+  // The predicates are defined as proto maps, and these seem to re-serialize into a different
+  // order. Hence we trim the maps to contain a single entry so they don't thwart our
+  // textual comparison below.
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("benchmark.http_4xx"));
```
mum4k: Similar question here - are all of these assertions here related to the new multi-target functionality?

Do we need to keep copying these into all new test cases?

api/client/options.proto
eric846: Added an example

include/nighthawk/client/options.h
eric846: Added comment

source/client/factories_impl.cc
eric846: Updated comment based on my understanding.

source/client/factories_impl.cc
eric846: Done

source/client/options_impl.cc
eric846: Done

test/integration/integration_test_fixtures.py
eric846: Done (added some comments to the class, but not sure what to say for the constructor apart from Args)

test/integration/integration_test_fixtures.py
eric846: Done

test/integration/integration_test_fixtures.py
eric846: Done

test/integration/test_integration_basics.py
eric846: The assertions involving 3 connections and the total number of responses are relevant to making sure multi-target is working correctly. I relaxed the ones about the total bytes, but we might still want check if there are more than zero bytes as a sanity check. Removed unneeded assertions about the number of stats.

test/options_test.cc
eric846: The assertion is not particularly relevant, but this "erase" command still needs to be done in order to do the following global proto diff. But we might still keep the assertion as a sanity check since the particular value I remove here is based on defaults that come into play, different from the main `AlmostAll` test.

PR #253, opened 2020-01-02
Push to envoyproxy/nighthawk-dev:latest

oschaaf:
This will make CI push snapshots of the latest and greatest from the `master` branch.


ci/docker/docker_push.sh
```diff
@@ -0,0 +1,8 @@
+set -e
+set +x
```
htuch: I recommend a comment similar to https://github.com/envoyproxy/envoy/blob/master/ci/docker_push.sh#L3 here.

oschaaf: Update this to work mostly like Envoy's version, thanks for pointing that out

PR #255, opened 2020-01-03
Labels option

oschaaf:
Adds a new `--label` option. Allows tagging executions with multiple
labels. These labels will be persisted in our structured output formats.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--jitter-uniform <duration>] [--open-loop]
+bazel-bin/nighthawk_client  [--label <<string>>] ... [--jitter-uniform
+<duration>] [--open-loop]
 [--experimental-h1-connection-reuse-strategy
 <mru|lru>] [--failure-predicate <<string,
 uint64_t>>] ... [--termination-predicate
@@ -69,6 +70,10 @@ format>
 
 
 Where:
+
+--label <<string>>  (accepted multiple times)
+Label. Allows specifying multiple labels which will be persisted in
+structured output formats.
 
 --jitter-uniform <duration>
 Add uniformly distributed absolute request-release timing jitter. For
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--label <<string>>] ... [--jitter-uniform
+bazel-bin/nighthawk_client  [--label <string>] ... [--jitter-uniform
 <duration>] [--open-loop]
 [--experimental-h1-connection-reuse-strategy
 <mru|lru>] [--failure-predicate <<string,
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -43,7 +43,8 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--jitter-uniform <duration>] [--open-loop]
+bazel-bin/nighthawk_client  [--label <<string>>] ... [--jitter-uniform
```
mum4k: Curious, why do we have some of the types with double and some with single angle brackets? Is this significant? Do we use ti to indicate that the option accepts multiple items?

source/client/output_formatter_impl.cc
```diff
@@ -184,8 +185,11 @@ const nighthawk::client::Statistic& FortioOutputFormatterImpl::getRequestRespons
 
 std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
   nighthawk::client::FortioResult fortio_output;
-
-  fortio_output.set_labels("Nighthawk");
+  std::string labels;
+  for (const auto& label : output.options().labels()) {
+    labels += label + " ";
```
mum4k: This feels like we are taking multiple labels and concatenating them, into one label. Is that the behavior? if yes, can we document it next to the new labels option?

Alternatively, maybe the FortioResult message expects a space separated list of labels? If that is so, can we document it here?

https://github.com/envoyproxy/nighthawk/blob/68e59719fe754401907c16e53156bf18cd0e4e65/api/client/transform/fortio.proto#L18

test/options_test.cc
```diff
@@ -133,6 +135,11 @@ TEST_F(OptionsImplTest, All) {
   EXPECT_EQ(cmd->jitter_uniform().nanos(), options->jitterUniform().count());
   EXPECT_EQ(cmd->experimental_h1_connection_reuse_strategy().value(),
             options->h1ConnectionReuseStrategy());
+  i = 0;
+  for (const auto& label : cmd->labels()) {
```
mum4k: This feels like we are re-implementing a matcher that might already exist. I am assuming we are using googletest under the hood, so we should be able to utilize the existing container matches, specifically `ElementsAre()`.

https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#matching-containers

That way we can directly compare two containers, the actual and the expected.

oschaaf: Nice; done via `ElementsAreArray()` in 3cb9f42354387ed6bca47bb241c87b64087a0dc7

source/client/output_formatter_impl.cc
oschaaf: Added a descriptive comment in the proto: 3cb9f42354387ed6bca47bb241c87b64087a0dc7

README.md
oschaaf: Hmm; double checking we pass this in via a TCLAP argument. It looks like that in some cases I wrapped the type description in `<>`, which is something TCLAP does by itself as well, leading up to double angle brackets. 
I backed it out here in 3cb9f42, and filed #256  for fixing this in other spots. Thanks for catching this.

PR #261, opened 2020-01-09
Update Envoy

oschaaf:
Update Envoy to 29b30911dbfb3f9760efeb28238ceac36e1a1a23

It is worth noting this upgrades Envoy proto api usage to `v3alpha`

Brings in the structured versioning API, which we need in #226 


source/server/README.md
```diff
@@ -114,6 +115,9 @@ bazel-bin/nighthawk_test_server  [--use-fake-symbol-table <bool>]
 
 Where:
 
+--disable-extensions <string>
```
mum4k: Are these changes related to this PR?

source/server/http_test_server_filter_config.cc
```diff
@@ -31,7 +31,7 @@ class HttpTestServerDecoderFilterConfig
     return Envoy::ProtobufTypes::MessagePtr{new nighthawk::server::ResponseOptions()};
   }
 
-  std::string name() override { return "test-server"; }
+  std::string name() const override { return "test-server"; }
```
mum4k: Similar question here, not sure I follow if this is related to updating Envoy. Do we want a separate PR with description that would explain these additional changes?

Alternatively, please help me understand why they are necessary here.

source/server/README.md
oschaaf: Yeah, this is auto generated and originates from the updated Envoy version which apparently has a new cli arg. As ‘nighthawk_test_server’ basically is Envoy + a static extension we may see this when updating the dependency.

source/server/http_test_server_filter_config.cc
oschaaf: > Can we hold off this PR for about a day? Eric is trying to import Nighthawk into our internal codebase. We are hoping to import just Nighthawk, not all of Envoy and this PR might break / delay that effort. We should be done with the import by tomorrow end of day.

Sure no problem

mum4k:
This is now ready to go, we have completed the import internally. Please help by resolving the merge conflict.

oschaaf:
Done in 23dc8ae6b5d6fe4a9e50af77100ef74959ebb88e

Issue #263, opened 2020-01-09
Investigate variability of TimingVerificationTest


eric846:
test/rate_limiter_test.cc:340
Expected equality of these values:
  getAcquisitionTimings(50_Hz, 1s)
    Which is: { 140, 200, 240, 360, 380, 400, 440, 460, 520, 540, 580, 600, 620, 640, 660, 680, 700, 740, 760, 780, 820, 840, 860, 880, 920, 940, 960, 980, 1000 }
  std::vector<int64_t>({120, 320, 380, 560, 580, 600, 620, 640, 660, 680, 700, 740, 760, 780, 840, 860, 880, 900, 920, 940, 960, 980, 1000})
    Which is: { 120, 320, 380, 560, 580, 600, 620, 640, 660, 680, 700, 740, 760, 780, 840, 860, 880, 900, 920, 940, 960, 980, 1000 }

Meanwhile the test is marked disabled.

oschaaf:
I suspect we're running into this: https://stackoverflow.com/questions/48730363/if-we-seed-c11-mt19937-as-the-same-on-different-machines-will-we-get-the-same
Not sure how to address that. @htuch do you have any suggestions here?

mum4k:
@oschaaf, now that we have a solution where we mock out the generator - do we still need to track this? Is there any downside to using the mock instead?

oschaaf:
@mum4k Thinking some more about it, you are right, this is 100% fine now. So as far as I am concerned we can close this.

mum4k:
Thank you @oschaaf.

PR #264, opened 2020-01-09
Temporary fix for TimingVerificationTest determinism

eric846:
Bypassed the random number generator in the test until we have a better solution (#263). (Now the fake distribution returns 500000 every time.) This test is for code that isn't wired in yet, so there is zero impact on users.

We need to update to the latest version of Nighthawk for a project that depends on new features, and this test is the only thing blocking the import.

Signed-off-by: eric846 <56563761+eric846@users.noreply.github.com>

### DIFF
--- a/test/rate_limiter_test.cc
+++ b/test/rate_limiter_test.cc
@@ -334,7 +334,7 @@ public:
   }
 };
 
-TEST_F(GraduallyOpeningRateLimiterFilterTest, TimingVerificationTest) {
+TEST_F(GraduallyOpeningRateLimiterFilterTest, DISABLED_TimingVerificationTest) {
   EXPECT_EQ(getAcquisitionTimings(50_Hz, 1s),
             std::vector<int64_t>({120, 320, 380, 560, 580, 600, 620, 640, 660, 680, 700, 740,
                                   760, 780, 840, 860, 880, 900, 920, 940, 960, 980, 1000}));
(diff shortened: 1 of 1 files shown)


eric846:
Unless there's a quick fix, can we go ahead and disable this test for now?

test/rate_limiter_test.cc
```diff
@@ -334,7 +334,7 @@ class GraduallyOpeningRateLimiterFilterTest : public Test {
   }
 };
 
-TEST_F(GraduallyOpeningRateLimiterFilterTest, TimingVerificationTest) {
+TEST_F(GraduallyOpeningRateLimiterFilterTest, DISABLED_TimingVerificationTest) {
```
oschaaf: Can we add something like:

```
// TODO(#263): Fix test determinism across all environments and re-enable
```

eric846: Done

test/rate_limiter_test.cc
```diff
@@ -295,7 +295,9 @@ class GraduallyOpeningRateLimiterFilterTest : public Test {
     std::uniform_int_distribution<uint64_t> dist(dist_min, dist_max);
```
mum4k: Do we need to keep the mt? This line:

```cc
std::mt19937_64 mt(1243);
```

It doesn't seem to be used anywhere.

eric846: Deleted for now, to be restored from history

PR #265, opened 2020-01-09
Add tracking of response body-size and header-size

### LINKED ISSUE #271: Add Fortio Sizes and HeaderSizes field
When `--output-format fortio` is specified, `Sizes` and `HeaderSizes` are missing.
Fortio tracks these as `DurationHistograms` named `Sizes` and `HeaderSizes`.
We need to start tracking these statistics, and wire them through to our output formats.

A sample Fortio json fragment:

```json
 "Sizes": {
    "Count": 3000,
    "Min": 116,
    "Max": 116,
    "Sum": 348000,
    "Avg": 116,
    "StdDev": 0,
    "Data": [
      {
        "Start": 116,
        "End": 116,
        "Percent": 100,
        "Count": 3000
      }
    ],
    "Percentiles": [
      {
        "Percentile": 50,
        "Value": 116
      }
    ]
  },
  "HeaderSizes": {
    "Count": 3000,
    "Min": 116,
    "Max": 116,
    "Sum": 348000,
    "Avg": 116,
    "StdDev": 0,
    "Data": [
      {
        "Start": 116,
        "End": 116,
        "Percent": 100,
        "Count": 3000
      }
    ],
    "Percentiles": [
      {
        "Percentile": 50,
        "Value": 116
      }
    ]
  },
```

oschaaf:
PR summary:

- Adds tracking of response body-size and header- size using the more lightweight `StreamingStatistic`. 
- Wires these through to all output formats. (Including the Fortio output, which is driving this effort).
- Adds tracking if `min`/`max` in our statistics, and wires that through to our output.
- Some small refactoring / enhancements in our statistics implementations
- Fix edge case in `StreamingStatistic::combine`
- Some new test and test enhancments, including a refresher of gold files.

The new statistics will show in the CLI as:

```
Response body size in bytes (20 samples)
min: 847 | mean: 847 | max: 847 | pstdev: 0

Response header size in bytes (20 samples)
min: 47282 | mean: 47308.3 | max: 47372 | pstdev: 20.6186
```

Assuming we don't need full blown histograms for this, this fixes #271
Note: I observed that the fortio seems to be tracking different things for size/headersize in fortio, depending on if the fast http client is used or not. As this seems like a bug, I didn't try to reproduce that specific behavior in NH.

----------

Waiting on:

# Prerequisites to merging this
- [X] #226 to go in
- [x] Address @mum4k his comments from https://github.com/envoyproxy/nighthawk/pull/226#discussion_r366631832 here
- [x] Discuss fields not used in `Fortio` visualization and `istio/tools`, but that actually are emitted in Fortio's json output: 
  - [x] SocketCount (only actually set for one of the two http clients of I remember correctly)
  - [x] AbortOn
  - [x] Exactly (-> target fixed amount of queries/replies, possibly could translate to NH termination 
    predicates, but not shown in visualization, not used in istio/tools)
- [x] Fortio's `RetCodes`
- [x] Discuss that Fortios `Sizes` and `HeaderSizes` [in its json output](https://github.com/fortio/fortio/blob/d98c066bd7acf65e12670ef6a6bea916ad01bd2c/fhttp/httprunner.go#L59
) are [used in `istio/tools`](https://github.com/istio/tools/blob/master/perf/benchmark/runner/fortio.py#L85) and missing in NH's output. However, they are only used to compute an error ratio (slightly different from how Fortio does that itself, and only precisely status 200 seems to count as succes), as well as compute an average response payload size for output in the csv transform.
In this PR we add simple tracking for the sizes (min/mean/max/stdev), and not full blown histogram tracking of sizes. It is worth nothing that Fortio seems to report different size values depending:

[1] 
- With -stdclient, Sizes reports response body size and 0-valued Header Sizes.

```json
  "Sizes": {
    ...
    "Avg": 14,
  },
  "HeaderSizes": {
    ...
    "Avg": 0,
  },
```

- Without -stdclient (which I think implies using the diy fastclient), Sizes reports body+header size for `Sizes` and body size for `HeaderSizes`. It's worth noting that it seems like Fortio may internally switch to stdclient when using https.

```json
  "Sizes": {
    ...
    "Avg": 185,
  },
  "HeaderSizes": {
    ...
    "Avg": 171,
  }
```


api/client/output.proto
```diff
@@ -13,17 +13,39 @@ message Counter {
 }
 
 message Percentile {
-  google.protobuf.Duration duration = 1;
+  oneof duration_type {
```
oschaaf: @mum4k I would love early feedback here with respect to tracking durations and response sizes in statistics like we discussed yesterday.
This change in `output.proto` is non-breaking; it adds a new `domain` field to indicate what we are tracking, and adds a bunch of `oneof`s / extra fields for the non-duration valued fields.
I followed up all the way to the CLI output formatter to get a feel for what this approach ends up like.
If this is acceptable, to round this up, I would need to update the other output formatters as well so we wire through the newly tracked statistics in all formats we emit.

mum4k: From what I know, this is a backward compatible change. I don't think we need to add the enums though, since the oneof itself can act as such. One can figure out which one of the oneof fields was set by calling `message.has_field()`.

Adding @htuch to get his feedback on whether this is a change we want to allow on the API, but looks good from my perspective.

api/client/output.proto
```diff
@@ -13,17 +14,39 @@ message Counter {
...
+    RAW = 0;      // When this is set, the raw_* uint64 valued fields will be populated.
+    DURATION = 1; // When this is set, the duration-valued fields will be populated.
+  }
   uint64 count = 1;
   string id = 2;
-  google.protobuf.Duration mean = 3;
-  google.protobuf.Duration pstdev = 4;
+  oneof mean_type {
+    google.protobuf.Duration mean = 3;
+    double raw_mean = 8;
```
mum4k: Not sure if "raw_" is the best prefix / name. Can we try to document what is the meaning of raw in each one of these cases and then we will see? Is it always bytes? Something else?

api/client/output.proto
htuch: From a gRPC/proto perspective, this is not a breaking wire change. It will break the Go language bindings, but that is probably OK for NIghthawk stuff today.

I'm unclear on what value the raw value is adding here.. why is `duration` not sufficient?

oschaaf: > From a gRPC/proto perspective, this is not a breaking wire change. It will break the Go language bindings, but that is probably OK for NIghthawk stuff today.
> 
> I'm unclear on what value the raw value is adding here.. why is `duration` not sufficient? Generally aiming for minimalism in wire representations.

In this PR we start tracking the number of bytes observed in response headers and bodies. Using the duration-valued fields for that as a storage backend didn't seem like the right thing to do to me.

api/client/output.proto
oschaaf: The definition of the `_raw` post-fixed fields map 1:1 to the return types of the corresponding getters from our internal `Statistics` c++ interface. In this PR we will use them to track the number of header bytes and response bytes, but possibly we may want to re-use these in the future to track things that aren't a "number of bytes"? (e.g. track number of requests over a single connection)

api/client/output.proto
htuch: I see. LGTM.

mum4k:
Please mark this again as waiting-for-review once the merge conflicts are resolved.

mum4k:
@eric846 please review and assign back to me when ready. Looks to be a lot of lines, but most of them are golden files.

source/common/statistic_impl.cc
```diff
@@ -9,23 +9,35 @@
...
-  statistic.mutable_mean()->set_seconds(nanos / 1000000000);
-  statistic.mutable_mean()->set_nanos(nanos % 1000000000);
-
-  nanos = count() == 0 ? 0 : static_cast<int64_t>(std::round(std::isnan(pstdev()) ? 0 : pstdev()));
-  statistic.mutable_pstdev()->set_seconds(nanos / 1000000000);
-  statistic.mutable_pstdev()->set_nanos(nanos % 1000000000);
+  if (domain == Statistic::SerializationDomain::DURATION) {
+    int64_t nanos;
+    nanos = count() == 0 ? 0 : static_cast<int64_t>(std::round(mean()));
+    statistic.mutable_mean()->set_seconds(nanos / 1000000000);
```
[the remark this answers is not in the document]
eric846: Also since this pattern with /1000000000 and %1000000000, maybe there should be a helper method for setting a proto duration, taking mutable Duration* and int64_t nanos

source/common/statistic_impl.cc
```diff
@@ -148,13 +171,21 @@ void HdrStatistic::addValue(uint64_t value) {
...
+// We override count for the Hdr statistics, because it may have dropped
+// out of range values. hence our own tracking may be inaccurate.
 uint64_t HdrStatistic::count() const { return histogram_->total_count; }
 double HdrStatistic::mean() const { return count() == 0 ? std::nan("") : hdr_mean(histogram_); }
 double HdrStatistic::pvariance() const { return pstdev() * pstdev(); }
 double HdrStatistic::pstdev() const { return count() == 0 ? std::nan("") : hdr_stddev(histogram_); }
+uint64_t HdrStatistic::min() const {
+  return count() == 0 ? UINT64_MAX : hdr_value_at_percentile(histogram_, 0);
+}
+uint64_t HdrStatistic::max() const { return hdr_value_at_percentile(histogram_, 100); }
```
eric846: Will max() blow up if count()==0?

source/common/statistic_impl.h
```diff
@@ -85,6 +110,7 @@ class InMemoryStatistic : public StatisticImpl {
     return streaming_stats_->resistsCatastrophicCancellation();
   }
   uint64_t significantDigits() const override { return streaming_stats_->significantDigits(); }
+  StatisticPtr createNewInstance() const override { return std::make_unique<InMemoryStatistic>(); };
```
eric846: +1 on the clever elimination of the factory. 

When I saw this being called, I assumed createNewInstance() would create a copy of the existing object -- the most obvious explanation for why it would be an instance method on an existing object. What about createNewInstanceOfSameType() to clarify the meaning?

source/common/statistic_impl.cc
oschaaf: We should be ok with respect to that (see `TypedStatisticTest::Empty`)

oschaaf:
@eric846 thanks for the review, comments addressed in  9aa9e98

PR #274, opened 2020-01-12
Facilitate SNI

### LINKED ISSUE #273: Use server name indication (sni)
Some https based load tests may fail. For example:

```
bazel-bin/nighthawk_client -v trace https://www.we-amp.com/
```

The logs then would show something like:

```
[12:13:19.042236][2848][D] [external/envoy/source/common/network/connection_impl.cc:563] [C0] connected
[12:13:19.042312][2848][D] [external/envoy/source/extensions/transport_sockets/tls/ssl_socket.cc:191] [C0] handshake expecting read
[12:13:19.141149][2848][T] [external/envoy/source/common/network/connection_impl.cc:466] [C0] socket event: 3
[12:13:19.141176][2848][T] [external/envoy/source/common/network/connection_impl.cc:554] [C0] write ready
[12:13:19.141202][2848][D] [external/envoy/source/extensions/transport_sockets/tls/ssl_socket.cc:198] [C0] handshake error: 5
[12:13:19.141212][2848][D] [external/envoy/source/extensions/transport_sockets/tls/ssl_socket.cc:226] [C0] 
```

To resolve this, one can configure the tls context to use server name indication.

```bash
bazel-bin/nighthawk_client -v trace --tls-context "{sni:\"www.we-amp.com\"}" https://www.we-amp.com/
````

I think it would be nice to automatically configure this in Nighthawk, and possibly have an option to opt-out of it doing that as an escape latch.

### PR DESCRIPTION
**Pull request by oschaaf**
For supporting load tests against servers that serve multiple hostnames
using a single ip address on ssl. Tries to automatically derive the
right sni host. This can be overridden by specifying a custom tls
context / transport socket  through configuration.

Fixes #273

### CODE REVIEW

File: source/client/process_impl.h [hunk 1]
```diff
@@ -68,6 +68,23 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
...
+   * Uri-derived sni host. If the passed in protocol is HTTP/2 or higher, next to host, :authority
+   * will be considered as well.
+   *
+   * @param parsed_uris The parsed target uris configured for the load test.
+   * @param request_headers Request headers to scan.
+   * @param protocol The anticipated protocol that will be used.
+   * @return std::string The sni-host derived from the configured load test target Uris and any
+   * host/authority request-headers found. Empty if no (unambiguous) sni host could be derived.
+   */
+  static std::string computeSniHost(const std::vector<UriPtr>& parsed_uris,
```
mum4k: Does this function in any way depend on the class we are attaching it to? If it doesn't can we place it in a pre-existing or a new relevant utility library or in an anonymous namespace within the same file? 

The latter would be preferred if nothing needs to call this from the outside, however we should keep the unit tests. Placing it outside of the class helps the reader understand that this doesn't depend on or modify the state of the class in any way.

oschaaf: Sure, done in 84f83b7

source/client/sni_utility.h
```diff
@@ -0,0 +1,34 @@
...
+#include <vector>
+
+#include "nighthawk/common/uri.h"
+
+#include "external/envoy/include/envoy/http/protocol.h"
+
+namespace Nighthawk {
+namespace Client {
+
+class SniUtility : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
```
mum4k: Curious, looks like we ended up defining a class for this anyways. I guess it is for the purposes of the logging mix-in. Is creation of a class the only way to use the Envoy logging facility? Or is there some other shape or form of it that we could use for plain utility functions that don't require to be on a class?

test/sni_utility_test.cc
```diff
@@ -0,0 +1,47 @@
...
+                               const std::vector<std::string>& request_headers,
+                               const Envoy::Http::Protocol protocol,
+                               absl::string_view expected_sni_host) {
+    std::vector<UriPtr> parsed_uris;
+    parsed_uris.reserve(uris.size());
+    for (const std::string& uri : uris) {
+      parsed_uris.push_back(std::make_unique<UriImpl>(uri));
+    }
+    std::string sni_host = SniUtility::computeSniHost(parsed_uris, request_headers, protocol);
+    EXPECT_EQ(sni_host, expected_sni_host);
```
mum4k: Calling these macros from helper functions tends to complicate test failure readability. I seem to remember that the test failures will point to this line and might not show a good view of the expected and actual values. Could you verify that by making one of these fail?

Would it be better to update this helper function to just construct the parsed_uri? We could then call both the function under test and the comparison macro (EXPECT_EQ) from the test body. That would also result in a more typical test.

oschaaf: amded in d915684fbbfd81630369b6fe99444df737a4ddc6

source/client/sni_utility.h
oschaaf: That is a good question. I think the mix-in is the way it is supposed to be used.
We could create one or more classes that mix these in just dedicated to logging?
e.g.
```c++
class MainLog : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
public:
 // bunch of helpers to log at various log levels, possibly shared from a base class
}
```

mum4k: As discussed, this is something we can explore in the future.

PR #281, opened 2020-01-15
rate-limiting: Add ScheduledStartingRateLimiter

oschaaf:
As per discussion [1], this is intended to be used as a replacement
for the way we offset worker request release timings today.

[1] https://github.com/envoyproxy/nighthawk/pull/219#discussion_r366537002

### DIFF
--- a/source/common/rate_limiter_impl.cc
+++ b/source/common/rate_limiter_impl.cc
@@ -52,6 +52,36 @@ void BurstingRateLimiter::releaseOne() {
   previously_releasing_ = absl::nullopt;
 }
 
+ScheduledStartingRateLimiter::ScheduledStartingRateLimiter(
+    RateLimiterPtr&& rate_limiter, Envoy::MonotonicTime scheduled_starting_time)
+    : ForwardingRateLimiterImpl(std::move(rate_limiter)),
+      scheduled_starting_time_(scheduled_starting_time) {
+  if (timeSource().monotonicTime() >= scheduled_starting_time_) {
+    throw NighthawkException("Scheduled starting time needs to be in the future");
+  }
+}
+
+bool ScheduledStartingRateLimiter::tryAcquireOne() {
+  if (timeSource().monotonicTime() < scheduled_starting_time_) {
+    aquisition_attempted_ = true;
+    return false;
+  }
+  // If we start forwarding right away on the first attempt that is remarkable, so leave a hint
+  // about this happening in the logs.
+  if (!aquisition_attempted_) {
+    aquisition_attempted_ = true;
+    ENVOY_LOG(warn, "First acquisition attempt was late");
+  }
+  return rate_limiter_->tryAcquireOne();
+}
+
+void ScheduledStartingRateLimiter::releaseOne() {
+  if (timeSource().monotonicTime() < scheduled_starting_time_) {
+    throw NighthawkException("Unexpected call to releaseOne()");
+  }
+  return rate_limiter_->releaseOne();
+}
+
 LinearRateLimiter::LinearRateLimiter(Envoy::TimeSource& time_source, const Frequency frequency)
     : RateLimiterBaseImpl(time_source), acquireable_count_(0), acquired_count_(0),
       frequency_(frequency) {
--- a/source/common/rate_limiter_impl.cc
+++ b/source/common/rate_limiter_impl.cc
@@ -70,7 +70,7 @@ bool ScheduledStartingRateLimiter::tryAcquireOne() {
   // about this happening in the logs.
   if (!aquisition_attempted_) {
     aquisition_attempted_ = true;
-    ENVOY_LOG(warn, "First acquisition attempt was late");
+    ENVOY_LOG(warn, "ScheduledStartingRateLimiter: first acquisition attempt was late");
   }
   return rate_limiter_->tryAcquireOne();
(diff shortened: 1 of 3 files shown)


mum4k:
@dubious90 please review and assign to me once done.

test/rate_limiter_test.cc
```diff
@@ -68,6 +68,58 @@ TEST_F(RateLimiterTest, BurstingRateLimiterTest) {
...
+    EXPECT_CALL(unsafe_mock_rate_limiter, timeSource)
+        .Times(AtLeast(1))
+        .WillRepeatedly(ReturnRef(time_system));
+    RateLimiterPtr rate_limiter = std::make_unique<ScheduledStartingRateLimiter>(
+        std::move(mock_rate_limiter), scheduled_starting_time);
+    EXPECT_CALL(unsafe_mock_rate_limiter, tryAcquireOne)
+        .Times(AtLeast(1))
+        .WillRepeatedly(Return(true));
+
+    if (starting_late) {
```
dubious90: I'm not sure if starting_late is adding any extra coverage for this test case.  This test should already be testing what happens when enough time has passed and when not enough time has passed.

test/rate_limiter_test.cc
```diff
@@ -68,6 +68,58 @@ TEST_F(RateLimiterTest, BurstingRateLimiterTest) {
...
+      time_system.sleep(schedule_delay);
+    }
+
+    // We should expect zero releases until it is time to start.
+    while (time_system.monotonicTime() < scheduled_starting_time) {
```
dubious90: If this test case doesn't succeed, it will run infinitely rather than fail.  Can you add in a tripping mechanism?

source/common/rate_limiter_impl.cc
```diff
@@ -52,6 +52,36 @@ void BurstingRateLimiter::releaseOne() {
...
+bool ScheduledStartingRateLimiter::tryAcquireOne() {
+  if (timeSource().monotonicTime() < scheduled_starting_time_) {
+    aquisition_attempted_ = true;
+    return false;
+  }
+  // If we start forwarding right away on the first attempt that is remarkable, so leave a hint
+  // about this happening in the logs.
+  if (!aquisition_attempted_) {
+    aquisition_attempted_ = true;
+    ENVOY_LOG(warn, "First acquisition attempt was late");
```
dubious90: Can we add in an identifier here, so this message makes more sense for a debugger who might not have context into this specific rate limiter?
Something like
```ScheduledStartingRateLimiter first acquisition attempt was late```

File: source/common/rate_limiter_impl.cc [hunk 4]
```diff
@@ -52,6 +52,36 @@ void BurstingRateLimiter::releaseOne() {
...
+  }
+  return rate_limiter_->tryAcquireOne();
+}
+
+void ScheduledStartingRateLimiter::releaseOne() {
+  if (timeSource().monotonicTime() < scheduled_starting_time_) {
```
dubious90: I'm probably missing something, but couldn't a cancel operation (not yet implemented) result in releasing before the time arrives?  What are we trying to prevent by failing here?

oschaaf: Right now, it is only valid to release what has been acquired. So when we hit this, it ought to be impossible to have a preceding successful acquisition because we're not due to start yet. The intent is to catch programming mistakes early when consuming this, but possibly this is overly paranoid.. shall I remove this?

source/common/rate_limiter_impl.cc
oschaaf: done

test/rate_limiter_test.cc
oschaaf: I'm not sure I follow; the intent here is to skip the scheduled delay entirely, so the first call to `acquireOne()` will hit the line that logs the warning that the "first acquisition attempt was late":

https://github.com/envoyproxy/nighthawk/pull/281/files/de66b5e48997d31728125385f33265bfb62094df#diff-698caf1e153397f66dc8ca94b64a391fR72

test/rate_limiter_test.cc
oschaaf: The simulated time system here is guaranteed to move forward in this loop with `1ms` steps we add via the call to `time_system.sleep(1ms)` below.. afaict this is guaranteed to terminate?

source/common/rate_limiter_impl.cc
dubious90: This explanation makes sense to me. I don't think it's necessary to remove this.

test/rate_limiter_test.cc
dubious90: Ah, okay. I didn't process the comments in line 74-76 properly when I read this code the first time.  I follow now.

Is there a way to assert that the log line was actually hit? Or is this just for line code coverage?

test/rate_limiter_test.cc
dubious90: Nevermind here. I was overcomplicated the situation in my head. Sorry about that!

test/rate_limiter_test.cc
oschaaf: This is just for coverage. Unfortunately we can't really mock the logging today and I couldn't find an efficient way to assert on hitting that line

Issue #280, opened 2020-01-15
nighthawk_client: handle termination signal


oschaaf:
It would be nice if nighthawk_client could gracefully and quickly terminate on `SIGTERM`.
One way to achieve that is by:
- making the signal handling currently in effect in `nighthawk_service` shareable.
- add it to `nighthawk_client`. 

upon signal reception, we could bump a counter.
we can add default termination predicates with a zero threshold to make the workers respond to that.

This gives CLI users a means to let a test run for a while and manually abort it, and still get results out.

PR #282, opened 2020-01-15
Extract SignalHandler class for re-use

oschaaf:
In preparation of sharing functionality for signal handling,
extract what we have right now into a `SignalHandler` class.

Preparation for #280


oschaaf:
@mum4k requesting early feedback on this; this extracts our signal handling code from `nighthawk_service` into a separate class for re-use later. In terms of design and fool proofing this we can probably enhance this, but a perfect design is not the initial goal: 
- I think this would be an improvement compared to what's on master today as we contain this functionality.
- `service_main.cc` gets much cleaner.

Would it make sense to shoot for landing this as-is?

source/common/signal_handler.h
```diff
@@ -0,0 +1,44 @@
...
+#include <functional>
+#include <memory>
+#include <thread>
+#include <vector>
+
+#include "external/envoy/source/common/common/logger.h"
+
+namespace Nighthawk {
+
+using SignalCallback = std::function<void()>;
```
mum4k, line 12: Since this is part of the header - can we document it?

source/common/signal_handler.h
```diff
@@ -0,0 +1,44 @@
...
+
+class SignalHandler final : public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
```
mum4k, line 14: Let's also add documentation for the class including an example of usage.

source/common/signal_handler.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+  SignalHandler(const SignalCallback& signal_callback);
```
mum4k, line 16: Can we add documentation for the constructor?

source/common/signal_handler.h
```diff
@@ -0,0 +1,44 @@
...
+  SignalHandler(SignalHandler const&) = delete;
```
mum4k, line 17: Can we add a single comment above these two to indicate the intention, something like:

// Not copyable or movable.

oschaaf:
@mum4k thanks, comments addressed in e665347

PR #285, opened 2020-01-17
Re-introduce phases

oschaaf:
This PR has two commits:
- Reverts 870d7c8888a6d6c8d0f32dcd7de69413fda58d68, re-introducing phases again
- Additional changes in 12cf789 

------ 

Fixes:

- Pass the computed start time around to all involved components, and use that to compute a consistent view of elapsed time.

Enhancements:

- Phase shift the timing of `LinearRateLimiter` to avoid acquisitions at second boundaries.
- Test queries / replies / latency samples counts in various configuration to make sure this
  stays put.

Unrelated changes:
- Bump the CI coverage threshold to slightly below what is passing.
- Fix a bug in the test code for correctly obtaining the global histogram.


source/common/sequencer_impl.cc
```diff
@@ -82,49 +81,50 @@ void SequencerImpl::updateStartBlockingTimeIfNeeded() {
 
 void SequencerImpl::run(bool from_periodic_timer) {
   ASSERT(running_);
+  // Update time so that our time source will yield an up-to-date monotonic time sample to work with
+  // during this function call.
+  dispatcher_.updateApproximateMonotonicTime();
```
mum4k: This feels a little bit counter intuitive. We now have a clock (time) abstraction that moves forward only when we tell it to. This might lead to programming errors and the reasoning isn't obvious?

If I understand fixes in this PR correctly, when compared to the first attempt on phases we are now passing the expected start time to all components. Why are we also caching time? Is it an unrelated performance improvement or required for phases? If it isn't required for phases, can we handle it in a separate PR and start by discussing the numbers that lead us to believe that caching is necessary?

oschaaf: I will back this out and create a separate PR. It’s not needed to address the problem I noticed, but rather a separate potential enhancement. For context, performance is not the goal of the caching here, but a side effect:, it’s consistency of “now” across the sequencer, rate limiters and termination predicates that I liked.  I thought this might be nice as it avoids subtle edge cases caused by small slight differences caused by querying the clock independently.

oschaaf: Back out in 048dd15daeed78a068d37e94b8d00c96948eab5d

PR #293, opened 2020-01-22
Introduce cached time source

oschaaf:
This helps achieve a consistent view of “now” across the sequencer and its associated rate limiter(s) and termination predicate(s). This should help avoiding subtle edge cases caused by small differences when these components all query the clock independently (and possibly TOCTOU).

------

Prerequisite:

- [x] Land #285, from which this change was split out.
- [x] Merge master


mum4k:
@dubious90 please review and assign to me when ready.

@htuch, this feels like a change where your feedback would be valuable.

mum4k:
@oschaaf please merge master back here after phases were merged to ease the review.

oschaaf:
> @oschaaf please merge master back here after phases were merged to ease the review.

@mum4k done!

source/common/sequencer_impl.cc
```diff
@@ -81,6 +81,9 @@ void SequencerImpl::updateStartBlockingTimeIfNeeded() {
 
 void SequencerImpl::run(bool from_periodic_timer) {
   ASSERT(running_);
+  // Update time so that our time source will yield an up-to-date monotonic time sample to work with
+  // during this function call.
+  dispatcher_.updateApproximateMonotonicTime();
```
pemor: if I understand this correctly, then real monotonicTime() is called in updateApproximateMonotonicTime, which means no optimization gain for this specific call. 
Can you please call out where the cached_time_source..monotonicTime() would be saving time.

source/client/client_worker_impl.h
```diff
@@ -48,6 +48,7 @@ class ClientWorkerImpl : public WorkerImpl, virtual public ClientWorker {
 private:
   void simpleWarmup();
 
+  std::unique_ptr<Envoy::TimeSource> cached_time_source_;
```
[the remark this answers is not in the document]
oschaaf: I opted for a smart pointer here so we can force the implementation to go through the (abstract) `Envoy::TimeSource` interface. E.g. maybe we'll some day want an escape latch option to use the `regular` time source` and then it's just a matter of initializing this differently?

source/common/sequencer_impl.cc
oschaaf: I added a more elaborate comment in 823c15eba8d6261edb832840479e1c69bdd1d7e8, does that look helpful?

source/client/client_worker_impl.h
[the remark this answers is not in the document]
pemor: make sense, sorry I missed it.

source/common/sequencer_impl.cc
pemor: Very helpful. Thanks.

PR #295, opened 2020-01-24
Simple warmup: default to opting out

oschaaf:
The current single request we issue for the simple warmup pollutes
some of the counters we emit in our output.

Looking forward, when landing generic phases, we might want to
default to opting out.
This new flag could then serve as an opt-in afterwards, for those
who want to preserve old behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--no-simple-warmup] [--request-source <uri
+bazel-bin/nighthawk_client  [--simple-warmup] [--request-source <uri
 format>] [--label <string>] ...
 [--multi-target-use-https]
 [--multi-target-path <string>]
@@ -78,7 +78,7 @@ format>
 
 Where:
 
---no-simple-warmup
+--simple-warmup
 Do not perform the simple warmup call.
 
 --request-source <uri format>
--- a/README.md
+++ b/README.md
@@ -79,7 +79,9 @@ format>
 Where:
 
 --simple-warmup
-Do not perform the simple warmup call.
+Perform a simple single warmup request (per worker) before starting
+execution. Note that this will be reflected in the counters that
+Nighthawk writes to the output. Default is false.
 
 --request-source <uri format>
 Remote gRPC source that will deliver to-be-replayed traffic. Each
--- a/README.md
+++ b/README.md
@@ -43,8 +43,9 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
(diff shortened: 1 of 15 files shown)


api/client/options.proto
```diff
@@ -187,4 +187,6 @@ message CommandLineOptions {
   repeated string labels = 28;
   // TransportSocket configuration to use in every request
   envoy.config.core.v3.TransportSocket transport_socket = 27;
+  // Do not perform the simple warmup call.
+  google.protobuf.BoolValue no_simple_warmup = 32;
```
mum4k: We have learned internally that naming boolean flags in the negative can end up being confusing. E.g:

--no-simple-warmup=False

Can we name the flag simple-warmup instead and set its default accordingly?

mum4k: On a similar topic - why wouldn't we make it an opt-in right away? What are the use cases to have this enabled by default?

source/client/client_worker_impl.h
```diff
@@ -31,7 +31,7 @@ class ClientWorkerImpl : public WorkerImpl, virtual public ClientWorker {
                    const RequestSourceFactory& request_generator_factory,
                    Envoy::Stats::Store& store, const int worker_number,
                    const Envoy::MonotonicTime starting_time,
-                   Envoy::Tracing::HttpTracerPtr& http_tracer);
+                   Envoy::Tracing::HttpTracerPtr& http_tracer, const bool simple_warmup);
```
mum4k: How will this API evolve? This is about my usual concern around readability of boolean flags. Is this related to phases and will this be removed in favor of some phase information? Or should we introduce an enum now?

api/client/options.proto
oschaaf: fixed in 5940d5601eaa9924782c57be5a67895a7a1dfdd5

oschaaf: I think that is a good idea, fixed in 5940d5601eaa9924782c57be5a67895a7a1dfdd5 (with a callout in the changelist)

source/client/client_worker_impl.h
oschaaf: See 5940d5601eaa9924782c57be5a67895a7a1dfdd5, let me know what you think.

README.md
```diff
@@ -77,6 +78,9 @@ format>
 
 Where:
 
+--simple-warmup
+Do not perform the simple warmup call.
```
mum4k: We should also update the comment next to the flag. Should now say something like:

Perform the simple warmup call.

While we are here, do we also want to expand this a bit to explain what simple warmup call is?

oschaaf: whoops, done in bb70db154224cb50d06d315eb7e3359b25a8be9f (thanks!)

PR #298, opened 2020-01-28
Fortio output: add bytes sent / received

oschaaf:
This adds two new fields to the fortio output:

```proto
  // Bytes sent should reflect the bytes sent after protocol serialization, compression, and
  // encryption are applied (i.e., the size of the buffers passed to the send() system call or its
  // equivalent).
  uint64 BytesSent = 18;

  // Bytes received should reflect the bytes read before decryption, de-compression, and protocol
  // de-serialization are applied (i.e., the size of bytes read from the recv() system call or its
  // equivalent).
  uint64 BytesReceived = 19;
```

### DIFF
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -213,17 +213,16 @@ FortioOutputFormatterImpl::getGlobalResult(const nighthawk::client::Output& outp
   throw NighthawkException("Nighthawk output was malformed, contains no 'global' results.");
 }
 
-const nighthawk::client::Counter&
-FortioOutputFormatterImpl::getCounterByName(const nighthawk::client::Result& result,
-                                            absl::string_view counter_name) const {
+bool FortioOutputFormatterImpl::visitCounter(
+    const nighthawk::client::Result& result, absl::string_view counter_name,
+    std::function<void(const nighthawk::client::Counter&)> callback) const {
   for (const auto& nh_counter : result.counters()) {
     if (nh_counter.name() == counter_name) {
-      return nh_counter;
+      callback(nh_counter);
+      return true;
     }
   }
-
-  throw NighthawkException(absl::StrCat(
-      "Nighthawk result was malformed, contains no counter with name: ", counter_name));
+  return false;
 }
 
 const nighthawk::client::Statistic*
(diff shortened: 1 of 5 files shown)


oschaaf:
This will miss out on bytes exchanged during the ssl handshake. We'd need to adjust how Envoy performs the ssl handshake to capture that; consensus is that this isn't worth it. This captures what we want for our use case over at `istio/tools`, so I'd like to take this forward as-is. Reading this up for review.

mum4k:
@eric846 please review and assign to me once ready.

oschaaf:
Talking to @htuch there might be a way to capture the ssl overhead after all, by adding a transport socket wrapper. Like https://github.com/envoyproxy/envoy/blob/5248a4fb7d4c2a3d1fa151f944d3a63f6b7a06cf/source/extensions/transport_sockets/tap/tap.cc. IMHO We can still land this, and then try to land an improvement in a follow up.

source/client/output_formatter_impl.cc
```diff
@@ -288,45 +287,58 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
...
   // Get the result that represents all workers (global)
-  const auto& nh_global_result = this->getGlobalResult(output);
+  const auto& nh_global_result = getGlobalResult(output);
 
   // Fill in the actual QPS based on the counters
-  const auto& nh_rq_counter = this->getCounterByName(nh_global_result, "upstream_rq_total");
-  const double actual_qps = static_cast<double>(
-      nh_rq_counter.value() / std::chrono::duration<double>(actual_duration).count());
-  fortio_output.set_actualqps(actual_qps);
+  if (!visitCounter(
```
eric846: We only throw an error about a missing counter for `upstream_rq_total`, but no others. Is there something that guarantees the other counters will always be present?

source/client/output_formatter_impl.cc
```diff
@@ -288,45 +287,58 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
...
-    // codes.. and possibly also factor in connection failures, etc.
-    fortio_output.mutable_retcodes()->at("200") = nh_2xx_counter.value() - number_of_workers;
-  } catch (const NighthawkException& e) {
-    // If this field doesn't exist, then there were no 2xx responses
-    fortio_output.mutable_retcodes()->at("200") = 0;
-  }
-
-  auto* statistic =
-      this->findStatistic(nh_global_result, "benchmark_http_client.request_to_response");
+  visitCounter(nh_global_result, "benchmark.http_2xx",
```
eric846: The behavior of the new code seems different -- if the counter isn't found, the old code used to insert a 0 under the "200" retcode. Is it safe to assume the `benchmark.http_2xx` counter will always be present?

source/client/output_formatter_impl.h
```diff
@@ -69,11 +69,11 @@ class FortioOutputFormatterImpl : public OutputFormatterImpl {
...
    * @param result a single Nighthawk result, preferably the global result
    * @param counter_name the name of the counter to return
-   * @return the corresponding counter
-   * @throws NighthawkException if counter with given name is not found
+   * @param callback The callback to run iff a counter is found, passing the counter as an argument.
+   * @return True iff a counter was found.
    */
-  const nighthawk::client::Counter& getCounterByName(const nighthawk::client::Result& result,
-                                                     absl::string_view counter_name) const;
+  bool visitCounter(const nighthawk::client::Result& result, absl::string_view counter_name,
```
eric846: Would it simplify the calling code to have a parameter that if set to true asks visitCounter to throw an error if the counter is not found? (Not sure if that would be mixing up concerns.)

oschaaf:
@nareddyt currently in the Fortio output transform we throw when Nighthawk's`upstream_rq_total` counter is not present in the input proto. This may happen, for example, when testing against a dead address. 

I was conserving that behavior here, but I'm wondering: what would be against not throwing and instead assuming 0 as value if the NH counter is not present?

Trying it out, we'd end up with something like:

```json
{
 "Labels": "",
 "RequestedQPS": 5,
 "ActualQPS": 0,
 "ActualDuration": 100380192,
 "NumThreads": 100,
 "DurationHistogram": {
  "Count": "0",
  "Data": [
   {
    "Start": 0,
    "End": 0,
    "Percent": 100,
    "Count": "0"
   }
  ],
  "Min": 0,
  "Max": 0,
  "Sum": 0,
  "Avg": 0,
  "StdDev": 0,
  "Percentiles": []
 },
 "RetCodes": {
  "200": "0"
 },
 "URL": "127.0.0.1:10",
 "Version": "0.3.0",
 "Jitter": false,
 "RunType": "HTTP",
 "Sizes": {
  "Count": "0",
  "Data": [],
  "Min": 1.8446744073709552e+19,
  "Max": 0,
  "Sum": "NaN",
  "Avg": "NaN",
  "StdDev": "NaN",
  "Percentiles": []
 },
 "HeaderSizes": {
  "Count": "0",
  "Data": [],
  "Min": 1.8446744073709552e+19,
  "Max": 0,
  "Sum": "NaN",
  "Avg": "NaN",
  "StdDev": "NaN",
  "Percentiles": []
 },
 "BytesSent": "0",
 "BytesReceived": "0",
 "StartTime": "2020-02-04T12:53:11.932916507Z",
 "RequestedDuration": "5s"
}
```

test/test_data/output_formatter.medium.fortio.gold
```diff
@@ -476,6 +476,8 @@
    }
   ]
  },
+ "BytesSent": "3528",
```
pamorgan: Is there a test case that covers this change?

oschaaf: Yes, the integration tests check rx/tx byte counts e.g: https://github.com/envoyproxy/nighthawk/blob/1ef11bb53556c856e043ce64a1d1108635a279ad/test/integration/test_integration_basics.py#L32

source/client/output_formatter_impl.cc
oschaaf: > The behavior of the new code seems different -- if the counter isn't found, the old code used to insert a 0 under the "200" retcode. Is it safe to assume the `benchmark.http_2xx` counter will always be present?

Well, I came across this, and noticed that it gets inserted in the line above unconditionally. So I figured it wasn't needed below, and intended to clean that up. Does that make sense?

eric846: Ohh I see, yes, this one should be ok.

source/client/output_formatter_impl.cc
```diff
@@ -288,45 +287,54 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
...
+          nh_global_result, "upstream_rq_total",
+          [&fortio_output, actual_duration](const nighthawk::client::Counter& nh_counter) {
+            const double actual_qps = static_cast<double>(
+                nh_counter.value() / std::chrono::duration<double>(actual_duration).count());
+            fortio_output.set_actualqps(actual_qps);
+          })) {
+    throw NighthawkException(
+        "Nighthawk result was malformed, contains no counter with name: upstream_rq_total");
+  };
+  visitCounter(nh_global_result, "upstream_cx_rx_bytes_total",
```
eric846: Is the intended behavior that `upstream_cx_rx_bytes_total` and `upstream_cx_tx_bytes_total` will be unset in the fortio output if those counters were missing?

oschaaf: > Is the intended behavior that `upstream_cx_rx_bytes_total` and `upstream_cx_tx_bytes_total` will be unset in the fortio output if those counters were missing?

No, thinking about it, it would probably be helpful to consumers to zero them. Thanks for catching that.

source/client/output_formatter_impl.cc
eric846: This was addressed in another thread.

source/client/output_formatter_impl.h
eric846: This doesn't seem to be needed, given that we only throw an error in the absence of one particular counter.

eric846:
LGTM modulo resolution of pamorgan's questions.

oschaaf:
@nareddyt kindly pinging this, I would love your input on the strategy with regard to exceptions

oschaaf:
(and @eric846 thanks for the review!)

nareddyt:
Sorry for the late response. I don't remember my initial reasoning for throwing an exception instead of defaulting to 0, it might have just been oversight.

It's probably fine to convert to 0. It will just show up as 0 in the fortio UI.

https://github.com/fortio/fortio/blob/f8f17fa310d60960f8267e1625231cf6e5ce459d/ui/uihandler.go#L350

oschaaf:
> Sorry for the late response. I don't remember my initial reasoning for throwing an exception instead of defaulting to 0, it might have just been oversight.
> 
> It's probably fine to convert to 0. It will just show up as 0 in the fortio UI.
> 
> https://github.com/fortio/fortio/blob/f8f17fa310d60960f8267e1625231cf6e5ce459d/ui/uihandler.go#L350

Grest, so if there's nothing depending on the behavior I will take this forward then. Thanks!

oschaaf:
@pamorgan Please see 197201173da7b3a5c8ac52d68594d7a0280f32c0, which rather simplifies things. We switch the strategy to do best-effort translation to Fortio's json format in case we miss counters or don't have any latency samples at all. This rids us of the visitor pattern that was introduced here, and instead introduces `getCounterValue(input, counter_name, default_value=0)`. Let me know if that looks better. 

Looking ahead, once this intermediate step lands, I'm working to figure out if and how we can add a wrapper for the sockets we use in NH for the purpose of accurately tracking tx/rx bytes. Doing so would allow us to report more truthfully when using other transports then plain text (e.g. ssl).

oschaaf:
@pamorgan kindly pinging this review, see https://github.com/envoyproxy/nighthawk/pull/298#issuecomment-583325403

PR #299, opened 2020-02-04
Update Envoy dependency

oschaaf:
- Update Envoy to 0d9f2d59f9cac9dc742e8bcd4588e71e9331f4ac
- Sync/tighten some clang-tidy check options
- Fixes to unbreak the build

### DIFF
--- a/source/client/benchmark_client_impl.h
+++ b/source/client/benchmark_client_impl.h
@@ -15,7 +15,7 @@
 #include "nighthawk/common/statistic.h"
 
 #include "external/envoy/source/common/common/logger.h"
-#include "external/envoy/source/common/http/http1/conn_pool.h"
+#include "external/envoy/source/common/http/http1/conn_pool_legacy.h"
 #include "external/envoy/source/common/http/http2/conn_pool.h"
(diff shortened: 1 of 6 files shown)


source/client/benchmark_client_impl.h
```diff
@@ -74,7 +74,7 @@ class Http1PoolImpl : public Envoy::Http::Http1::ProdConnPoolImpl {
 // --max-requests-per-connection may help as well, as doing periodically initiating new connections
 // may help the benchmark target by giving it an opportunity to rebalance.
 class Http2PoolImpl : public Envoy::Http::ConnectionPool::Instance,
-                      public Envoy::Http::ConnPoolImplBase {
+                      public Envoy::Http::Legacy::ConnPoolImplBase {
```
pamorgan: Do we have a bug to remove legacy?

oschaaf: Not yet -- but let me add one. It looks like some exciting stuff landed in Envoy :-)

oschaaf:
@pamorgan added https://github.com/envoyproxy/nighthawk/issues/302 for tracking that we should switch to the new pool implementations / stop using the legacy ones.

PR #303, opened 2020-02-09
Nighthawk socket wrapper

oschaaf:
This registers our own `SocketFactory`. Doing so allows our own new socket implementation to wrap around the configured transport sockets. In turn that allows us to observe the underlying transport sockets, which is going to help with accomplishing accurate tracking of bytes written and received.

The intent is that this also works with arbitrary `--transport-socket` configurations.

Part of #304, needs changes on the Envoy side (`SslSocket`) to complete.

-----

**State draft / needs early feedback**. The principle has been manually verified, but
it would be good to confirm this doesn't break anything. The NH socket implementation
currently writes bytes read/written as observed on the socket to stderr.

--

- [x] Discuss when Harvey is back


oschaaf:
Pushed a change which goes a long way towards finalizing this change; but when amending test expectations I observed that we're not out of the woods with this yet: the socket wrapper also doesn't get to see what happens with regard to tls transport.

It seems all the action is directly on the fd owned by SslSocket, which doesn't bubble up i/o in the `Envoy::Network::IoResult`s returned by `SslSockets`'s doRead() / doWrite(). 

So we end up with the same byte counts as the connection, though we do add tracking of  the number of reads and writes.

oschaaf:
@envoyproxy/nighthawk-reviewers 
This needs additional testing to fix coverage issues, but I think that is a small amount of work to add.
However, unfortunately this doesn't address the problem it intended to solve on its own. We're still not capturing the SSL overhead. To make that happen, we would need changes to Envoy's `SslSocket` implementation. So I think the value this adds is:
- tracing the number of reads and writes
- if `SslSocket` would be changed to truthfully return bytes handled for reads and writes, the counters introduced here would capture that (including the ssl handshake, which wouldn't be captured at the connection level still).

It would be good to assess if that's enough value to land this. I'm up for it, it's relatively little work, but it's not strictly needed to accomplish with we need over at `istio/tools`

mum4k:
@eric846 please review and assign back to me when ready.

source/common/socket.cc
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Nighthawk {
+
+class SocketConfigFactoryImpl : public UpstreamSocketConfigFactory {};
+
+Envoy::ProtobufTypes::MessagePtr UpstreamSocketConfigFactory::createEmptyConfigProto() {
+  return std::make_unique<nighthawk::transport_socket::TransportSocket>();
+}
+
+Envoy::Network::TransportSocketFactoryPtr UpstreamSocketConfigFactory::createTransportSocketFactory(
```
eric846: I think this method needs an overview comment and also a comment that explains each line. I didn't understand the meaning of "inner" and "outer" at first, and we are dealing with both `auto& inner_config_factory` and `MessagePtr inner_factory_config`. 

Let me see if I followed this:

- We register `UpstreamSocketConfigFactory` with Envoy
- `UpstreamSocketConfigFactory` provides the name "NighthawkSocket" for Envoy's registry
- Envoy calls `UpstreamSocketConfigFactory::createTransportSocketFactory` when it needs a transport socket factory, passing the transport socket config `Message message` from the bootstrap
- We extract the "outer config" `nighthawk::transport_socket::TransportSocket` proto from `message`
- Contained within the outer config is the transport socket "inner config" proto the user specified, or a "raw buffer" transport socket config if they didn't specify it
- We look up the transport socket factory-factory of the inner config in the Envoy registry
- We create the inner transport socket factory, passing the inner config *again*
- We return a new outer transport socket factory that wraps the inner transport socket factory

Now that Envoy has the outer transport socket factory, Envoy will call it:
- the outer transport socket factory calls the inner transport socket factory to get the real transport socket
- the outer transport socket factory creates a `Socket`, passing the real transport socket
- `Socket` increments counters in some methods, and in every method delegates to the real transport socket

source/common/socket.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+class SocketConfigFactory
+    : public virtual Envoy::Server::Configuration::TransportSocketConfigFactory {
+public:
+  ~SocketConfigFactory() override = default;
+  std::string name() const override { return "NighthawkSocket"; }
+};
+
+class UpstreamSocketConfigFactory
```
eric846: If I saw an interface called `TransportSocketConfigFactory` I would expect it to return something of type `TransportSocketConfig`. But this one returns `TransportSocketFactoryPtr`. 

Ideally it would be named `ProtoConfigTransportSocketFactoryFactory`.

Is there any way to use our class name to mitigate Envoy's naming, without also confusing people familiar with Envoy?

source/common/socket.h
```diff
@@ -0,0 +1,79 @@
...
+
+struct SocketStats {
+  ALL_SOCKET_STATS(GENERATE_COUNTER_STRUCT)
+};
+
+class SocketConfigFactory
+    : public virtual Envoy::Server::Configuration::TransportSocketConfigFactory {
+public:
+  ~SocketConfigFactory() override = default;
+  std::string name() const override { return "NighthawkSocket"; }
```
eric846: If `NighthawkSocket` only appears here, how does Envoy's registry locate the factory? Does it use the proto type name?

source/common/socket.cc
```diff
@@ -0,0 +1,95 @@
...
+                                                       context.messageValidationVisitor(),
+                                                       inner_config_factory);
+  auto inner_transport_factory =
+      inner_config_factory.createTransportSocketFactory(*inner_factory_config, context);
+  return std::make_unique<SocketFactory>(outer_config, SocketConfigFactoryImpl(),
+                                         context.scope().createScope("upstream_socket."),
+                                         std::move(inner_transport_factory));
+}
+
+SocketFactory::SocketFactory(const nighthawk::transport_socket::TransportSocket&,
```
eric846: @mum4k I don't have enough experience to review the `&&`

source/common/socket.cc
```diff
@@ -0,0 +1,95 @@
...
+      outer_config.transport_socket());
+  Envoy::ProtobufTypes::MessagePtr inner_factory_config =
+      Envoy::Config::Utility::translateToFactoryConfig(outer_config.transport_socket(),
+                                                       context.messageValidationVisitor(),
+                                                       inner_config_factory);
+  auto inner_transport_factory =
+      inner_config_factory.createTransportSocketFactory(*inner_factory_config, context);
+  return std::make_unique<SocketFactory>(outer_config, SocketConfigFactoryImpl(),
+                                         context.scope().createScope("upstream_socket."),
+                                         std::move(inner_transport_factory));
```
eric846: This looks like a double move -- is that ok?

source/common/socket.cc
```diff
@@ -0,0 +1,95 @@
...
+}
+
+SocketFactory::SocketFactory(const nighthawk::transport_socket::TransportSocket&,
+                             SocketConfigFactory&&, Envoy::Stats::ScopePtr&& scope,
```
eric846: Are `const nighthawk::transport_socket::TransportSocket&` and `SocketConfigFactory&&` unused? Can we just remove these parameters? If they are required or used, I didn't understand how.

source/client/process_impl.cc
```diff
@@ -275,6 +277,16 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v3::Boo
     if (options_.transportSocket().has_value()) {
       *cluster->mutable_transport_socket() = options_.transportSocket().value();
     }
+    // Now we wrap whatever Envoy-native transport socket config we came up up with into our own
+    // transport socket extension configuration from socket.h.
+    nighthawk::transport_socket::TransportSocket socket;
```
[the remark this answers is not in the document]
oschaaf: Yes that is the case

api/client/socket.proto
```diff
@@ -0,0 +1,14 @@
...
+package nighthawk.transport_socket;
+
+import "envoy/config/core/v3/base.proto";
+
+import "validate/validate.proto";
+
+// [#extension: nighthawk.transport_sockets.transport_socket]
+message TransportSocket {
+  // The underlying transport socket being wrapped.
+  envoy.config.core.v3.TransportSocket transport_socket = 2
```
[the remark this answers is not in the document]
oschaaf: No, fix in 77da075a5e3fc322192bbd6aaac4bedb1e882e72

source/common/socket.cc
```diff
@@ -0,0 +1,95 @@
...
+  auto& inner_config_factory = Envoy::Config::Utility::getAndCheckFactory<
+      Envoy::Server::Configuration::UpstreamTransportSocketConfigFactory>(
+      outer_config.transport_socket());
+  Envoy::ProtobufTypes::MessagePtr inner_factory_config =
+      Envoy::Config::Utility::translateToFactoryConfig(outer_config.transport_socket(),
+                                                       context.messageValidationVisitor(),
+                                                       inner_config_factory);
+  auto inner_transport_factory =
+      inner_config_factory.createTransportSocketFactory(*inner_factory_config, context);
+  return std::make_unique<SocketFactory>(outer_config, SocketConfigFactoryImpl(),
```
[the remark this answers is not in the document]
oschaaf: Maybe I should add a comment like:

```c++
When reading this code, when trying to make a distinction between our own and Envoy's
classes, it is helpful to remember that everything sourced from Envoy is prefixed with "Envoy:: ..."
```
What that help? I'm trying to figure out something better, at first I used a Nighthawk
prefix for these, but I stripped that when I realized it was kind of redundant...

oschaaf:
@eric846 Thanks for the review; talking to @mum4k and @dubious90 we agreed to hold of updates to this for now, as such I have labelled this as `shelved`. This does not achieve its intended goal: capturing the ssl overhead. With that there is no pressure landing this. As you noticed, at the code-level this somehow didn't turn out trivial; whereas functionally it's easy to describe: wrap the socket and wrap the socket configuration. It would be great to collect more input on if this can be simplified / made easier to grok.

Good points to continue working to land this would be:
- When Envoy's SslSocket has started truthfully returning bytes read/written, because then this would add value right away by propagating that to the new counters.
- If another feature we'd like to see added would benefit from a socket wrapper

mum4k:
@oschaaf can we close this for now to keep the PR dashboard focused on active PRs?

Issue #272, opened 2020-01-11
CI flaking on asan runs


oschaaf:
Recently CI asan jobs started flaking. Log inspection shows that it looks like jobs are getting killed,
suggesting we're short on memory. I *suspect* that this may have started since the last Envoy update, but I am not sure.
This needs looking into.

oschaaf:
Error message for reference

Use --sandbox_debug to see verbose messages from the sandbox
clang: error: unable to execute command: Killed
clang: error: clang frontend command failed due to signal (use -v to see invocation)
clang version 9.0.0 (https://github.com/llvm-mirror/llvm c62b24f070c9a4bb1a76409e623042a740cac4cd)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/llvm/bin
clang: note: diagnostic msg: PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
clang: note: diagnostic msg: 
********************
‘’’

oschaaf:
Re-opening, as this was merged on master, and unfortunately the asan job ended with:

```
clang: error: unable to execute command: Killed
...
PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/mocks-564137.cpp
clang: note: diagnostic msg: /tmp/mocks-564137.sh
clang: note: diagnostic msg: 
```

It looks like it was working on compiling our mocks.

oschaaf:
My current hypothesis is that splitting up the build of our mocks into multiple files may help here.

oschaaf:
Ugh, this still happened with the latest merge to master [1]
What is weird is that bazel has 8 jobs running in parallel, while I thought we asked it to
only run 6.

[1]

```
[128 / 143] Compiling external/envoy/test/mocks/server/mocks.cc; 137s processwrapper-sandbox ... (8 actions running)
ERROR: /root/project/BUILD:29:1: Linking of rule '//:nighthawk_test_server' failed (Exit 254): clang failed: error executing command 
  (cd /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/sandbox/processwrapper-sandbox/3259/execroot/nighthawk && \
  exec env - \
    BAZEL_LINKLIBS=-l%:libstdc++.a \
[...]

PR #305, opened 2020-02-11
Move MockOptions into its own compilation unit.

oschaaf:
Part of an effort to attempt to reduce memory usage during
the build, to address ASAN flakes in CI.

### DIFF
--- a/include/nighthawk/client/options.h
+++ b/include/nighthawk/client/options.h
@@ -9,6 +9,8 @@
 #include "envoy/config/cluster/v3/cluster.pb.h"
 #include "envoy/config/core/v3/base.pb.h"
 
+#include "nighthawk/common/termination_predicate.h"
+
 #include "api/client/options.pb.h"
 
 #include "absl/types/optional.h"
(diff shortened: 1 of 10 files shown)


oschaaf:
(cross referencing associated issue: #272)

mum4k:
@dubious90 please review and assign back to me when ready.

include/nighthawk/client/options.h
```diff
@@ -9,6 +9,8 @@
 #include "envoy/config/cluster/v3/cluster.pb.h"
 #include "envoy/config/core/v3/base.pb.h"
 
+#include "nighthawk/common/termination_predicate.h"
```
dubious90: A little surprised to see this include with no other corresponding changes.  Why was this suddenly necessary?

oschaaf: this is an include-what-you-use fix; the need for which became apparent when isolating the compilation unit. Apparently, before doing that, the include leaded into scope via another include (the original mock_options.h) in a test which needed it.

PR #308, opened 2020-02-20
Envoy update

oschaaf:
Update Envoy to 5be15b31aa46e3c3776a15fd3d399c67ac432598

These things had to change on the NH side:

- Update code to build against and work with refactors in Envoy:
  - `HeaderMap` became a base for various specializations
  - `StreamDecoder` became a base for `ResponseDecoder` and `RequestDecoder`
- Make fix-format work again
- Fixes for new assert on request headers without a method
- Fixes for time system initialization in test api usage. The default test time system
  seems to have changed to simulated time, which led up to some tests mixing both
  real time and simulated time.
- Pass a fake remote address to avoid a trace finalization segfault
- Work around the fact that we now lack of a copy constructor for new `RequestHeaderMapImpl`,
  by performing an extra explicit copy call.

No functional changes

- [x] Prerequisite: Land a tweak in Envoy to avoid clang-tidy warnings about virtual function calls during construction of test header maps: https://github.com/envoyproxy/envoy/pull/10121. Once that lands we can update the dependency here, and the clang-tidy CI failure should go away.


mum4k:
@dubious90 please review and assign back to me once ready.

source/common/request_stream_grpc_client_impl.cc
```diff
@@ -67,9 +67,10 @@ void RequestStreamGrpcClientImpl::onRemoteClose(Envoy::Grpc::Status::GrpcStatus
 }
 
 RequestPtr ProtoRequestHelper::messageToRequest(
-    const Envoy::Http::HeaderMap& base_header,
+    const Envoy::Http::RequestHeaderMap& base_header,
     const nighthawk::request_source::RequestStreamResponse& message) {
-  auto header = std::make_shared<Envoy::Http::HeaderMapImpl>(base_header);
+  auto header = std::make_shared<Envoy::Http::RequestHeaderMapImpl>();
+  header->copyFrom(*header, base_header);
```
dubious90: do you think this instantiation is something we're going to do in the future too?  Would it make sense to create a convenience method for this?  It happens twice in this PR, but if we expected it to happen in the future, too, it might make things easier for us.

oschaaf: I suspect the inconvience was introduced on purpose in Envoy: drop this constructor to discourage copying. But that is a wild guess.. maybe we can bring up re-introducing that / ask why it was dropped? Otherwise we could consider proposing explicit ::clone() method(s)?

PR #313, opened 2020-02-24
Move all mock factories into their own compilation units.

oschaaf:
Part of an effort to attempt to reduce memory usage during
the build, to address ASAN flakes in CI.

- This has no functional changes.
- Has a small refactor to decouple BenchmarkClient and Sequencer:
  instead of passing in the BenchmkarClient interface to Sequencer,
  we pass in the SequencerTarget.

This is the last of the series of PRs to move mocks into their own
compilation units. Hopefully we will see less OOM kills with of the
asan build process in CI upon landing this.


mum4k:
@eric846 please review and assign back to me when done.

Issue #317, opened 2020-03-10
TSAN failure in CI


oschaaf:
Observed a CI run associated to https://github.com/envoyproxy/nighthawk/pull/316

```
[ RUN      ] ClientTest.AutoConcurrencyRun
[14:11:50.442223][31224][C] [source/client/process_impl.cc:124] assert failure: shutdown_. Details: shutdown not called before destruction.
[14:11:50.448097][31224][C] [bazel-out/k8-dbg/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:83] Caught Aborted, suspect faulting address 0x79f8
[14:11:50.448206][31224][C] [bazel-out/k8-dbg/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:70] Backtrace (use tools/stack_decode.py to get line numbers):
[14:11:50.449640][31224][C] [bazel-out/k8-dbg/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:71] Envoy version: 0/1.14.0-dev/redacted/DEBUG/BoringSSL
[14:11:50.670944][31224][C] [bazel-out/k8-dbg/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:75] #0: Envoy::SignalAction::sigHandler() [0x683abe7]
[14:11:50.866130][31224][C] [bazel-out/k8-dbg/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:75] #1: __tsan::CallUserSignalHandler() [0x2789350]
external/bazel_tools/tools/test/test-setup.sh: line 310: 31224 Aborted                 "${TEST_PATH}" "$@" 2>&1
```

Full log: https://circleci.com/api/v1.1/project/github/envoyproxy/nighthawk/11758/output/102/0?file=true&allocation-id=5e679b96b55f4d48eecdfc16-0-build%2F5ECCDF5B

The assert may or may not be related to the Envoy update in #316, it has been first observed there.

The assert that gets triggered resides here: https://github.com/envoyproxy/nighthawk/blob/b631a17bfbba1c1b46ac7b3b1bf5adb4b05719d7/source/client/process_impl.cc#L123

[...]

PR #321, opened 2020-04-06
Update Envoy to f84440dc4f95890f14e2e0686b07258f030b54b3

oschaaf:
- Trace id generation broke. This was amended by directly
using the new request id uid implementation. It would be
nice to leverage the new request id extension capability
later for this.
- Amended calls to get counters by name
- Synced .bazelrc
- Histogram value formatting subtly changed in some cases, probably caused by an update
  to Envoy's dependency on `libfmt`. 
  Introduce an explicit formatting specifier in the affected output formatting  code to address that.
- Tweaks for resolving new OOM issues with ASAN/TSAN in CI that started with the updated Envoy 
   revision. This affected both the build process as well as the testing process itself. Notably, the
   python-based integration tests would get OOM-killed in asan runs.


oschaaf:
For this PR, ASAN failed for times in a row with:

```
Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/server/jvm.out')
```

oschaaf:
Readying this up for review. I plan to respin the tests a couple of times to make sure, but I think
the new out-of-memory issues in CI are addressed in here sufficiently now.

ci/do_ci.sh
```diff
@@ -73,18 +73,21 @@ function do_asan() {
...
+    run_bazel build ${BAZEL_TEST_OPTIONS} -c dbg --config=clang-asan -- //test/mocks/... && \
+    run_bazel build ${BAZEL_TEST_OPTIONS} -c dbg --config=clang-asan -- //test/... && \
+    run_bazel test ${BAZEL_TEST_OPTIONS} -c dbg --config=clang-asan --jobs=6 //test/...
 }
 
 function do_tsan() {
     echo "bazel TSAN debug build with tests"
     echo "Building and testing envoy tests..."
     cd "${SRCDIR}"
+    [ -z "$CIRCLECI" ] || export BAZEL_BUILD_OPTIONS="${BAZEL_TEST_OPTIONS} --local_ram_resources=12288"
```
mum4k, line 90: Curios: what was the default value of the local_ram_resources flag?

source/client/output_formatter_impl.cc
```diff
@@ -189,7 +190,7 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
               "{}.microseconds: {}", percentile_prefix,
               Envoy::Protobuf::util::TimeUtil::DurationToMicroseconds(percentile.duration()));
         } else {
-          ss << fmt::format("{}.value: {}", percentile_prefix, percentile.raw_value());
+          ss << fmt::format("{}.value: {:.{}f}", percentile_prefix, percentile.raw_value(), 0);
```
mum4k, line 193: Not sure I follow this, what is the ", 0" for ?

ci/do_ci.sh
oschaaf: From https://docs.bazel.build/versions/master/user-manual.html:

```
By default Bazel will estimate amount of RAM and number of CPU cores directly
from system configuration.
```

source/client/output_formatter_impl.cc
oschaaf: Basically the intent is to avoid trailing `.0` by specifying zero precision. However, I changed this to use a static cast to an integer instead, which I'm hoping is more intuitive?

oschaaf:
Ugh:

ASAN ran out of memory again. It seems this now happens when building the Envoy
dependency:

```
-o bazel-out/k8-dbg/bin/external/envoy/test/mocks/server/_objs/server_mocks/mocks.pic.o)
Execution platform: @local_config_platform//:host

Use --sandbox_debug to see verbose messages from the sandbox
clang: error: unable to execute command: Killed
```

It should be possible to scale back --jobs some more to resolve this.

TSAN reported a different kind of failure. Looking into that, we're tracking that one over at https://github.com/envoyproxy/nighthawk/issues/317.

PR #324, opened 2020-04-17
CircleCI: reduce ASAN parallelism

oschaaf:
Another stab at https://github.com/envoyproxy/nighthawk/issues/272
Since the last Envoy update the ASAN build runs out of memory more often then not.

I noticed that the earlier attempt at reducing `NUM_CPUS`/`--jobs` didn't yield expected
results in CI, as bazel would still be running 8 jobs in parallel.

Upon investigation, I noticed that my earlier attempt had issues:
- it set the wrong variable
- upon fixing that, the script would end up handing two `--jobs=x` arguments to bazel

The approach here sets `NUM_CPUS` early on in the script, and doesn't attempt to mess
with the build flags later on.

Hopefully fixes https://github.com/envoyproxy/nighthawk/issues/272

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -73,7 +73,6 @@ function do_asan() {
     echo "bazel ASAN/UBSAN debug build with tests"
     echo "Building and testing envoy tests..."
     cd "${SRCDIR}"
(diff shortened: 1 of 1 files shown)


oschaaf:
Yeah, checking the CI build logs, bazel seems to run the expected amount of parallel actions (5) for ASAN. Marking this is ready for review.

PR #325, opened 2020-04-17
release prep: push docker image upon tagging

oschaaf:


oschaaf:
Unfortunately I can't really truly test this end to end (or I don't know how). 
But I think this should work. I guess we'll find out when tagging the release.

PR #327, opened 2020-04-20
Log exception in ProcessImpl::run()

oschaaf:
Related to https://github.com/envoyproxy/nighthawk/issues/317

Log any exceptions and rethrow. This is to assist with
diagnosis once it happens again in the CI env.

### DIFF
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -377,6 +377,82 @@ void ProcessImpl::addRequestSourceCluster(
   socket_address->set_port_value(uri.port());
 }
 
+bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriPtr>& uris,
+                              const UriPtr& request_source_uri, const UriPtr& tracing_uri) {
+  int number_of_workers = determineConcurrency();
+  shutdown_ = false;
+  const std::vector<ClientWorkerPtr>& workers = createWorkers(number_of_workers);
+  tls_.registerThread(*dispatcher_, true);
+  store_root_.initializeThreading(*dispatcher_, tls_);
+  runtime_singleton_ = std::make_unique<Envoy::Runtime::ScopedLoaderSingleton>(
+      Envoy::Runtime::LoaderPtr{new Envoy::Runtime::LoaderImpl(
+          *dispatcher_, tls_, {}, *local_info_, init_manager_, store_root_, generator_,
+          Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_)});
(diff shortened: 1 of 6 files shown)


source/client/process_impl.cc
```diff
@@ -414,75 +414,82 @@ bool ProcessImpl::run(OutputCollector& collector) {
...
-  createBootstrapConfiguration(bootstrap, uris, request_source_uri, number_of_workers);
-  if (tracing_uri != nullptr) {
-    setupTracingImplementation(bootstrap, *tracing_uri);
-    addTracingCluster(bootstrap, *tracing_uri);
-  }
-  ENVOY_LOG(debug, "Computed configuration: {}", bootstrap.DebugString());
-  cluster_manager_ = cluster_manager_factory_->clusterManagerFromProto(bootstrap);
-  maybeCreateTracingDriver(bootstrap.tracing());
-  cluster_manager_->setInitializedCb([this]() -> void { init_manager_.initialize(init_watcher_); });
+  try {
```
mum4k, line 417: This was already a fairly long function to begin with. How hard would it be to refactor the content of the new try { } into a separate private method? That way we could separate the exception handling (the outer run method) from the business logic (the new private method) and make this a bit easier to follow.

oschaaf: Done in 9c89e74453724491aec1fbfe479a871545db6553

oschaaf:
Coverage is flaking because of a single line sometimes not getting hit:

https://12132-180498819-gh.circle-artifacts.com/0/coverage/coverage/proc/self/cwd/source/client/client_worker_impl.cc.html#L55

https://12111-180498819-gh.circle-artifacts.com/0/coverage/coverage/proc/self/cwd/source/client/client_worker_impl.cc.html#L55

mum4k:
Is the flaky coverage something we should address / track?

Also can you help me by triggering it again so this becomes able to merge?

oschaaf:
> Is the flaky coverage something we should address / track?
> 
> Also can you help me by triggering it again so this becomes able to merge?

I'm trying to address this in-place, it should be possible to reliably hit the line I mentioned.

oschaaf:
I punted the coverage flake to https://github.com/envoyproxy/nighthawk/issues/328
Also, added https://github.com/envoyproxy/nighthawk/pull/327/commits/e28e1851c168a3e2db7f7d5a4b4bfc4a299afc28 which has cleanup but should be a no-op. At this time I can't understand how the line I pointed out above could not be covered.

Issue #268, opened 2020-01-10
Doc comments check


oschaaf:
Some of the public methods in the interface headers are missing doc comments; it would be good
to go over them and supplement where needed. 
(Self assigning)

PR #330, opened 2020-04-27
Supplement doc comments in interface headers.

oschaaf:
Fixes https://github.com/envoyproxy/nighthawk/issues/268

### DIFF
--- a/include/nighthawk/client/factories.h
+++ b/include/nighthawk/client/factories.h
@@ -32,7 +32,7 @@ public:
    * @param dispatcher supplies the owning thread's dispatcher.
    * @param scope stats scope for any stats tracked by the benchmark client.
    * @param cluster_manager Cluster manager preconfigured with our target cluster.
-   * @param http_tracer http tracer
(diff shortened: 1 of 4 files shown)


include/nighthawk/client/factories.h
```diff
@@ -24,19 +24,44 @@ namespace Client {
...
   virtual ~BenchmarkClientFactory() = default;
+
+  /**
+   * Constructs a BenchmarkClient
+   *
+   * @param api reference to the Api object.
+   * @param dispatcher supplies the owning thread's dispatcher.
+   * @param scope stats scope for any stats tracked by the benchmark client.
+   * @param cluster_manager Cluster manager preconfigured with our target cluster.
+   * @param http_tracer http tracer
```
mum4k, line 35: (nit) Do we have anything more interesting to say about this, maybe what it is used for pr what it means in the context of this function? The current comment is kind of redundant.

oschaaf: yeah, that wasn't a very helpful doc comment :-)
done in 28487f0

PR #331, opened 2020-04-27
Update Envoy to 888e0e28900a470df448c65d7b99d8065fd60251

oschaaf:
This Envoy update is a bit more involved, because it forced moving away from the
legacy connection pools it had, which got eliminated from the code base. So we migrate
to Envoy's new pool models.

- The h1 pool migrates seamlessly to the new model with minor changes.

- The experimental h2 connection pool that supports multiple connections gets
dropped. We now offer multiple connection support based on the new H2 pool.
Behavior has changed, but there was a warning about that in the description of 
the `--experimental` flag (see changes to the test associated to our earlier experimental h2
pool)

It should now be possible to add new options to add better support for h2 & multiple connections,
and deprecate or remove the `--experimental` flag for that.


source/client/benchmark_client_impl.cc
```diff
@@ -200,9 +157,12 @@ void BenchmarkClientHttpImpl::onPoolFailure(Envoy::Http::ConnectionPool::PoolFai
   case Envoy::Http::ConnectionPool::PoolFailureReason::Overflow:
     benchmark_client_stats_.pool_overflow_.inc();
     break;
-  case Envoy::Http::ConnectionPool::PoolFailureReason::ConnectionFailure:
+  case Envoy::Http::ConnectionPool::PoolFailureReason::LocalConnectionFailure:
+  case Envoy::Http::ConnectionPool::PoolFailureReason::RemoteConnectionFailure:
     benchmark_client_stats_.pool_connection_failure_.inc();
     break;
+  case Envoy::Http::ConnectionPool::PoolFailureReason::Timeout:
```
oschaaf, line 164: I will add a TODO to track these, as well as add separate counters for the combined connection failure counter above.

oschaaf:
I had to amend test expectations in 571c3990d8329a0f617ec2a1a7a0c3ab467e5e78 to allow for a behavior that showed up in the earlier CI tsan run. I've been looking into that, and this looks like it is stock Envoy behavior to exceed the configured connection limits in some scenarios.

oschaaf:
As a next step to this, to enhance support for supporting multiple http/2 connections, we could consider exposing the max concurrent streams setting + warn/deprecate the experimental flag which just caps that to `1`.  (And then some day it would be nice to support/unify the connection re-use strategies we have in place for h1).

PR #335, opened 2020-05-15
Update Envoy to 31128e7dc22355876020188bc8feb99304663041

oschaaf:
- As of this change we needed to reduce parallelism in CI
   to avoid OOM during coverage runs. This probably can
   be traced back to a new way of running coverage being
   used in Envoy, which trickled down to Nighthawk.
- Removed our own `io_bazel_rules_python` as we can now 
  piggy-back on Envoy for that.
- Removes a file that accidentally crept in (lorem_ipsum.txt)


oschaaf:
The updated coverage seems to be a lot more memory hungry. I need to figure out why that is
and how to constrain it.  The `--local_ram_resources` we had in place for running in CircleCI clearly
doesn't cut it.

oschaaf:
Optimistically marked this as ready for review.
I just updated this to the latest Envoy revision, which didn't need any
further changes. This also triggers another CI coverage run, which should not run into OOM
now that we have the parallelism tuned for it.

Issue #245, opened 2019-12-20
Formalize / automate release performance stability testing 


oschaaf:
In #232 I added a code-level todo regarding on of the release steps, which is meant to
ensure that we don't introduce major performance and/or accuracy changes between releases

Filing this issue to discuss and track.

PR #337, opened 2020-05-31
Dockerized benchmarking framework

oschaaf:
This PR has changes that make it easy to consume our integration testing as a benchmarking
tool framework for external repositories, allowing those to write custom tests for running in CI.

This probably needs to be split up into multiple PR's, but this allows reviewers to
have an early peek at what such PR's would lead up to.

The `README.md` in here has more context, but the primary goals are te be able to:
- run the tests against arbitrary Envoy revisions
- persist profile dumps, flamegraphs, and latency numbers per test
- run the nighthawk tools via docker
- offer stock tests, but also allow scaffolding consumer-specific tests

Example of a fully dockerized flow (more example in `README.md`):

```
# This script runs the dockerized benchmarking framework, which in
# turn will pull Nighthawk and Envoy in via docker. 

set -eo pipefail
set +x
set -u

# The benchmark logs and artifacts will be dropped here
OUTDIR="/home/oschaaf/code/envoy-perf-vscode/nighthawk/benchmarks/tmp/"
# Used to map the test that we want to see executed into the docker container
TEST_DIR="/home/oschaaf/code/envoy-perf-vscode/nighthawk/benchmarks/test/"

# Rebuild the docker in case something changed.
./docker_build.sh && 
docker run -it --rm \
  -v "/var/run/docker.sock:/var/run/docker.sock:rw" \
  -v "${OUTDIR}:${OUTDIR}:rw" \
  -v "${TEST_DIR}:/usr/local/bin/benchmarks/benchmarks.runfiles/nighthawk/benchmarks/external_tests/" \
  --network=host \
  --env NH_DOCKER_IMAGE="envoyproxy/nighthawk-dev:latest" \
  --env ENVOY_DOCKER_IMAGE_TO_TEST="envoyproxy/envoy-dev:f61b096f6a2dd3a9c74b9a9369a6ea398dbe1f0f" \
  --env TMPDIR="${OUTDIR}" \
  oschaaf/benchmark-dev:latest ./benchmarks --log-cli-level=info -vvvv 
```

Part of #245


benchmarks/BUILD
```diff
@@ -1,20 +1,33 @@
...
-    name = "benchmarks",
+    name = "test_benchmarks",
     srcs = [
         "benchmarks.py",
-        "test_benchmark.py",
+        "infra.py",
+        "test/test_discovery.py",
+    ],
+    data = [
+        "configurations/envoy_proxy.yaml",
```
htuch, line 27: Is it possible to depend on `:benchmarks` directly here to avoid having to redeclare things?

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
+# WIP - Benchmarking testsuite
+
+The NH benchmark test suite builds on top Nighthawk's integration test framework, and
+can be used to scaffold tests and obtain latency reports as well as flamegraphs.
+
+## Immediate goals
+
+The goal is te be able to:
```
htuch, line 8: Nit: to be able to

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+- run the suite against arbitrary Envoy revisions
+- persist profile dumps, flamegraphs, and latency numbers per test
+- run the nighthawk tools via docker
+- offer stock tests, but also allow scaffolding consumer-specific tests
+
+## Example: Docker based execution, scavaging benchmark/
+
+This scripts shows how to use the benchmarking suite.
+It will run a selection of an example [benchmarks](test/test_discovery.py)
+scavenged from `/benchmarks`, which injects Envoy between the
```
htuch, line 18: Maybe say "extracted" or "sourced" rather than "scavenged" :)

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+The NH benchmark test suite builds on top Nighthawk's integration test framework, and
+can be used to scaffold tests and obtain latency reports as well as flamegraphs.
+
+## Immediate goals
+
+The goal is te be able to:
+- run the suite against arbitrary Envoy revisions
+- persist profile dumps, flamegraphs, and latency numbers per test
+- run the nighthawk tools via docker
+- offer stock tests, but also allow scaffolding consumer-specific tests
```
htuch, line 12: Maybe point out that this is also intended for the use in A/B testing by developers and integration with GitHub CI flows (via the Docker support).

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+# Envoy is called 'Envoy' in the Envoy docker image.
+export ENVOY_PATH="envoy"
+
+# run all tests starting with test_http_h1_small in benchmarks/
+bazel-bin/benchmarks/benchmarks --log-cli-level=info -vvvv -k test_http_h1_small benchmarks/
+```
+
+## Example: running with binaries
+
+This will build the Nighthawk binaries from the c++ code, and use those to
```
htuch, line 43: Nit: s/c++/C++/g

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+The NH benchmark test suite builds on top Nighthawk's integration test framework, and
+can be used to scaffold tests and obtain latency reports as well as flamegraphs.
+
+## Immediate goals
+
+The goal is te be able to:
+- run the suite against arbitrary Envoy revisions
+- persist profile dumps, flamegraphs, and latency numbers per test
+- run the nighthawk tools via docker
```
htuch, line 11: Nit: s/docker/Docker/g

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+```bash
+git clone https://github.com/oschaaf/nighthawk.git benchmark-test
+cd benchmark-test
+bazel test --test_summary=detailed --test_output=all --test_arg=--log-cli-level=info --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_env=HEAPPROFILE= --test_env=HEAPCHECK= --cache_test_results=no --compilation_mode=opt --cxxopt=-g --cxxopt=-ggdb3 //benchmarks:*
+```
+
+## Example: fully dockerized flow
+
+The framework can be run via docker and used that way to execute
+python benchmarks scripts not sourced from this repository, but
```
htuch, line 58: Nit: s/python/Python/g

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Rebuild the docker in case something changed.
+./docker_build.sh && 
+docker run -it --rm \
+  -v "/var/run/docker.sock:/var/run/docker.sock:rw" \
+  -v "${OUTDIR}:${OUTDIR}:rw" \
+  -v "${TEST_DIR}:/usr/local/bin/benchmarks/benchmarks.runfiles/nighthawk/benchmarks/external_tests/" \
+  --network=host \
+  --env NH_DOCKER_IMAGE="envoyproxy/nighthawk-dev:latest" \
+  --env ENVOY_DOCKER_IMAGE_TO_TEST="envoyproxy/envoy-dev:f61b096f6a2dd3a9c74b9a9369a6ea398dbe1f0f" \
```
htuch, line 83: How do you specify a custom Envoy binary in the Docker flow?

benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+- Copy out the artifacts and push those to a gcp bucket. Current status:
+   - cpu profiles are dumped to tmp per test (named according to the test). ideally we'd
+     also dump flamegraph svg's
+- A UI -- though we may be able to get by with just a uri structure conventioned around the envoy
+  sha. e.g. http://perf-ci-host/gcpsync/[envoy-sha]-[timestamp]/ to link CI, and directory index the 
+  artifacts.
+
+# FUTURE
+
+- profiling / flamegraphing via perf/bcc tools; include the proxy-wasm flamegraphing research r&d
```
htuch, line 99: I think the most useful things to add are (in my $0.02 stack rank):
1. The ability to repeat the runs multiple times and obtain stats, e.g. how much variance there is, mean, etc.
2. The ability to obtain flamegraphs.
3. The ability to do A/B testing, similar to https://github.com/envoyproxy/envoy-perf/blob/master/siege/siege.py#L3.

But, I'm sure there are other folks who have other priorities here ;-)

benchmarks/configurations/envoy_proxy.yaml
```diff
@@ -0,0 +1,44 @@
...
+              dynamic_stats: false
+  clusters:
+  - name: local_service
+    connect_timeout: 0.25s
+    type: strict_dns
+    lb_policy: round_robin
+    hosts:
+    - socket_address:
+        address: $server_ip
+        port_value: $server_port
```
htuch, line 44: In addition to A/B tests, perhaps we could have A/B/C at some point, where A/B are the before/after Envoy binaries corresponding to some change and C is direct NH client to test server, to demonstrate the available source/sink capacity.

benchmarks/docker/docker_build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# NOTE: explicit no -x for verbose commands. Because this is run in CI, doing so may result in
+# publishing sensitive information into public CI logs if someone makes a change in a
+# consuming script that is off guard.
+
+DOCKER_NAME="benchmark"
+DOCKER_IMAGE_PREFIX="oschaaf/${DOCKER_NAME}"
+BAZEL_BIN="$(bazel info bazel-bin)"
+WORKSPACE="$(bazel info workspace)"
+bazel build //benchmarks:benchmarks
```
htuch, line 17: Why not use `BAZEL_BIN` here?

benchmarks/docker/docker_build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Stop on errors.
+set -e
+
+# NOTE: explicit no -x for verbose commands. Because this is run in CI, doing so may result in
+# publishing sensitive information into public CI logs if someone makes a change in a
+# consuming script that is off guard.
+
+DOCKER_NAME="benchmark"
+DOCKER_IMAGE_PREFIX="oschaaf/${DOCKER_NAME}"
```
htuch, line 14: Do you want to use `${USER}` here instead of `oschaaf`?

benchmarks/docker/docker_build.sh
```diff
@@ -0,0 +1,39 @@
...
+BAZEL_BIN="$(bazel info bazel-bin)"
+WORKSPACE="$(bazel info workspace)"
+bazel build //benchmarks:benchmarks
+TMP_DIR="${WORKSPACE}/tmp-docker-build-context"
```
htuch, line 18: Why not use `mkdtemp`? Then you don't have to `rm -rf` in the next line.

benchmarks/docker/test.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Rebuild the docker in case something changed.
+./docker_build.sh && 
+docker run -it --rm \
+  -v "/var/run/docker.sock:/var/run/docker.sock:rw" \
+  -v "${OUTDIR}:${OUTDIR}:rw" \
+  -v "${TEST_DIR}:/usr/local/bin/benchmarks/benchmarks.runfiles/nighthawk/benchmarks/external_tests/" \
+  --network=host \
+  --env NH_DOCKER_IMAGE="envoyproxy/nighthawk-dev:latest" \
+  --env ENVOY_DOCKER_IMAGE_TO_TEST="envoyproxy/envoy-dev:f61b096f6a2dd3a9c74b9a9369a6ea398dbe1f0f" \
```
htuch, line 23: When and who will update this SHA?

benchmarks/BUILD
oschaaf: I improved this, but could not eliminate all of the duplication, ptal

benchmarks/README.md
oschaaf: This is documented in the flow with binaries. You don't specify a docker image, but instead just use `ENVOY_PATH` to point out your Envoy binary.

benchmarks/configurations/envoy_proxy.yaml
oschaaf: I think if we would have a way to annotate tests with group metadata, then we can go all out here in another layer of python scripting, which performs post-processing based on that metadata. E.g. we can treat a group of tests annotated with the same tag as a/b/c/... and crunch the numbers + visualize like that.

benchmarks/docker/docker_build.sh
oschaaf: I think `mkdtemp` is a python thing, whereas this is bash? At least my system can't find that binary :-)

benchmarks/docker/test.sh
oschaaf: This is actually an accidental test script that got committed. Maybe it can serve a purpose though, as a NH CI task which ought to succeed. The Envoy sha could then be sourced from our dependency or some such. Regardless it should at least be a variable, maybe defaulting to `latest`

benchmarks/test/test_discovery.py
```diff
@@ -0,0 +1,90 @@
...
+import json
+import pytest
+import os
+from test.integration.integration_test_fixtures import (http_test_server_fixture,
+                                                        https_test_server_fixture)
+from test.integration.utility import *
+from envoy_proxy import (inject_envoy_http_proxy_fixture, proxy_config)
+
+
+def run_with_cpu_profiler(fixture,
```
htuch, line 18: Why is this in discovery? Isn't this just general execution ochestration?

test/integration/integration_test.py
```diff
@@ -23,6 +23,7 @@
           test_selection_arg,  # Passed in via BUILD/py_test()
           "-x",
           path,
+          "--log-cli-level=debug",
```
htuch, line 26: ?

test/integration/test_integration_zipkin.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import pytest
 
-from integration_test_fixtures import (http_test_server_fixture)
+from integration_test_fixtures import (http_test_server_fixture, server_config)
```
htuch, line 4: Doesn't look like `server_config` is used here?

test/integration/test_remote_execution.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from test.integration.integration_test_fixtures import http_test_server_fixture
+from test.integration.integration_test_fixtures import (http_test_server_fixture, server_config)
```
htuch, line 5: Ditto

test/integration/test_integration_zipkin.py
oschaaf: It doesn't look like, but `http_test_server_fixture` relies on it. If we don't explicitly import `server_config` here, we will get an error [1]. I think that is kind of silly, but it's also not
something I can do anything about, this seems to be related to something in the way that
pytest works. I've spend some time looking into this, and so far the easiest resolution seems to
be to just go with the flow and import the `server_config` fixture in scripts that use fixtures that rely on it. I'll add a comment about it for clarification.

[1]
```
  @pytest.fixture(params=determineIpVersionsFromEnvironment())
  def http_test_server_fixture(request, server_config):
E       fixture 'server_config' not found
>       available fixtures: capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, http_test_server_fixture, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, testrun_uid, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.
```

benchmarks/README.md
oschaaf: That certainly is possible, but it result in a large image (I assume the binaries would need to be unstripped and build with all the right flags for profiling?) .

test/integration/test_integration_zipkin.py
htuch: Ack, thanks for the explanation.

benchmarks/README.md
htuch: Yeah. I don't think it needs to be in this PR, but it's a pretty useful feature IMHO. One thing that I wanted to do is run the exact same Docker image on my local workstation and different sized Cloud VMs to see what performance range is observed.

oschaaf: Added this to the list of things that need to be done in `README.md`

benchmarks/README.md
oschaaf: Also moved these into the future list in README.md

htuch:
CC @adisuissa

benchmarks/test/test_discovery.py
oschaaf: I may want to pull this out of this PR and host it as an example on a private repo or some such. WDYT?

benchmarks/docker/docker_build.sh
oschaaf: I'm not sure what you mean here, could you elaborate?

benchmarks/test/test_discovery.py
htuch: Refactoring makes sense, or are you saying the whole file is moved somewhere?

benchmarks/docker/docker_build.sh
htuch: I think I was confused by `BAZEL_BIN` above. This is the `bazel-bin` dir, not the bazel binary..

benchmarks/docker/test.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# This script runs the dockerized benchmarking framework, which in
+# turn will pull Nighthawk and Envoy in via docker. 
+
+set -eo pipefail
+set +x
+set -u
+
+# The benchmark logs and artifacts will be dropped here
+OUTDIR="/home/oschaaf/code/envoy-perf-vscode/nighthawk/benchmarks/tmp/"
```
htuch, line 11: These are hardcoded paths, I think I may have said something about `$USER` in another thread.

benchmarks/docker/test.sh
htuch: I think drop it, but you can use fragments in a doc perhaps.

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,82 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
-      config = config.substitute(self.parameters)
-      logging.info("Parameterized server configuration: %s", config)
-
-    with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".yaml") as tmp:
+    self.docker_image = os.getenv("NH_DOCKER_IMAGE", "")
+    self.tmpdir = os.path.join(os.getenv("TMPDIR", "/tmp/nighthawk_benchmark/"), tag + "/")
+    self.parameters["tmpdir"] = self.tmpdir
+    self.parameters["tag"] = tag
+
+    def substitute_yaml_values(obj, params):
```
htuch, line 47: Would it be possible to feature freeze and land the PR as is before adding new things like this? This will make it easier to land I reckon.

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,82 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
+          if obj[0] == '$':
+            return Template(obj).substitute(params)
+          elif obj.startswith(INJECT_RUNFILE_MARKER):
+            r = runfiles.Create()
+            with open(r.Rlocation(obj[len(INJECT_RUNFILE_MARKER):].strip()), 'r') as file:
+              return file.read()
+      return obj
+
+    r = runfiles.Create()
+    with open(r.Rlocation(self.config_template_path)) as f:
```
htuch, line 67: Why not just make the entire config a giant `string.Template` or Python f-string? That way you don't need to do YAML traversal.

oschaaf: Done

test/integration/nighthawk_test_server.py
oschaaf: Backed this out in favor of hard coding the cert in the configs.

oschaaf:
@htuch added a benchmark test in CI for NH itself, artifacts, [example](https://app.circleci.com/pipelines/github/envoyproxy/nighthawk/1597/workflows/dc20d1ae-2cab-4fcc-8104-1add64c87091/jobs/12559/artifacts)

Right now the most interesting artifacts are probably the html plots, e.g.:
[direct](https://12559-180498819-gh.circle-artifacts.com/0/2020-06-05-11-47-18/http_h1_small_request_small_reply_direct_IpVersion.IPV4-nighthawk_test_integration_configurations_nighthawk_http_origin.yaml/simple_plot.html) or [via our test server as a proxy](
https://12559-180498819-gh.circle-artifacts.com/0/2020-06-05-11-47-18/http_h1_small_request_small_reply_via_IpVersion.IPV4-nighthawk_test_integration_configurations_nighthawk_http_origin.yaml-nighthawk_benchmarks_configurations_envoy_proxy.yaml/simple_plot.html)

The collected profiles from Envoy's built-in profiler are also in there, but these are probably not very useful at this time; one would need access to the binaries and libraries involved to be able to use them properly.

benchmarks/docker/test.sh
oschaaf: I'll move this into README.md. In the future, we could move this and turn it into a ci task: we fetch a simple 101 test via http, and check that it works like we expect, to at least have some check in place this doesn't break.

benchmarks/test/test_discovery.py
oschaaf: I was suggesting the possibility of moving it out of here and hosting it somewhere else. But after thinking about it: let's keep it in here, I'll see if I can move it into a more sensible place then. We will eventually need a location that hosts "foreign" tests for testing the full docker flow though.

htuch:
@oschaaf thanks. I'm hoping we can stabilize and land what is in the PR right now so we don't have it grow too much more :)

benchmarks/run_benchmark.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -eo pipefail
+set +x
+set -u
+
+function cleanup() {
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        docker rm -f $(docker ps -a -q) || true
```
htuch, line 9: This seems dangerous; what if other Docker images are running?

benchmarks/run_benchmark.sh
```diff
@@ -0,0 +1,36 @@
...
+"${BAZEL}" build //benchmarks:benchmarks
+
+export ENVOY_IP_TEST_VERSIONS=v4only
+export ENVOY_PATH="envoy"
+export TMPDIR="$(pwd)/benchmarks/tmp"
+export NH_DOCKER_IMAGE="envoyproxy/nighthawk-dev:latest"
+export ENVOY_DOCKER_IMAGE_TO_TEST="envoyproxy/envoy-dev:latest"
+
+# run all tests starting with test_http_h1_small
+bazel-bin/benchmarks/benchmarks --log-cli-level=info -vvvv benchmarks/
```
htuch, line 36: Any concerns over the impact of the level of verbosity and benchmark performance?

benchmarks/test/test_discovery.py
```diff
@@ -0,0 +1,103 @@
...
+  if hasattr(fixture, "proxy_server"):
+    with open(os.path.join(fixture.test_server.tmpdir, "proxy_version.txt"), "w") as f:
+      f.write(fixture.proxy_server.getCliVersionString())
+  r = runfiles.Create()
+  with open(r.Rlocation("nighthawk/benchmarks/test/templates/simple_plot.html"), "r") as r:
+    txt = r.readlines()
+    with open(os.path.join(fixture.test_server.tmpdir, "simple_plot.html"), "w") as w:
+      # This will source nighthawk.json over http from the same dir.
+      # TODO(oschaaf): consider injecting the json directly into a html script source.
+      w.write(str.join("", txt))
```
htuch, line 82: Are you just trying to copy the file here? Why not use direct utils for this such as `shutil.copyfile`?

ci/do_ci.sh
```diff
@@ -90,6 +97,33 @@ function do_tsan() {
...
 }
 
+function cleanup_benchmark_artifacts {
+    # TODO(oschaaf): we clean the tmp dir above from uninteresting stuff
+    # that crept into the tmp/output directory. The cruft gets in there because
+    # other tooling also responds to the TMPDIR environment variable, which in retrospect
+    # was a bad choice.
+    # Consider using a different environment variable for the benchmark tooling
+    # to use for this.
+    rm -rf ${TMPDIR}/tmp.*
```
htuch, line 107: This seems very scary of `${TMPDIR}` isn't set. It might be a good idea to verify that first.

test/integration/configurations/nighthawk_https_origin.yaml
```diff
@@ -1,40 +1,146 @@
...
+
+                      M8MD3RDSq/90FSiME4vbyNEyTmj0
+
+                      -----END CERTIFICATE-----
+
+                      "
+                  private_key:
+                    inline_string: "-----BEGIN RSA PRIVATE KEY-----
+
+                      MIIEpAIBAAKCAQEAuvPdQdmwZongPAgQho/Vipd3PZWrQ6BKxIb4l/RvqtVP321I
```
htuch, line 94: Generally not a fan of inlining these, since when the certs expire, it's hard to know how to regenerate them. All the test certs in the Envoy tree have associated scripts to rebuild them.

oschaaf: So to make sure I get it right, you'd like to see https://github.com/envoyproxy/nighthawk/pull/337/commits/605e5ce39369a2ed93505fb8c0498849cf82bf70 back in, but then instead of parsing yaml we do a string based substitution?

benchmarks/test/templates/simple_plot.html
```diff
@@ -0,0 +1,150 @@
...
+            return r;
+        }
+        function getTraces(results, name, stat, linex) {
+            var a = [];
+            for (var i = 0; i < results.length; i++) {
+                var result = results[i];
+                var statistic = findStatisticsById(result, stat);
+                var percentiles = statistic["percentiles"];
+                var points = getPoints(percentiles);
+                //var samples = statistic["count"]
```
snowp, line 57: remove?

oschaaf: done

.circleci/config.yml
```diff
@@ -1,5 +1,5 @@
 references:
-  envoy-build-image: &envoy-build-image
+  envoy-build-image: &envoy-build-image
```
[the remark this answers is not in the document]
oschaaf: removed

ci/do_ci.sh
oschaaf: I safeguarded this for a bit by ensuring tmpdir is > 4 characters and the directory exists, wdyt?

benchmarks/test/test_discovery.py
oschaaf: I anticipated we'd might want to run replacements here, but so far we don't, so yeah, followed up and using `shutil.copyfile` now.

benchmarks/run_benchmark.sh
oschaaf: No the log verbosity won't impact benchmark performance. There's only some things pre and post execution that are getting logged.

benchmarks/run_benchmark.sh
oschaaf: Removed this. The OSX variant will need a little bit of love anyway it broke somewhere along the way. I can do that in another iteration - after we land this PR - if someone needs that.

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+
+  def __init__(self, ip_version, server_config, proxy_config):
+    """See base class."""
+    super(InjectHttpProxyIntegrationTestBase, self).__init__(ip_version, server_config)
+    self.proxy_config = proxy_config
+
+  def setUp(self):
+    super(InjectHttpProxyIntegrationTestBase, self).setUp()
+    logging.info("injecting envoy proxy ...")
+    # TODO(oschaaf): how should this interact with multiple backends?
```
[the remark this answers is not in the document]
oschaaf: Great point; added this to the TODO list in `README.md`. This may or may not be a doc issue, possibly this can already be made to work with the status quo and just needs a clear example of how to do it. (The NH integration tests does have tests involving multiple origins).

oschaaf:
@snowp thanks for having a look, I addressed your comments
@htuch I addressed most of your comments, modulo the one about hard-coding ssl certs, I had a question there instead.

ci/do_ci.sh
htuch: SGTM

test/integration/configurations/nighthawk_https_origin.yaml
htuch: Yeah. My main issue with that PR was that it was doing sophisticated YAML parsing; I think that since YAML is a textual format, you can just substitute directly into it. If you use something like Python's string.Template format in the YAML, there's no need to even write a custom substituter.

oschaaf: @htuch following our Slack discussion, I pushed 345972b

oschaaf:
Thanks for the review @htuch 
@mum4k are you comfortable with me merging this?

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+
+class EnvoyProxyServer(NighthawkTestServer):
+  """
+  Envoy proxy server abstraction. Note that it derives from NighthawkTestServer, as that
+  is implemented as a customized Envoy, which is convenient here: the CLI and admin interface
+  mechanics that we rely on are the same. So all we do here, is specialize so we can override
+  the docker image and binary name.
+  """
+
+  def __init__(self, config_template_path, server_ip, ip_version, parameters=dict(), tag=""):
```
mum4k, line 24: Can we add a comment for the initializer describing its arguments?

The same comment applies to all public functions, classes and methods.

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+from test.integration.nighthawk_test_server import NighthawkTestServer
+
+
+class EnvoyProxyServer(NighthawkTestServer):
+  """
+  Envoy proxy server abstraction. Note that it derives from NighthawkTestServer, as that
+  is implemented as a customized Envoy, which is convenient here: the CLI and admin interface
+  mechanics that we rely on are the same. So all we do here, is specialize so we can override
+  the docker image and binary name.
+  """
```
mum4k, line 22: Can we also document any public attributes inside an "Attributes" section of the class docstring?

mum4k, line 22: Would it make sense to add an example of use inside the class docstring?

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.fixture()
+def proxy_config():
+  yield "nighthawk/benchmarks/configurations/envoy_proxy.yaml"
+
+
+class InjectHttpProxyIntegrationTestBase(HttpIntegrationTestBase):
+  """
+  Fixture which spins up a Nighthawk test server which listens on plain http,
+  and set up an Envoy instances to proxy to that, als listening on plain http.
```
mum4k, line 44: (nit) Sentence doesn't read well around "als listening".

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+  """
+
+  def __init__(self, ip_version, server_config, proxy_config):
+    """See base class."""
+    super(InjectHttpProxyIntegrationTestBase, self).__init__(ip_version, server_config)
+    self.proxy_config = proxy_config
```
mum4k, line 50: Similarly in this class, let's document any public attributes. Alternatively we could make them private if they don't have to be public.

self._proxy_config

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+        ip=proxy_server.server_ip, port=proxy_server.server_port))
+    self.proxy_server = proxy_server
+
+  def tearDown(self):
+    super(InjectHttpProxyIntegrationTestBase, self).tearDown()
+    assert (self.proxy_server.stop() == 0)
+
+  def getTestServerRootUri(self):
+    """See base class."""
+    r = super(InjectHttpProxyIntegrationTestBase, self).getTestServerRootUri()
```
mum4k, line 75: Can we use a more descriptive name? E.g. "root_uri".

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+    # TODO(oschaaf): fix, kind of a hack.
+    r = r.replace(":%s" % self.test_server.server_port, ":%s" % self.proxy_server.server_port)
+    return r
+
+
+@pytest.fixture(params=determineIpVersionsFromEnvironment())
+def inject_envoy_http_proxy_fixture(request, server_config, proxy_config):
+  '''
+  Injects an Envoy proxy.
```
mum4k, line 84: Can we expand the doc comment by adding a section describing the arguments and the returned / yielded value?

benchmarks/envoy_proxy.py
```diff
@@ -0,0 +1,89 @@
...
+  '''
+  f = InjectHttpProxyIntegrationTestBase(request.param, server_config, proxy_config)
```
mum4k, line 86: Similarly here, can we use a more descriptive name?

benchmarks/test/templates/simple_plot.html
```diff
@@ -0,0 +1,168 @@
+<head>
```
mum4k, line 1: Can we document this file and the functions in it?

benchmarks/test/templates/simple_plot.html
```diff
@@ -0,0 +1,168 @@
...
+      for (var i = 0; i < results.length; i++) {
+        if (results[i].name == "global") {
+          return results[i];
+        }
+      }
+      return null;
+    }
+    function findWorkersResults(data) {
+      var results = data["results"];
+      var a = [];
```
mum4k, line 21: Can we find a more descriptive name?

Applies to multiple locations.

benchmarks/test/test_discovery.py
```diff
@@ -0,0 +1,117 @@
...
+integration test framework to run benchmark executions.
+"""
+
+import logging
+import json
+import pytest
+import os
+from test.integration.integration_test_fixtures import (http_test_server_fixture,
+                                                        https_test_server_fixture)
+from test.integration.utility import *
```
mum4k, line 14: Can we import specific packages / modules?

oschaaf:
@mum4k thanks, I think I addressed everything. In addition, I added a small tweak to way docker gets run to avoid breakage in the docker flows when they use the latest Envoy revisions:
https://github.com/envoyproxy/nighthawk/pull/337/commits/c24062fb4f5f80896d2264696916b51cdb13f7fe#diff-281f9acd4aa73a1b04f644ae61f98febR94

oschaaf:
Another thing I should explicitly call out is that for `benchmark/` I added some more python flake8 checks in CI.
Generally check more things, and specifically add docstring (format) checks.

https://github.com/envoyproxy/nighthawk/pull/337/files#diff-ae86159524d9bc451d3e3cc04dc48bb2

test/integration/integration_test_fixtures.py
```diff
@@ -17,6 +17,8 @@
 from test.integration.nighthawk_test_server import NighthawkTestServer
 from test.integration.nighthawk_grpc_service import NighthawkGrpcService
 
+TIMESTAMP = time.strftime('%Y-%m-%d-%H-%M-%S')
```
mum4k, line 20: (nit) Can we either make this private (i.e. _TIMESTAMP), or document it?

test/integration/integration_test_fixtures.py
```diff
@@ -45,24 +47,28 @@ class IntegrationTestBase():
   This class will be refactored (https://github.com/envoyproxy/nighthawk/issues/258).
   """
 
-  def __init__(self, ip_version, backend_count=1):
+  def __init__(self, ip_version, server_config, backend_count=1):
```
mum4k, line 50: Can we document server_config in the Args: section of the docstring?

test/integration/integration_test_fixtures.py
```diff
@@ -45,24 +47,28 @@ class IntegrationTestBase():
...
     Args:
       ip_version: a single IP mode that this instance will test: IpVersion.IPV4 or IpVersion.IPV6
       backend_count: number of Nighthawk Test Server backends to run, to allow testing MultiTarget mode
     """
     super(IntegrationTestBase, self).__init__()
-    self.test_rundir = os.path.join(os.environ["TEST_SRCDIR"], os.environ["TEST_WORKSPACE"])
-    self.nighthawk_test_server_path = os.path.join(self.test_rundir, "nighthawk_test_server")
-    self.nighthawk_test_config_path = None
-    self.nighthawk_client_path = os.path.join(self.test_rundir, "nighthawk_client")
+    self.confdir = "nighthawk/test/integration/configurations/"
```
mum4k, line 57: We are adding some public attributes, can we document them in the class docstring? It will ideally have an "Attributes:" section.

This is partially pre-existing since none of the other attributes are documented there. Up to you if we document all now, or leave a todo for the pre-existing ones. We should still prefer making anything that can be private.

test/integration/integration_test_fixtures.py
```diff
@@ -84,15 +90,22 @@ def setUp(self):
     """
     Performs sanity checks and starts up the server. Upon exit the server is ready to accept connections.
     """
-    assert (os.path.exists(self.nighthawk_test_server_path))
-    assert (os.path.exists(self.nighthawk_client_path))
-    test_id = os.environ.get('PYTEST_CURRENT_TEST').split(':')[-1].split(' ')[0].replace(
-        "[", "_").replace("]", "")
-    self.parameters["test_id"] = test_id
+    if os.getenv("NH_DOCKER_IMAGE", "") == "":
+      assert (os.path.exists(self.nighthawk_test_server_path))
```
mum4k, line 94: Can we amend these assertions with some error message that would help the receiver of this error message understand what wen't wrong?

Something like:
assert False, "custom error message"

test/integration/integration_test_fixtures.py
```diff
@@ -84,15 +90,22 @@ def setUp(self):
...
+      assert (os.path.exists(self.nighthawk_client_path))
+
+    self.test_id = os.environ.get('PYTEST_CURRENT_TEST').split(':')[-1].split(' ')[0].replace(
```
mum4k, line 97: Ideally attributes on classes are all initialized (and documented if public) inside the __init__ method. Crating attributes "dynamically" in other methods makes it harder to find the complete list (API) of the class.

Can we initialize self.test_id to some useful value inside __init__? If nothing else, we can set it to the zero value of the same type (an empty string?).

test/integration/integration_test_fixtures.py
```diff
@@ -84,15 +90,22 @@ def setUp(self):
...
+        "[", "_").replace("]", "").replace("/", "_")[5:]
+    self.tag = "{timestamp}/{test_id}".format(timestamp=TIMESTAMP, test_id=self.test_id)
+
     for i in range(self.backend_count):
-      test_server = NighthawkTestServer(self.nighthawk_test_server_path,
-                                        self.nighthawk_test_config_path, self.server_ip,
-                                        self.ip_version, self.parameters)
+      test_server = NighthawkTestServer(
```
mum4k, line 102: This setUp method is getting fairly long. Any chance we can extracts some logical sub-parts from it and move it into well named private functions (or methods if necessary) ?

test/integration/integration_test_fixtures.py
```diff
@@ -214,26 +233,43 @@ def runNighthawkClient(self, args, expect_failure=False, timeout=30, as_json=Tru
       assert (client_process.returncode == 0)
     return output, logs
 
+  def transformNighthawkJson(self, json, format="human"):
```
mum4k, line 236: Can we document this public method?

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,85 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
     self.server_binary_config_path_arg = server_binary_config_path_arg
-
     self.parameters["server_ip"] = self.server_ip
-    with open(self.config_template_path) as f:
-      config = Template(f.read())
-      config = config.substitute(self.parameters)
-      logging.info("Parameterized server configuration: %s", config)
-
-    with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".yaml") as tmp:
+    self.docker_image = os.getenv("NH_DOCKER_IMAGE", "")
```
mum4k, line 42: Note about documenting public attributes applies here as well.

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,85 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
+    self.tmpdir = os.path.join(os.getenv("TMPDIR", "/tmp/nighthawk_benchmark/"), tag + "/")
+    self.parameters["tmpdir"] = self.tmpdir
+    self.parameters["tag"] = tag
+
+    def substitute_yaml_values(obj, params):
```
mum4k, line 47: Can we define this as a standalone (global scope) function instead to make the __init__ method shorter and easier to follow?

Once we do, we should also document it.

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,85 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
+        for i in range(len(obj)):
+          obj[i] = substitute_yaml_values(obj[i], params)
+      else:
+        if isinstance(obj, str):
+          # Inspect string values and substitute where applicable.
+          INJECT_RUNFILE_MARKER = '@inject-runfile:'
+          if obj[0] == '$':
+            return Template(obj).substitute(params)
+          elif obj.startswith(INJECT_RUNFILE_MARKER):
+            r = runfiles.Create()
```
mum4k, line 61: Here and below, can we use a better name than "r" ?

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,85 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
+
+    with tempfile.NamedTemporaryFile(
+        mode="w", delete=False, suffix=".adminport", dir=self.tmpdir) as tmp:
       self.admin_address_path = tmp.name
 
   def serverThreadRunner(self):
-    args = [
+    args = []
+    if self.docker_image != "":
+      # TODO(XXX): As of https://github.com/envoyproxy/envoy/commit/e8a2d1e24dc9a0da5273442204ec3cdfad1e7ca8
```
mum4k, line 91: Did we mean to replace XXX with a value?

test/integration/nighthawk_test_server.py
```diff
@@ -30,41 +33,85 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
...
+      return obj
+
+    r = runfiles.Create()
+    with open(r.Rlocation(self.config_template_path)) as f:
+      data = yaml.load(f, Loader=yaml.FullLoader)
+      data = substitute_yaml_values(data, self.parameters)
+
+    Path(self.tmpdir).mkdir(parents=True, exist_ok=True)
+
+    with tempfile.NamedTemporaryFile(
```
mum4k, line 73: Are there any parts of this function that we could move out so that we make __init__ shorter and easier to follow?

test/integration/nighthawk_test_server.py
```diff
@@ -131,6 +179,18 @@ def __init__(self,
...
                server_ip,
                ip_version,
-               parameters=dict()):
+               parameters=dict(),
+               tag=""):
     super(NighthawkTestServer, self).__init__(server_binary_path, config_template_path, server_ip,
-                                              ip_version, "--config-path", parameters)
+                                              ip_version, "--config-path", parameters, tag)
+
+  def getCliVersionString(self):
```
mum4k, line 187: Can we document this method?

test/integration/test_grpc_service.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import pytest
 
-from test.integration.integration_test_fixtures import http_test_server_fixture
+from test.integration.integration_test_fixtures import (http_test_server_fixture, server_config)
```
mum4k, line 4: (optional for this PR) comment about only importing modules / packages applies here too.

Optional, because it is pre-existing and we probably don't want to increase the scope of this PR.

test/integration/test_integration_basics.py
```diff
@@ -30,7 +30,7 @@ def test_http_h1(http_test_server_fixture):
   assertCounterEqual(counters, "upstream_cx_rx_bytes_total", 3400)
   assertCounterEqual(counters, "upstream_cx_total", 1)
   assertCounterEqual(counters, "upstream_cx_tx_bytes_total",
-                     1400 if http_test_server_fixture.ip_version == IpVersion.IPV6 else 1500)
+                     1400 if http_test_server_fixture.ip_version == IpVersion.IPV6 else 1450)
```
mum4k, line 33: The two literals (1400, 1450) would benefit from being well named constants with some explanation.

oschaaf: The change here is triggered by that we now connect to 0.0.0.0 instead of 127.0.0.1, resulting in a decrease of `upstream_cx_tx_bytes_total` as the request header gets slightly smaller.

oschaaf: After discussing on Slack, we concluded this can stay as is; it's not as if this has anything to do with MTU sizes :-)

test/integration/nighthawk_test_server.py
```diff
@@ -18,61 +18,79 @@
...
+                  'r') as file:
+          return file.read()
+  return obj
+
+
 class TestServerBase(object):
   """
     Base class for running a server in a separate process.
+
+    Arguments:
```
mum4k, line 45: (nit) The "Args:" section of a class docstring actually belongs to the docstring of the initializer (__init__). Would you mind moving it there?

I am also happy to do this in a follow-up PR since this is minor, just let me know.

oschaaf: I prefer to do it in a follow up: this file exists out of `benchmarks/`, I'm hoping to tighten up the format checking so that I can get all the python to comply with the check format rules that are now specific to `benchmark/`. Being backed by that makes it easier, as the doc string format isn't in my muscle memory yet. Added a note to https://github.com/envoyproxy/nighthawk/issues/371 as a reminder.

oschaaf: Forgot to mark this is ready for review after posting the comment above  - just did so.

PR #338, opened 2020-06-01
Add echo_request_headers to the test server filter ResponseOptions.

efimki:
If `echo_request_headers` option is set to `true`, then request headers dump is appended to the response body.

### DIFF
--- a/source/server/http_test_server_filter.h
+++ b/source/server/http_test_server_filter.h
@@ -73,6 +73,7 @@ private:
   Envoy::Http::StreamDecoderFilterCallbacks* decoder_callbacks_;
   nighthawk::server::ResponseOptions base_config_;
   absl::optional<std::string> error_message_;
+  absl::optional<std::string> request_headers_dump_;
 };
 
 } // namespace Server
--- a/test/server/http_test_server_filter_integration_test.cc
+++ b/test/server/http_test_server_filter_integration_test.cc
@@ -191,15 +191,17 @@ TEST_P(HttpTestServerIntegrationTest, TestHeaderConfig) {
 
 TEST_P(HttpTestServerIntegrationTest, TestEchoHeaders) {
   Envoy::BufferingStreamDecoderPtr response = makeSingleRequest(
-      lookupPort("http"), "GET", "/echoheaders", "", downstream_protocol_, version_, "foo.com", "",
+      lookupPort("http"), "GET", "/somepath", "", downstream_protocol_, version_, "foo.com", "",
       [](Envoy::Http::RequestHeaderMapImpl& request_headers) {
         request_headers.addCopy(Envoy::Http::LowerCaseString("gray"), "pidgeon");
         request_headers.addCopy(Envoy::Http::LowerCaseString("red"), "fox");
+        request_headers.addCopy(Nighthawk::Server::TestServer::HeaderNames::get().TestServerConfig,
+                                "{echo_request_headers: true}");
       });
   ASSERT_TRUE(response->complete());
   EXPECT_EQ("200", response->headers().Status()->value().getStringView());
   EXPECT_THAT(response->body(), HasSubstr(R"(':authority', 'foo.com')"));
-  EXPECT_THAT(response->body(), HasSubstr(R"(':path', '/echoheaders')"));
+  EXPECT_THAT(response->body(), HasSubstr(R"(':path', '/somepath')"));
   EXPECT_THAT(response->body(), HasSubstr(R"(':method', 'GET')"));
   EXPECT_THAT(response->body(), HasSubstr(R"('gray', 'pidgeon')"));
   EXPECT_THAT(response->body(), HasSubstr(R"('red', 'fox')"));
--- a/test/server/http_test_server_filter_integration_test.cc
+++ b/test/server/http_test_server_filter_integration_test.cc
@@ -190,21 +190,26 @@ TEST_P(HttpTestServerIntegrationTest, TestHeaderConfig) {
 }
 
 TEST_P(HttpTestServerIntegrationTest, TestEchoHeaders) {
-  Envoy::BufferingStreamDecoderPtr response = makeSingleRequest(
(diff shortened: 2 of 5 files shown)


efimki:
@htuch PTAL when you have a chance.

efimki:
> This looks like a great feature to me, some thoughts:

Excellent, thanks for your suggestions!

> * Maybe an option to allow people to configure `/echoheaders` would be nice-to-have.
>   (It would also allow an escape latch in the (highly unlikely) event someone needs this path to function like it did before this change)

The current implementation is inspired by Chromium's Embedded Test Server [1] that is mostly used in net and cronet [2] unit tests. 

Your suggestion made me realize that in Nighthawk it will be better to configure special handlers using config instead of pre-defined request path. This way echo behaviors can be configured per individual request, applied to any request path, and combined if necessary (for example, echo headers and echo metadata). I'll update the PR.

> * A note in `README.md` would help make this feature discoverable by others.

Absolutely! I just wanted to flesh out the implementation before updating the docs.

[1] https://source.chromium.org/chromium/chromium/src/+/master:net/test/embedded_test_server/embedded_test_server.h
[2] https://source.chromium.org/chromium/chromium/src/+/master:components/cronet/test/test_server.h

efimki:
Thanks for the suggestion! 
I've added an example to README.md.
WDYT?

efimki:
On related note, after this PR I'd like to add more response options that I think might be useful:

* `echo_request_metadata` - include the dump of request metadata in the response body
* `echo_request_body` - include the body of `POST` request in the response body
* `status_code` - set http status code to arbitrary value, like `204` or `302`

WDYT?

source/server/http_test_server_filter.h
```diff
@@ -73,6 +73,7 @@ class HttpTestServerDecoderFilter : public Envoy::Http::StreamDecoderFilter {
   Envoy::Http::StreamDecoderFilterCallbacks* decoder_callbacks_;
   nighthawk::server::ResponseOptions base_config_;
   absl::optional<std::string> error_message_;
+  absl::optional<std::string> request_headers_dump_;
```
mum4k, line 76: I am not familiar with this portion of the code - can you confirm that the lifetime of the HttpTestServerDecoderFilter class is per single request?

test/server/http_test_server_filter_integration_test.cc
```diff
@@ -189,6 +189,24 @@ TEST_P(HttpTestServerIntegrationTest, TestHeaderConfig) {
   EXPECT_EQ(std::string(10, 'a'), response->body());
 }
 
+TEST_P(HttpTestServerIntegrationTest, TestEchoHeaders) {
```
mum4k, line 192: Can we add a test case with multiple requests and different headers in them to confirm the lifetime question above? The headers should be reflected just to a single response related to a single request.

htuch:
@efimki LGTM modulo the comments from @mum4k. Re: future directions, do you want to eventually have a more structured body. E.g. make it a proto defined JSON format or something like that? To what extent is this going to be human interpreted vs. machine read?

test/server/http_test_server_filter_integration_test.cc
efimki: I've added unique_header to verify that echoed headers vary per request. 
Unfortunately I don't know how to make those requests parallel instead of sequential. Do you have any pointers?

efimki:
> @efimki LGTM modulo the comments from @mum4k. Re: future directions, do you want to eventually have a more structured body. E.g. make it a proto defined JSON format or something like that? To what extent is this going to be human interpreted vs. machine read?

Maybe? I haven't really thought this through yet. 
The chromium test server that I know is very simple and doesn't do any special formatting.
My current use case is fine with plain text dump.
I would follow YAGNI spirit and don't decide on this until we have a clear use case for something more elaborate.

mum4k:
Thank you @efimki for addressing my comments, this looks good in regards to test coverage.

I think @htuch raises a good point. I do agree that we shouldn't implement anything prematurely, however this might be the only time to introduce any structure, since we have a backward compatibility guarantee in the Nighthawk repository. Should we consider at least adding an inner structure (proto definition?) that will for now contain these headers only?

mum4k:
@efimki can you please push a commit (or two) to kick the CI jobs (automated check)? We will want them to be green before we proceed.

efimki:
Hopefully this will do. If not, I'll try again later.

oschaaf:
FWIW I suspect CircleCI is hogged, we may just need to wait for that to clear up. Other PRs also seem subject to this now

efimki:
Trying to close / reopen to get ci unstuck.

mum4k:
If nothing else helps, you can consider closing this PR and creating a new one with the same content. Since we seem to be out of ideas :(

efimki:
FWIW I don't seek the fame of authoring this PR. :)
Maybe somebody else could create a new PR based on it that will be not be ignored by CI?

oschaaf:
I mirrored this PR into #349, and CI is running fine over there.
@efimki I'm sorry this is such a hassle. 
I have checked everything that I have access to with respect to project-settings in github and CircleCI, but couldn't find anything I could relate to CI not running here. So I still can only speculate about potential causes.

oschaaf:
So .. all tests are now passing here

efimki:
Oh, nice, thanks for your help! 
I think I've addressed @mum4k change request, but I don't know whether I should somehow explicitly mark it as addressed. Any ideas!

oschaaf:
I think the last step is that @mum4k needs to explicitly add an approving review before this PR becomes mergeable. Apparently the CI run of my clone PR was able to green-light CI for this one by flagging the last (shared) commit as OK for all jobs.

mum4k:
Magic, thanks @oschaaf.

@efimki would you mind closing all the related duplicates?

efimki:
I've closed obsolete PRs #345 and #346 and opened #351 to discuss structure of additional echo responses.

PR #357, opened 2020-06-11
integration-tests: tweaks to stabilize sanitizer runs

oschaaf:
This PR has changes to deflake integration tests for sanitizer runs.

- Allow servers more time to start up
- Slow down some tests to make them less sensitive to timing issues 
- It is possible for some counter values to slightly exceed a configured termination threshold
  for them in edge cases. Update expectations to match that. 

Fixes #332

PR #358, opened 2020-06-12
Coverage testing and docker flow fixes

oschaaf:
Following some digging into stability of our coverage testing, creating this PR for testing CI
with some fixes / enhancements, as well as covering a few more lines in our
CLI tooling exception paths. There seems to be something racy going on where sometimes the coverage originating from our python integration tests is not counted.

Eventually I learned that things might stabilise when we switch the toolchain to clang 10. We follow Envoy's lead on that, so for now stopping further efforts on this.

### Description of the changes in this PR:

- Increase timeouts of ASAN/TSAN runs
- Make `envoy_build_sha.sh`, which apparently broke in one of the Envoy updates
- Cover some of the outer exception handling paths for Nighthawk's CLIs
- Integration test for nighthawk_output_transform
- Add coverage for code lines where it seemed trivial to do so (e.g. NullStatistic)
- Fix `run_nighthawk_bazel_coverage.sh` to work on my machine
- Stricter bash options in some scripts + changes to make that work

### Prerequisites

- [x] Needs #357 to go in first.

---------

Leaving some notes/learnings here:

- there seems to be something racy going on where sometimes the coverage originating from
   our python integration tests is not counted. This is relatively infrequent in CI.
   - On my dev machine this structurally happens. This might be a toolchain/version issue.
   - On my dev machine, using the docker flow, I have difficulties: the docker container exits when running the full test suite, without leaving a trace why. bash traps in the container don't fire, no useful logging from docker. Eventually I was able to get past bazel fetching its deps, and then build just the python integration tests. As long as a single test is involved, things seem to work / the container gets to finish. It's remarkable that CI doesn't have this problem. Coverage structurally looks as expected in this scenario.
- The link step is super memory hungry. Separating that step might be a ci perf opportunity. Coverage is the slowest CI job we have.


oschaaf:
Now that #357 is in, readying this one up for review. This should further stabilize CI.

test/integration/test_grpc_service.py
```diff
@@ -42,3 +42,25 @@ def test_grpc_service_stress(http_test_server_fixture):
   counters = http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
   assertGreaterEqual(counters["benchmark.http_2xx"], 5000)
   assertEqual(counters["requestsource.internal.upstream_rq_200"], 4)
+
+
+def run_service_with_args(args):
```
mum4k, line 47: Can we add documentation for this helper? E.g.

```py
def run_service_with_args(args):
  """Executes the Nighthawk service with the provided arguments.

  Args:
    args: A string, the command line arguments for the service, e.g. "--foo --bar".

  Returns:
    A tuple in the form (exit_code, output), where exit_code is the code the Nighthawk
    service terminated with and the output is its standard output.
  """
```

Comment about adding documentation applies to all new non-test public functions. We could consider making some of them private instead.

test/integration/test_grpc_service.py
```diff
@@ -42,3 +42,25 @@ def test_grpc_service_stress(http_test_server_fixture):
...
+  return run_binary_with_args("nighthawk_service", args)
+
+
+def test_grpc_service_help():
+  (exit_code, output) = run_service_with_args("--help")
+  assert (exit_code == 0)
```
mum4k, line 53: Since these are tests, we should likely use self.assertTrue instead of assert. I think assert just terminates the execution while self.assertFoo correctly records the test failure using the test framework.

test/integration/test_output_transform.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+import pytest
+
+from test.integration.utility import *
```
mum4k, line 4: Can we import specific modules or packages explicitly? It helps when debugging.

oschaaf:
> Thank you very much for improving the stability, this has been hitting us on multiple PRs. As an overall comment - can we try to enumerate the fixes we are making here? It makes it a bit hard to review the PR and relate the various changes to intentions.

Ah yes, sure: I updated the PR description to list what I did here. Let me know if that looks better.

I will address the other comments later today.

test/integration/test_grpc_service.py
oschaaf: Well, we're using `pytest`, I think this ought te be ok. quoting from the docs:
```
Due to pytest’s detailed assertion introspection, only plain assert statements are used. 
```

Posting the output of a sample failure (introduced on purpose):

```
    def test_output_transform_help():
      (exit_code, output) = _run_output_transform_with_args("--help")
>     assert (exit_code == 1)
E     assert 0 == 1
E       -0
E       +1

exit_code  = 0
output     = ('WARNING: Perftools heap leak checker is active -- Performance may suffer\n'
 '\n'
 'USAGE: \n'
.... <snip> ...
```


[1] https://docs.pytest.org/en/stable/

test/integration/test_grpc_service.py
oschaaf: I made added this comment, slightly modified, to `run_service_with_args`. And then made this function and others that reuse it private.

test/integration/test_grpc_service.py
oschaaf: Having said that, we do have some helpers in `utilities.py`, like `assertEqual`. Switched to use that, for consistency with the rest of the test code.

test/integration/test_output_transform.py
oschaaf: Amended this in some places. But we may want to track this as tech debt in an issue, because this occurs in other places as well. I filed https://github.com/envoyproxy/nighthawk/issues/371 for that (as `tech-debt`)

oschaaf:
@mum4k thanks, I left some remarks, and also pushed f0596d4278a0c20eb81a982ae2516cd1f4bb71f1 which hopefully addresses your comments.

test/integration/test_grpc_service.py
```diff
@@ -2,7 +2,8 @@
 import pytest
 
 from test.integration.integration_test_fixtures import http_test_server_fixture
-from test.integration.utility import *
+from test.integration.utility import (isSanitizerRun, assertEqual, assertIn, assertGreaterEqual,
```
mum4k, line 5: Looks like this now imports functions from the utility module. What we really should be importing is the utility module itself.

from test.integration import utility

And then calls will look like.
utility.assertEqual

Ideally utility would have a better more descriptive name, but we can address that separately. The Python style we are following internally suggests to only import packages or modules, that way it is easy to say where some function is defined. If the reader just sees "assertEqual", they have to do some digging.

oschaaf: That makes sense, in the benchmark PR I noticed pep8 had trouble with the * imports too.
Changed what was relevant to this PR in 66f658553658b31ad93fac5f1dfc4b7871045265
Filed https://github.com/envoyproxy/nighthawk/issues/371 earlier to track doing a pass over
the all the other code, to change the rest.

mum4k: Thank you for filing the issue.

Issue #360, opened 2020-06-15
idea: track and propagate test server timings for comparison


oschaaf:
The test server could add some extra timing information to its response headers:

- absolute time of request start from the test server PoV. (Probably only useful in tests where the client and server run on the same machine.
- latency between the test server's obervations of request start and its own reply time. This gets a sense of time spend generating a reply. 
- latencies between consecutive inbound request start times. 

Nighthawk's client could then add opt-in tracking for these values in new histograms with relatively little effort.

Having insight into these timings may be valuable as they assist in tracking (distortion of) request-release timings as well as give some more insight into components that make up the overall latency profiles.

colt415:
+1

This timing info would be very useful for debugging with NH test server.

PR #361, opened 2020-06-15
[DRAFT] Tracking test server origin timings

oschaaf:
Consolidating an old draft here as a companion to #360 

Primarily to gauge interest, if there is none we can close this


mum4k:
@oschaaf Can we close this for now to keep the PR dashboard focused on active PRs?

PR #363, opened 2020-06-16
gRPC service logging bugfix

### LINKED ISSUE #289: gRPC service logs first ExecutionStream request to STDERR, but not further requests
After starting up the gRPC service, it will log ONLY the first request, which can make it harder to debug issues going on with the service.

We should log every request, though I'm not sure STDERR is the correct place to log any of these.

Also, while addressing this, we should make sure that it doesn't extend to other behaviors that may need to run on every request, but do not do so.

oschaaf:
This makes the gRPC service log more then one ExecutionStream requests.
The log level used will be controlled by the options associated to
ExecutionStream requests (possible candidate for improvement later on).

This change also incorporates a change to avoid creating multiple instances
of Envoy::ProcessWide in the Nighthawk gRPC service flow.
This doesn't seem to address any know issues, but it seems to be more along
the way Envoy::ProcessWide was intended to be used.

Fixes #289


source/client/service_impl.cc
```diff
@@ -33,11 +33,7 @@ void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionReque
     return;
   }
 
-  ProcessImpl process(*options, time_system_);
```
oschaaf, line 36: Note to reviewers: the fix is to not do this here. We do this once now at construction time, and hold on to the logging context throughout the lifetime of the `ServiceImpl`.
Note that `ProcessImpl` instantiations will change the log level per inbound configured options. We only allow one `ProcessImpl` instantiation at a time today, so this might be allright for now.

source/client/process_impl.cc
```diff
@@ -92,8 +92,11 @@ class ClusterManagerFactory : public Envoy::Upstream::ProdClusterManagerFactory
   bool prefetch_connections_{};
 };
 
-ProcessImpl::ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system)
-    : time_system_(time_system), stats_allocator_(symbol_table_), store_root_(stats_allocator_),
+ProcessImpl::ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system,
+                         const std::shared_ptr<Envoy::ProcessWide>& process_wide)
+    : process_wide_(process_wide == nullptr ? std::make_shared<Envoy::ProcessWide>()
```
mum4k, line 97: Unsure about the context - do we need to worry about thread safety here or is ProcessImpl guaranteed to be constructed from a single thread?

source/client/process_impl.cc
```diff
@@ -92,8 +92,11 @@ class ClusterManagerFactory : public Envoy::Upstream::ProdClusterManagerFactory
   bool prefetch_connections_{};
 };
 
-ProcessImpl::ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system)
-    : time_system_(time_system), stats_allocator_(symbol_table_), store_root_(stats_allocator_),
+ProcessImpl::ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system,
```
mum4k, line 95: Is there any way how we can add test coverage for this change to keep the fixed functionality working?

source/client/process_impl.h
```diff
@@ -46,7 +46,8 @@ class ClusterManagerFactory;
  */
 class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
 public:
-  ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system);
+  ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_system,
```
mum4k, line 49: Can we add a doc comment? We could also explain the optional process_wide argument and say what happens if it isn't provided.

source/client/service_impl.h
```diff
@@ -26,6 +28,10 @@ class ServiceImpl final : public nighthawk::client::NighthawkService::Service,
                           public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
 
 public:
+  ServiceImpl() : process_wide_(std::make_shared<Envoy::ProcessWide>()) {
```
mum4k, line 31: Can we add a doc comment for the public constructor?

source/client/process_impl.cc
oschaaf: Actually, yes. Lacking C++ mocks for the logging, we can't easily unit-test this issue. 
But I could add an end-to-end integration test for this, see 619c05e

source/client/process_impl.cc
oschaaf: All `ProcessImpl` clients should ensure [there's only a single active instance at any given time](https://github.com/envoyproxy/nighthawk/blob/master/source/client/process_impl.h#L40). We don't have to worry about it.

oschaaf:
@mum4k thanks for the review; comments addressed!

test/integration/nighthawk_grpc_service.py
```diff
@@ -39,6 +39,7 @@ def __init__(self,
     assert ip_version != IpVersion.UNKNOWN
     self.server_port = 0
     self.server_ip = server_ip
+    self.log_lines = None
```
mum4k, line 42: (nit) Does this need to be public or would self._log_lines do? If public, can we document it above in the Attributes: section of the class docstring?

oschaaf: Well, it gets consumed over at https://github.com/envoyproxy/nighthawk/pull/363/files#diff-9620177d1b84e578a841abdc925dcdf4R33. So it needs to be public, and documented. I'll add that.

oschaaf: Done in 593fe3c (also moved the line that resets the log to `None` to a better place).

PR #365, opened 2020-06-16
Avoid log warning for tls_context if the user didn't set it

### LINKED ISSUE #364: CLI: log spam with warning about deprecated tls_context being used
When using `nighthawk_client`, a warning-level log line [1] will always appear about `nighthawk.client.CommandLineOptions.tls_context` being used -- even while it's not actively being set as a configuration argument.

The warning should only emerge when it's being used actively configured. 

[1] sample
```
[2020-06-16 20:51:43.610][79803][warning][misc] Deprecated field: type nighthawk.client.CommandLineOptions Using deprecated option 'nighthawk.client.CommandLineOptions.tls_context' from file options.proto. This configuration will be removed from Envoy soon. 
Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

oschaaf:
Fixes #364


source/client/options_impl.cc
```diff
@@ -671,7 +671,14 @@ CommandLineOptionsPtr OptionsImpl::toCommandLineOptionsInternal() const {
       request_options->mutable_request_body_size()->set_value(requestBodySize());
     }
   }
-  *(command_line_options->mutable_tls_context()) = tls_context_;
+
+  // Only set the tls context if it looks non-empty, to avoid a warning being logged about field
+  // deprecation. Ideally this would follow the way transport_socket uses absl::optional below.
+  // But as this field is about to get eliminated this minimal effort shortcut may be more suitable.
+  if (!Envoy::MessageUtil()(tls_context_,
```
mum4k, line 678: Can we add a comment explaining how we are using Envoy::MessageUtil to verify that the tls_context_ message is non-empty?

(optional) On another note - isn't there a more direct way of checking if it is non-empty? E.g. proto2 used to have a method that could be used for that - my_proto.ByteSizeLong() == 0. However that call tends to be very inefficient on messages with many nested fields and I am unsure if it works in proto3 also.

oschaaf: I found that theres also `Envoy::Protobuf::util::MessageDifferencer::Equivalent` these days, which I think is easier to grok. See 6b8a0889c10225a3f036c8c098571832f03b900c, does that look better?

source/client/options_impl.cc
```diff
@@ -672,11 +672,11 @@ CommandLineOptionsPtr OptionsImpl::toCommandLineOptionsInternal() const {
...
   }
 
-  // Only set the tls context if it looks non-empty, to avoid a warning being logged about field
-  // deprecation. Ideally this would follow the way transport_socket uses absl::optional below.
+  // Only set the tls context if needed, to avoid a warning being logged about field deprecation.
+  // Ideally this would follow the way transport_socket uses absl::optional below.
   // But as this field is about to get eliminated this minimal effort shortcut may be more suitable.
-  if (!Envoy::MessageUtil()(tls_context_,
-                            envoy::extensions::transport_sockets::tls::v3::UpstreamTlsContext())) {
+  if (!Envoy::Protobuf::util::MessageDifferencer::Equivalent(
```
mum4k, line 678: Note - equivalent is not the same as equal. This would say that two messages are the same if one has a value set to zero and the other one has it unset where the default is zero.

Considering we are using this to determine if the message is unset, is that behavior desired?

oschaaf: That is a good catch, thanks. Thinking this through, you're right, this isn't what we need here. I switched this to use the  `ByteSizeLong()` method you pointed out. I'm not super concerned about it not being the fastest way to pull this off, as it is going to be eliminated at some point. Plus this isn't in the hot path. Addressed in f467e86

Issue #344, opened 2020-06-05
Add functionality to flush metrics periodically and a few metrics (latency, total requests sent, ...)


qqustc:
Currently Nighthawk does not flush statistics and metrics in real time during its run. We plan to add the functionality to flush Nighthawk metrics periodically (at a configured time interval) to Envoy stats Sinks following the example of what has been done in Envoy.

We also plan to add new counters and histograms to monitor
number of total requests sent  		(counter)
request latency 				(histogram)
request/response header/body size		(histogram)
number of total requests blocked  		(counter)

/cc @mum4k

oschaaf:
This sounds awesome. Some thoughts that are potentially relevant:

- I found this counter intuitive, but the tls counters aren't actually thread local. So at this time we explicitly use a namespace to isolate them per worker. At the time I did have - from the top of my head - a derived Envoy counter implementation which would track both shared and local counters. This would avoid having to use the namespaces to isolate. This may (or may not) be worth revisiting.
- Another thing is that there's also an [open issue](https://github.com/envoyproxy/nighthawk/issues/115), where we concluded we'd like to switch from HdrHistogram to libcirclhist. Envoy uses libcirclhist these days, so actually we could just switch to Envoy's internal histogram implementation. Upon doing so, we'd have a similar issue with the per worker namespace we use, like above with the counters. I have not looked into what it takes to address that.
- A cross cutting concern might be how we sync histograms. Do we want to shoot for a unified way of sync/merging histograms when we think about threads, processes, machines, etc? If so, the way Envoy facilitates merging of histograms to the master thread might seem low hanging fruit -- but it might not be ideal when also considering the other scenarios.

qqustc:
Thanks Otto for the information!

- We plan to define per-worker Envoy counters and histograms by using the scope/namespace (similar to http_2xx counter in benchmark_client_impl.h), so the issue about tls counters aren't actually thread local should not matter here.

- Yes, our plan is to use Envoy's internal histogram implementation at per-worker level.

- Haven't decided on how we sync histograms but maybe it is natural to follow the way Envoy sync histograms.

qqustc:
Thanks Otto for the information!

- We plan to define per-worker Envoy counters and histograms by using the scope/namespace (similar to http_2xx counter in benchmark_client_impl.h), so the issue about tls counters aren't actually thread local should not matter here.

- Yes, our plan is to use Envoy's internal histogram implementation at per-worker level.

[...]

PR #366, opened 2020-06-16
Add "no-duration" to facilitate infinite load generation

### LINKED ISSUE #333: Allow --duration=0 or infinite duration
It can be useful to continually apply load. For these cases, an infinite duration would be nice. Currently setting duration=0 provides an error:

```
$ docker run envoyproxy/nighthawk-dev:59683b759eb8f8bd8cce282795c08f9e2b3313d4 nighthawk_client google.com --duration 0
[2020-05-01 17:01:34.204][1][warning][misc] [external/envoy/source/common/protobuf/utility.cc:198] Using deprecated option 'nighthawk.client.CommandLineOptions.tls_context' from file options.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated for details.
[2020-05-01 17:01:34.204][1][debug][misc] [external/envoy/source/common/protobuf/utility.cc:255] Proto validation error; throwing Proto constraint validation failed (CommandLineOptionsValidationError.Duration: ["value must be greater than or equal to " "1ns"]): requests_per_second {
```

oschaaf:
Adds `--no-duration`, which will be treated as a request to infinitely
continue load generation.  Other default predicates will still exist, so\
execution may still stop when connection failures or error response codes
are observed.

This leaves some small tech debt:

- We need test coverage, but that's much easier to add once
  #280 or #344 goes in

Fixes #333

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -102,8 +102,8 @@ message CommandLineOptions {
       [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
   // The maximum allowed number of concurrent connections per event loop. HTTP/1 only. Default: 100.
   google.protobuf.UInt32Value connections = 2 [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
-  // The number of seconds that the test should run. Default: 5.
-  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 1];
+  // The number of seconds that the test should run. Default: 5. 0 will be treated as infinite.
+  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 0];
   // Connection connect timeout period in seconds. Default: 30.
   google.protobuf.Duration timeout = 4 [(validate.rules).duration.gte.seconds = 0];
   // Use HTTP/2
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -102,8 +102,12 @@ message CommandLineOptions {
       [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
   // The maximum allowed number of concurrent connections per event loop. HTTP/1 only. Default: 100.
   google.protobuf.UInt32Value connections = 2 [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
-  // The number of seconds that the test should run. Default: 5. 0 will be treated as infinite.
-  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 0];
+  oneof oneof_duration_options {
+    // The number of seconds that the test should run. Default: 5.
+    google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 0];
+    // Request infinite execution. Note that the default failure predicates will still be added..
+    google.protobuf.BoolValue no_duration = 33;
+  }
   // Connection connect timeout period in seconds. Default: 30.
   google.protobuf.Duration timeout = 4 [(validate.rules).duration.gte.seconds = 0];
   // Use HTTP/2
--- a/source/client/factories_impl.cc
+++ b/source/client/factories_impl.cc
@@ -178,7 +178,7 @@ TerminationPredicateFactoryImpl::create(Envoy::TimeSource& time_source, Envoy::S
   if (options_.duration() > 0s) {
     root_predicate = std::make_unique<DurationTerminationPredicateImpl>(
         time_source, options_.duration(), scheduled_starting_time);
-  } 
+  }
 
   TerminationPredicate* current_predicate = root_predicate.get();
   current_predicate = linkConfiguredPredicates(*current_predicate, options_.failurePredicates(),
--- a/source/client/factories_impl.cc
+++ b/source/client/factories_impl.cc
@@ -174,12 +174,13 @@ TerminationPredicateFactoryImpl::TerminationPredicateFactoryImpl(const Options& 
 TerminationPredicatePtr
 TerminationPredicateFactoryImpl::create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
                                         const Envoy::MonotonicTime scheduled_starting_time) const {
-  TerminationPredicatePtr root_predicate = std::make_unique<NullTerminationPredicateImpl>();
-  if (options_.duration() > 0s) {
+  TerminationPredicatePtr root_predicate;
+  if (options_.noDuration()) {
+    root_predicate = std::make_unique<NullTerminationPredicateImpl>();
+  } else {
     root_predicate = std::make_unique<DurationTerminationPredicateImpl>(
         time_source, options_.duration(), scheduled_starting_time);
   }
-
   TerminationPredicate* current_predicate = root_predicate.get();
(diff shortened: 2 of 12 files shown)


source/client/factories_impl.cc
```diff
@@ -172,15 +174,24 @@ TerminationPredicateFactoryImpl::TerminationPredicateFactoryImpl(const Options&
...
                                         const Envoy::MonotonicTime scheduled_starting_time) const {
-  TerminationPredicatePtr duration_predicate = std::make_unique<DurationTerminationPredicateImpl>(
-      time_source, options_.duration(), scheduled_starting_time);
-  TerminationPredicate* current_predicate = duration_predicate.get();
+  TerminationPredicatePtr root_predicate;
+  if (options_.duration() > 0s) {
+    root_predicate = std::make_unique<DurationTerminationPredicateImpl>(
+        time_source, options_.duration(), scheduled_starting_time);
+  } else {
+    // TODO(oschaaf): clean this up. Refactor linkConfiguredPredicates so we can easily omit
```
oschaaf, line 182: Note to self, alternatively a `NullPredicate` may serve a purpose here too

source/client/factories_impl.cc
```diff
@@ -172,15 +174,19 @@ TerminationPredicateFactoryImpl::TerminationPredicateFactoryImpl(const Options&
 TerminationPredicatePtr
 TerminationPredicateFactoryImpl::create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
                                         const Envoy::MonotonicTime scheduled_starting_time) const {
-  TerminationPredicatePtr duration_predicate = std::make_unique<DurationTerminationPredicateImpl>(
-      time_source, options_.duration(), scheduled_starting_time);
-  TerminationPredicate* current_predicate = duration_predicate.get();
+  TerminationPredicatePtr root_predicate = std::make_unique<NullTerminationPredicateImpl>();
+  if (options_.duration() > 0s) {
```
mum4k, line 178: This may be a bit surprising to unaware users. Especially if zero is the default.

Additionally what would happen in the gRPC world if we send a request to the NH server with an infinite duration and how will the client terminate such a run and get the results? The same question likely applies to local runs - how do we terminate and get results?

Do we want to do some or all of the following?
a) print out on the screen + log saying that this is a run with an infinite duration and an instruction of how to terminate it.
b) should we use -1 instead of zero to avoid the "runs infinitely by default" situation? It would also help to make this change backwards compatible.
c) should we improve the documentation of the option indicating how does one terminate an infinite run and get results?

source/client/factories_impl.cc
```diff
@@ -172,15 +174,19 @@ TerminationPredicateFactoryImpl::TerminationPredicateFactoryImpl(const Options&
...
+    root_predicate = std::make_unique<DurationTerminationPredicateImpl>(
```
mum4k, line 179: Readability wise - it may be better to have an explicit if<->else rather then set a value and modify it after. The latter may confuse the reader if they don't notice the modification under the if statement.

Consider (pseudo code):
root_predicate = NULL
if (finite_run) {
  root_predicate = TerminateAt...
} else {
  root_predicate = NeverTerminate
}

source/common/termination_predicate_impl.h
```diff
@@ -27,6 +27,14 @@ class TerminationPredicateBaseImpl : public TerminationPredicate {
   TerminationPredicatePtr linked_child_;
 };
 
+/**
+ * Predicate which always returns the passed in termination status.
```
mum4k, line 31: Doesn't seem to accept any parameters, does the doc comment need an update?

test/run_nighthawk_bazel_coverage.sh
```diff
@@ -41,7 +41,7 @@ COVERAGE_VALUE=${COVERAGE_VALUE%?}
```
mum4k, line 41: Can we (for now) add a TODO pointing to an issue about the missing test coverage? Not sure if this is the right file, so feel free to add it in the appropriate location.

source/client/factories_impl.cc
oschaaf: > This may be a bit surprising to unaware users. Especially if zero is the default.
> 

Our default duration actually is 5 seconds, but that gets defined elsewhere in the options implementation header. That part doesn't change in this PR.

> Additionally what would happen in the gRPC world if we send a request to the NH server with an infinite duration and how will the client terminate such a run and get the results? The same question likely applies to local runs - how do we terminate and get results?
> 

For local runs, there's no way to terminate (gracefully) today. But this PR purposely didn't worry about that, as the feature was requested in a context where [results are irrelevant](https://github.com/envoyproxy/nighthawk/issues/333#issuecomment-645011251). So this PR helps by proving a way to just generate infinite load. 
The gRPC service would behave identical. 

However, I did have [similar thoughts](https://github.com/envoyproxy/nighthawk/issues/333#issuecomment-644989540) before starting on this, so I also [added a follow up](https://github.com/envoyproxy/nighthawk/pull/367) to that teaches the CLI about handling signals and wires through a way for `ProcessImpl` to quickly and gracefully exit and yield results. `RemoteProcessImpl` still has a `TODO` for that, but the path I had in mind there would send a gRPC message to request cancellation of the currently active `ProcessImpl` over there, which would then be propagated to the active `ProcessImpl` running within this service -- if any (we can only have a single one active at a time today). 

> Do we want to do some or all of the following?
> a) print out on the screen + log saying that this is a run with an infinite duration and an instruction of how to terminate it.

Right now the CLI will print that execution will target 0 seconds via an informational log line. I will make that print out "infinite" instead.

> b) should we use -1 instead of zero to avoid the "runs infinitely by default" situation? It would also help to make this change backwards compatible.

Alternatively, we could also try to add an explicit `--no-duration` flag, and make this value an internal detail. If we back out the current change, zero would be suitable for that, as we guard against that value being set by the user. I can go either way.

> c) should we improve the documentation of the option indicating how does one terminate an infinite run and get results?
That probably belongs in the follow up I mentioned above?

oschaaf:
@mum4k Comments addressed in f3a7ee8. As proposed on Slack, that commit changes this PR to use `--no-duration` instead of a magical value.

api/client/options.proto
```diff
@@ -102,8 +102,12 @@ message CommandLineOptions {
       [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
   // The maximum allowed number of concurrent connections per event loop. HTTP/1 only. Default: 100.
   google.protobuf.UInt32Value connections = 2 [(validate.rules).uint32 = {gte: 1, lte: 1000000}];
-  // The number of seconds that the test should run. Default: 5. 0 will be treated as infinite.
-  google.protobuf.Duration duration = 3 [(validate.rules).duration.gte.nanos = 0];
+  oneof oneof_duration_options {
+    // The number of seconds that the test should run. Default: 5.
```
mum4k, line 106: Should this be a oneof in the proto, or we don't do / support that?

mum4k: I was looking at the right revision, but my eyes betrayed me. Nothing to see here, all good ;)

PR #367, opened 2020-06-17
CLI - cancel execution upon signal reception

### LINKED ISSUE #280: nighthawk_client: handle termination signal

oschaaf:
Teach the CLI to handle SIGTERM/SIGINT, and handle those as a request to
gracefully cancel execution.

Partially resolves #280 

---- 

- [x] Prerequisite: Land #282


mum4k:
@eric846 please review and assign to me once ready.

source/client/process_impl.cc
```diff
@@ -385,44 +390,50 @@ void ProcessImpl::addRequestSourceCluster(
...
-  createBootstrapConfiguration(bootstrap, uris, request_source_uri, number_of_workers);
-  if (tracing_uri != nullptr) {
-    setupTracingImplementation(bootstrap, *tracing_uri);
-    addTracingCluster(bootstrap, *tracing_uri);
-  }
-  ENVOY_LOG(debug, "Computed configuration: {}", bootstrap.DebugString());
-  cluster_manager_ = cluster_manager_factory_->clusterManagerFromProto(bootstrap);
-  maybeCreateTracingDriver(bootstrap.tracing());
-  cluster_manager_->setInitializedCb([this]() -> void { init_manager_.initialize(init_watcher_); });
+  {
```
oschaaf, line 397: Note to reviewers: The "hide whitespace" toggle in Github makes this a lot more fun to review.
(The scope introduced for the lock guard introduces a lot of whitespace noise).

test/integration/test_integration_basics.py
```diff
@@ -654,6 +658,33 @@ def test_http_request_release_timing(http_test_server_fixture, qps_parameterizat
...
+  ]
+  client_process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  Thread(target=(lambda: send_sigterm(client_process))).start()
+  stdout, stderr = client_process.communicate()
+  client_process.wait()
+  output = stdout.decode('utf-8')
+  assert (client_process.returncode == 0)
+  parsed_json = json.loads(output)
+  counters = http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
+  assertCounterEqual(counters, "graceful_stop_requested", 2)
```
eric846, line 685: Might also want to assert that the 2xx counters are >0, so we know tsan didn't slow it down so much that it only exercised the early exit in ProcessImpl::runInternal when cancelled_ is already true.

test/process_test.cc
```diff
@@ -32,12 +33,34 @@ class ProcessTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
         options_(TestUtility::createOptionsImpl(
             fmt::format("foo --duration 1 -v error --rps 10 https://{}/", loopback_address_))){};
-  void runProcess(RunExpectation expectation) {
+  void runProcess(RunExpectation expectation, bool do_cancel = false) {
     ProcessPtr process = std::make_unique<ProcessImpl>(*options_, time_system_);
     OutputCollectorImpl collector(time_system_, *options_);
+    std::thread cancel_thread;
+    if (do_cancel) {
+      cancel_thread = std::thread([&process] {
+        sleep(5);
```
eric846, line 42: Would it be worth also testing the case where the sigterm comes before the threads started?

test/integration/test_integration_basics.py
oschaaf: Sure; done in 1fddbec526a464606577a569cf8a8a464d175025

test/process_test.cc
oschaaf: I agree that would be a good thing to have. But it seems challenging to add such a test (without adding code to the implementation to support it). I need to think about it some more.

oschaaf: I think I misunderstood, added a test in 20e610867dbbbeb48d07952fa0de45b1bad62d52

oschaaf: Whoops @eric846 , my comment crossed yours :-)

oschaaf: @eric846 IMHO your suggestion did yield a meaning full test which turned out easy to add. So thanks!

Still, thinking more about it, I feel this isn't complete. A repeated and non deterministic test which fuzzes timing to hopefully strike all possible flows with regard to cancellation might be good to have too at some point. But this isn't trivial to add today.

oschaaf:
Thanks @eric846 for the review. Just marked this as ready for another pass

source/client/client.cc
```diff
@@ -73,7 +74,12 @@ bool Main::run() {
   }
   OutputFormatterFactoryImpl output_formatter_factory;
   OutputCollectorImpl output_collector(time_system, *options_);
-  const bool res = process->run(output_collector);
+  bool res;
```
mum4k, line 77: (nit) If we can, we should try to avoid abbreviations in variable names (res).

source/client/remote_process_impl.cc
```diff
@@ -54,5 +54,11 @@ bool RemoteProcessImpl::run(OutputCollector& collector) {
   return false;
 }
 
+bool RemoteProcessImpl::requestExecutionCancellation() {
+  ENVOY_LOG(error, "Remote process cancellation not supported yet");
+  // TODO(XXX): Send a cancel request to the gRPC service.
```
mum4k, line 59: Did we want to replace XXX with a value?

source/client/process_impl.h
```diff
@@ -99,7 +101,7 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
   void maybeCreateTracingDriver(const envoy::config::trace::v3::Tracing& configuration);
 
   void configureComponentLogLevels(spdlog::level::level_enum level);
-  const std::vector<ClientWorkerPtr>& createWorkers(const uint32_t concurrency);
+  void createWorkers(const uint32_t concurrency);
```
mum4k, line 104: (optional) Is this a good time to document the method?

test/integration/test_integration_basics.py
```diff
@@ -654,6 +658,34 @@ def test_http_request_release_timing(http_test_server_fixture, qps_parameterizat
     assertCounterEqual(counters, "benchmark.http_2xx", (total_requests))
 
 
+def send_sigterm(p):
+  # Sleep for a while, under tsan the client needs a lot of time
```
mum4k, line 662: Can we add a Python style docstring, since the function is public.

test/integration/test_integration_basics.py
```diff
@@ -654,6 +658,34 @@ def test_http_request_release_timing(http_test_server_fixture, qps_parameterizat
     assertCounterEqual(counters, "benchmark.http_2xx", (total_requests))
 
 
+def send_sigterm(p):
```
mum4k, line 661: Can we choose a more readable name than "p".

test/integration/test_integration_basics.py
```diff
@@ -654,6 +658,34 @@ def test_http_request_release_timing(http_test_server_fixture, qps_parameterizat
...
+      http_test_server_fixture.nighthawk_client_path, "--concurrency", "2",
+      http_test_server_fixture.getTestServerRootUri(), "--duration", "1000", "--output-format",
+      "json"
+  ]
+  client_process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  Thread(target=(lambda: send_sigterm(client_process))).start()
+  stdout, stderr = client_process.communicate()
+  client_process.wait()
+  output = stdout.decode('utf-8')
+  assert (client_process.returncode == 0)
```
mum4k, line 682: Can we add a custom error message giving some context?

Or should this be a regular test assertion like below?

test/process_test.cc
```diff
@@ -32,12 +33,45 @@ class ProcessTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+
+  void runProcess(RunExpectation expectation, bool do_cancel = false,
+                  bool terminate_right_away = false) {
     ProcessPtr process = std::make_unique<ProcessImpl>(*options_, time_system_);
     OutputCollectorImpl collector(time_system_, *options_);
+    std::thread cancel_thread;
+    if (do_cancel) {
+      cancel_thread = std::thread([&process, terminate_right_away] {
+        if (!terminate_right_away) {
+          sleep(5);
```
mum4k, line 45: Should we add a comment explaining the sleep and the choice of 5?

source/client/remote_process_impl.cc
oschaaf: I filed https://github.com/envoyproxy/nighthawk/issues/380 (and updated `XXX` to `#380`) in the changes I'm about to push.

oschaaf:
@mum4k thanks for the review. I pushed 3198623 to address your feedback.

PR #368, opened 2020-06-17
Fix benchmark_http_client_test.cc and add test cases.

qqustc:
Fix benchmark_http_client_test.cc which has a bug in function testBasicFunctionality().

For example, if we have a test fixture with testBasicFunctionality(max_pending=0,connection_limit=15,amount_of_request=10),
then EXPECT_FALSE(client_->tryStartRequest(f)) and
EXPECT_EQ(max_pending == 0 ? 1 : max_pending, inflight_response_count) will fail.

For test fixture with
testBasicFunctionality(max_pending=2,connection_limit=3,amount_of_request=10),
EXPECT_EQ(max_pending == 0 ? 1 : max_pending, inflight_response_count) will fail.

Added 2 test cases to verify the expected behavior fixed by this change.

Tested:
Unit test.

### DIFF
--- a/test/benchmark_http_client_test.cc
+++ b/test/benchmark_http_client_test.cc
@@ -99,14 +99,15 @@ public:
       }
     }
 
-    // If max_pending is set to 0, and we queued up work, we shouldn't be able to add more.
-    if (max_pending == 0 && amount > 0) {
+    const uint64_t max_in_flight_allowed = max_pending + connection_limit;
+    // If amount_of_request >= max_in_flight_allowed, we are not able to add more request.
+    if (amount >= max_in_flight_allowed) {
       EXPECT_FALSE(client_->tryStartRequest(f));
     }
 
     dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
-    // If max pending is set > 0, we expect in_flight to be equal to max_pending.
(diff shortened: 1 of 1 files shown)


test/benchmark_http_client_test.cc
```diff
@@ -261,6 +275,7 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
   };
 
   // Note we we explicitly do not expect encodeData to be called.
+  EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(0);
```
oschaaf, line 278: (Optional) now that we actually have an explicit expectation in place instead of an implied one, we could probably do without the comment above this line?

qqustc: Done.

Thank you Otto for the review. 
Verified the change pass all unit tests except termination_predicate_test which is failing at master branch.

bazel test test:all

oschaaf: Great, thank you, this looks good to me. As for the termination_predicate_test failure, could you paste the details here or in a separate issue? Maybe it is flaking (I can't find an occurrence of that, neither on master).

qqustc: Thanks Otto. 
Opened an issue here
https://github.com/envoyproxy/nighthawk/issues/372

Issue #362, opened 2020-06-15
Coverage flakes


oschaaf:
Sometimes coverage dips far below the threshold (90.x% instead of 98.x%). When this happens it is always the python integration test runs not being considered when it comes to computing
the coverage. Some more information on this can be found [here](https://github.com/envoyproxy/nighthawk/pull/358#issue-433909675
). 

I'm filing this issue because it just happened on a CI run following a merge to master:
https://app.circleci.com/pipelines/github/envoyproxy/nighthawk/1640/workflows/6823a1ea-bb7d-4bce-a42c-ae6c6b38de5d/jobs/12923/steps

TLDR; we probably need to wait for clang-10 to get used in the docker image we use to test, and we source that image from Envoy.
If that doesn't resolve it, more looking into this is warranted.

oschaaf:
Cross pasting the tail of the CI output in case it gets wiped because of storage policies 30 days ahead:

```
Overall coverage rate:
  lines......: 90.7% (3134 of 3455 lines)
  functions..: 86.9% (344 of 396 functions)
+ COVERAGE_VALUE=90.7%
+ COVERAGE_VALUE=90.7
+ [[ -z '' ]]
+ '[' true == true ']'
+ COVERAGE_THRESHOLD=98.2
++ echo '90.7<98.2'
++ bc
+ COVERAGE_FAILED=1
+ test 1 -eq 1
+ echo Code coverage 90.7 is lower than limit of 98.2
Code coverage 90.7 is lower than limit of 98.2
+ exit 1
```

oschaaf:
Bad news, clang-10 didn't fix this, as https://github.com/envoyproxy/nighthawk/pull/367 just hit this again:
https://app.circleci.com/pipelines/github/envoyproxy/nighthawk/1714/workflows/12b06f69-d7a3-443d-9899-1b65de5859fb/jobs/13535/steps

The python integration tests failed to yield coverage.dat

```
 /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/bazel-out/k8-fastbuild/testlogs/test/output_transform_main_test/coverage.dat
//test:platform_util_test                                                PASSED in 19.7s
  /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/bazel-out/k8-fastbuild/testlogs/test/platform_util_test/coverage.dat
[...]

PR #376, opened 2020-06-19
Update Envoy to 12b3d2c2ffa582507e5d6dd34632b2b990f1b195

oschaaf:
Notable changes in this Update:

- Changes related to header construction
- CI moves from clang-9 to clang-10
- Changes related to satisfy the new clang-tidy version:
  - Fix superfluous initializers in the rate limiter code.
  - Suppress modernize-avoid-bind to get sequencer_test.cc to pass checks.
  (clang-tidy's auto generated suggested fix fails to build)
- Move to use typed configs in `AddFilter()` in the server filter tests. 
  This avoids asserts being triggered by the integration test framework
  as deprecated fields for which support was dropped get generated
  otherwise.
- Drop the coverage threshold by 0.1%

Hopefully addresses #362 by producing more reliable and accurate
coverage reports.

---

#### prerequisite before merge ####

- [x] Analyse why coverage dropped 0.1%
- [x] Check https://github.com/envoyproxy/envoy/pull/9663 to analyse
       impact on NH, if any. Possibly we need to move away some code 
       from deprecated cluster field usage. (filed https://github.com/envoyproxy/nighthawk/issues/377)


oschaaf:
The coverage threshold had to be lowered because of small changes, probably related to clang-10.

For example, compare [the new factories impl coverage](https://13416-180498819-gh.circle-artifacts.com/0/coverage/source/client/factories_impl.cc.gcov.html) with [the latest from master](https://13363-180498819-gh.circle-artifacts.com/0/coverage/source/client/factories_impl.cc.gcov.html).
Both count these as missing coverage, but only the new version reports that the `default` switch label isn't hit:

```
(new)
     102           0 :   default:
     103           0 :     NOT_REACHED_GCOVR_EXCL_LINE;
```

```
(old)
     102          75 :   default:
     103           0 :     NOT_REACHED_GCOVR_EXCL_LINE;
```

I think that makes sense, and that the new version is more accurate. So lowering the coverage threshold seems
the right thing to do.

PR #385, opened 2020-06-24
Ignore: trigger ci for 381

oschaaf:
Hack to trigger CI, as both /retest and CircleCI's interface don't help here.
Let's see if this forces the last commit from #381 to get retested.


oschaaf:
This has served its purpose, closing

PR #386, opened 2020-06-26
Update Envoy to 5c553261d14730af5a2c290c4b3687ae667a65db

### LINKED ISSUE #377: Move away from using hidden_envoy_deprecated_hosts
In https://github.com/envoyproxy/nighthawk/pull/376 I had to make some changes in tests, to avoid asserts being triggered by deprecations in https://github.com/envoyproxy/envoy/pull/9663.

This reminded me, Nighthawk is still using `hidden_envoy_deprecated_hosts` in some places.
It is probably a good idea to stop doing that.

oschaaf:
Notable changes:

- Fixes our Http1PoolImpl to work with the updated Envoy
  code.
- Tweaks the cluster setup: we specify at least an empty
  http2_protocol_options proto config. Without doing this
  passing --h2 to the nighthawk client would fail as caught
  by tests.
- Moves away from using hidden_envoy_deprecated_hosts.
  Side effect of looking into the bevavioral change listed
  above. This resolves #377. (strictly not needed for this
  update).

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -1,7 +1,7 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-ENVOY_COMMIT = "7f251daa2e488587bc7335f91faceed420f162c4"  # June 19th, 2020
-ENVOY_SHA = "bc14914c9639eca9c5b2b09117b812796067c17c7f1d0d3f7ad622d554323d94"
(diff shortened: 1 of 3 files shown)


oschaaf:
While getting this to work, I noticed that there's some duplication going on for the cluster setup code.
I filed https://github.com/envoyproxy/nighthawk/issues/387 for that.

Issue #82, opened 2019-05-06
Take ownership of the fault-filter delay configuration 


oschaaf:
We should add the delay as proto field, and synthesize a request header from that which the fault-filter will understand.

oschaaf:
This was closed via #390

PR #390, opened 2020-06-29
Test server dynamic delay

### LINKED ISSUE #389: Test server: controlling parallelism via dynamic delays
For simulating overloaded backends it would be useful to be able to have a means
to increase the injected delay as the number of requests behing handled grows.

There's the `downstream_rq_active` gauge that looks like it could be used as an
input to this. One way to achieve this would be to add configuration settings that
control this computation to allow one to have latency computed based on the delta
between the current and target value of a gauge [1]

The final delay would then look like `(target-value-of-the-gauge - current-value-of-the-gauge) * delay_step_us`.
To trigger the fault filter, the test server extension could synthesize an `x-envoy-fault-delay-request: <computed value>` header.

[1] example
```
          - name: test-server
            config:
              stats-based-delay: 
                  - gauge: "downstream_rq_active":
                     target-gauge-value: 100
                     delay-step-us: 1000
```

It's tempting to generalize this more by integrating the CEL expression engine, but it looks like that is an order 
of magnitude more work (I may be wrong about that, but that is my gut feeling after skimming some code).

oschaaf:
Add an extension which implements dynamic delay injection based on the number
of concurrent requests being handled, as well as static delays.

Fixes #389 and also #82. We may consider removing the direct fault filter configs
from our docs as well as emit a log warning when detecting usage to deprecate.
It would be nice if we could add the fault filter extension in code instead of imposing
configuration of that upon the end user, making that an implementation detail.

The following shows a per-request config example. 
This header-provided configuration gets merged with any provided static configuration to
determine a final configuration for each request:

```
curl -H "x-nighthawk-test-server-config:{static_delay:\"1s\"}"  -vv 127.0.0.1:10000
```

Full yaml config for the http_filters section:

```yaml
                http_filters:
                  - name: dynamic-delay
                    config:
                      concurrency_based_delay:
                        minimal_delay: 0.05s
                        concurrency_delay_factor: 0.010s                  
                  - name: test-server
                    config:
                      response_body_size: 10
                      response_headers:
                        - { header: { key: "x-nh", value: "1" } }
```

-------

State: first stab, manually verified to work. needs
- [x] discussion 
- [x] unit tests
- [x] a bit of cleanup and 
- [x] end to end test, to prove the number of active filter instances is tracked correctly and interaction with the fault filter works as intended (filed #393 to track).
- [x] docs.


oschaaf:
@mum4k @jmarantz opening this up as a draft to allow for early review. (See the issue description for state), but prevent merging (or the impression of this being in that shape yet).

I would like to invite you to assess the high level approach as well as the algorithm/parameterization to achieve the goal we're after here: being able to induce a sustained/fixed level of parallelism on the downstream side (e.g. a proxy running in front of the test server). 

While working on this PR I was mostly focussed on wiring in the overall mechanics of the desired type of functionality,
but not so much the algorithm itself.

api/server/response_options.proto
```diff
@@ -5,6 +5,15 @@ package nighthawk.server;
 import "google/protobuf/wrappers.proto";
 import "validate/validate.proto";
 import "envoy/api/v2/core/base.proto";
+import "google/protobuf/duration.proto";
+
+message GaugeBasedDelay {
```
jmarantz, line 10: I feel like this name focuses on the mechanism of how it's implemented, and it would be more intuitive to instead focus on the value it provides. Something more like: `QueuingDelay` if we want to be very concise, or `DelayComputedFromQueueSize` for a more verbose version.

On the other hand I see you are generically allowing computing a delay from any gauge, and I'm curious if you've thought of a use-case other than queued requests, for this generality to be helpful?

source/server/http_dynamic_delay_filter.cc
```diff
@@ -0,0 +1,87 @@
...
+  }
+  absl::optional<int64_t> delay;
+  if (base_config_.has_static_delay()) {
+    delay = Envoy::Protobuf::util::TimeUtil::DurationToMilliseconds(base_config_.static_delay());
+  } else if (base_config_.has_gauge_based_delay()) {
+    auto& gauge_based_delay = base_config_.gauge_based_delay();
+    const std::string gauge_name = gauge_based_delay.gauge_name();
+    const uint64_t gauge_value =
+        config_->statsScope()
+            .gaugeFromString(gauge_name, Envoy::Stats::Gauge::ImportMode::Uninitialized)
```
jmarantz, line 49: can you capture the Gauge& in the config so you don't have to do the name-based lookup on every request?

Also, I'd use Scope::findGauge() instead so you can report an error during config if the gauge doesn't exist.

Another thing that's concerning here is due to dynamic vs symbolic stat names. We don't really have a robust mechanism to look up a stat by string in prod now, due to dynamic vs symbolic stat name segments. To do the lookup we need to either know what segments of the stat name are dynamic vs symbolic, or we need to iterate through all stat names, serialize them to strings, and do string compares.

This is likely to work the way you have it because most (but not all) stat names are purely symbolic. But there is no guarantee.

In tests we keep a string->stat map (very expensive) to make this easy. In prod I think we will have to add:

Scope::slowlyFind(Counter|Gauge|Histogram|TextReadout)ByName(absl::string_view name) which takes time O(num stats). We would definitely want to only do that during config time.

What's worse, some of the stats may exist at the time filter is existed; not sure about that. Might have to lazy-init with an atomic protector or something, maybe using AtomicPtr : https://github.com/envoyproxy/envoy/blob/7ea1f24bd53522408d9bc55316dff9ed1701bc73/source/common/common/thread.h#L156

api/server/response_options.proto
```diff
@@ -5,6 +5,15 @@ package nighthawk.server;
...
+  // Name of the gauge providing the target value.
+  string gauge_name = 1;
+  int64 gauge_target_value = 2;
+  // The delay is computed as (guauge_target_value - gauge_value) * delta_delay;
```
jc1e100, line 14: Considering a use case where we want to inject latency proportional to the in-flight request count, how would we implement that using this config?

It looks like this models a gauge that starts as a positive value and decreases toward zero, but a gauge like downstream_rq_active does the opposite. If negative values for delta_delay or gauge_target_value are supported, could you state that explicitly?

Also, I think we need the first request to incur some latency, otherwise it will take pretty severe load before the in-flight count grows at all. This tells me that the model needs to allow an offset value - i.e. a baseline latency when the gauge is at the target.

jc1e100: To put it another way: right now I think you're proposing an equation like y = max(0, m * (x_0 - x)) with unspecified constraints on the RHS variables. I think it would be easier to reason about y = max(0, m * x + b) where m and b can be any real values.

oschaaf: @jc1e100 Currently the delay linearly shrinks to zero as the gauge approaches the target.
This is expressed as follows in the code:

```c++
    // Substracts one from the gauge value to avoid inclusion of the the current request
    delay = (gauge_target_value - (gauge_value - 1)) * delay_delta;
    if (delay <= 0) {
      delay = absl::nullopt;
    }
```

What went wrong here is that I misunderstood, and thought we had been aiming to get a steady number of in-flight downstream requests.

I'll make changes based on your input as well as Josh his earlier comment to simplify this.

oschaaf:
@jmarantz @jc1e100 I pushed changes to partially address comments in 277c7860011bf21a198653c11311b696d00f7e62, and updated the PR description.
I couldn't directly stick an atomic into the shared filter config, as that implicitly deletes the copy constructor, and prevents the build from succeeding with that. So it is static at this time. Let me know if this looks better.

api/server/response_options.proto
oschaaf: The delay now looks like below (any negative computed delays are ignored / not propagated):

```
    auto& concurrency = base_config_.concurrency_based_delay();
    const uint64_t current_value = config_->approximateInstances();
    delay = Envoy::Protobuf::util::TimeUtil::DurationToMilliseconds(
        concurrency.minimal_delay() + (current_value * concurrency.concurrency_delay_factor()));
```

api/server/response_options.proto
```diff
@@ -14,4 +24,11 @@ message ResponseOptions {
   uint32 response_body_size = 2 [(validate.rules).uint32 = {lte: 4194304}];
   // If true, then echo request headers in the response body.
   bool echo_request_headers = 3;
+
+  oneof oneof_delay_options {
```
jc1e100, line 28: Is there a universal convention for whether file-based config settings have precedence over request header settings or vice versa? If not, could you document those here?

oschaaf: good point, let me document that. The test-server extension assumes the static/file-based config is the default. The request-level config will get merged, and override. As these extensions share the same header / proto, I think they should share the same convention. SG?

api/server/response_options.proto
```diff
@@ -5,6 +5,16 @@ package nighthawk.server;
...
 import "envoy/api/v2/core/base.proto";
+import "google/protobuf/duration.proto";
+
+message ConcurrencyBasedDelay {
+  // Minimal delay to add to replies.
+  google.protobuf.Duration minimal_delay = 1 [(validate.rules).duration.gte.nanos = 0];
+  // Factor to use when adding latency as concurrency increases.
+  google.protobuf.Duration concurrency_delay_factor = 2 [(validate.rules).duration.gte.nanos = 0];
+  // Multiplier.
+  uint64 concurrency_delay_multiplier = 3;
```
jc1e100, line 16: This appears to be unused.

oschaaf: good catch, that needs to go indeed. FYI, I'm currently working to add tests and docs, to get this PR in shape

api/server/response_options.proto
jc1e100: SGTM, thanks.

source/server/http_dynamic_delay_filter.cc
```diff
@@ -0,0 +1,77 @@
...
+#include "envoy/server/filter_config.h"
+
+#include "server/common.h"
+
+#include "absl/strings/str_cat.h"
+
+namespace Nighthawk {
+namespace Server {
+
+std::atomic<uint64_t> HttpDynamicDelayDecoderFilterConfig::instances_(0);
```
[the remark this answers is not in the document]
oschaaf: I couldn't directly stick an atomic into the shared filter config, as doing so implicitly deletes the copy constructor, which breaks the build. I'll lazy init it.

source/server/http_dynamic_delay_filter.cc
```diff
@@ -0,0 +1,77 @@
...
+        static_cast<Envoy::Http::Code>(500),
+        fmt::format("dynamic-delay didn't understand the request: {}", *error_message_), nullptr,
+        absl::nullopt, "");
+    return Envoy::Http::FilterHeadersStatus::StopIteration;
+  }
+  absl::optional<int64_t> delay;
+  if (base_config_.has_static_delay()) {
+    delay = Envoy::Protobuf::util::TimeUtil::DurationToMilliseconds(base_config_.static_delay());
+  } else if (base_config_.has_concurrency_based_delay()) {
+    auto& concurrency = base_config_.concurrency_based_delay();
```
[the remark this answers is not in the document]
oschaaf: sure; I have extracted this into a static helper method for making testing easier, I'll make sure naming is clarified.

oschaaf:
This is ready for another pass; I punted a bug and a missing functional test to #392 and #393.

oschaaf:
Linking the [CI coverage report](https://13723-180498819-gh.circle-artifacts.com/0/coverage/source/server/index.html) for convenience, this seems ok to me (combined with the issue that tracks that we need covering actual end-to-end functionality).

mum4k:

source/server/http_dynamic_delay_filter.h
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * @return uint64_t the approximate number of active instances.
+   */
+  uint64_t approximateInstances() const { return instances(); }
+
+private:
+  const nighthawk::server::ResponseOptions server_config_;
+  static std::atomic<uint64_t>& instances() {
+    // We lazy-init the atomic to avoid static initialization / a fiasco.
+    static std::atomic<uint64_t> a(0);
```
jmarantz, line 44: MUTABLE_CONSTRUCT_ON_FIRST_USE(std::atomic<uint64>, 0)

What you have is still a fiasco on shutdown.

test/server/http_dynamic_delay_filter_integration_test.cc
```diff
@@ -0,0 +1,186 @@
...
+#include "api/server/response_options.pb.h"
+
+#include "server/common.h"
+#include "server/http_dynamic_delay_filter.h"
+
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+using namespace testing;
```
jmarantz, line 14: I bet you can remove this. If you do need to use something from `testing` and you don't want to qualify it in place, you can put the more specific using directive(s) here instead.

e.g. `using testing::HasSubstr;`

source/server/http_dynamic_delay_filter.h
oschaaf: Ugh great catch I’ll amend

api/server/response_options.proto
```diff
@@ -5,13 +5,31 @@ package nighthawk.server;
...
 import "validate/validate.proto";
 import "envoy/api/v2/core/base.proto";
+import "google/protobuf/duration.proto";
 
-// Options that control the test server response.
+message ConcurrencyBasedDelay {
+  // Minimal delay to add to replies.
+  google.protobuf.Duration minimal_delay = 1 [(validate.rules).duration.gte.nanos = 0];
+  // Factor to use when adding latency as concurrency increases.
+  google.protobuf.Duration concurrency_delay_factor = 2 [(validate.rules).duration.gte.nanos = 0];
```
htuch, line 14: This seems to be a linear model for latency. I wonder if ultimately you want something more like a network queuing model where you see a hyperbolic delay as requests back up significantly. OTOH, you also have some in-built delay from the fact that the test server itself will have some queuing effects.

What's the ultimate goal here? To model overload when the test server isn't fully in overload?

source/server/http_dynamic_delay_filter.h
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * @return uint64_t the approximate number of active instances.
+   */
+  uint64_t approximateInstances() const { return instances(); }
+
+private:
+  const nighthawk::server::ResponseOptions server_config_;
+  static std::atomic<uint64_t>& instances() {
+    // We lazy-init the atomic to avoid static initialization / a fiasco.
+    MUTABLE_CONSTRUCT_ON_FIRST_USE(std::atomic<uint64_t>, 0); // NOLINT
```
htuch, line 44: As an alternative to this, I think you could examine the stats gauge maintained by Envoy tracking the number of currently active requests.

source/server/http_dynamic_delay_filter.cc
```diff
@@ -0,0 +1,82 @@
...
+        base_config_.concurrency_based_delay();
+    const uint64_t current_value = config_->approximateInstances();
+    delay_ms = computeDelayMilliseconds(current_value, concurrency_config.minimal_delay(),
+                                        concurrency_config.concurrency_delay_factor());
+  }
+  if (delay_ms.has_value() && delay_ms > 0) {
+    // Emit header to communicate the delay we desire to the fault filter extension.
+    const Envoy::Http::LowerCaseString key("x-envoy-fault-delay-request");
+    headers.setCopy(key, absl::StrCat(*delay_ms));
+  }
```
dubious90, line 62: I think it'd help the readability here if you abstracted this section into its own function, as this is the part of the function that is actually setting the delays, rather than just decoding headers.

source/server/http_dynamic_delay_filter.h
```diff
@@ -0,0 +1,98 @@
...
+class HttpDynamicDelayDecoderFilter : public Envoy::Http::StreamDecoderFilter {
+public:
+  HttpDynamicDelayDecoderFilter(HttpDynamicDelayDecoderFilterConfigSharedPtr);
+  ~HttpDynamicDelayDecoderFilter() override;
+
+  // Http::StreamFilterBase
+  void onDestroy() override;
+
+  // Http::StreamDecoderFilter
+  Envoy::Http::FilterHeadersStatus decodeHeaders(Envoy::Http::RequestHeaderMap&, bool) override;
```
dubious90, line 68: These appear to be overriding, and then just passing a CONTINUE, almost as if they're a no-op.  Could you add a comment explaining why they are doing that?

source/server/http_dynamic_delay_filter.h
```diff
@@ -0,0 +1,98 @@
...
+  Envoy::Http::FilterDataStatus decodeData(Envoy::Buffer::Instance&, bool) override;
+  Envoy::Http::FilterTrailersStatus decodeTrailers(Envoy::Http::RequestTrailerMap&) override;
+  void setDecoderFilterCallbacks(Envoy::Http::StreamDecoderFilterCallbacks&) override;
+
+  /**
+   * Compute the delay in milliseconds based on the parameters provided.
+   *
+   * @param concurrency indicating the number of active requests.
+   * @param minimal_delay gets unconditoinally included in the return value.
```
dubious90, line 77: nit: unconditionally

source/server/http_dynamic_delay_filter_config.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "api/server/response_options.pb.h"
+#include "api/server/response_options.pb.validate.h"
+
+#include "server/http_dynamic_delay_filter.h"
+
+namespace Nighthawk {
+namespace Server {
+namespace Configuration {
+
+class HttpDynamicDelayDecoderFilterConfig
```
dubious90, line 16: I'm confused.  Why are we defining a class with public methods in a .cc file, rather than a .h file?

source/server/http_dynamic_delay_filter.h
oschaaf: For now this is implemented like with an atomic, for context see  https://github.com/envoyproxy/nighthawk/pull/390/files/3ec87c082f960576a489a696dbb2911877271653#diff-3b51180e80dfb0f4e9f97a9596c93bc1 and https://github.com/envoyproxy/envoy/issues/11799.

api/server/response_options.proto
oschaaf: The goal here is to model response times associated to an overloaded server, without the server actually being overloaded.

What you are saying here sounds like something me and @jmarantz discussed on Slack; we figured the current linear model would suffice for the concrete use-case backing this, but I think what you're proposing would be a more accurate simulation, and it reflects something that was discussed on Slack. 

Maybe we should incorporate the word "linear" in the name here, that way if someone needs a hyperbolic delay there's no ambiguity, and it's clear what happens here. And it would be very easy to add it.

source/server/http_dynamic_delay_filter.h
oschaaf: We're implementing a pure virtual interface [1], so there's no real choice here, we must override / implement.
Since we don't have anything useful to do in these implementations, we indicate we'd like to continue to the state machine driving this. Indeed a no-op, but I'm also not sure I can add a real helpful comment about it?

[1] https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/filter.h#L515

source/server/http_dynamic_delay_filter_config.cc
oschaaf: I got confused as well when looking closer at this; but more because we had two classes named `HttpDynamicDelayDecoderFilterConfig` in two namespaces. 
Renamed this one to `HttpDynamicDelayDecoderFilterConfigFactory` which is more accurate.

As for your question on the class with public methods in the `.cc` file: when done like that, it allows other code within the same `.cc` to consume its public methods. But not publishing it in the .h prevents anything else from doing so.

source/server/common.h
```diff
@@ -0,0 +1,50 @@
...
+} // namespace TestServer
+
+class Utility {
+public:
+  /**
+   * Merges a json string containing configuration into a ResponseOptions instance.
+   *
+   * @param json Json-formatted seralization of ResponseOptions to merge into the configuration.
+   * @param config The target that the json string should be merged into.
+   * @param error_message Will contain an error message iff an error occurred.
```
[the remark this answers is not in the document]
oschaaf: The problem described there is:

```
The first option probably doesn’t do what you want. 
If someone passes a Foo into MyFunc, the Foo will be copied by value into the absl::optional<Foo>, 
which will then be passed by reference into the function. 
If your goal was to avoid copying the Foo, you haven’t.
```

My intent here was that that we actually do copy by value, and that the optional takes ownership of the string we return.
This in turn makes it easy to manage the lifetime of the (dynamic) error string synthesized in the function itself.
If was hoping that by using the optional, it would also be intuitive that a caller may or may not get an error message.
I'm hoping that the error flow, where this copying will occur, will be a low-frequency event, and as such not hurt performance in the hot path.

With that as some more context, does that change how you feel about it?

source/server/README.md
```diff
@@ -68,15 +75,15 @@ admin:
 
 ## Response Options config
 
-The ResponseOptions proto can be used in the test-server filter config or passed in `x-nighthawk-test-server-config``
-request header.
+The [ResponseOptions proto](/api/server/response_options.proto) can be used in the test-server
+and dynamic-delay filter config or passed in `x-nighthawk-test-server-config`` request header.
```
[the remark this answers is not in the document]
oschaaf: This is a real good point, and I'm still mulling over how to best improve this. 

Some context: this is an artifact of my first stab at implementing this, where I used a single extension and injected it in two places. For technical reasons, that failed, I ran into trouble with what seemed Envoy getting confused about it - or I may have gotten confused myself as it complicated the extension more then I had hoped for. As we rely on the stock fault filter extension today for actually inducing latency, we need to be able to insert ourselves into two places in the filter pipeline: before and after the fault filter. Imho, ideally, for the end user, a response can be specified as a single entity, and how fragments of that apply to various parts within the test server (and its extensions) are implementation details. Right now we're sitting somewhere in the middle: request unifies it, but the static/file based configuration splits it, as more or less imposed by Envoy. 
I rewrote the docs part for a bit [1], but I'd love to see if we can improve the static/file-based configuration approach; yet I need to think some more about how (suggestions welcome!). Let me get back to this after thinking about it some more.

[1] 
```markdown
## Response Options config

The [ResponseOptions proto](/api/server/response_options.proto) is shared by
the `Test Server` and `Dynamic Delay` filter extensions. Each filter will
interpret the parts that are relevant to it. This allows specifying a what
a response should look like in a single message, which can be done at request
time via the optional `x-nighthawk-test-server-config` request-header.

### Test Server

- `response_body_size` - number of 'a' characters repeated in the response body.
- `response_headers` - list of headers to add to response. If `append` is set to
  `true`, then the header is appended.
- `echo_request_headers` - if set to `true`, then append the dump of request headers to the response
  body.

The response options above could be used to test and debug proxy or server configuration, for example, to verify request headers that are added by intermediate proxy:

<snip example req/resp>

This example shows that intermediate proxy has added `x-forwarded-proto` and
`x-forwarded-for` request headers.

### Dynamic Delay

The Dynamic Delay interprets the `oneof_delay_options` part in the [ResponseOptions proto](/api/server/response_options.proto). If specified, it can be used to:

- Configure a static delay
- Configure delays which linearly increase as the number of active requests grows, representing a simplified model of an overloaded server.
```

api/server/response_options.proto
oschaaf: Renamed proto to mention 'linear'  to disambiguate with potential future additions/enhancements, and made it clear that we're running a simplified model in the docs in c40e9ce65c82b62a6f9eab47ac31327a7605a6f4

source/server/README.md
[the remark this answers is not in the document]
oschaaf: Looking at the fault filter, the actual code that handles the delay injecting can be minimally expressed as:

```c++
// Set up delay
Http::FilterHeadersStatus FaultFilter::decodeHeaders(Http::RequestHeaderMap& headers, bool) {
  absl::optional<std::chrono::milliseconds> duration = delayDuration(headers);
  if (duration.has_value()) {
    delay_timer_ = decoder_callbacks_->dispatcher().createTimer(
        [this, &headers]() -> void { postDelayInjection(headers); });
    delay_timer_->enableTimer(duration.value(), &decoder_callbacks_->scope());
    return Http::FilterHeadersStatus::StopIteration;
  }
}

void FaultFilter::postDelayInjection(const Http::RequestHeaderMap& request_headers) {
  resetTimerState();
  decoder_callbacks_->continueDecoding();
}

```

We could consider taking ownership of that by copying it. 
Potential advantages: 

- we can inject into a single place
- easier setup for the end-user, single config/extension
- Have a relatively easy and surefire path to generically map/route qs params and/or headers to proto config fields
- It may also allow us to switch to sub-ms granularity with a small tweak

Possible disadvantages:

- We own that code
- Need to add unit tests to support it, can no longer rely on Envoy for that
- Fault filter may still need to be pulled in for other functionality (fault injection)
- Looking at the code, the fault filter is more flexible in configuration, in that it supports route-level
configuration and maybe more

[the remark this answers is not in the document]
oschaaf: Another idea: could we use the fault filter as a base and derive from it?

api/server/response_options.proto
htuch: One other question while we're here. If the goal is to model an overloaded server without being overloaded, why make use of actual concurrent requests at all? Couldn't you just inject a fake number of concurrent requests and use that to compute delay?

source/server/README.md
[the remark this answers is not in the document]
htuch: I think if the code is as simple as you point to, there's no good reason to be doing a lot of sharing with the fault injection filter. If you want to more generally extend it, then you could consider this pattern. Keep in mind that subclassing may expose your code to fragility, since there is no hard contract on the interface in the Envoy repo.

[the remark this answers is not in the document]
oschaaf: I looked some more into this, and the fault filter has evolved for quite a bit since I last looked at it, and learned new cool tricks.

Also, I made a small trip to explore taking the fault filter as a base class;  [Here's a diff](https://github.com/oschaaf/nighthawk/compare/test-server-dynamic-delay...oschaaf:explore-fault-filter-as-a-base?expand=1), but a high level description of the code-level contact surface would be:

- The constructor -- we take NHs own `ResponseOptions` proto, and translate that into a minimal configuration for the fault filter base class which it takes in its constructor to set it up like the way we need it. 
- We override decodeHeaders() and synthesize the request header that the fault filter understands to trigger the actual delay and forward the call to the base via  `Envoy::Extensions::HttpFilters::Fault::FaultFilter::decodeHeaders(headers, end_stream)` when our own override is done, which is the only difference with what we have right now.
- We override `HttpDynamicDelayDecoderFilter::setDecoderFilterCallbacks()` to grab a copy of the callbacks which we'll only use in case we didn't understand the per-request configuration to short circuit and send out an error response. We forward the call to the base, obviously, after grabbing the copy. This would be relatively fragile, as the Fault filter may not anticipate us directly replying. We could instead not do that and log an error about bad configs to make it more robust. We no longer need to override `setDecoderFilterCallbacks()`.

Challenges:

- Includes aren't routed through external/envoy (see code level todo in the diff above), requires an upstream change in the build system to make that happen (probably similar to what we did earlier).
- Need to check the stats_prefix_ argument that the fault filter takes & verify counter behaviour is allright.

I feel similar in that imho it depends on how we envision the test server functionality to expand in the future. For example, one of the fault filters new tricks (to me at least) is throttling streams; I wouldn't be surprised if similar functionality is going to be requested at some point for Nighthawk's test server. If we want to extend / modify behavior for that as well, using the fault filter as a base might be convenient -- iff it turns out fragility isn't too bad. Otoh, if the fault filter moves in a direction that would enforce us to re-implement some things, it's good take we do take ownership of the proto api, as that limits impact to the end-user: The fault filter as a base becomes an implementation detail.

@dubious90 / @mum4k do you have any thoughts on this?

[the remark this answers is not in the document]
oschaaf: also cc @jmarantz ^^

api/server/response_options.proto
oschaaf: if one is testing behaviour of a load balancer running in front of multiple of these servers that are simulating overload, 
having the delay based on actual concurrency may help verify the load balancer's balancing strategy.

source/server/README.md
```diff
@@ -75,21 +75,23 @@ admin:
 
 ## Response Options config
 
-The [ResponseOptions proto](/api/server/response_options.proto) can be used in the test-server
-and dynamic-delay filter config or passed in `x-nighthawk-test-server-config`` request header.
+The [ResponseOptions proto](/api/server/response_options.proto) is shared by
+the `Test Server` and `Dynamic Delay` filter extensions. Each filter will
+interpret the parts that are relevant to it. This allows specifying a what
```
dubious90, line 80: nit: specifying what, remove extra a

source/server/http_dynamic_delay_filter_config.cc
dubious90: Ah, okay.  Would it be okay if we moved it within an anonymous namespace?  That seems to be the google convention for indicating what you're talking about.  If there's a reason not to do that, totally understand.

source/server/http_dynamic_delay_filter.h
dubious90: understood.  thanks for explaining.  i think what i'm looking for is a just a quick comment to help readers know that they don't have to spend their time understanding what these are, but if you don't think that's necessary, i'll follow your lead.

source/server/README.md
oschaaf: done

source/server/http_dynamic_delay_filter_config.cc
oschaaf: Ooh, yeah, good catch. Done.

source/server/http_dynamic_delay_filter.h
oschaaf: Done;

source/server/http_dynamic_delay_filter.cc
oschaaf: Done, let me know if that looks like what you had in mind (see 8ee78db003c13b96290837815803a75322bd3a38)

source/server/common.h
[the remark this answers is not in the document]
dubious90: My understanding is that nighthawk follows envoy style, and envoy largely follows the google-published C++ style guide.
See https://github.com/envoyproxy/nighthawk/pull/390/files#r450301523.

We try to save absl::optional for use in optional input parameters, and pointers are used for output parameters.

(In this case, for me reading the code, I was confused why the code was accepting an optional error message and then immediately overriding it.  I had to jump to this comment to understand it was an output error message.)

source/server/README.md
```diff
@@ -68,15 +75,15 @@ admin:
...
 
 The following parameters are available:
 
-* `response_body_size` - number of 'a' characters repeated in the response body.
-* `response_headers` - list of headers to add to response. If `append` is set to
+- `response_body_size` - number of 'a' characters repeated in the response body.
+- `response_headers` - list of headers to add to response. If `append` is set to
   `true`, then the header is appended.
-* `echo_request_headers` - if set to `true`, then append the dump of request headers to the response
+- `echo_request_headers` - if set to `true`, then append the dump of request headers to the response
```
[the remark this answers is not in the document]
dubious90: I think you missed this, unless this is dependent on the result of the larger thread about the use of the fault filter.

source/server/README.md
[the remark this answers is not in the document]
dubious90: My understanding of this is limited, so I'm going to defer to you, Harvey, and Jakub on what makes sense here.

dubious90:
LGTM, after the two quick comments are resolved.  @mum4k to approve.

source/server/http_dynamic_delay_filter.cc
dubious90: I was actually thinking grab lines 49 - 62 here, so including the part where it determines what the delay is.  But I defer to you on what makes the most sense.

oschaaf: Done, let me know if that looks better.

source/server/README.md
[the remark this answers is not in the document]
mum4k: Thank you for the detailed analysis here @oschaaf, this is very useful. I would suggest we err on the side of minimizing maintenance cost and code churn and implement this without the inheritance, keeping ourselves to the external/envoy imports.

Correct me if I am wrong, but I believe we don't have use cases for the test server that would require us to modify the behavior of the fault filter. Even if we find some, it might be more beneficial to push such modifications upstream by adding options or improving the fault filter itself. Once we find a use case that is very specific to Nighthawk, we can reopen this discussion.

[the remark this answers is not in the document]
oschaaf: So I’d be happy to back out the basing-off of the fault filter that I merged earlier today, and switch this  to copy the relevant code so we own it, see https://github.com/envoyproxy/nighthawk/pull/390#discussion_r450503227

That takes the fault filter out of the equation, and allows us to provide our own configuration,  but also has a side effect: while right now  we just need to prove we wire up correctly to the fault filter, and can  rely on Envoy for testing that that does what it says on the tin with respect to delays , now we’d also need to test that ourselves to do it right.

IMHO upstreaming might not *always* be an option, as goals are not guaranteed to align; which is why we are here I guess? :-) so maybe copying the delay code is a middle ground then?

source/server/common.h
```diff
@@ -0,0 +1,50 @@
...
+class HeaderNameValues {
+public:
+  const Envoy::Http::LowerCaseString TestServerConfig{"x-nighthawk-test-server-config"};
+};
+
+using HeaderNames = Envoy::ConstSingleton<HeaderNameValues>;
+
+} // namespace TestServer
+
+class Utility {
```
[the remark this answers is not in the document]
oschaaf: Done in 9be53343674899bc8a2e962ebbee192a96ff31f9, let me know if that looks better.

source/server/common.h
```diff
@@ -0,0 +1,50 @@
...
+#include "external/envoy/source/common/singleton/const_singleton.h"
+
+#include "api/server/response_options.pb.h"
+
+namespace Nighthawk {
+namespace Server {
+
+namespace TestServer {
+
+class HeaderNameValues {
```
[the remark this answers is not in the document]
oschaaf: Like for above, see 9be53343674899bc8a2e962ebbee192a96ff31f9. However, I kept the Envoy construct, because this is shared/reused across the two extensions, and having the `LowerCaseString` is convenient -- that is the form that consumers are probably going to end up needing when they want access to this.

source/server/common.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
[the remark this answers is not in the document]
oschaaf: There are pre-existing unit tests for this. For example, see https://github.com/envoyproxy/nighthawk/blob/06a6a027adeb4970c853ad2f646acf28027e3fb6/test/server/http_test_server_filter_integration_test.cc#L303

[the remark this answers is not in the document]
oschaaf: Yeah, you're right, sorry about that. Will extract these mechanical refactors out into a separate PR next time.

oschaaf:
@mum4k I pushed a enhancements based on your feedback in 9be5334 and 
ad5e34c. Let me know of those two commits sufficiently address it.

Issue #115, opened 2019-08-15
Evaluate the possibility of switching to libcirclhist


htuch:
Envoy uses https://github.com/circonus-labs/libcircllhist, the original design discussion was in https://github.com/envoyproxy/envoy/issues/1965. 

In the interest of minimizing total dependencies, we should answer "is there a strong reason to be using HDR histogram vs. this log-linear library?". This library may have already been considered, if we can articulate our rationale in this issue we can have some design history.

oschaaf:
So reading up, what may be for hdrhistogram is improved accuracy. But seeing is believing, I feel it's worth doing some tests for comparison. On a side-note: one fairly recent observation is that the merge we with perform after the test run to compute the aggregated latencies with HdrHistogram massively slows down when using lots of workers (80 on a 40 cpu (HT-enabled) system with 128GB of ram).  I didn't look into that, but it seems very reproducible, so I am  making note of it here so I don't forget when getting back to this.

htuch:
Also worth evaluating is the project health of the various options, some might have more active communities and contributions than others.

HeinrichHartmann:
Hi @htuch, the qualitiative differences to HDR Histograms are as follows:

- circllhist mandates all bin and range choices. This means two circllhists from different sources are always mergable. But also you have to live with 2 digit decimal precision (usually not a big issue).

- circllhist is generally more optimized (since it does less). In particular the C implementation is highly performant when it comes to insertion (ns-range) and merge (ns..us-range) operations.

- HDR-Histogram uses base-2 arithmetic behind the scenes, circllhist uses base 10. So you get precise counts at "human readable" float numbers like 0.24, 120, 1800, as opposed to `0.09375=0.00011 base-2`.

I ran some superficial benchmarks for a recent conference talk:
https://nbviewer.jupyter.org/github/HeinrichHartmann/Statistics-for-Engineers/blob/master/2019-10-02-SRECon-Dublin/2.4%20Mergeable%20Percentile%20Aggregation%20Comparison.ipynb

Here is the summary table:
```
-------------  -----------------  ----------------------  ---------------------  --------------  -------------  ------------------
method         p90                err                     rel-err-%              latency insert  latency merge  latency percentile
-              35.77457865293068  0                       0                      0               0              0
HDR Histogram  35.775             0.00042134706932017707  0.0011777834573759823  3.59            0.00341        0.00332
[...]

PR #391, opened 2020-07-01
Add new NH statistic class CircllhistStatistic and SinkableCircllhist…

qqustc:
Add new NH statistic class CircllhistStatistic and SinkableCircllhistStatistic.

CircllhistStatistic uses Circllhist under the hood to compute statistics where Circllhist is used in the implementation of Envoy Histograms. Compared to HdrHistogram Circllhist trades precision for fast performance in merge and insertion according to #115.

SinkableCircllhistStatistic wraps the Envoy::Stats::Histogram interface and is used to flush histogram value to downstream Envoy stats Sinks.

Linked Issues: #344

Testing: unit tests

### DIFF
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -253,7 +253,7 @@ nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain do
   if (count() == 0)
     return proto;
 
-  std::vector<double> quantiles{0, 0.25, 0.5, 0.75, 0.90, 0.95, 0.99, 0.995, 0.999, 1};
+  std::vector<double> quantiles{0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.775, 0.8, 0.825, 0.85, 0.875, 0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1};
   std::vector<double> computed_quantiles(quantiles.size(), 0.0);
   hist_approx_quantile(histogram_, quantiles.data(), quantiles.size(), computed_quantiles.data());
   for (size_t i = 0; i < quantiles.size(); i++) {
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -256,8 +256,8 @@ double CircllhistStatistic::pstdev() const {
 StatisticPtr CircllhistStatistic::combine(const Statistic& statistic) const {
   auto combined = std::make_unique<CircllhistStatistic>();
   const auto& stat = dynamic_cast<const CircllhistStatistic&>(statistic);
-  hist_accumulate(combined->histogram_, &this->histogram_, /*histogram_count=*/1);
-  hist_accumulate(combined->histogram_, &stat.histogram_, /*histogram_count=*/1);
+  hist_accumulate(combined->histogram_, &this->histogram_, /*cnt=*/1);
+  hist_accumulate(combined->histogram_, &stat.histogram_, /*cnt=*/1);
 
   combined->min_ = std::min(this->min(), stat.min());
   combined->max_ = std::max(this->max(), stat.max());
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -253,7 +253,9 @@ nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain do
   if (count() == 0)
     return proto;
 
-  std::vector<double> quantiles{0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.55, 0.6, 0.65, 0.7, 0.75, 0.775, 0.8, 0.825, 0.85, 0.875, 0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1};
+  std::vector<double> quantiles{0,    0.1,   0.2,  0.3,   0.4,  0.5,   0.55,  0.6,
+                                0.65, 0.7,   0.75, 0.775, 0.8,  0.825, 0.85,  0.875,
+                                0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1};
   std::vector<double> computed_quantiles(quantiles.size(), 0.0);
   hist_approx_quantile(histogram_, quantiles.data(), quantiles.size(), computed_quantiles.data());
   for (size_t i = 0; i < quantiles.size(); i++) {
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -236,4 +236,39 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
   return proto;
 }
 
+StatisticPtr CircllhistStatistic::combine(const Statistic& statistic) const {
+  auto combined = std::make_unique<CircllhistStatistic>();
+  const auto& b = dynamic_cast<const CircllhistStatistic&>(statistic);
+  hist_accumulate(combined->histogram_, &this->histogram_, 1);
+  hist_accumulate(combined->histogram_, &b.histogram_, 1);
+
+  combined->min_ = std::min(this->min(), b.min());
+  combined->max_ = std::max(this->max(), b.max());
+  combined->count_ = this->count() + b.count();
+  return combined;
+}
+
+nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
+  nighthawk::client::Statistic proto = StatisticImpl::toProto(domain);
+  if (count() == 0)
+    return proto;
+
+  std::vector<double> quantiles{0, 0.25, 0.5, 0.75, 0.90, 0.95, 0.99, 0.995, 0.999, 1};
+  std::vector<double> computed_quantiles(quantiles.size(), 0.0);
+  hist_approx_quantile(histogram_, quantiles.data(), quantiles.size(), computed_quantiles.data());
+  for (size_t i = 0; i < quantiles.size(); i++) {
+    nighthawk::client::Percentile* percentile = proto.add_percentiles();
+    if (domain == Statistic::SerializationDomain::DURATION) {
+      setDurationFromNanos(*percentile->mutable_duration(),
+                           static_cast<int64_t>(computed_quantiles[i]));
+    } else {
+      percentile->set_raw_value(computed_quantiles[i]);
+    }
+    percentile->set_percentile(quantiles[i]);
+    percentile->set_count(hist_approx_count_below(histogram_, computed_quantiles[i]));
+  }
+
+  return proto;
+}
+
 } // namespace Nighthawk
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -250,8 +250,9 @@ StatisticPtr CircllhistStatistic::combine(const Statistic& statistic) const {
 
 nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
   nighthawk::client::Statistic proto = StatisticImpl::toProto(domain);
-  if (count() == 0)
+  if (count() == 0) {
     return proto;
+  }
 
   std::vector<double> quantiles{0,    0.1,   0.2,  0.3,   0.4,  0.5,   0.55,  0.6,
                                 0.65, 0.7,   0.75, 0.775, 0.8,  0.825, 0.85,  0.875,
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -236,16 +236,37 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
   return proto;
 }
 
+CircllhistStatistic::CircllhistStatistic() {
+  histogram_ = hist_alloc();
+  ASSERT(histogram_ != nullptr);
+}
+
+CircllhistStatistic::~CircllhistStatistic() { hist_free(histogram_); }
+
+void CircllhistStatistic::addValue(uint64_t value) {
+  hist_insert_intscale(histogram_, value, 0, 1);
+  StatisticImpl::addValue(value);
+}
+double CircllhistStatistic::mean() const { return hist_approx_mean(histogram_); }
+double CircllhistStatistic::pvariance() const { return pstdev() * pstdev(); }
+double CircllhistStatistic::pstdev() const {
+  return count() == 0 ? std::nan("") : hist_approx_stddev(histogram_);
+}
+
 StatisticPtr CircllhistStatistic::combine(const Statistic& statistic) const {
   auto combined = std::make_unique<CircllhistStatistic>();
-  const auto& b = dynamic_cast<const CircllhistStatistic&>(statistic);
-  hist_accumulate(combined->histogram_, &this->histogram_, 1);
-  hist_accumulate(combined->histogram_, &b.histogram_, 1);
+  const auto& stat = dynamic_cast<const CircllhistStatistic&>(statistic);
+  hist_accumulate(combined->histogram_, &this->histogram_, /*histogram_count=*/1);
+  hist_accumulate(combined->histogram_, &stat.histogram_, /*histogram_count=*/1);
 
-  combined->min_ = std::min(this->min(), b.min());
-  combined->max_ = std::max(this->max(), b.max());
-  combined->count_ = this->count() + b.count();
+  combined->min_ = std::min(this->min(), stat.min());
+  combined->max_ = std::max(this->max(), stat.max());
+  combined->count_ = this->count() + stat.count();
   return combined;
+}
+
+StatisticPtr CircllhistStatistic::createNewInstanceOfSameType() const {
+  return std::make_unique<CircllhistStatistic>();
 }
 
 nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
@@ -254,9 +275,11 @@ nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain do
     return proto;
   }
 
-  std::vector<double> quantiles{0,    0.1,   0.2,  0.3,   0.4,  0.5,   0.55,  0.6,
-                                0.65, 0.7,   0.75, 0.775, 0.8,  0.825, 0.85,  0.875,
-                                0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1};
+  // List of quantiles is based on hdr_proto_json.gold which lists the quantiles provided by
+  // HdrHistogram.
+  const std::vector<double> quantiles{0,    0.1,   0.2,  0.3,   0.4,  0.5,   0.55,  0.6,
+                                      0.65, 0.7,   0.75, 0.775, 0.8,  0.825, 0.85,  0.875,
+                                      0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1};
   std::vector<double> computed_quantiles(quantiles.size(), 0.0);
   hist_approx_quantile(histogram_, quantiles.data(), quantiles.size(), computed_quantiles.data());
   for (size_t i = 0; i < quantiles.size(); i++) {
@@ -272,6 +295,44 @@ nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain do
   }
 
   return proto;
+}
+
+SinkableStatistic::SinkableStatistic(Envoy::Stats::Scope& scope, absl::optional<int> worker_id)
+    : Envoy::Stats::HistogramImplHelper(scope.symbolTable()), scope_(scope), worker_id_(worker_id) {
+}
+
+SinkableStatistic::~SinkableStatistic() {
+  // We must explicitly free the StatName here in order to supply the
+  // SymbolTable reference.
+  MetricImpl::clear(scope_.symbolTable());
+}
+
+Envoy::Stats::Histogram::Unit SinkableStatistic::unit() const {
+  return Envoy::Stats::Histogram::Unit::Unspecified;
+}
+
+Envoy::Stats::SymbolTable& SinkableStatistic::symbolTable() { return scope_.symbolTable(); }
+
+SinkableHdrStatistic::SinkableHdrStatistic(Envoy::Stats::Scope& scope,
+                                           absl::optional<int> worker_id)
+    : SinkableStatistic(scope, worker_id) {}
+
+void SinkableHdrStatistic::recordValue(uint64_t value) {
+  addValue(value);
+  // Currently in Envoy Scope implementation, deliverHistogramToSinks() will flush the histogram
+  // value directly to stats Sinks.
+  scope_.deliverHistogramToSinks(*this, value);
+}
+
+SinkableCircllhistStatistic::SinkableCircllhistStatistic(Envoy::Stats::Scope& scope,
+                                                         absl::optional<int> worker_id)
+    : SinkableStatistic(scope, worker_id) {}
+
+void SinkableCircllhistStatistic::recordValue(uint64_t value) {
+  addValue(value);
+  // Currently in Envoy Scope implementation, deliverHistogramToSinks() will flush the histogram
+  // value directly to stats Sinks.
+  scope_.deliverHistogramToSinks(*this, value);
 }
 
 } // namespace Nighthawk
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -275,8 +275,7 @@ nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain do
     return proto;
   }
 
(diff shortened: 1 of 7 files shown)


qqustc:
@oschaaf hi Otto, it seems all CI checks failed again due to same error as last time. I will try to trigger it to rerun, meanwhile please go ahead review the change. Thanks!

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+  std::string name() const override { return id(); }
+  std::string tagExtractedName() const override { return id(); }
+
+  const absl::optional<int> worker_id() { return worker_id_; }
+
+private:
+  // This is used for delivering the histogram data to sinks.
+  Envoy::Stats::Scope& scope_;
+  // worker_id can be used in downstream stats Sinks as the stats tag.
+  absl::optional<int> worker_id_;
```
oschaaf, line 234: I wonder if we'll still need worker_id? I suspect that if we use the scope to sink recorded values, then the prefix will automagically turn out to be OK and include the worker id, as that gets supplied by the consuming code.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+    hist_insert_intscale(histogram_, value, 0, 1);
+    StatisticImpl::addValue(value);
+  }
+  double mean() const override { return hist_approx_mean(histogram_); }
+  double pvariance() const override { return pstdev() * pstdev(); }
+  double pstdev() const override {
+    return count() == 0 ? std::nan("") : hist_approx_stddev(histogram_);
+  }
+  StatisticPtr combine(const Statistic& statistic) const override;
+  uint64_t significantDigits() const override { return 1; }
```
oschaaf, line 178: This seems a low guarantee; I wonder, can we improve? (HdrStatistic manages 4 significant digits in the way it we set it up).

oschaaf:
(I have asked for help with CI in the Slack nighthawk-maintainers channel;  I'm still clueless as to what is causing the CI trouble here, and me creating mirror PRs and closing them once we merge isn't sustainable)

source/common/statistic_impl.h
oschaaf: please ignore the comment above, you're right, it looks like we're going to need that worker_id.

test/statistic_test.cc
```diff
@@ -269,6 +272,19 @@ TEST(StatisticTest, StreamingStatProtoOutputLargeValues) {
   EXPECT_EQ(proto.pstdev().nanos(), 0);
 }
 
+TEST(StatisticTest, CircllhistStatisticProtoOutputLargeValues) {
+  CircllhistStatistic a;
+  uint64_t value = 100ul + 0xFFFFFFFF;
```
oschaaf, line 277: I would find a named const instead of `0xFFFFFFFF` slightly easier to read, but that is probably personal, so feel free to ignore :-)

test/statistic_test.cc
```diff
@@ -269,6 +272,19 @@ TEST(StatisticTest, StreamingStatProtoOutputLargeValues) {
   EXPECT_EQ(proto.pstdev().nanos(), 0);
 }
 
+TEST(StatisticTest, CircllhistStatisticProtoOutputLargeValues) {
+  CircllhistStatistic a;
```
oschaaf, line 276: s/a/statistic/? (this may diverge from other places in this source file, but we're trying to consistently use meaningful names these days)

test/statistic_test.cc
```diff
@@ -289,16 +305,39 @@ TEST(StatisticTest, HdrStatisticPercentilesProto) {
   EXPECT_EQ(json, golden_json);
 }
 
+TEST(StatisticTest, CircllhistStatisticPercentilesProto) {
```
oschaaf, line 308: Would it be hard to duplicate the code here with `StatisticTest.HdrStatisticPercentilesProto`?

test/test_data/circllhist_proto_json.gold
```diff
@@ -0,0 +1,60 @@
+{
```
oschaaf, line 1: I wonder if we could somehow unify this with the hdr histogram and have a single gold file / avoid having two different ones.. but probably we're going to see minor differences because of the 2-based vs 10-based implementation details, right?

oschaaf:
I did the magic trick again, creating a PR from my own repo with the same commits to trigger CI. Let's see what happens.

qqustc:
Thanks Otto for the help! Sorry I'm on vacation till the end of this week,
I will try to fix the CI check (currently this is too much hassle :( ) and
update the PR next week.

source/common/statistic_impl.h
qqustc: Discussed offline. Confirmed that circllhist only has 2 digit decimal precision as a result of base 10 algorithm. Since this is out of current work scope, decided to leave the discussion of potential impacts for future work when we try to switch to circllhist

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+
+  // Envoy::Stats::Histogram
+  void recordValue(uint64_t value) override {
+    addValue(value);
+    // Currently in Envoy Scope implementation, deliverHistogramToSinks() will flush the histogram
+    // value directly to stats Sinks.
+    scope_.deliverHistogramToSinks(*this, value);
+  }
+  Envoy::Stats::Histogram::Unit unit() const override {
+    return Envoy::Stats::Histogram::Unit::Unspecified;
```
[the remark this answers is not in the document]
qqustc: Added a comment here.

Currently Envoy Histogram Unit supports {Unspecified, Bytes, Microseconds, Milliseconds}. Since Nighthawk Statistic uses nanosecond as the unit of latency histograms at all places by default, I think it is better to isolate Nighthawk Statistic from the Envoy Histogram Unit concept to avoid confusion.

test/statistic_test.cc
qqustc: If we were to have a shared utility function e.g. `loadGoldFile(parsed_json_proto, gold_file_path)`, it would only dedup 2 lines of code which seems not providing much benefit. I slightly prefer current form since it makes the test setup logic locally in the test case (easier to navigate).

test/test_data/circllhist_proto_json.gold
qqustc: Yes, since the 2 histogram implementation has different precision, the values in the gold file will be different.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+  nighthawk::client::Statistic toProto(SerializationDomain domain) const override;
+
+private:
+  histogram_t* histogram_;
+};
+
+/**
+ * In order to be able to flush histogram value to downstream Envoy stats Sinks, Per worker
+ * SinkableCircllhistStatistic takes the Scope reference in the constructor and wraps the
+ * Envoy::Stats::Histogram interface.
```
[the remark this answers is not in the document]
qqustc: Done

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+    return std::make_unique<CircllhistStatistic>();
+  }
+  nighthawk::client::Statistic toProto(SerializationDomain domain) const override;
+
+private:
+  histogram_t* histogram_;
+};
+
+/**
+ * In order to be able to flush histogram value to downstream Envoy stats Sinks, Per worker
```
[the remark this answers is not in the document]
qqustc: Done

source/common/statistic_impl.cc
```diff
@@ -236,4 +236,39 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
...
+  combined->count_ = this->count() + b.count();
+  return combined;
+}
+
+nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
+  nighthawk::client::Statistic proto = StatisticImpl::toProto(domain);
+  if (count() == 0)
+    return proto;
+
+  std::vector<double> quantiles{0, 0.25, 0.5, 0.75, 0.90, 0.95, 0.99, 0.995, 0.999, 1};
```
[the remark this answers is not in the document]
qqustc: Added a few quantile points
`{0, 0.1, 0.2,  0.3, 0.4, 0.5, 0.55,  0.6, 0.65, 0.7, 0.75, 0.775, 0.8,  0.825, 0.85,  0.875,  0.90, 0.925, 0.95, 0.975, 0.99, 0.995, 0.999, 1}`
which now has mostly the same quantiles as in hdr_proto_json.gold (plus 0.99, 0.995, 0.999 percentiles which are more interesting)

test/statistic_test.cc
```diff
@@ -269,6 +272,19 @@ TEST(StatisticTest, StreamingStatProtoOutputLargeValues) {
...
 
+TEST(StatisticTest, CircllhistStatisticProtoOutputLargeValues) {
+  CircllhistStatistic a;
+  uint64_t value = 100ul + 0xFFFFFFFF;
+  a.addValue(value);
+  a.addValue(value);
+  const nighthawk::client::Statistic proto = a.toProto(Statistic::SerializationDomain::DURATION);
+
+  EXPECT_EQ(proto.count(), 2);
+  Helper::expectNear(((1.0 * proto.mean().seconds() * 1000 * 1000 * 1000) + proto.mean().nanos()),
```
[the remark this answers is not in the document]
qqustc: Done

source/common/statistic_impl.h
```diff
@@ -150,4 +151,87 @@ class HdrStatistic : public StatisticImpl {
...
+  }
+  Envoy::Stats::Histogram::Unit unit() const override {
+    return Envoy::Stats::Histogram::Unit::Unspecified;
+  };
+  bool used() const override { return count() > 0; }
+  Envoy::Stats::SymbolTable& symbolTable() override { return scope_.symbolTable(); }
+  // Overriding name() and tagExtractedName() method in Envoy::Stats::MetricImpl to return
+  // Statistic::id().
+  std::string name() const override { return id(); }
+  std::string tagExtractedName() const override { return id(); }
```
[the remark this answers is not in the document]
qqustc: Discussed offline. Decided to add ASSERT here to currently prohibit the usage of tagExtractedName() in sinkable Nighthawk Statistic since currently it is not needed.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,124 @@ class HdrStatistic : public StatisticImpl {
...
+    addValue(value);
+    // Currently in Envoy Scope implementation, deliverHistogramToSinks() will flush the histogram
+    // value directly to stats Sinks.
+    scope_.deliverHistogramToSinks(*this, value);
+  }
+  bool used() const override { return count() > 0; }
+  // Overriding name() to return Nighthawk::Statistic::id().
+  std::string name() const override { return id(); }
+  std::string tagExtractedName() const override {
+    RELEASE_ASSERT(false, "tagExtractedName() should not be called in Nighthawk Statistic");
```
oschaaf, line 269: Sorry -- maybe we can replace this and the return with `NOT_IMPLEMENTED_GCOVR_EXCL_LINE`, e.g.

```c++
std::string tagExtractedName() const override { NOT_IMPLEMENTED_GCOVR_EXCL_LINE; }
```

This expresses that we don't expect this to be called, and should assert when that does happen.
It may be wishful thinking, but hopefully someday our coverage will actually also exclude these lines from its report.

test/statistic_test.cc
oschaaf: makes sense, sounds good to me then.

source/common/statistic_impl.h
qqustc: Thanks Otto. Changed to NOT_IMPLEMENTED_GCOVR_EXCL_LINE.

Also lower the coverage threshold from 98.6 to 98.4 since there is no meaningful way to test the lines that are missing coverage.

oschaaf: makes sense, thanks

test/statistic_test.cc
```diff
@@ -289,25 +315,39 @@ TEST(StatisticTest, HdrStatisticPercentilesProto) {
...
+
+  for (int i = 1; i <= 10; i++) {
+    statistic.addValue(i);
+  }
+
+  Envoy::MessageUtil util;
+  util.loadFromJson(Envoy::Filesystem::fileSystemForTest().fileReadToEnd(
+                        TestEnvironment::runfilesPath("test/test_data/circllhist_proto_json.gold")),
+                    parsed_json_proto, Envoy::ProtobufMessage::getStrictValidationVisitor());
+  // Instead of comparing proto's, we perform a string-based comparison, because that emits a
```
mum4k, line 330: A more idiomatic approach would be to compare the proto and output a textual diff only if the comparison fails. That way we will be comparing via the binary representations that are the public API. The string are only for humans to read.

Note that EXPECT_EQ takes in an optional stream which can contain a custom error message - this is where we can inject the readable diff:

https://github.com/google/googletest/blob/master/googletest/docs/primer.md#assertions

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
\ No newline at end of file
+/**
+ * CircllhistStatistic uses Circllhist under the hood to compute statistics.
+ * Circllhist is used in the implementation of Envoy Histograms, compared to HdrHistogram it trades
+ * precision for fast performance in merge and insertion. For more info, please see
+ * https://github.com/circonus-labs/libcircllhist
+ */
+class CircllhistStatistic : public StatisticImpl {
+public:
+  CircllhistStatistic() {
```
mum4k, line 162: Can we place the implementation of the constructor in the .cc file? It helps to keep the header files more readable and focused on the API.

Same comment applies to the destructor and the other methods and class.

source/common/statistic_impl.h
mum4k: Once we move the implementation into the .cc file, we should add a comment explaining the single digit precision.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
+  }
+  StatisticPtr combine(const Statistic& statistic) const override;
+  uint64_t significantDigits() const override { return 1; }
+  StatisticPtr createNewInstanceOfSameType() const override {
+    return std::make_unique<CircllhistStatistic>();
+  }
+  nighthawk::client::Statistic toProto(SerializationDomain domain) const override;
+
+private:
+  histogram_t* histogram_;
```
mum4k, line 185: (optional) Using a raw pointer here opens us to memory leaks and all the usual problems of raw pointers.

Would it be possible to move this to a smart pointer (std::unique_ptr) with a custom deleter?

https://www.bfilipek.com/2016/04/custom-deleters-for-c-smart-pointers.html

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
+ * SinkableStatistic takes the Scope reference in the constructor and wraps the
+ * Envoy::Stats::HistogramHelper interface. Implementation of sinkable Nighthawk Statistic class
+ * will inherit from this class.
+ */
+class SinkableStatistic : public Envoy::Stats::HistogramImplHelper {
+public:
+  // Calling HistogramImplHelper(SymbolTable& symbol_table) constructor to construct an empty
+  // MetricImpl. This is to bypass the complicated logic of setting up SymbolTable/StatName in
+  // Envoy.
+  SinkableStatistic(Envoy::Stats::Scope& scope, const absl::optional<int> worker_id)
```
mum4k, line 199: The "const" in front of absl::optional<int> probably doesn't have any useful effect here. We could mark a reference const if we wanted to show that this function isn't allowed to modify the caller's object. However since these are ints, there is nothing to modify.

We can just change it to plain absl::optional<int>.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
+
+  // Currently Envoy Histogram Unit supports {Unspecified, Bytes, Microseconds, Milliseconds}. By
+  // default, Nighthawk::Statistic uses nanosecond as the unit of latency histograms, so Unspecified
+  // is returned here to isolate Nighthawk Statistic from Envoy Histogram Unit.
+  Envoy::Stats::Histogram::Unit unit() const override {
+    return Envoy::Stats::Histogram::Unit::Unspecified;
+  };
+  Envoy::Stats::SymbolTable& symbolTable() override { return scope_.symbolTable(); }
+
+  const absl::optional<int> worker_id() { return worker_id_; }
```
mum4k, line 217: Since this one isn't an override and is public, we should document it. We should also explicitly say why does it return absl::optional and what does it mean when the value isn't present.

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
+
+protected:
```
mum4k, line 219: Can you help me understand why this is protected rather than public?

source/common/statistic_impl.h
```diff
@@ -150,4 +151,118 @@ class HdrStatistic : public StatisticImpl {
...
+  bool used() const override { return count() > 0; }
+  // Overriding name() to return Nighthawk::Statistic::id().
+  std::string name() const override { return id(); }
+  std::string tagExtractedName() const override { NOT_IMPLEMENTED_GCOVR_EXCL_LINE; }
+};
+
+// Implementation of sinkable Nighthawk Statistic with Circllhist Histogram.
+class SinkableCircllhistStatistic : public SinkableStatistic, public CircllhistStatistic {
+public:
+  SinkableCircllhistStatistic(Envoy::Stats::Scope& scope,
```
mum4k, line 251: Can we document the constructor?

source/common/statistic_impl.cc
```diff
@@ -236,4 +236,42 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
   return proto;
 }
 
-} // namespace Nighthawk
\ No newline at end of file
+StatisticPtr CircllhistStatistic::combine(const Statistic& statistic) const {
+  auto combined = std::make_unique<CircllhistStatistic>();
+  const auto& b = dynamic_cast<const CircllhistStatistic&>(statistic);
```
mum4k, line 241: Let's find a better (more expressive) variable name than just "b".

source/common/statistic_impl.cc
```diff
@@ -236,4 +236,42 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
...
+  hist_accumulate(combined->histogram_, &this->histogram_, 1);
```
mum4k, line 242: Without seeing the API of the histogram, it is hard to figure out what does the literal 1 stand for. Is there an useful name we can give it by either making it a constant or placing an inline comment?

```cc
// Either:
const int good_name = 1;
hist_accumulate(combined->histogram_, &this->histogram_, good_name);

// Or:
hist_accumulate(combined->histogram_, &this->histogram_, /* good_name = */ 1);
```

source/common/statistic_impl.cc
```diff
@@ -236,4 +236,42 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
...
+  return combined;
+}
+
+nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
+  nighthawk::client::Statistic proto = StatisticImpl::toProto(domain);
+  if (count() == 0) {
+    return proto;
+  }
+
+  std::vector<double> quantiles{0,    0.1,   0.2,  0.3,   0.4,  0.5,   0.55,  0.6,
```
mum4k, line 257: Can we add a comment indicating how did we select these quantile values?

mum4k, line 257: (optional) If this and the next vector can be const, marking them so helps the reader understand that they aren't going to be modified further down.

test/run_nighthawk_bazel_coverage.sh
```diff
@@ -43,7 +43,7 @@ COVERAGE_VALUE=${COVERAGE_VALUE%?}
 
 if [ "$VALIDATE_COVERAGE" == "true" ]
 then
-  COVERAGE_THRESHOLD=98.6
+  COVERAGE_THRESHOLD=98.4
```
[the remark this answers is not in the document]
qqustc: https://14282-180498819-gh.circle-artifacts.com/0/coverage/source/common/statistic_impl.h.gcov.html

The lines not covered are `std::string tagExtractedName() const override { NOT_IMPLEMENTED_GCOVR_EXCL_LINE; }`, where I added test `EXPECT_DEATH(stat.tagExtractedName(), ".*");` but it does not count into coverage. So after talked to Otto, we decided to lower the coverage.

test/statistic_test.cc
qqustc: Changed to compare the proto. PTAL.

source/common/statistic_impl.h
qqustc: Done. PTAL

source/common/statistic_impl.h
qqustc: Thanks Jakub for the reference. It seems the [`hist_free(histogram_t *hist)`](https://github.com/circonus-labs/libcircllhist/blob/18f1b97603757d5e65baefa65b90d6ec182e8a30/src/circllhist.c#L1447) is quite complicated, so I will not move it to a smart pointer for now

source/common/statistic_impl.h
qqustc: Done

source/common/statistic_impl.h
qqustc: Done.

source/common/statistic_impl.h
qqustc: Done. PTAL

source/common/statistic_impl.h
qqustc: Ah, there is no special reason other than to make its access more restrictive. So changed it to public.

source/common/statistic_impl.cc
qqustc: changed it to "stat".

source/common/statistic_impl.cc
qqustc: Thanks Jakub for the advice. Added an inline comment.

source/common/statistic_impl.cc
qqustc: Done.

qqustc: Done.

source/common/statistic_impl.h
mum4k: This comment may have been missed. Even if we end up not moving this to the .cc file, we should add a comment explaining the choice of "1" as the precision.

source/common/statistic_impl.h
mum4k: I apologize, I made a typo in the comment. I meant to say "rather than private". We should always prefer private methods and fields as they minimize the public API and allow us to execute faster (internal changes).

Can this be private instead?

source/common/statistic_impl.cc
```diff
@@ -236,4 +236,103 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
...
+  return std::make_unique<CircllhistStatistic>();
+}
+
+nighthawk::client::Statistic CircllhistStatistic::toProto(SerializationDomain domain) const {
+  nighthawk::client::Statistic proto = StatisticImpl::toProto(domain);
+  if (count() == 0) {
+    return proto;
+  }
+
+  // List of quantiles is based on hdr_proto_json.gold which lists the quantiles provided by
```
oschaaf, line 278: Note: I think HdrHistogram will emit more percentiles as the number of (differently-valued) data points added grows. So it's not a fixed set with HdrHistogram. 
I am not sure if there is an equivalent way to iterate the available buckets with libcircllhist to get more detail out, and I am also not sure this is really a problem, but I thought it would be good to call out the difference here. 

(The only direct implication I figure it would have today is that the Fortio plots would be drawn with less data points)

source/common/statistic_impl.h
qqustc: Thank you mum4k@.
Added a comment here.

qqustc:
> Thank you @qqustc, just last two nits.

Thank you mum4k@ for the review.
Made scope_ private, exposed it through a public function method `scope()` for child class SinkableHdrStatistic to use.
Previously member scope_ is protected in SinkableStatistic, it is because scope_ is used in the child class SinkableHdrStatistic.

source/common/statistic_impl.cc
qqustc: Thanks Otto for the information! Updated the comment to just `List of quantiles is based on hdr_proto_json.gold.`
From the APIs in [circllhist.h](https://github.com/circonus-labs/libcircllhist/blob/master/src/circllhist.h), there is no easy way to iterate the quantiles. If the listed data points turn out to be not enough in the future, we can always add more here.

Issue #371, opened 2020-06-18
Replace occurrences of "importing * from .." in python code 


oschaaf:
In some places we have code that looks like:

```python
from test.integration.utility import *
```

We should change those to explicitly import only what we need, for example:

```python
from test.integration.utility import assertEquals
```

mum4k:
As discussed in #358, we should:

- ensure we only import packages or modules rather than functions, objects. E.g.:

```py
from test.integration import utility
```

- We should also rename "utility" into something more specific and meaningful so that statements like this read well. E.g.

```py
from test.integration import test

test.assertEqual
```

oschaaf:
Linking https://github.com/envoyproxy/nighthawk/pull/337#discussion_r445789198 as a reminder the doc comment there should needs enhancement.

PR #395, opened 2020-07-01
Stricter and more checks for python code

oschaaf:
In https://github.com/envoyproxy/nighthawk/pull/337 some review comments related
to doc comments have been postponed to a followup. 

Adds more flake8 convention checks.  This broadens the checks we  introduced in the
benchmarking flow to cover the global code base, and exclude a few
files that we either don't own or care about.

Also updates yapf & flake8 version requirements, which enhances automatic formatting,
as well as fixes a python 3.8 compatibility bug.

Should fix https://github.com/envoyproxy/nighthawk/issues/371
Should fix https://github.com/envoyproxy/nighthawk/issues/400

Summary of compliance failures under the proposed checks before this change:

```
13    D100 Missing docstring in public module
2     D101 Missing docstring in public class
9     D102 Missing docstring in public method
46    D103 Missing docstring in public function
4     D107 Missing docstring in __init__
29    D200 One-line docstring should fit on one line with quotes
9     D202 No blank lines allowed after function docstring
32    D205 1 blank line required between summary line and description
3     D208 Docstring is over-indented
1     D209 Multi-line docstring closing quotes should be on a separate line
1     D210 No whitespaces allowed surrounding docstring text
1     D300 Use """triple double quotes"""
39    D400 First line should end with a period
16    D401 First line should be in imperative mood
1     E124 closing bracket does not match visual indentation
2     E125 continuation line with same indent as next logical line
7     E126 continuation line over-indented for hanging indent
1     E401 multiple imports on one line
2     E714 test for object identity should be 'is not'
4     F403 'from test.integration.utility import *' used; unable to detect undefined names
152   F405 'assertCounterEqual' may be undefined, or defined from star imports: test.integration.utility
2     F841 local variable 'logs' is assigned to but never used
8     W291 trailing whitespace
1     W293 blank line contains whitespace
3     W504 line break after binary operator
1     W605 invalid escape sequence '\:'
389
```


oschaaf:
@mum4k requested a review, it would be great to have your thoughts on the broadened
checking introduced here, before following up and making the code compliant with that.

oschaaf:
This now fails check_format with:

```
9     D100 Missing docstring in public module
2     D101 Missing docstring in public class
9     D102 Missing docstring in public method
42    D103 Missing docstring in public function
4     D107 Missing docstring in __init__
```

So just the the missing docstrings are left to work in to complete this.

test/integration/asserts.py
```diff
@@ -0,0 +1,98 @@
+"""A library with assertation helpers for unit tests."""
```
mum4k, line 1: I wonder - do we have a good reason to maintain our own library as compared to using something like:

https://docs.python.org/2/library/unittest.html

Note - it would be ok to address this in a follow-up.

test/integration/asserts.py
```diff
@@ -0,0 +1,98 @@
...
+  assert a in b
+
+
+def assertBetweenInclusive(a, min_value, max_value):
+  """Assert that the passed value is between min_value and max_value, inclusive."""
+  assertGreaterEqual(a, min_value)
+  assertLessEqual(a, max_value)
+
+
+def assertCounterEqual(counters, name, value):
```
mum4k, line 40: Do we need assert helpers like this? Assuming we do switch to the standard unittest library, we could simply assert on:

self.assertEquals(counters[name], value)

And if the counter isn't in the dictionary, it would still "correctly" fail.

test/integration/common.py
```diff
@@ -1,12 +1,22 @@
...
 
 
 class NighthawkException(Exception):
+  """Base exception class for Nighthawk's python code.
+
+  Args:
+      Exception (string): exception message.
+  """
+
   pass
```
mum4k, line 13: (nit, optional) "pass" is no longer needed now that we have a docstring.

test/integration/common.py
```diff
@@ -1,12 +1,22 @@
+"""Miscellaneous utilities used in the integration tests."""
+
 from enum import Enum
 
 
 class NighthawkException(Exception):
+  """Base exception class for Nighthawk's python code.
+
+  Args:
```
mum4k, line 9: Class docstring doesn't really have Args section, since classes don't take arguments. We would typically put that in the docstring of the __init__ method, but that one isn't needed here.

We can just say:

```py
class NighthawkException(Exception):
  """Base exception class for Nighthawk's python code."""
```

test/integration/integration_test_fixtures.py
```diff
@@ -21,6 +19,15 @@
 
 
 def determineIpVersionsFromEnvironment():
+  """Determine the IP version(s) for test execution from the environment.
+
+  Raises:
```
mum4k, line 24: We should only add the Raises section for exceptions raised by the function directly. Otherwise this is a futile effort since every line and character of Python code can raise at least one exception (SyntaxError, MemoryError, RuntimeError, ...).

This functions doesn't have the raise statement, so it shouldn't need the Raises section.

test/integration/integration_test_fixtures.py
```diff
@@ -21,6 +19,15 @@
...
+      NighthawkException: raised when no ip version could be determined, or
+      an invalid one was encountered.
+
+  Returns:
+      test.integration.common.IpVersion: Array with ip versions obtained from the ENVOY_IP_TEST_VERSIONS environment variable.
```
mum4k, line 29: (nit) Return values are typically formatted as a set of sentences, without the "type:", e.g:

```py
Returns:
  A list of test.integration.common.IpVersion with ip versions obtained from the 
  ENVOY_IP_TEST_VERSIONS environment variable.
```

Applies in a few places.

test/integration/integration_test_fixtures.py
```diff
@@ -141,34 +150,26 @@ def _tryStartTestServers(self):
...
+    """Get the global histograms from the json indexed by id."""
     return {
         statistic["id"]: statistic for statistic in self.getGlobalResults(parsed_json)["statistics"]
     }
 
   def getTestServerRootUri(self, https=False):
-    """
-    Utility for getting the http://host:port/ that can be used to query the server we started in setUp()
-    """
+    """Get the http://host:port/ that can be used to query the server we started in setUp()."""
```
mum4k, line 172: (nit) The one-line docstring should fit on a single line (80 chars).

If it doesn't (i.e. we need to say more) we should still choose an abstract that fits into 80 charts and write the rest into the body of the docstring.

Applies in a few places.

test/integration/utility.py
```diff
@@ -1,67 +1,20 @@
+"""Packages utility methods for tests."""
```
mum4k, line 1: (nit, pre-existing) We may want to choose a better name, "utility" is a catch all and doesn't help the reader much. Even something like "test_utility" would be better.

tools/format_python_tools.sh
```diff
@@ -15,11 +15,15 @@ echo "Running Python format check..."
...
 echo "Running Python3 flake8 check..."
-EXCLUDE="--exclude=../benchmarks/tmp/*,*/venv/*"
+cd ..
+EXCLUDE="--exclude=benchmarks/tmp/*,*/venv/*,tools/format_python_tools.py,tools/gen_compilation_database.py,bazel-*"
 flake8 . ${EXCLUDE} --count --select=E901,E999,F821,F822,F823 --show-source --statistics
 # We raise the bar higher for benchmarks/ overall, but especially when it comes to docstrings.
 # Check everything, except indentation and line length for now.
-# Also, we ignore unused imports and redefinitions of unused, as those seems to raise false flags in test definitions.
-flake8 ../benchmarks/ ${EXCLUDE} --docstring-convention pep257 --ignore=E114,E111,E501,F401,F811 --count --show-source --statistics
+# We ignore unused imports and redefinitions of unused, as those seems to raise false flags in test definitions.
```
mum4k, line 23: Is this worth investigating? Can you help by providing some details / examples?

tools/format_python_tools.sh
```diff
@@ -15,11 +15,15 @@ echo "Running Python format check..."
...
+# We ignore formatting isssues E124/E125/E126 because they conflict with the automtic fix format feature.
```
mum4k, line 24: Are we using the wrong auto formatter? Which formatting issues are we ignoring here?

tools/format_python_tools.sh
```diff
@@ -15,11 +15,15 @@ echo "Running Python format check..."
...
+flake8 . ${EXCLUDE} --ignore=E114,E111,E501,F401,F811,E124,E125,E126,D --count --show-source --statistics
```
mum4k, line 25: What is this and the following excludes excluding?

test/integration/asserts.py
oschaaf: This came to be out of unawareness of that library existing. I agree it would be good to move to that.
I would prefer to punt this to an issue though, to not increase size of this PR (and keep it a no-op)

test/integration/integration_test_fixtures.py
oschaaf: `determineIpVersionsFromEnvironment()` has:

```
  else:
    raise NighthawkException("Unknown ip version: '%s'" % versions)

```

So I think in this instance it's OK?

tools/format_python_tools.sh
oschaaf: This ignores/excludes:

- ignored, because of conflict with the automatic fix format script. This reformats according to https://github.com/envoyproxy/nighthawk/blob/master/tools/.style.yapf. We followed Envoy's lead
there, but maybe this is a good time to evaluate that:
E111 Indentation is not a multiple of four
E114 Indentation is not a multiple of four (comment) 
E501 Line too long (82 &gt; 79 characters)
E124 Closing bracket does not match visual indentation
E125 Continuation line with same indent as next logical line
E126 Continuation line over-indented for hanging indent
- False positives because of what I think are pytest peculiarities [1]. Taking a look at this again, I do think it's worth following up, because there's some actually valid complaints about unused imports too it seems. But there's also real false positives; e.g. importing pytest is something we really must do.
F401 Module imported but unused
F811 Redefinition of unused name from line n
- doc conventions
D = doc comment related checks (we do those below, checking both p257 AND google conventions as both seem to have different strengths in their checks; and so far I haven't experienced any conflicts)

[1] 
```
./test/integration/test_integration_zipkin.py:3:1: F401 'pytest' imported but unused
import pytest
^
./test/integration/test_integration_zipkin.py:7:1: F401 'integration_test_fixtures.http_test_server_fixture' imported but unused
from integration_test_fixtures import (http_test_server_fixture, server_config)
^
./test/integration/test_integration_zipkin.py:7:1: F401 'integration_test_fixtures.server_config' imported but unused
from integration_test_fixtures import (http_test_server_fixture, server_config)
^
./test/integration/test_integration_zipkin.py:11:1: F811 redefinition of unused 'http_test_server_fixture' from line 7
def test_tracing_zipkin(http_test_server_fixture):
```

oschaaf:
> This is awesome, would you mind updating the PR description with the latest list of remaining findings?

I updated the description -- but I kept the list, mentioning that these where the issues before the changes in here.
(that might be convenient as a "lookup table" to get descriptions from numbers while we're working on this)

oschaaf:
@mum4k I added be8f7ced7a7a950a14365617168d08f0d714060d to update the flake8/yapf versions. This should fix the python 3.8 compatibility [issue](https://github.com/envoyproxy/nighthawk/issues/400) that recently came up, as well as enhances automatic formatting.
In the next commit, def550c4586d9529b0c8c66a7272ad0c340807f3, I ran `ci/do_ci.sh fix_format` to get the code to comply. Also had to rename a variable called `l` to avoid violating a new check. 

Let me know if you're OK with having these changes in here, we can also split that out to a follow up (but having them in flight separately at the same time will be challenging because of lots of merge conflict opportunities).

mum4k:
Thank you @oschaaf, it is fine to have them here. I will review them in the next few days.

tools/format_python_tools.sh
mum4k: Thank you for the explanations and detailed list. I think it is generally fine for us to remain aligned with Envoy. I don't see anything that would be very concerning in the list above. This PR will significantly help on future reviews of Python code, thank you for these improvements.

Would you mind copying the explanation of the exceptions from this discussion thread into an inline comment next to the exceptions for posterity?

oschaaf: Done: 1e162bc15fc8d434da3a5c859699a9de2f94bcd4

oschaaf:
@mum4k I filed https://github.com/envoyproxy/nighthawk/issues/407 as a sink for leftovers in here.

oschaaf:
(merged this, as this was approved & passing CI)

Issue #382, opened 2020-06-23
Add CI setup that builds and tests Nighthawk using gcc


qqustc:
error msg:

```
/external/envoy/test/mocks/config/BUILD:11:1: C++ compilation of rule '@envoy//test/mocks/config:config_mocks' failed (Exit 1) gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ...

Use --sandbox_debug to see verbose messages from the sandbox
In file included from bazel-out/k8-fastbuild/bin/external/envoy/include/envoy/upstream/_virtual_includes/cluster_manager_interface/envoy/upstream/cluster_manager.h:15,
                 from bazel-out/k8-fastbuild/bin/external/envoy/include/envoy/server/_virtual_includes/factory_context_interface/envoy/server/factory_context.h:26,
                 from bazel-out/k8-fastbuild/bin/external/envoy/include/envoy/server/_virtual_includes/filter_config_interface/envoy/server/filter_config.h:10,
                 from bazel-out/k8-fastbuild/bin/external/envoy/include/envoy/config/_virtual_includes/config_provider_manager_interface/envoy/config/config_provider_manager.h:6,
                 from external/envoy/test/mocks/config/mocks.h:3,
                 from external/envoy/test/mocks/config/mocks.cc:1:
bazel-out/k8-fastbuild/bin/external/envoy/include/envoy/config/_virtual_includes/grpc_mux_interface/envoy/config/grpc_mux.h:75:16: error: 'virtual void Envoy::Config::GrpcMux::pause(std::vector<std::__cxx11::basic_string<char> >)' was hidden [-Werror=overloaded-virtual]
   75 |   virtual void pause(const std::vector<std::string> type_urls) PURE;
      |                ^~~~~
In file included from external/com_google_googletest/googlemock/include/gmock/gmock.h:61,
                 from external/envoy/test/test_common/utility.h:35,
                 from external/envoy/test/mocks/config/mocks.h:15,
                 from external/envoy/test/mocks/config/mocks.cc:1:
[...]

PR #397, opened 2020-07-03
CI: Test that we can produce artifacts with gcc

oschaaf:
- First iteration on testing gcc support in CI. This restricts the
task to just building Nighthawk, and not testing it yet, as that
needs a change in Envoy to go in first.
- Small cleanup, and a bugfix for those who use the script to run
locally, which would be detected as running in CircleCI.

Related issue: https://github.com/envoyproxy/nighthawk/issues/382
(needs a follow up to close, left in as a code-level todo).

----- 

- [x] Prerequisite: #394


oschaaf:
I plan to restart the CI tasks in here to get a sense of stability, as there are changes in here that impact memory requirements.

oschaaf:
It looks like that when building the asan task with 3 concurrent jobs we get “Too long with no output (exceeded 10m0s): context deadline exceeded”fairly structurally  instead of it being oom killed occasionally :( the link steps are very memory hungry, maybe we need to lower parallelism  further. It might then also be worth looking into a way to wrap ‘ld’ to serialize just the link calls and have the compile steps run at higher concurrency, as  that may save some time. That’s quite hacky though. Maybe there’s also a larger build VM available that  we can use for asan. Memory seems very tight for asan; it may take a long time otherwise (approx 1.5-2hours).

PR #398, opened 2020-07-09
Adaptive Load Controller protos

eric846:
All protos, including plugin-specific config protos for all plugins we plan to ship initially.

Works on #416


mum4k:
@oschaaf please assign to me once you are done reviewing this PR (after Eric implements the latest round of changes).

eric846:
Hi Otto, 
This should be ready to review now. Renamed Adaptive RPS Controller to Adaptive Load Controller to reflect the general capability. 

The step_controller_impl.proto configs now take a NighthawkFieldSelector (defined in step_controller.proto) enum input that tells the controller what single field within CommandLineOptions it should adjust dynamically. I included every numeric field in the enum that I didn't know to rule out. It's possible that some of the fields I included in NighthawkFieldSelector don't actually make sense and should be removed.

Future numerical fields added to CommandLineOptions would need a corresponding value added to NighthawkFieldSelector.

We will ship only basic single-variable StepControllers at first, but they will already work equally well for any field, not just RPS.

Advanced multivariable optimization StepControllers could still use NighthawkFieldSelector fields to specify the multiple variables.

Thanks!

eric846:
Updated again to remove some of the hard-coded stuff for setting multiple field and instead use plugins, which actually simplifies the design.

api/adaptive_load/metric_spec.proto
```diff
@@ -0,0 +1,50 @@
...
+
+// Specifies a simple upper or lower threshold, or configures a scoring
+// function.
+message ThresholdSpec {
+  // A threshold for evaluating a metric value.
+  oneof threshold_spec {
+    // Simple upper threshold:
+    //   metric <= upper_threshold -> WITHIN_THRESHOLD
+    //   metric > upper_threshold -> OUTSIDE_THRESHOLD
+    google.protobuf.DoubleValue upper_threshold = 1;
```
oschaaf, line 26: do we need the wrappers for the double valued fields here?

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
...
+  // controller will also be configured to overwrite at least one of the
+  // numerical fields during the search, such as requests_per_second, so any
+  // value of those fields specified here will be ignored.
+  //
+  // All other fields in |nighthawk_traffic_options| are passed through to the
+  // Nighthawk Service.
+  //
+  // Note that |concurrency| in |nighthawk_traffic_options| multiplies the total
+  // RPS sent.
+  nighthawk.client.CommandLineOptions nighthawk_traffic_template = 4;
```
oschaaf, line 44: would `[(validate.rules).message.required = true];` be good to have here? (may also apply to a couple of other spots)

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
...
+  // The duration of each short benchmark during the adjusting stage. Must be
+  // set. Reasonable value: 10 seconds.
+  google.protobuf.Duration measuring_period = 5 [(validate.rules).duration.gt.seconds = 0];
```
oschaaf, line 47: just a random thought -- we're also able to terminate NH benchmarks on other conditions then a max duration, e.g. when a NH counter hits a threshold. I can't see direct value in that in the context of this PR at this time -- but I haven't spend time trying to see it either. So I'm just thinking out loud: maybe it is worth considering not pinning step termination to a duration, but leaving an escape latch in here for if someone ever would like to terminate steps based on other premises? Or maybe no conditional at all, and leave it up to an external sigterm? 
This also brings another thought: do we need to allow handling of error conditions, like should it be weighed somehow if there's a certain 5xx replies observed? (In the current code base these concepts are sort of similar/related in the way they are implemented, e.g. there's termination predicates, and failure predicates, hence me associating these things in a single comment). 
Possibly this can be handled as-is, but I'm just curious if failing to find what we're looking for, based on conditions other then the convergence deadline, is also an option, and what that would look like concretely in configuration.

samples/adaptive_load/example_session_spec.textproto
```diff
@@ -0,0 +1,109 @@
...
+testing_stage_duration {
+  seconds: 30
+}
+
+step_controller_config {
+  name: "exponential-search"
+  typed_config {
+    [type.googleapis.com/nighthawk.adaptive_load.ExponentialSearchStepControllerConfig] {
+      input_variable_setter {
+        name: "rps"
```
oschaaf, line 103: I'm not suggesting I think it would be better, just floating a thought:  I wonder if a  templating approach could be convenient here. E.g. the end-user provides something like:

```
{
 "options": {
  ...
  "duration": "10s",
  "rps": @ExponentialSearchStepControllerConfig.computedRps(),
 ....
 }, 
```

api/adaptive_load/metric_spec.proto
eric846: The reason for wrapping these is to be able to check which of the oneof options is set. Theoretically the default double 0.0 could be the threshold, and a threshold could be positive or negative, so in general can't inspect the value to tell if the field was set. At least not as far as I know with proto3.

api/adaptive_load/adaptive_load.proto
eric846: Oh yes, will add that in many places, it will cut out a lot of C++ validation code.

api/adaptive_load/adaptive_load.proto
eric846: That's very interesting. I think we can actually fully support this as-is:

You can put any termination predicates in nighthawk_traffic_template so that it doesn't keep going for the full 10 second measuring period after causing fatal 5xxs.

If you anticipate 5xxs, you would also want to add a threshold on a builtin metric for something like success-rate, which is 2xx count / total count, so that the next RPS will be reduced on the basis of causing 5xx, even if latency still looked ok. We might recommend always including a threshold for success-rate.

I already implemented success-rate, but maybe there are other metrics that could be derived from NH counters. This code is coming a few PRs from now.

samples/adaptive_load/example_session_spec.textproto
eric846: I see, so you would put a special symbol in any fields of the CommandLineOptions where the controller should vary the value. This would be a very convenient way to express rps and even fault injection headers and the test server response size header. But other cases I had in mind were URI length and header count, and that might open us to needing things like @ExponentialSearchStepControllerConfig.generatedUriString() which then someone would have to implement in C++, or the ability to write an expression like {{'/'+'a'*num}}, or a more complex template mechanism that can replicate a nested proto a certain number of times. The immediate solution to all such problems is C++ code that manipulates the whole CommandLineOption proto, e.g. setting the URI to /aaaaaaaaaa... of length N or inserting N headers.

api/adaptive_load/metric_spec.proto
oschaaf: I see, makes sense

api/adaptive_load/adaptive_load.proto
oschaaf: Sounds great

api/adaptive_load/BUILD
```diff
@@ -0,0 +1,24 @@
+load("@envoy_api//bazel:api_build_system.bzl", "api_cc_py_proto_library")
+
+licenses(["notice"])  # Apache 2
+
+api_cc_py_proto_library(
+    name = "base",
```
mum4k, line 6: Can we use a more descriptive name here? E.g. adaptive_load_proto or similar.

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
```
mum4k, line 1: Since this introduces a few proto files, can we add a file level comment to each one of them indicating (summarizing) what it contains?

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
...
+message AdaptiveLoadSessionSpec {
+  // Settings for MetricsPlugins that obtain metrics from outside sources.
+  // An entry is required for every plugin referred to by metric_thresholds,
+  // other than the "builtin" plugin.
+  repeated MetricsPluginConfig metrics_plugin_configs = 1;
+  // Metrics and thresholds that determine load adjustments.
+  repeated MetricSpecWithThreshold metric_thresholds = 2 [(validate.rules).repeated .min_items = 1];
+  // Metrics that are collected and included in the output but not taken into
+  // account when adjusting the load.
+  repeated MetricSpec informational_metric_specs = 3;
```
mum4k, line 29: What is this going to be used for in practice?

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
...
+  // value of those fields specified here will be ignored.
+  //
+  // All other fields in |nighthawk_traffic_options| are passed through to the
+  // Nighthawk Service.
+  //
+  // Note that |concurrency| in |nighthawk_traffic_options| multiplies the total
+  // RPS sent.
+  nighthawk.client.CommandLineOptions nighthawk_traffic_template = 4;
+  // The duration of each short benchmark during the adjusting stage. Must be
+  // set. Reasonable value: 10 seconds.
```
mum4k, line 46: Should we set the reasonable value as the default instead of commenting it?

Applies to multiple locations with the goal of "out of the box" usability.

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,68 @@
...
+import "google/rpc/status.proto";
+import "validate/validate.proto";
+
+// Parameters describing the adjusting and testing stages of an adaptive load
+// session, which consists of a series of Nighthawk benchmarks probing for
+// the optimal load on the system, followed by a longer benchmark to validate
+// the values. Load adjustments are calculated by the selected StepController
+// plugin. Metrics can come from Nighthawk stats and counters or
+// platform-specific data sources via MetricsPlugins.
+message AdaptiveLoadSessionSpec {
```
mum4k, line 20: For all the fields meant to be provided by the user- can we add an explicit comment whether a field is considered Optional or Required?

api/adaptive_load/benchmark_result.proto
```diff
@@ -0,0 +1,55 @@
...
+
+import "api/adaptive_load/metric_spec.proto";
+import "api/client/output.proto";
+
+// Binary result of checking a metric value against a threshold.
+enum SimpleThresholdStatus {
+  // Value should only occur when the field is unset.
+  UNKNOWN_THRESHOLD_STATUS = 0;
+  // The metric was below an upper threshold or above a lower threshold.
+  WITHIN_THRESHOLD = 1;
```
mum4k, line 13: Why are we choosing to do both the binary result and the threshold score? I thought we were going towards the score only.

api/adaptive_load/benchmark_result.proto
```diff
@@ -0,0 +1,55 @@
...
+
+// Summary of a single Nighthawk Service benchmark session with evaluation
+// results.
+message BenchmarkResult {
+  // Raw Nighthawk Service output. Includes start/end times and full Nighthawk
+  // Service input spec. May contain an error status.
+  nighthawk.client.Output nighthawk_service_output = 1;
+  // Status of all declared metrics during this benchmark session. Not present
+  // in the event of Nighthawk errors.
+  repeated MetricEvaluation metric_evaluations = 3;
```
mum4k, line 54: Should we say anything about the order? Even if it is just that the order or not specified?

api/adaptive_load/input_variable_setter.proto
```diff
@@ -0,0 +1,18 @@
...
+import "google/protobuf/any.proto";
+
+// Selects and configures an InputVariableSetter plugin, which encapsulates
+// knowledge of how to apply a varying numeric input value to a
+// CommandLineOptions proto. For example, RequestsPerSecondInputVariableSetter
+// aka "rps" writes the value to the CommandLineOptions |requests_per_second|
+// field. Some InputVariableSetter plugins such as "rps" will not define a
+// config proto, in which case typed_config should be unset.
+message InputVariableSetterConfig {
+  // Name used to locate the plugin in the Envoy registry.
```
mum4k, line 14: Wherever we refer to a plugin name - it would be good to also include some indication where the user can find a list of available plugins of the kind.

api/adaptive_load/input_variable_setter_impl.proto
```diff
@@ -0,0 +1,18 @@
...
+
+package nighthawk.adaptive_load;
+
+// Note: RequestsPerSecondInputVariableSetter aka "rps" does not define a
+// custom config proto.
+
+// Configuration for an InputVariableSetter plugin that sets an HTTP header
+// within CommandLineOptions to a numeric value being varied by a
+// StepController.
+message NumericHttpHeaderInputVariableSetterConfig {
```
mum4k, line 13: I would suggest we focus on implementing the RPS plugin only in this set of PRs. We can add other plugin types and features in other set of PRs which will help us to stay focused and keep the reviews and test surface smaller.

api/adaptive_load/metrics_plugin_impl.proto
```diff
@@ -0,0 +1,13 @@
+// Plugin-specific config protos for MetricsPlugin.
+
+syntax = "proto3";
+
+package nighthawk.adaptive_load;
+
+// Example plugin-specific config proto, representing connection info for an
+// outside source of metrics. Part of an example showing how to create and
+// register a MetricsPlugin.
+message ExampleMetricsPluginConfig {
```
mum4k, line 10: What is our goal of submitting this example as part of our public API? Would it be better to just include these examples inside documentation, or at least in some non-public part of the codebase?

mum4k, line 10: Additionally as per my other comment - I would suggest we keep ourselves focused on the minimal viable product and keep implementation of any advanced example into another chain of PRs.

api/adaptive_load/scoring_function.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package nighthawk.adaptive_load;
+
+import "google/protobuf/any.proto";
+
+// Configures a ScoringFunction plugin that measures proximity to a threshold
+// according to an arbitrary formula.
+message ScoringFunctionConfig {
```
mum4k, line 9: (optional) Is it worthwhile to create a separate file for each of these SomethingConfig proto messages? Do we expect to be adding more into these files? Or should we just have a config.proto file with all of them?

samples/adaptive_load/example_session_spec.textproto
```diff
@@ -0,0 +1,109 @@
+# proto-file: adaptive_load/adaptive_load.proto
```
mum4k, line 1: What is our long-term plan with this example?

api/adaptive_load/adaptive_load.proto
eric846: Done.

api/adaptive_load/adaptive_load.proto
eric846: This could be useful for both debugging and visualization and comes at almost no added cost. Originally I was going to have comments pointing out that defining a metric with a dummy threshold and weight 0 could be used for this purpose, but then realized it could be explicit.

api/adaptive_load/adaptive_load.proto
eric846: Done.

api/adaptive_load/adaptive_load.proto
eric846: Done.

api/adaptive_load/input_variable_setter_impl.proto
eric846: Agreed, removing this.

api/adaptive_load/metrics_plugin_impl.proto
eric846: Deleted this proto. But I kept the file with no protos so that we don't have to document how to create the file and add it to the build later.

api/adaptive_load/scoring_function.proto
eric846: Moved all to config.proto.

samples/adaptive_load/example_session_spec.textproto
eric846: Deleting this for now. The codelab to be developed later can probably do a better job of illustrating the options, and may or may not have a checked-in sample file.

api/adaptive_load/metrics_plugin_impl.proto
eric846: Agreed.

api/adaptive_load/adaptive_load.proto
eric846: The builtin metrics I'm planning could be interesting to track as informational even if they are not used to control the RPS:
- attempted RPS
- achieved RPS
- success ratio (2xx responses / total responses)
- send rate (how much of the attempted RPS Nighthawk actually managed to send)

api/adaptive_load/BUILD
eric846: Done.

mum4k:
@htuch this PR contains the public API for the adaptive load feature. Considering the visibility and the impact on Nighthawk's architecture, we would love to get your feedback too.

Thank you for the review.

api/adaptive_load/benchmark_result.proto
eric846: I'm now leaning toward a more general approach from the start, where we get rid of the binary result entirely, even for basic cases. It would simplify the code quite a bit by removing branches.

ExponentialSearchStepController can look at the sign of the score and ignore the magnitude.

To avoid confusion, we can provide StepFunctionScoringFunction that returns 1.0 for within threshold, -1.0 otherwise, for use with ExponentialSearchStepController. This would be better than making the user choose LinearScoringFunction with a dummy k that has no effect.

api/adaptive_load/benchmark_result.proto
eric846: Sorry missed this comment. I'll add a comment that the order of metrics is not significant.

eric846:
@mum4k PTAL, simplified protos as discussed

api/adaptive_load/benchmark_result.proto
eric846: (Added the comment on the input side)

api/adaptive_load/config.proto
```diff
@@ -0,0 +1,57 @@
...
+// Selects and configures a StepController plugin, which computes the next load
+// amount (e.g. RPS) and checks for convergence to the optimal load, based on
+// feedback metrics.
+//
+// For available plugin names, see step_controller_impl.proto.
+message StepControllerConfig {
+  // Name used to locate the plugin in the Envoy registry. Required.
+  string name = 1;
+  // Plugin-specific config. Required.
+  google.protobuf.Any typed_config = 2;
```
htuch, line 19: I recommend using `TypedExtensionConfig` for all your extension configs.

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,74 @@
...
+  // value of those fields specified here will be ignored.
+  //
+  // All other fields in |nighthawk_traffic_options| are passed through to the
+  // Nighthawk Service.
+  //
+  // Note that |concurrency| in |nighthawk_traffic_options| multiplies the total
+  // RPS sent.
+  //
+  // Required.
+  nighthawk.client.CommandLineOptions nighthawk_traffic_template = 4
```
htuch, line 49: This seems the wrong level of abstraction. E.g. there are NH CLI options that don't relate to traffic shaping for example, in the future we want to add new options that aren't 1:1. I recommend a standalone proto for this, which is probably some strict subset of NH CLI to begin with.

api/adaptive_load/benchmark_result.proto
```diff
@@ -0,0 +1,41 @@
...
+
+// Records the status of a single metric during a benchmark session.
+message MetricEvaluation {
+  // Original metric specification (plugin name, metric name).
+  MetricSpec metric_spec = 1;
+  // Numerical value of the metric measured during this benchmark session.
+  double metric_value = 2;
+  // Threshold that was used to check this metric value. Not set if the metric
+  // was only informational.
+  ThresholdSpec threshold_spec = 3;
```
htuch, line 20: Would there be benefit in simplifying here if metrics/thresholds are named?

api/adaptive_load/metric_spec.proto
```diff
@@ -0,0 +1,40 @@
...
+import "api/adaptive_load/config.proto";
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+
+// Identifies a feedback metric.
+message MetricSpec {
+  // Name of the metric to evaluate. For the set of built-in metric names, see
+  // source/adaptive_load/metrics_plugin_impl.cc. Required.
+  string metric_name = 1 [(validate.rules).string.min_len = 1];
+  // Name of the MetricsPlugin providing the metric ("builtin" for built-in).
```
htuch, line 16: I recommend adopting a convention such as `nighthawk.foo` for builtin extension `foo`. This is what we do in Envoy.

api/adaptive_load/config.proto
eric846: Will do. "I did not realize that was an option!"

api/adaptive_load/adaptive_load.proto
eric846: I see what you mean. CommandLineOptions is actually thought of as a misnamed version of this standalone proto [1]. But some CLI-specific info not relevant to Nighthawk Service has crept in: OutputFormat [2]. Nighthawk Service-specific info is also possible in the future.

My actual goal in this initial implementation is to collect exactly the information needed to call Nighthawk Service, since internally it needs to be able to do that. In the future, we might move away from Nighthawk Service but still need the underlying traffic spec.

@dubious90 has some plans to fix the entire issue. It may be a lower cost to do this PR with CommandLineOptions as-is and clean up everything at once. We also have some time pressure for adaptive load so we can't necessarily afford to do cleanups at the beginning.

[1] https://github.com/envoyproxy/nighthawk/blob/677b78359748a24b72ce492bcec1fde66bb61c7b/api/client/options.proto#L97

[2] https://github.com/envoyproxy/nighthawk/blob/677b78359748a24b72ce492bcec1fde66bb61c7b/api/client/options.proto#L127

api/adaptive_load/benchmark_result.proto
eric846: The metric spec proto can be boiled down to a single string <plugin name>.<metric name>. I think we can actually just omit ThresholdSpec entirely here. It would be exactly the same verbose proto every time. From just the metric (plugin name, metric name), you can figure out the ThresholdSpec since they have a 1:1 association in the input.

If we don't include the ThresholdSpec inline in each result, we might then want to reproduce the list of (metric spec, threshold spec) at the top level in the output proto, so that the output proto can still be fully interpreted without referring back to the input.

api/adaptive_load/metric_spec.proto
eric846: Will do.

api/adaptive_load/benchmark_result.proto
eric846: Went with `<plugin name>/<metric name>` because the plugin name will have dots in it.

api/adaptive_load/config.proto
eric846: Done

api/adaptive_load/metric_spec.proto
eric846: Done

mum4k:
@pamorgan please review and assign back to me when done.

api/adaptive_load/metric_spec.proto
```diff
@@ -0,0 +1,41 @@
...
+
+import "envoy/config/core/v3/extension.proto";
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+
+// Identifies a feedback metric.
+message MetricSpec {
+  // Name of the metric to evaluate. For the set of built-in metric names, see
+  // source/adaptive_load/metrics_plugin_impl.cc. Required.
+  string metric_name = 1 [(validate.rules).string.min_len = 1];
```
htuch, line 15: How do plugins and metric names relate? Does each plugin only export out a fixed number of metric names? Or is metric name an opaque ID?

eric846: Each plugin will support its own fixed set of metric names. We will query each plugin for its metric names at startup in order to validate the MetricSpec in the adaptive load session input proto.

For nighthawk.builtin:
- latency-min
- latency-mean
- latency-mean-plus-1stdev
- latency-mean-plus-2stdev
- latency-mean-plus-3stdev
- latency-max
- achieved-rps
- attempted-rps
- send-rate
- success-rate

mum4k:
fyi pending review by @pamorgan. Please assign to me once done.

api/adaptive_load/metric_spec.proto
htuch: SGTM. I'm going to be OOTO next two weeks, so please go ahead with this set of protos. I think as long as we're the only consumers, which is highly likely for the near future, it's fine to iterate as needed.

api/adaptive_load/BUILD
```diff
@@ -0,0 +1,21 @@
+load("@envoy_api//bazel:api_build_system.bzl", "api_cc_py_proto_library")
```
pamorgan, line 1: I have some nits but LGTM

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,78 @@
...
+message AdaptiveLoadSessionOutput {
+  // Overall status of the session with error detail.
+  google.rpc.Status session_status = 1;
+  // Results of each short benchmark performed during the adjusting stage.
+  repeated BenchmarkResult adjusting_stage_results = 2;
+  // Result of the single benchmark of the testing stage.
+  BenchmarkResult testing_stage_result = 3;
+  // Metrics and thresholds that were used to determine load adjustments, as referenced in the
+  // BenchmarkResults.
+  repeated MetricSpecWithThreshold metric_thresholds = 4;
```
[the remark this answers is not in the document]
eric846: It's copied directly from metric_thresholds in the input. I included it in the output to preserve the context when dumping the output proto to an archive. Otherwise the archive would only have the value and the score, but not the threshold. If somebody just archived the full input proto alongside the output proto, this field would be redundant.

api/adaptive_load/benchmark_result.proto
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Summary of a single Nighthawk Service benchmark session with evaluation
+// results.
+message BenchmarkResult {
+  // Raw Nighthawk Service output. Includes start/end times and full Nighthawk
+  // Service input spec.
+  nighthawk.client.Output nighthawk_service_output = 1;
+  // Status of this Nighthawk Service benchmark session.
+  google.rpc.Status status = 2;
```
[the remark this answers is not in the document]
eric846: Clarified the comment on this field, which applies narrowly to failures to call the Nighthawk Service and internal errors returned by Nighthawk Service.

Also extended the comment in AdaptiveLoadSessionOutput where things like convergence status are recorded.

api/adaptive_load/metric_spec.proto
```diff
@@ -0,0 +1,41 @@
...
+  // Name of the MetricsPlugin providing the metric ("nighthawk.builtin" for built-in).
+  // Required.
+  string metrics_plugin_name = 2 [(validate.rules).string.min_len = 1];
```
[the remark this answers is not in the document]
eric846: Done

api/adaptive_load/adaptive_load.proto
```diff
@@ -0,0 +1,78 @@
...
+      [(validate.rules).message.required = true];
+  // The duration of each short benchmark during the adjusting stage. Optional, default 10 seconds.
+  google.protobuf.Duration measuring_period = 5 [(validate.rules).duration.gt.seconds = 0];
+  // Maximum amount of time the adjusting stage should wait for convergence
+  // before returning an error. Optional, default 300 seconds.
+  google.protobuf.Duration convergence_deadline = 6 [(validate.rules).duration.gt.seconds = 0];
+  // The duration of the single benchmark session of the testing stage to
+  // confirm the performance at the level of load found in the adjusting stage.
+  // Required.
+  google.protobuf.Duration testing_stage_duration = 7
```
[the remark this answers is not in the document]
eric846: Done

PR #403, opened 2020-07-14
Add missing dependencies to run tools/gen_compilation_database.py 

pamorgan:
Add missing dependencies to run tools/gen_compilation_database.py successfully. This command will create the files needed for CLang intellisense to work.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,5 @@ pytest
 pytest-dependency
 pytest-xdist
 pyyaml
+zipp
+importlib_metadata


oschaaf:
LGTM, only the DCO check is failing. See https://github.com/envoyproxy/nighthawk/pull/403/checks?check_run_id=870849751 and https://github.com/envoyproxy/nighthawk/blob/master/CONTRIBUTING.md#dco-sign-your-work for how to best fix that.

PR #401, opened 2020-07-13
Update Envoy to 7abb0e0bbed4f6b6304403b93762614ad385f80d

oschaaf:
Update Envoy + run gcc-built tests in CI because now we can.

Fixes #382

-----

Prerequisite: 

- [x] Land and merge https://github.com/envoyproxy/envoy/pull/12057

Signed-off-by: Otto van der Schaaf <oschaaf@we-amp.com>

PR #406, opened 2020-07-15
Vscode bootstrap

oschaaf:
Initial VSCode workspace.

- Minimal workspace 
- Common tasks pre-setup. 
- Refreshes the compilation database generation script with Envoy's version.
- Sample debug launch task
- Python linting / autoformatting configuration

------

Prerequisite:

- [x] Land #401


mum4k:

oschaaf:
> I think that the tasks portion was working well. I think we might need more documentation indicating that the intellisense in this version only seems to work with the microsoft cpp plugin rather than with clangd, and that you have to run tools/gen_compilation.py and what to do in case vscode doesn't have the pop up. Also, there were those two or three paths that I needed to add to the c_cpp_properties.json. I'm not sure if we should edit the script to include those, or edit the instructions to indicate that those might be desireable.

Shall we punt the docs on intellisense to an issue then? That part is intentionally left that out of here. 
Ideally we doc at least the clang language server and its plugin (and potentially other ones, like the ms c++ one) in one go.
For now we could omit this part, and leave it up to the user to configure the IDE correctly on that front; 
(though we do offer a way to generate compile_commands.json (which we also need for clang-tidy)).

wjuan-AFK:
Ok. That sounds good then. 

> > I think that the tasks portion was working well. I think we might need more documentation indicating that the intellisense in this version only seems to work with the microsoft cpp plugin rather than with clangd, and that you have to run tools/gen_compilation.py and what to do in case vscode doesn't have the pop up. Also, there were those two or three paths that I needed to add to the c_cpp_properties.json. I'm not sure if we should edit the script to include those, or edit the instructions to indicate that those might be desireable.
> 
> Shall we punt the docs on intellisense to an issue then? That part is intentionally left that out of here.
> Ideally we doc at least the clang language server and its plugin (and potentially other ones, like the ms c++ one) in one go.
> For now we could omit this part, and leave it up to the user to configure the IDE correctly on that front;
> (though we do offer a way to generate compile_commands.json (which we also need for clang-tidy)).

oschaaf:
Filed https://github.com/envoyproxy/nighthawk/issues/423 to track documentation and/or configuration of plugins.

oschaaf:
@mum4k Added a `README.md` to `.vscode`: https://github.com/oschaaf/nighthawk/blob/vscode-bootstrap/.vscode/README.md

oschaaf:
Looks like the updated compilation database generation script made us catch a small clang-tidy nit after
rebasing to master:
```
clang-tidy -extra-arg-before=-xc++ -p=/root/project -quiet /build/tmp/_bazel_bazel/f85b6fb5740e6e8c7efea142eec4b6e8/execroot/nighthawk/include/nighthawk/adaptive_load/scoring_function.h
include/nighthawk/adaptive_load/scoring_function.h:5:10: error: 'envoy/common/pure.h' file not found [clang-diagnostic-error]
#include "envoy/common/pure.h"
```

Added af5ff2aee08c446d9b8ce2b4bf795bbbe1960a69 to resolve. /cc @eric846

.vscode/tasks.json
```diff
@@ -0,0 +1,253 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
```
mum4k, line 2: Does this comment format work in JSON files? I.e. was this tested after the addition of the comment?

oschaaf: Yes that works. json on steroids ;-)

oschaaf:
@eric846 because clang-tidy didn't check the adaptive controller interface headers earlier - it only starts doing that with the changes in here - we missed some deps earlier; I added them in the last couple of commits. clang-tidy is happy with the way things are here, it might be good to incorporate those in your outstanding PRs (or wait for this to land and merge master into them afterwards).

PR #408, opened 2020-07-15
Adding a multiple path test to confirm that expected paths are hit

wjuan-AFK:
Adding an additional unit test for client_benchmark that has multiple path. Important because we want to add a request source that supports multiple path locally.
Signed-off-by: William Juan <66322422+wjuan-AFK@users.noreply.github.com>


dubious90:
(testing repokitteh permissions)

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
 }
-
+TEST_F(BenchmarkClientHttpTest, MultipleRequestsDifferentPath) {
+  std::vector<HeaderMapPtr> requests_vector;
+  std::initializer_list<std::pair<std::string, std::string>> header1{{":scheme", "http"},
+                                                                     {":method", "GET"},
+                                                                     {":path", "/a"},
+                                                                     {":host", "localhost"},
+                                                                     {"Content-Length", "1313"}};
+  std::initializer_list<std::pair<std::string, std::string>> header2{{":scheme", "http"},
```
pamorgan, line 310: I don't think you need the initializer_list you can pass in the const array directly to the TestRequestHeaderMapImpl constructor.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+                                                                     {":method", "GET"},
+                                                                     {":path", "/b"},
+                                                                     {":host", "localhost"},
+                                                                     {"Content-Length", "1313"}};
+  requests_vector.push_back(std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header1));
```
pamorgan, line 315: returned_request_vector would be a better name for this to clarify what's that this is the out

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+  requests_vector.push_back(std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header2));
+  std::vector<HeaderMapPtr>::iterator request_iterator;
+  request_iterator = requests_vector.begin();
+  RequestGenerator request_generator = [&]() {
+    auto item = *request_iterator;
```
pamorgan, line 320: This test is okay but by updating the RequestGenerator we missed the testing the construction logic and the consumption of the  options proto.
We need a nother test that mocks the remoterequestsource so when we make the updates in RequestSourceFactoryImpl::create(, we can be sure that the old functionality is not broken.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+    return std::make_unique<RequestImpl>(item);
+  };
+  std::vector<absl::flat_hash_map<std::string, std::string>> expected_requests_vector;
+  expected_requests_vector.push_back(
+      getTestRecordedProperties(Envoy::Http::TestRequestHeaderMapImpl(header1)));
+  expected_requests_vector.push_back(
+      getTestRecordedProperties(Envoy::Http::TestRequestHeaderMapImpl(header2)));
+
+  EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(2);
+  testBasicFunctionality(1, 1, 2, request_generator, expected_requests_vector);
```
pamorgan, line 331: some comment that explains that the in and out headers check will be done by the testBasicFunctionality which is the main test case.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
   };
 
   EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(0);
-  testBasicFunctionality(1, 1, 1);
+  testBasicFunctionality(1, 1, 1, request_generator);
   EXPECT_EQ(1, getCounter("http_2xx"));
 }
-
+TEST_F(BenchmarkClientHttpTest, MultipleRequestsDifferentPath) {
```
pamorgan, line 303: preferably start method names with a verb

test/benchmark_http_client_test.cc
```diff
@@ -121,14 +132,25 @@ class BenchmarkClientHttpTest : public Test {
     decoders_.clear();
     dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
     EXPECT_EQ(0, inflight_response_count);
```
pamorgan, line 134: I'm no expert on this but can you double check that EXPECT_FALSE(true) would fail here, I remember that EXPECT_* don't work in methods and has to be in the main test case.

test/benchmark_http_client_test.cc
```diff
@@ -121,14 +132,25 @@ class BenchmarkClientHttpTest : public Test {
...
+    // If we have no expectations, then we don't test.
+    if (header_expectations.size() != 0) {
```
pamorgan, line 136: it's not clear from the method signature that empty header_expectation would simply ignore the check instead of validating that the returned header were empty

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,52 @@ class BenchmarkClientHttpTest : public Test {
...
           std::initializer_list<std::pair<std::string, std::string>>(
               {{":scheme", "http"}, {":method", "GET"}, {":path", "/"}, {":host", "localhost"}}));
       return std::make_unique<RequestImpl>(header);
     };
+    return request_generator;
   }
-
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  void testBasicFunctionality(
```
pamorgan, line 61: This utility method deserves a comment and better name :)

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,52 @@ class BenchmarkClientHttpTest : public Test {
...
-    request_generator_ = []() {
+        .WillRepeatedly([&](const Envoy::Tracing::Config& config, const Envoy::Http::HeaderMap&,
+                            const Envoy::StreamInfo::StreamInfo&,
+                            const Envoy::Tracing::Decision) -> Envoy::Tracing::Span* {
+          EXPECT_EQ(Envoy::Tracing::OperationName::Egress, config.operationName());
+          auto* span = new NiceMock<Envoy::Tracing::MockSpan>();
+          return span;
+        });
+  }
+  RequestGenerator getDefaultRequestGenerator() {
```
pamorgan, line 52: this method can be const
 RequestGenerator getDefaultRequestGenerator() const {}

pamorgan: Maybe this doesn't need to be a member in the class and can be an utility function

test/benchmark_http_client_test.cc
wjuan-AFK: It does fail here.

test/benchmark_http_client_test.cc
wjuan-AFK: I'm not sure what this comment is applying to. Is this the MultipleRequestsDifferentPath name?

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +297,38 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
 
   EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(0);
-  testBasicFunctionality(1, 1, 1);
+  testBasicFunctionality(1, 1, 1, request_generator);
   EXPECT_EQ(1, getCounter("http_2xx"));
 }
-
+TEST_F(BenchmarkClientHttpTest, MultipleRequestsDifferentPath) {
+  std::vector<HeaderMapPtr> requests_vector;
+  std::initializer_list<std::pair<std::string, std::string>> header1{{":scheme", "http"},
```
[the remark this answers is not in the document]
wjuan-AFK: Done

test/benchmark_http_client_test.cc
```diff
@@ -121,6 +130,17 @@ class BenchmarkClientHttpTest : public Test {
...
     dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
     EXPECT_EQ(0, inflight_response_count);
+    // If we have no expectations, then we don't test.
+    if (header_expectations.size() != 0) {
+      EXPECT_THAT(header_expectations, UnorderedElementsAreArray(called_headers));
+    }
+  }
+
+  absl::flat_hash_map<std::string, std::string>
+  getTestRecordedProperties(const Envoy::Http::RequestHeaderMap& header) {
```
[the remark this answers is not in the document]
wjuan-AFK: Done.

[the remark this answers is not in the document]
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
```diff
@@ -121,6 +130,17 @@ class BenchmarkClientHttpTest : public Test {
     decoders_.clear();
     dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
     EXPECT_EQ(0, inflight_response_count);
+    // If we have no expectations, then we don't test.
+    if (header_expectations.size() != 0) {
+      EXPECT_THAT(header_expectations, UnorderedElementsAreArray(called_headers));
+    }
+  }
+
+  absl::flat_hash_map<std::string, std::string>
```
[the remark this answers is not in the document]
wjuan-AFK: I wanted to use a map because I thought we might extend this function later when we found more fields that should be matched.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

mum4k:
Looks like this is waiting for changes (some comments aren't addressed yet), please flip the labels if I misunderstood.

test/benchmark_http_client_test.cc
wjuan-AFK: I don't think that works. I can make it a const initializer_list, which is done. But const array is not accepted. There is no default conversion between initializer_list and array that I am aware of.

test/benchmark_http_client_test.cc
wjuan-AFK: After consideration, this will be in a separate PR.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Add clarifying comments on the method, not sure if there's a better way to do that.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
+  void setupBenchmarkClient(const RequestGenerator& request_generator) {
+    client_ = std::make_unique<Client::BenchmarkClientHttpImpl>(
+        *api_, *dispatcher_, store_, std::make_unique<StreamingStatistic>(),
+        std::make_unique<StreamingStatistic>(), std::make_unique<StreamingStatistic>(),
+        std::make_unique<StreamingStatistic>(), false, cluster_manager_, http_tracer_, "benchmark",
+        request_generator, true);
+  }
+  /// Primary testing method. Confirms that connection limits are met and number of requests are
+  /// correct. If specified, also checks the header expectations, if not specified, it is ignored.
+  void testBenchmarkClientFunctionality(
```
pamorgan, line 79: Prefer that the new name describes the behavior of the function, maybe VerifyBenchmarkClientCanProcessInflightRequests ?

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
           std::initializer_list<std::pair<std::string, std::string>>(
               {{":scheme", "http"}, {":method", "GET"}, {":path", "/"}, {":host", "localhost"}}));
       return std::make_unique<RequestImpl>(header);
     };
+    return request_generator;
   }
-
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  /// Helper function to get properties in a map that should be verified during the test
```
pamorgan, line 62: nit: this function doesn't need to be a member in the class. it's better to move it outside of the class for readability.

test/benchmark_http_client_test.cc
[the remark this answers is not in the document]
pamorgan: it's better to start with the simplest data structure and change it when the need rises.

test/benchmark_http_client_test.cc
pamorgan: test names names should be action-condition-result.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
-    request_generator_ = []() {
+        .WillRepeatedly([&](const Envoy::Tracing::Config& config, const Envoy::Http::HeaderMap&,
+                            const Envoy::StreamInfo::StreamInfo&,
+                            const Envoy::Tracing::Decision) -> Envoy::Tracing::Span* {
+          EXPECT_EQ(Envoy::Tracing::OperationName::Egress, config.operationName());
+          auto* span = new NiceMock<Envoy::Tracing::MockSpan>();
+          return span;
+        });
+  }
+  //Default function for request generator when the content doesn't matter.
```
pamorgan, line 52: nit: please leave a space after //.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  /// Helper function to get properties in a map that should be verified during the test
+  absl::flat_hash_map<std::string, std::string>
+  getTestRecordedProperties(const Envoy::Http::RequestHeaderMap& header) {
+    absl::flat_hash_map<std::string, std::string> properties_map;
+    properties_map["uri"] = std::string(header.getPathValue());
+    return properties_map;
+  }
+  /// Used to set up benchmarkclient. Especially from within the testBenchmarkClientFunctionality.
```
pamorgan, line 69: nit: one line comments should only start with // not ///.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
+  void setupBenchmarkClient(const RequestGenerator& request_generator) {
+    client_ = std::make_unique<Client::BenchmarkClientHttpImpl>(
+        *api_, *dispatcher_, store_, std::make_unique<StreamingStatistic>(),
+        std::make_unique<StreamingStatistic>(), std::make_unique<StreamingStatistic>(),
+        std::make_unique<StreamingStatistic>(), false, cluster_manager_, http_tracer_, "benchmark",
+        request_generator, true);
+  }
+  /// Primary testing method. Confirms that connection limits are met and number of requests are
+  /// correct. If specified, also checks the header expectations, if not specified, it is ignored.
```
pamorgan, line 78: nit: I think you mean if empty. it's always specified as this is a const reference.

nit: We could add another boolean that controls the header validation instead of relying on the empty value.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +41,70 @@ class BenchmarkClientHttpTest : public Test {
...
+      const RequestGenerator& request_generator,
+      const std::vector<absl::flat_hash_map<std::string, std::string>>& header_expectations =
+          std::vector<absl::flat_hash_map<std::string, std::string>>()) {
     if (client_ == nullptr) {
-      setupBenchmarkClient();
+      setupBenchmarkClient(request_generator);
       cluster_info().resetResourceManager(connection_limit, max_pending, 1024, 0, 1024);
     }
-
+    // this is where we store the properties of headers that are passed to the stream encoder. We
```
pamorgan, line 88: nit: typo This

wjuan-AFK: Done. That was careless of me.

test/benchmark_http_client_test.cc
wjuan-AFK: Updated this, hopefully the new name is better.

test/benchmark_http_client_test.cc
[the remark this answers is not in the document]
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: done

test/benchmark_http_client_test.cc
wjuan-AFK: done

test/benchmark_http_client_test.cc
wjuan-AFK: done

test/benchmark_http_client_test.cc
wjuan-AFK: First one is done. Is it better to have a separate boolean? I vaguely feel like passing more parameters like that means I'm doing something wrong, but I'm not sure what the correct approach is.

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,26 @@ using namespace testing;
 
 namespace Nighthawk {
 
+namespace BenchmarkClientTestHelperFunctions {
```
pamorgan, line 29: please make it anonymous namespace. No name is needed for this namespace. It helps readers to know that this won't be referenced anywhere outside this file.

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,26 @@ using namespace testing;
...
+// Default function for request generator when the content doesn't matter.
```
pamorgan, line 30: nit: please leave an empty line after the namespace.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +61,55 @@ class BenchmarkClientHttpTest : public Test {
...
+        request_generator, true);
+  }
+  // Primary testing method. Confirms that connection limits are met and number of requests are
+  // correct. If not empty, also checks the header expectations, if empty, it is ignored.
+  void TestBenchmarkClientProcessesExpectedInflightRequests(
+      const uint64_t max_pending, const uint64_t connection_limit, const uint64_t amount_of_request,
+      const RequestGenerator& request_generator,
+      const std::vector<absl::flat_hash_set<std::string>>& header_expectations =
+          std::vector<absl::flat_hash_set<std::string>>()) {
     if (client_ == nullptr) {
```
pamorgan, line 88: Can we remove the if condition. The client_ state is not observable by the caller, and they might not understand why the generator lambda is not taking effect.

test/benchmark_http_client_test.cc
```diff
@@ -41,41 +61,55 @@ class BenchmarkClientHttpTest : public Test {
...
+        std::make_unique<StreamingStatistic>(), std::make_unique<StreamingStatistic>(),
+        std::make_unique<StreamingStatistic>(), false, cluster_manager_, http_tracer_, "benchmark",
+        request_generator, true);
+  }
+  // Primary testing method. Confirms that connection limits are met and number of requests are
+  // correct. If not empty, also checks the header expectations, if empty, it is ignored.
+  void TestBenchmarkClientProcessesExpectedInflightRequests(
+      const uint64_t max_pending, const uint64_t connection_limit, const uint64_t amount_of_request,
+      const RequestGenerator& request_generator,
+      const std::vector<absl::flat_hash_set<std::string>>& header_expectations =
```
pamorgan, line 86: After thinking about, I think a nullable pointer seem more appropriate for this. absl::optional is usually used for something like this.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +311,45 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
   };
 
   EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(0);
-  testBasicFunctionality(1, 1, 1);
+  TestBenchmarkClientProcessesExpectedInflightRequests(1, 1, 1, request_generator);
   EXPECT_EQ(1, getCounter("http_2xx"));
 }
```
pamorgan, line 316: nit: Please add an empty line.

test/benchmark_http_client_test.cc
```diff
@@ -276,8 +311,45 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
-
+TEST_F(BenchmarkClientHttpTest, RequestGeneratorProvidingDifferentPathsSendsRequestsOnThosePaths) {
+  std::vector<HeaderMapPtr> requests_for_generator_to_send;
+  const std::initializer_list<std::pair<std::string, std::string>> first_header_map_to_send{
```
pamorgan, line 319: nit: if I understand correctly then header_map_for_first_request might be a more meaningfull name.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: I'm afraid that I don't really understand what this bit means. Also, not sure if it's still relevant now that I made it nullable.

wjuan-AFK: Now that I understand your comment, I agree with it, but because it affects other existing tests, I will be making that change in a separate PR to isolate it.

oschaaf:
I think we need https://github.com/envoyproxy/nighthawk/pull/421 to resolve the CI problems, also occurs on other PRs and the last master merge.

test/benchmark_http_client_test.cc
```diff
@@ -310,8 +349,48 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+      {":path", "/b"},
+      {":host", "localhost"},
+      {"Content-Length", "1313"}};
+  requests_for_generator_to_send.push_back(
+      std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header_map_for_first_request));
+  requests_for_generator_to_send.push_back(
+      std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header_map_for_second_request));
+  std::vector<HeaderMapPtr>::iterator request_iterator;
+  request_iterator = requests_for_generator_to_send.begin();
+  RequestGenerator request_generator = [&]() {
```
oschaaf, line 377: I'm not a huge fan of `[&]` as things may creep into that during maintenance; maybe use `[&request_iterator]`?

test/benchmark_http_client_test.cc
```diff
@@ -310,8 +349,48 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+    auto item = *request_iterator;
+    request_iterator++;
+    return std::make_unique<RequestImpl>(item);
```
oschaaf, line 380: collapse this to just `return std::make_unique<RequestImpl>(*request_iterator++);` ?

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,27 @@ using namespace testing;
...
+    auto header = std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(
+        std::initializer_list<std::pair<std::string, std::string>>(
+            {{":scheme", "http"}, {":method", "GET"}, {":path", "/"}, {":host", "localhost"}}));
+    return std::make_unique<RequestImpl>(header);
+  };
+  return request_generator;
+}
+// Helper function to get headers in a set that should be verified during the test.
+absl::flat_hash_set<std::string>
+getTestRecordedProperties(const Envoy::Http::RequestHeaderMap& header) {
```
oschaaf, line 43: I'm a bit confused by purpose and/or naming here. I'm wondering why does returns a `absl::flat_hash_set` which only has a single entry (path?), and create a `std::vector<absl::flat_hash_set>` later on? 
If the purpose is to record all paths requested throughout a test, would storing all the paths in a single `absl::flat_hash_set` instance be more convenient?

test/benchmark_http_client_test.cc
```diff
@@ -46,41 +67,52 @@ class BenchmarkClientHttpTest : public Test {
...
+                            const Envoy::Tracing::Decision) -> Envoy::Tracing::Span* {
+          EXPECT_EQ(Envoy::Tracing::OperationName::Egress, config.operationName());
+          auto* span = new NiceMock<Envoy::Tracing::MockSpan>();
+          return span;
+        });
   }
 
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  // void setupBenchmarkClient(const RequestGenerator& request_generator) {
```
pamorgan, line 79: Please remove the commented method.

test/benchmark_http_client_test.cc
```diff
@@ -46,41 +67,52 @@ class BenchmarkClientHttpTest : public Test {
...
+  //   client_ = std::make_unique<Client::BenchmarkClientHttpImpl>(
+  //       *api_, *dispatcher_, store_, std::make_unique<StreamingStatistic>(),
+  //       std::make_unique<StreamingStatistic>(), std::make_unique<StreamingStatistic>(),
+  //       std::make_unique<StreamingStatistic>(), false, cluster_manager_, http_tracer_,
+  //       "benchmark", request_generator, true);
+  // }
+  // Primary testing method. Confirms that connection limits are met and number of requests are
+  // correct. If not empty, also checks the header expectations, if empty, it is ignored.
```
pamorgan, line 87: extra "If not empty".

test/benchmark_http_client_test.cc
```diff
@@ -46,41 +67,52 @@ class BenchmarkClientHttpTest : public Test {
...
+  void TestBenchmarkClientProcessesExpectedInflightRequests(
```
pamorgan, line 88: Comment should change to if null instead of "if empty,"

test/benchmark_http_client_test.cc
```diff
@@ -46,41 +67,52 @@ class BenchmarkClientHttpTest : public Test {
...
+      const uint64_t max_pending, const uint64_t connection_limit, const uint64_t amount_of_request,
+      const RequestGenerator& request_generator,
+      const std::vector<absl::flat_hash_set<std::string>>* header_expectations = nullptr) {
```
pamorgan, line 91: I usually prefer using absl::Optional for pointers so the caller knows that a null is acceptable.

test/benchmark_http_client_test.cc
pamorgan: let's file a bug instead for this.

test/benchmark_http_client_test.cc
```diff
@@ -310,8 +349,48 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+      {":host", "localhost"},
+      {"Content-Length", "1313"}};
+  requests_for_generator_to_send.push_back(
+      std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header_map_for_first_request));
+  requests_for_generator_to_send.push_back(
+      std::make_shared<Envoy::Http::TestRequestHeaderMapImpl>(header_map_for_second_request));
+  std::vector<HeaderMapPtr>::iterator request_iterator;
+  request_iterator = requests_for_generator_to_send.begin();
+  RequestGenerator request_generator = [&]() {
+    auto item = *request_iterator;
```
pamorgan, line 378: if you can use the type instead of auto here, it would be more readable.

wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: I think that optional is not recommended for this, and a const pointer is preferred. This is to avoid making unnecessary copies.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: That makes sense. I had originally wanted to do this with more properties, but in the end, it wasn't necessary.

test/benchmark_http_client_test.cc
wjuan-AFK: Thanks, that was sloppy of me. Done.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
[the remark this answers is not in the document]
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,27 @@ using namespace testing;
 
 namespace Nighthawk {
 
+namespace {
+
+// Default function for request generator when the content doesn't matter.
+RequestGenerator getDefaultRequestGenerator() {
```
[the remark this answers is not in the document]
wjuan-AFK: Done.

wjuan-AFK:
Sorry it took me so long to get to this. I will do better moving forward.

test/benchmark_http_client_test.cc
```diff
@@ -310,8 +343,46 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
+  expected_requests.insert(
+      getTestRecordedProperty(Envoy::Http::TestRequestHeaderMapImpl(header_map_for_first_request)));
+  expected_requests.insert(getTestRecordedProperty(
+      Envoy::Http::TestRequestHeaderMapImpl(header_map_for_second_request)));
+
+  EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(2);
+
+  // Most of the testing happens inside of this call. Will confirm that the requests received match
+  // the expected requests vector.
+  TestBenchmarkClientProcessesExpectedInflightRequests(1, 1, 2, request_generator,
```
oschaaf, line 384: probably out of scope, but now that we have c++17 it looks like we could potentially
have named arguments [1], which could enhance readability of the numbers we pass to these calls. 

[1] https://www.fluentcpp.com/2018/12/14/named-arguments-cpp/

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,27 @@ using namespace testing;
...
+
+// Default function for request generator when the content doesn't matter.
+RequestGenerator getDefaultRequestGenerator() {
+  RequestGenerator request_generator = []() {
+    return std::make_unique<RequestImpl>(getDefaultHeaderMap());
+  };
+  return request_generator;
+}
+// Helper function to get headers in a set that should be verified during the test.
+std::string getTestRecordedProperty(const Envoy::Http::RequestHeaderMap& header) {
```
oschaaf, line 45: Perhaps consider updating this to `getPathFromRequest()` and update the comment above, to match  it to the narrowed down functionality?

test/benchmark_http_client_test.cc
```diff
@@ -46,41 +67,46 @@ class BenchmarkClientHttpTest : public Test {
...
 
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  // Primary testing method. Confirms that connection limits are met and number of requests are
+  // correct. If header expectations is not null, also checks the header expectations, if null, it
+  // is ignored.
+  void TestBenchmarkClientProcessesExpectedInflightRequests(
+      const uint64_t max_pending, const uint64_t connection_limit, const uint64_t amount_of_request,
+      const RequestGenerator& request_generator,
+      const absl::flat_hash_set<std::string>* header_expectations = nullptr) {
```
oschaaf, line 85: Optional as far as I'm concerned, but we could pass a reference instead of a pointer here by creating an overload which doesn't have the expectations, and then calls this with an empty set. Pointers are mostly avoided, and doing this would help consistency.

test/benchmark_http_client_test.cc
```diff
@@ -126,12 +152,17 @@ class BenchmarkClientHttpTest : public Test {
     decoders_.clear();
     dispatcher_->run(Envoy::Event::Dispatcher::RunType::Block);
     EXPECT_EQ(0, inflight_response_count);
+    // If we have no expectations, then we don't test.
+    if (header_expectations != nullptr) {
+      EXPECT_THAT((*header_expectations), UnorderedElementsAreArray(called_headers));
```
oschaaf, line 157: Not sure if applicable here, but it looks like Envoy does have test helpers these days in this area, e.g. see `IsSupersetOfHeaders` over at  https://github.com/envoyproxy/envoy/blob/6dabdb393e695928f7e6777445d722fe2f625413/test/extensions/filters/http/cache/cache_filter_test.cc#L97

test/benchmark_http_client_test.cc
wjuan-AFK: I originally had the empty set thing, but I felt like nullable was clearer, since empty set might mislead people to think that we're testing that we get an empty set of headers, rather than that we don't test it.

oschaaf: Fair enough

test/benchmark_http_client_test.cc
wjuan-AFK: done

test/benchmark_http_client_test.cc
wjuan-AFK: done.

test/benchmark_http_client_test.cc
wjuan-AFK: I refactored this to reduce the number of arguments instead. Figured that this would be cleaner.

test/benchmark_http_client_test.cc
wjuan-AFK: This seems to be very useful. Thanks otto. For now, I think that it's fine to just save and compare the paths for now. We can change to using the superset conditioning when we want to extend this further.

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,37 @@ using namespace testing;
 
 namespace Nighthawk {
 
+namespace {
+const std::shared_ptr<Envoy::Http::RequestHeaderMap>& getDefaultHeaderMap() {
```
mum4k, line 30: Even though optional, can we add at least a single line of comment explaining what this function does?

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,37 @@ using namespace testing;
...
+  RequestGenerator request_generator = []() {
+    return std::make_unique<RequestImpl>(getDefaultHeaderMap());
+  };
+  return request_generator;
+}
+// Helper function to get headers in a set that should be verified during the test.
+std::string getPathFromRequest(const Envoy::Http::RequestHeaderMap& header) {
+  return std::string(header.getPathValue());
+}
+struct ClientSetupParameters {
```
mum4k, line 48: Documenting this struct, its constructor and members will help future readers / editors of the test.

test/benchmark_http_client_test.cc
```diff
@@ -26,6 +26,37 @@ using namespace testing;
 
 namespace Nighthawk {
 
+namespace {
+const std::shared_ptr<Envoy::Http::RequestHeaderMap>& getDefaultHeaderMap() {
+  static auto* header =
```
mum4k, line 31: While this is a known idiom when static storage on globals is required, I am trying to understand if it is necessary here.

Can we instead create the `std::shared_ptr<Envoy::Http::RequestHeaderMap>` instance as a member of the `BenchmarkClientHttpTest` test fixture? We could do this inside its constructor or inside the SetUp() method. Either of those would introduce simpler code for the posterity.

test/benchmark_http_client_test.cc
mum4k: Seeing a const pointer as an argument is fairly unusual. If the intent is to make this an optional argument, we should use `absl::optional` instead and the value can then be the expected const reference (or non-const reference). Otherwise the reader of this code has to guess why we are using a pointer here.

There are already a few places in the Nighthawk codebase that use `absl::optional` and can serve as an example.

test/benchmark_http_client_test.cc
```diff
@@ -310,8 +359,46 @@ TEST_F(BenchmarkClientHttpTest, BadContentLength) {
...
   EXPECT_CALL(stream_encoder_, encodeData(_, _)).Times(0);
-  testBasicFunctionality(1, 1, 1);
+  auto client_setup_parameters = ClientSetupParameters(1, 1, 1, request_generator);
+  TestBenchmarkClientProcessesExpectedInflightRequests(client_setup_parameters);
   EXPECT_EQ(1, getCounter("http_2xx"));
 }
 
+TEST_F(BenchmarkClientHttpTest, RequestGeneratorProvidingDifferentPathsSendsRequestsOnThosePaths) {
+  std::vector<HeaderMapPtr> requests_for_generator_to_send;
+  const std::initializer_list<std::pair<std::string, std::string>> header_map_for_first_request{
```
mum4k, line 369: If I remember correctly, we discussed that we will try to avoid using the `std::initializer_list` explicitly. What changed our plans?

test/benchmark_http_client_test.cc
```diff
@@ -46,50 +77,56 @@ class BenchmarkClientHttpTest : public Test {
...
+          return span;
+        });
   }
 
-  void testBasicFunctionality(const uint64_t max_pending, const uint64_t connection_limit,
-                              const uint64_t amount_of_request) {
+  // Primary testing method. Confirms that connection limits are met and number of requests are
+  // correct. If header expectations is not null, also checks the header expectations, if null, it
+  // is ignored.
+  void TestBenchmarkClientProcessesExpectedInflightRequests(
```
mum4k, line 92: The name of this function seems inconsistent with the rest of the style in this file. Most function names start with a lowercase letter. We likely cannot use a name starting with lowercase `test` since the testing framework might interpret this as a name of a test method. Can we choose a different name that will adhere to our style, e.g.:

`verifyBenchmarkClientProcessesExpectedInflightRequests`

test/benchmark_http_client_test.cc
wjuan-AFK: I think the problem is that the constructor specifically takes an initializer list, and it makes sense to declare it as an initializer list for that. I think I can actually use auto here, but I'm not sure that it would be an improvement.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Done.

test/benchmark_http_client_test.cc
wjuan-AFK: Removing this function and replacing with a member.

test/benchmark_http_client_test.cc
wjuan-AFK: done.

PR #415, opened 2020-07-19
CI: lower test_gcc parallellism.

oschaaf:
Compared to clang, gcc uses more memory. Lower parallelism to
avoid the build running out of memory.


oschaaf:
Increased the proposed parallelism from 4 to 6, to avoid a timeout from
being triggered because of the build output staying silent for too long.
So we'll have to balance between that and running out of memory.

PR #418, opened 2020-07-22
Add CLI flags "stats-sinks" and "stats-flush-interval" to support configure stat sinks in NH

qqustc:
Add CLI flags 
- `--stats-sinks` - Stats sinks (in json or compact yaml) where Nighthawk metrics will be flushed. This argument is intended to be specified multiple times.
- `--stats-flush-interval` - Time interval (in seconds) between flushes to configured stats sinks. Default: 5 (seconds).

The flags' values will be used to update [`stats_sinks `](https://github.com/envoyproxy/envoy/blob/54652f03ea0b3504bd4d64e8ae31985be2892a95/api/envoy/config/bootstrap/v3/bootstrap.proto#L129) and [`stats_flush_interval `](https://github.com/envoyproxy/envoy/blob/54652f03ea0b3504bd4d64e8ae31985be2892a95/api/envoy/config/bootstrap/v3/bootstrap.proto#L139) in `bootstrap.proto`.

Linked Issues: #344
Testing: unit tests
Docs Changes: README.md


mum4k:
@eric846 please review and assign back to me when done.

api/client/options.proto
```diff
@@ -195,4 +196,10 @@ message CommandLineOptions {
   // execution. Note that this will be reflected in the counters that
   // Nighthawk writes to the output. Default is false.
   google.protobuf.BoolValue simple_warmup = 32;
+  // Optional set of stats sinks where Nighthawk metrics will be flushed to.
+  repeated envoy.config.metrics.v3.StatsSink stats_sinks = 34;
```
eric846, line 200: Can the two new fields be 33, 34 or is there a conflicting PR with 33?

source/client/options_impl.cc
```diff
@@ -390,6 +406,19 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
   TCLAP_SET_IF_SPECIFIED(labels, labels_);
   TCLAP_SET_IF_SPECIFIED(simple_warmup, simple_warmup_);
   TCLAP_SET_IF_SPECIFIED(no_duration, no_duration_);
+  if (stats_sinks.isSet()) {
+    envoy::config::metrics::v3::StatsSink sink;
```
eric846, line 410: I would move this inside the `for` loop to emphasize that it's not shared across iterations.

api/client/options.proto
qqustc: 33 has been used here
https://github.com/envoyproxy/nighthawk/blob/c57811f3626b53f997f0241aaa482366dfe790c0/api/client/options.proto#L109

source/client/options_impl.cc
qqustc: Done.

qqustc:
> LGTM modulo one nit

Thank you Eric for the quick review.

api/client/options.proto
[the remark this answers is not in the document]
qqustc: Yes, it is possible to configure a stats flush interval without specifying any sinks (it just does nothing since there is no sink configured).

qqustc:
> one question, otherwise looks good to me

Thanks Otto for the quick review.

api/client/options.proto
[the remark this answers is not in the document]
mum4k: I think @oschaaf has a point. If the user configures a flag that does nothing - that can end up being confusing. Can we add a validation where specifying interval without sink results in a command line error?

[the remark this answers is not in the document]
qqustc: thanks @mum4k . Added a check that if --stats-flush-interval is set, then --stats-sinks must also be set. Also verified that the check is not affected by the default value of  --stats-flush-interval.

PR #421, opened 2020-07-24
Update Envoy to c04c605a9a34139235de67f8027f257f3eec18d8

oschaaf:
Adds a foo implementation of version linkstamp constants
Tweaks python integration test requirements to unbreak CI. 


mum4k:

qqustc:
> @qqustc please review and assign back to me once done.

LGTM. (Though I'm not familiar with the change in this PR )
Assigned back to @mum4k for review.

external/envoy.patch
```diff
@@ -0,0 +1,24 @@
+diff --git source/common/version/BUILD source/common/version/BUILD
+index 9d72656..bc79c6a 100644
+--- source/common/version/BUILD
++++ source/common/version/BUILD
+@@ -81,3 +81,19 @@ envoy_basic_cc_library(
+     }),
+     alwayslink = 1,
+ )
++
++envoy_basic_cc_library(
```
mum4k, line 10: A few questions to help me understand:

1) what made this break, did Envoy recently make this private? Or did we just start depending on this?
2) Assuming tis hat Envoy doesn't want to make this public - how complicated would it be do define out own linkstamp target with the same constants? (I am assuming we are just missing the constant definition which we could get by having a nighthawk_version.cc with the linkstamp build flags.

oschaaf: Envoy shuffled some things and made the visibility of the link stamp cc private to avoid
having to rebuild a lot of things ever time. I'm not sure why, but at my first attempt I didn't
succeed by just adding the implementation for the constants as you suggested.
However, trying again, I do not run into problems doing so. So I was able to back the patch
out and do this more cleanly. The constants themselves are stubbed as we don't use them.
It's a separate target because we don't want to link this into our test server executable, as doing
that would result in multiple definitions. 
Let me know if this looks better.

Additionally, something broke with respect to python requirements, which must be related to 
something that changed in the CI env. I added some changes in here that hopefully resolve that,
and hopefully will also resolve the CI issues that the other PRs seem to have run into simultaneously.
Apparently we're doing something wrong, I don't really understand what happened here and why
certain requirements all of a sudden are no longer available.

oschaaf: I think f24824e fixes the problem; that should work consistently across my machine and CI.
Let's see if CI agrees. This should also fix the broken CI on master.

oschaaf:
This works; however coverage flaked. Restarted it.

mum4k:
@oschaaf can you please update the PR description to reflect the latest changes?

PR #425, opened 2020-07-31
Minimize the stock extension set for the test server

oschaaf:


oschaaf:
(quick experiment to check how this affects CI build time, cuts down the number of build actions from ~ 6K to 4K)

oschaaf:
cool, this seems to speed up the build CI task 2X:

https://app.circleci.com/pipelines/github/envoyproxy/nighthawk/1940/workflows/0c0da3ce-2e7b-47e0-b3e3-4b96251e4d56/jobs/15852

https://app.circleci.com/pipelines/github/envoyproxy/nighthawk/1941/workflows/0d5b191c-d6a3-4a24-a4e3-0634fd2f9f9a/jobs/15856

oschaaf:
Requesting a high level review for this first; as this needs tidying up before this is suitable for landing.
But first I'd like to hear if there's interest in landing this, there seems to be some benefit to it:
- (much) faster build times
- smaller binary
- "reduced surface": we explicitly list what we like to offer for the test server, instead of 
  having extension "creep in" through envoy updates. 

It might also be possible to diverge in different build types to produce a release with all extensions
enabled, but run CI with just the minimal set we need / the use-cases we have in mind.

oschaaf:
Some more CI numbers. The slowest running tests are asan and coverage.

- `asan` goes from ~ 1.5 hours to 1 hour
- `coverage` goes from ~ 1h05m to 50m 
- `test` goes from ~30min to ~20min
- `docker` image goes from ~30min to ~15min

oschaaf:
Made some changes as discussed in an offline chat: 
- minimised the number of extensions for the stock build
- documented usage

As there was consensus that there are multiple upsides to this, and this is
something we should do, this is now ready for code level review.

oschaaf:
CI time taken per task with the minimised extension set as of d3c6a7bf393f0b8b780565c9614d7a7bbd9f95e7:

```
tsan 27m 40s
test_gcc 35m 32s
test 32m 10s
docker 22m 45s
coverage 47m 24s
clang_tidy 26m 27s
build 10m 58s
benchmark 19m 35s
asan 57m 37s
```

mum4k:

source/server/README.md
```diff
@@ -8,6 +8,11 @@ A test-server filter which is capable of generating test responses.
 bazel test -c dbg //test/server:http_test_server_filter_integration_test
 ```
 
+It is possible to enable
```
dubious90, line 11: Should this be in the build section below, rather than the testing section?

dubious90, line 11: nit: can we say "enable additional envoy extensions"?

extensions_build_config.bzl
```diff
@@ -0,0 +1,9 @@
+# See https://github.com/envoyproxy/envoy/blob/master/bazel/README.md for details on how this system works.
```
dubious90, line 1: The link here provides an avalanche of information.  Do you have objections to using this link instead, which I think is to the more relevant section?
https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#disabling-extensions

extensions_build_config.bzl
```diff
@@ -0,0 +1,9 @@
...
+EXTENSIONS = {
+    "envoy.filters.http.router": "//source/extensions/filters/http/router:config",
+    "envoy.filters.http.fault": "//source/extensions/filters/http/fault:config",
+    "envoy.filters.listener.tls_inspector": "//source/extensions/filters/listener/tls_inspector:config",
+    "envoy.filters.network.http_connection_manager": "//source/extensions/filters/network/http_connection_manager:config",
+    "envoy.tracers.zipkin": "//source/extensions/tracers/zipkin:config",
+    "envoy.transport_sockets.raw_buffer": "//source/extensions/transport_sockets/raw_buffer:config",
```
dubious90, line 8: This is just my own ignorance of this, but why don't we need to include our test server filter here?

oschaaf:
@dubious90 review feedback applied in 06dc01e7f4fd346197bdf19130f0f0d9e810cc94

extensions_build_config.bzl
oschaaf: It's a good question, that didn't occur to me before. Tests are passing, so I think it is safe to assume that "foreign" extensions don't play along with this feature. We could look into proposing a feature over at the Envoy repository if that is desired?

oschaaf: (with "foreign", I meant extensions not residing in Envoy's own code base)

dubious90: My area of interest with this is partially on our own documentation around it.  If others want to add in their own test server equivalent building on what we already have, how would they do that?  But that's only tangentially related to this. So it's probably best to move on.

extensions_build_config.bzl
```diff
@@ -0,0 +1,9 @@
+# See https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#disabling-extensions for details on how this system works.
+EXTENSIONS = {
```
mum4k, line 2: Is there a complete list of extensions somewhere? I would like to take a look at what we are disabling and cross-check it with the functionality we are promising and using.

oschaaf: The full set is contained over at https://github.com/envoyproxy/envoy/blob/master/source/extensions/extensions_build_config.bzl

oschaaf:
@mum4k I added 7cdda88 in here, which updates Envoy and fixes build breakage against the latest;  what `extensions_build_config.bzl` should look like apparently is a fast moving target.

oschaaf:
In e863df4a5e54942793f3645e36d4e267c09ef7fa there's a change that leaves in all the extensions that can be enabled as comments. That might be helpful, but it also puts a burden on us to maintain that list. @dubious90 @mum4k wdyt?

Also, optimized CI parallellism, complete CI cycles now complete close to 4X faster. clang-tidy speeds up 5X because of the update to gen_compilation_database.py: we no longer actually build the project, as we don't need that.

oschaaf:
Summarising CI timings for master vs current state of this PR

| CI job | new duration | old duration |
| ---- | ---- | ---- |
tsan | 24m 38s | 39m 59s
test_gcc | 32m 39s | 48m 35s
test | 18m 9s | 27m 2s
format | 1m 17s | 1m 8s
docker | 16m 17s | 32m 42s
coverage | 27m 20s | 1h 12m 37s
clang_tidy | 10m 17s | 52m 13s
build | 13m 35s | 23m 58s
benchmark | 18m 18s | 36m 26s
asan | 32m 25s | 1h 29m 28s

oschaaf:
I respun CI a few times and I think this helps stabilize coverage and asan; we should see less flakes upon merging this. Also, running clang-tidy locally is going to be a much more frictionless thing.

dubious90:
Having trouble responding directly to your comment, but re: https://github.com/envoyproxy/nighthawk/pull/425#issuecomment-669133317

I think that it is much more desirable to have your original behavior, rather than this, because of the lower overhead of mainenance, and it is in my opinion, harder to read for pertinent information, in my opinion.

oschaaf:
@dubious90 backed out the full commented list of available extensions in e198c0b0633145072c931836eb46a4b7e2b3ada9

PR #427, opened 2020-07-31
Adaptive Load scoring functions

eric846:


source/adaptive_load/scoring_function_impl.cc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+REGISTER_FACTORY(SigmoidScoringFunctionConfigFactory, ScoringFunctionConfigFactory);
+
+SigmoidScoringFunction::SigmoidScoringFunction(
+    const nighthawk::adaptive_load::SigmoidScoringFunctionConfig& config)
+    : threshold_{config.threshold()}, k_{config.k()} {}
+
+double SigmoidScoringFunction::EvaluateMetric(double value) const {
+  return 1.0 - 2.0 / (1.0 + exp(-k_ * (value - threshold_)));
```
oschaaf, line 85: do we need to worry about overflow with respect to the call to `exp()`?

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,89 @@
...
+public:
+  explicit SigmoidScoringFunction(
+      const nighthawk::adaptive_load::SigmoidScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
+
+private:
+  // The target value of the metric.
+  double threshold_;
+  // Tuning constant: k in 1 - 2 / (1 + exp(-k(value - threshold))). k should
+  // be around the same size as 1/threshold.
```
oschaaf, line 76: it would be helpful (at least to me) if the comment would explain the rationale for why k should be around 1/threshold.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,89 @@
...
+// -1.0 for values far above the threshold.
+class SigmoidScoringFunction : public ScoringFunction {
+public:
+  explicit SigmoidScoringFunction(
+      const nighthawk::adaptive_load::SigmoidScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
+
+private:
+  // The target value of the metric.
+  double threshold_;
```
oschaaf, line 74: const

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,89 @@
...
+  // Tuning constant: k in 1 - 2 / (1 + exp(-k(value - threshold))). k should
+  // be around the same size as 1/threshold.
+  double k_;
```
oschaaf, line 77: const

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,231 @@
...
+namespace {
+
+TEST(BinaryScoringFunctionConfigFactoryTest, GeneratesEmptyConfigProto) {
+  ScoringFunctionConfigFactory& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<ScoringFunctionConfigFactory>(
+          "nighthawk.binary");
+  Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+
+  nighthawk::adaptive_load::BinaryScoringFunctionConfig expected_config;
+  EXPECT_EQ(message->DebugString(), expected_config.DebugString());
```
oschaaf, line 17: I learned that with proto3 default values may not get serialised. I'm not 100% sure that equivalence here guarantees actual equivalence with respect to set/not-set/default-value. Therefore, to make sure no edge cases slip past, maybe it makes sense to also double down with `MessageUtil()(message1, message1)`. This won't produce a helpful test message, so I'd put it after comparing the debug strings (which should produce a nice diff), maybe in a helper method.

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  EXPECT_LT(scoring_function.EvaluateMetric(1001.0), 0.0);
+}
+
+TEST(SigmoidScoringFunctionTest, ReturnsValueCloseToOneForValueFarBelowThreshold) {
+  nighthawk::adaptive_load::SigmoidScoringFunctionConfig config;
+  config.set_threshold(1000.0);
+  config.set_k(0.001);
+  SigmoidScoringFunction scoring_function(config);
+  const double score = scoring_function.EvaluateMetric(-10000.0);
+  EXPECT_GT(score, 0.99);
```
oschaaf, line 215: There's also EXPECT_NEAR. Not sure if it's needed here, but in the statistics test we also have a helper which facilitates verifying expected precision against output of a reference implementation:

https://github.com/envoyproxy/nighthawk/blob/master/test/statistic_test.cc#L180

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,231 @@
...
+
+TEST(LinearScoringFunctionTest, ReturnsNegativeValueForValueAboveThreshold) {
+  nighthawk::adaptive_load::LinearScoringFunctionConfig config;
+  config.set_threshold(10.0);
+  config.set_k(1.0);
+  LinearScoringFunction scoring_function(config);
+  EXPECT_EQ(scoring_function.EvaluateMetric(12.0), -2.0);
+}
+
+TEST(SigmoidScoringFunctionTest, ReturnsZeroForValueEqualToThreshold) {
```
oschaaf, line 185: For the sigmoid implementation especially, I wonder if it would make sense to generate a gold file with scipy's version of it, with a bunch of random numbers and some explicit edge cases as inputs, and verify that?
Also floating an idea -- I'm out of my league here -- but reading up [1] on the internets, I wonder if offering variants of it is worth something. Of course, that could be an excellent subject for experimentation in another plugin :-)

[1] https://stats.stackexchange.com/questions/330559/why-is-tanh-almost-always-better-than-sigmoid-as-an-activation-function

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,89 @@
...
+// Supports upper or lower threshold or both.
+class BinaryScoringFunction : public ScoringFunction {
+public:
+  explicit BinaryScoringFunction(
+      const nighthawk::adaptive_load::BinaryScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
+
+private:
+  // Upper threshold for the metric.
+  double upper_threshold_;
```
oschaaf, line 19: const here and below

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,89 @@
...
+// proportional to the difference from the threshold.
+class LinearScoringFunction : public ScoringFunction {
+public:
+  explicit LinearScoringFunction(
+      const nighthawk::adaptive_load::LinearScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
+
+private:
+  // The target value of the metric.
+  double threshold_;
```
oschaaf, line 45: const here and below

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,231 @@
+#include "external/envoy/source/common/config/utility.h"
+
+#include "adaptive_load/scoring_function_impl.h"
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+namespace {
```
oschaaf, line 8: If using an anonymous namespace here is what we want, we may want to file an issue to retrofit it everwhere for consistency.

test/adaptive_load/scoring_function_test.cc
eric846: Done, filed #433 to create the helper once all these small PRs are done.

source/adaptive_load/scoring_function_impl.h
eric846: Done

source/adaptive_load/scoring_function_impl.h
eric846: Done

source/adaptive_load/scoring_function_impl.cc
eric846: Ended up deleting this function -- we won't need it for months, and we might need to put more thought into it.

source/adaptive_load/scoring_function_impl.h
eric846: I actually don't know much about this, I just found that through trial and error. Removing the function entirely for now, since we won't need it yet.

source/adaptive_load/scoring_function_impl.h
eric846: Done

test/adaptive_load/scoring_function_test.cc
eric846: Will come back to this in #433

source/adaptive_load/scoring_function_impl.h
eric846: Done

test/adaptive_load/scoring_function_test.cc
eric846: Filed #431

test/adaptive_load/scoring_function_test.cc
eric846: Deleted these tests, will remember EXPECT_NEAR.

mum4k:
@pamorgan please review and assign back to me once done.

source/adaptive_load/scoring_function_impl.cc
```diff
@@ -0,0 +1,64 @@
+#include "adaptive_load/scoring_function_impl.h"
+
+#include <cmath>
+
+#include "envoy/registry/registry.h"
+
```
pamorgan, line 6: nit: not sure if you meant for the extra spacing between headers.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+  const double lower_threshold_;
+};
+
+// Factory that creates a BinaryScoringFunction from a BinaryScoringFunctionConfig proto.
+// Registered as an Envoy plugin.
+class BinaryScoringFunctionConfigFactory : public ScoringFunctionConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  ScoringFunctionPtr createScoringFunction(const Envoy::Protobuf::Message& message) override;
```
pamorgan, line 30: Can you comment that this function will throw if the message cannot be unpacked to BinaryScoringFunctionConfig.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+
+// ScoringFunction that calculates a metric score as k * (threshold - value), where k is a scaling
+// constant. The score is 0.0 when the value exactly equals the threshold, positive below the
+// threshold (meaning load should increase), and negative above the threshold. The score is
+// proportional to the difference from the threshold.
+class LinearScoringFunction : public ScoringFunction {
+public:
+  explicit LinearScoringFunction(
+      const nighthawk::adaptive_load::LinearScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
```
pamorgan, line 41: Is it possible to class enum instead of the double?

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+
+private:
+  // The target value of the metric.
+  const double threshold_;
+  // Scaling constant: k in k * (threshold - value). Use this in combination
+  // with step controller constants to produce reasonable load increments for
+  // reasonable differences from the threshold.
+  const double k_;
```
pamorgan, line 49: can we rename k_ to a more readable name like scaling_constant_?

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// Factory that creates a LinearScoringFunction from a LinearScoringFunctionConfig proto.
+// Registered as an Envoy plugin.
+class LinearScoringFunctionConfigFactory : public ScoringFunctionConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  ScoringFunctionPtr createScoringFunction(const Envoy::Protobuf::Message& message) override;
```
pamorgan, line 58: Please comment on the throwing expectation.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// Factory that creates a BinaryScoringFunction from a BinaryScoringFunctionConfig proto.
+// Registered as an Envoy plugin.
+class BinaryScoringFunctionConfigFactory : public ScoringFunctionConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  ScoringFunctionPtr createScoringFunction(const Envoy::Protobuf::Message& message) override;
+};
```
pamorgan, line 31: Can you please add DECLARE_FACTORY(BinaryScoringFunctionConfigFactory) and comment on how the factory is expected to be used.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,61 @@
...
+
+// Factory that creates a LinearScoringFunction from a LinearScoringFunctionConfig proto.
+// Registered as an Envoy plugin.
+class LinearScoringFunctionConfigFactory : public ScoringFunctionConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  ScoringFunctionPtr createScoringFunction(const Envoy::Protobuf::Message& message) override;
+};
```
pamorgan, line 59: same comment, we might need a DECLARE_FACTORY

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,180 @@
...
+
+TEST(BinaryScoringFunctionConfigFactoryTest, GeneratesEmptyConfigProto) {
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<ScoringFunctionConfigFactory>(
+          "nighthawk.binary");
+  const Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
```
pamorgan, line 14: message seem like a generic variable name, can we name this a more specific name like empty_config

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,180 @@
...
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<ScoringFunctionConfigFactory>(
+          "nighthawk.linear");
+  const Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+  const nighthawk::adaptive_load::LinearScoringFunctionConfig expected_config;
+  EXPECT_EQ(message->DebugString(), expected_config.DebugString());
+  EXPECT_TRUE(Envoy::MessageUtil()(*message, expected_config));
+}
+
+TEST(BinaryScoringFunctionConfigFactoryTest, CreatesBinaryScoringFunctionFactory) {
```
pamorgan, line 30: can we test the negative case were it throws.

source/adaptive_load/scoring_function_impl.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "external/envoy/source/common/protobuf/protobuf.h"
+
+#include "api/adaptive_load/scoring_function_impl.pb.h"
+
+namespace Nighthawk {
+
+std::string BinaryScoringFunctionConfigFactory::name() const { return "nighthawk.binary"; }
```
pamorgan, line 13: nighthawk.binary doesn't seem to be an descriptive enough name, we might need to add nighthawk.binary_scoring.

source/adaptive_load/scoring_function_impl.cc
```diff
@@ -0,0 +1,64 @@
...
+    : upper_threshold_{config.has_upper_threshold() ? config.upper_threshold().value()
+                                                    : std::numeric_limits<double>::infinity()},
+      lower_threshold_{config.has_lower_threshold() ? config.lower_threshold().value()
+                                                    : -std::numeric_limits<double>::infinity()} {}
+
+double BinaryScoringFunction::EvaluateMetric(double value) const {
+  return value <= upper_threshold_ && value >= lower_threshold_ ? 1.0 : -1.0;
+}
+
+std::string LinearScoringFunctionConfigFactory::name() const { return "nighthawk.linear"; }
```
pamorgan, line 40: same comment "nighthawk.linear" need a more descriptive name since it has to be unique.

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,180 @@
...
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<ScoringFunctionConfigFactory>(
+          "nighthawk.binary");
+  const Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+  const nighthawk::adaptive_load::BinaryScoringFunctionConfig expected_config;
+  EXPECT_EQ(message->DebugString(), expected_config.DebugString());
+  EXPECT_TRUE(Envoy::MessageUtil()(*message, expected_config));
+}
+
+TEST(LinearScoringFunctionConfigFactoryTest, GeneratesEmptyConfigProto) {
```
pamorgan, line 20: Do we need the extra Test in the name?

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,180 @@
+#include "external/envoy/source/common/config/utility.h"
+
+#include "adaptive_load/scoring_function_impl.h"
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+namespace {
+
+TEST(BinaryScoringFunctionConfigFactoryTest, GeneratesEmptyConfigProto) {
```
pamorgan, line 10: can you include the method you are testing in the test name. From only the name I understand the result but not what you are testing

eric846: Done, actually renamed every test here.

test/adaptive_load/scoring_function_test.cc
eric846: Removed

source/adaptive_load/scoring_function_impl.cc
eric846: Done

source/adaptive_load/scoring_function_impl.cc
eric846: Done

test/adaptive_load/scoring_function_test.cc
eric846: Done

source/adaptive_load/scoring_function_impl.h
eric846: Added in the interface header.

source/adaptive_load/scoring_function_impl.h
eric846: Good to know, I will add this throughout.

source/adaptive_load/scoring_function_impl.h
eric846: Done

test/adaptive_load/scoring_function_test.cc
eric846: Done

source/adaptive_load/scoring_function_impl.h
eric846: Added comment in the interface header

source/adaptive_load/scoring_function_impl.h
eric846: Apparently enum class can only be based on an integer type.

source/adaptive_load/scoring_function_impl.h
eric846: Done

source/adaptive_load/scoring_function_impl.cc
eric846: Yes, if you remove the lines, they come back after running ci/do_ci.sh fix_format. I think we are seeing are 5 1-item groups of includes, based on group prefixes in tools/check_format.sh.

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,204 @@
...
+
+TEST(BinaryScoringFunction, EvaluateMetricReturnsOneForValueEqualToUpperThresholdOfRange) {
+  nighthawk::adaptive_load::BinaryScoringFunctionConfig config;
+  config.mutable_lower_threshold()->set_value(5.0);
+  config.mutable_upper_threshold()->set_value(7.0);
+  BinaryScoringFunction scoring_function(config);
+  EXPECT_EQ(scoring_function.EvaluateMetric(7.0), 1.0);
+}
+
+TEST(BinaryScoringFunction, EvaluateMetricReturnsNegativeOneForValueBelowThresholdRange) {
```
oschaaf, line 162: One last remark: there's a fair bit of repetitiveness here, which could be condensed via `TEST_P`.
We'd loose the helpful specificity of the test names, which indicate what is being tested. Maybe a hybrid of `TEST_P` and named constants could provide a middle ground?

eric846: Done, it's much simpler with TEST_P.

mum4k:
Looks like we're still missing an LGTM with @pamorgan, please complete the review with Peter and assign back to me once done.

eric846:
The CI environment appears to be broken at the moment. 

ci/do_ci.sh test does compile and pass locally, so the CI issue is probably unrelated to this PR.

oschaaf:
> The CI environment appears to be broken at the moment.
> 
> ci/do_ci.sh test does compile and pass locally, so the CI issue is probably unrelated to this PR.

Yeah, I don't think it's related to this. I'm trying some things over at https://github.com/envoyproxy/nighthawk/pull/451,
I don't understand what caused it yet.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,79 @@
...
+ * ScoringFunction that calculates a metric score as k * (threshold - value), where k is a scaling
+ * constant. The score is 0.0 when the value exactly equals the threshold, positive below the
+ * threshold (meaning load should increase), and negative above the threshold. The score is
+ * proportional to the difference from the threshold.
+ */
+class LinearScoringFunction : public ScoringFunction {
+public:
+  explicit LinearScoringFunction(
+      const nighthawk::adaptive_load::LinearScoringFunctionConfig& config);
+  double EvaluateMetric(double value) const override;
```
pamorgan, line 53: I was thinking of a using a class enum and then if/else in the implementation that uses the magic double for each case.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
```
mum4k, line 1: Since this file contains multiple classes, let's add a file level comment.

source/adaptive_load/scoring_function_impl.h
eric846: Filed issue #473 to discuss this -- we could potentially refactor this.

source/adaptive_load/scoring_function_impl.h
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Nighthawk {
+
+/**
+ * ScoringFunction that returns 1.0 when a metric is within thresholds and -1.0 otherwise.
+ * Supports upper or lower threshold or both.
+ */
+class BinaryScoringFunction : public ScoringFunction {
+public:
+  explicit BinaryScoringFunction(
```
mum4k, line 17: Let's document the constructors.

Here and below.

test/adaptive_load/scoring_function_test.cc
```diff
@@ -0,0 +1,177 @@
...
+#include "adaptive_load/scoring_function_impl.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include <cmath>
+
+namespace Nighthawk {
+
+namespace {
+
+using nighthawk::adaptive_load::BinaryScoringFunctionConfig;
```
mum4k, line 14: Can we fully qualify these using statements? It's the recommended practice to improve code maintainability, since partially qualified statements might accidentally match a relative namespace path.

source/adaptive_load/scoring_function_impl.h
eric846: Done.

source/adaptive_load/scoring_function_impl.h
eric846: Done.

test/adaptive_load/scoring_function_test.cc
eric846: Done.

PR #428, opened 2020-07-31
Adaptive Load input variable setter

eric846:
Adaptive Load InputVariableSetter plugin implementation files.

Starting with one plugin, RequestsPerSecondInputVariableSetter, which simply sets the requests_per_second field in the input.


test/adaptive_load/input_variable_setter_test.cc
```diff
@@ -0,0 +1,45 @@
...
+  nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig config;
+  Envoy::ProtobufWkt::Any config_any;
+  config_any.PackFrom(config);
+
+  InputVariableSetterConfigFactory& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<InputVariableSetterConfigFactory>(
+          "nighthawk.rps");
+  InputVariableSetterPtr plugin = config_factory.createInputVariableSetter(config_any);
+
+  EXPECT_NE(dynamic_cast<RequestsPerSecondInputVariableSetter*>(plugin.get()), nullptr);
```
oschaaf, line 32: Maybe check the plugin's `name()` yields "nighthawk.rps"?

test/adaptive_load/input_variable_setter_test.cc
```diff
@@ -0,0 +1,45 @@
...
+}
+
+TEST(RequestsPerSecondInputVariableSetterTest, SetsCommandLineOptionsRpsValue) {
+  nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig config;
```
oschaaf, line 36: const

source/adaptive_load/input_variable_setter_impl.cc
```diff
@@ -0,0 +1,38 @@
...
+  command_line_options.mutable_requests_per_second()->set_value(
+      static_cast<unsigned int>(input_value));
+}
+
+std::string RequestsPerSecondInputVariableSetterConfigFactory::name() const {
+  return "nighthawk.rps";
+}
+
+Envoy::ProtobufTypes::MessagePtr
+RequestsPerSecondInputVariableSetterConfigFactory::createEmptyConfigProto() {
```
oschaaf, line 23: I think this could have been a `const` method, but updating that probably requires changing the interface which already landed.

source/adaptive_load/input_variable_setter_impl.h
```diff
@@ -0,0 +1,28 @@
...
+};
+
+// A factory that creates an RequestsPerSecondInputVariableSetter from an
+// RequestsPerSecondInputVariableSetterConfig proto.
+class RequestsPerSecondInputVariableSetterConfigFactory : public InputVariableSetterConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  InputVariableSetterPtr
+  createInputVariableSetter(const Envoy::Protobuf::Message& message) override;
```
oschaaf, line 25: const override?

test/adaptive_load/input_variable_setter_test.cc
```diff
@@ -0,0 +1,45 @@
...
+namespace {
+
+TEST(RequestsPerSecondInputVariableSetterConfigFactoryTest, GeneratesEmptyConfigProto) {
+  InputVariableSetterConfigFactory& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<InputVariableSetterConfigFactory>(
+          "nighthawk.rps");
+
+  Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+
+  nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig expected_config;
```
oschaaf, line 17: const

source/adaptive_load/input_variable_setter_impl.cc
eric846: Yes, opened #430

source/adaptive_load/input_variable_setter_impl.h
eric846: The compiler won't let me, but this will be covered by #430.

test/adaptive_load/input_variable_setter_test.cc
eric846: Done

test/adaptive_load/input_variable_setter_test.cc
eric846: Done

test/adaptive_load/input_variable_setter_test.cc
eric846: Added another test for the factory name. I still want to test the final created class because the implementations of these factories are always copy-pasted from other factories that all return `InputVariableSetterPtr`.

source/adaptive_load/input_variable_setter_impl.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "api/adaptive_load/input_variable_setter_impl.pb.h"
+
+namespace Nighthawk {
+
+// An InputVariableSetter that sets the |requests_per_second| field in the CommandLineOptions proto.
+class RequestsPerSecondInputVariableSetter : public InputVariableSetter {
+public:
+  RequestsPerSecondInputVariableSetter(
+      const nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig& config);
```
oschaaf, line 13: Maybe `RequestsPerSecondInputVariableSetter(
      const nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig& config) = default;` and drop the (empty) implementation from the .cc?

test/adaptive_load/input_variable_setter_test.cc
```diff
@@ -0,0 +1,57 @@
...
+
+namespace {
+
+TEST(RequestsPerSecondInputVariableSetterConfigFactoryTest, GeneratesEmptyConfigProto) {
+  InputVariableSetterConfigFactory& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<InputVariableSetterConfigFactory>(
+          "nighthawk.rps");
+
+  Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+
```
oschaaf, line 16: superfluous newlines? (personally I'm fine with this as is -- but at some point it was encouraged and we reached consensus to avoid being liberal with newlines, which is what I think a lot of the other codebase is leaning towards. Though it might not be super consistent as we don't enforce it).

eric846: Cleaned that up now.



I'll be on the lookout...

source/adaptive_load/input_variable_setter_impl.h
eric846: It doesn't seem to like that, maybe it doesn't work when the constructor takes an arg that isn't this class.

......input_variable_setter_impl.h:13:86: error: only special member functions may be defaulted
      const nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig&) = default;

oschaaf: Oh, sorry, yes that makes sense, I missed that. Sorry about the noise.

mum4k:

source/adaptive_load/input_variable_setter_impl.cc
```diff
@@ -0,0 +1,38 @@
+#include "adaptive_load/input_variable_setter_impl.h"
+
+#include "envoy/registry/registry.h"
+
+#include "external/envoy/source/common/protobuf/protobuf.h"
+
+namespace Nighthawk {
+
+RequestsPerSecondInputVariableSetter::RequestsPerSecondInputVariableSetter(
+    const nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig&) {}
```
qqustc, line 10: just currious, if RequestsPerSecondInputVariableSetterConfig is not used, do you still need this constructor?

source/adaptive_load/input_variable_setter_impl.cc
```diff
@@ -0,0 +1,38 @@
...
+RequestsPerSecondInputVariableSetterConfigFactory::createEmptyConfigProto() {
+  return std::make_unique<nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig>();
+}
+
+InputVariableSetterPtr RequestsPerSecondInputVariableSetterConfigFactory::createInputVariableSetter(
+    const Envoy::Protobuf::Message& message) {
+  const auto& any = dynamic_cast<const Envoy::ProtobufWkt::Any&>(message);
+  nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig config;
+  Envoy::MessageUtil::unpackTo(any, config);
+  return std::make_unique<RequestsPerSecondInputVariableSetter>(config);
```
qqustc, line 32: since config is not used in the above constructor, will it be used in the future?

qqustc:
> LGTM

Assigned back to @mum4k for review

source/adaptive_load/input_variable_setter_impl.cc
eric846: I'm a bit torn about this. Filed #438 to consider deleting the proto and this 1-arg constructor in a followup PR.

source/adaptive_load/input_variable_setter_impl.cc
eric846: Good point, see #438.

source/adaptive_load/input_variable_setter_impl.cc
qqustc: SG, thanks Eric.

source/adaptive_load/input_variable_setter_impl.h
```diff
@@ -0,0 +1,28 @@
...
+#include "nighthawk/adaptive_load/input_variable_setter.h"
+
+#include "api/adaptive_load/input_variable_setter_impl.pb.h"
+
+namespace Nighthawk {
+
+// An InputVariableSetter that sets the |requests_per_second| field in the CommandLineOptions proto.
+class RequestsPerSecondInputVariableSetter : public InputVariableSetter {
+public:
+  RequestsPerSecondInputVariableSetter(
```
mum4k, line 12: Can we document the constructor?

test/adaptive_load/input_variable_setter_test.cc
```diff
@@ -0,0 +1,51 @@
...
+          "nighthawk.rps");
+  InputVariableSetterPtr plugin = config_factory.createInputVariableSetter(config_any);
+  EXPECT_NE(dynamic_cast<RequestsPerSecondInputVariableSetter*>(plugin.get()), nullptr);
+}
+
+TEST(RequestsPerSecondInputVariableSetterTest, SetsCommandLineOptionsRpsValue) {
+  const nighthawk::adaptive_load::RequestsPerSecondInputVariableSetterConfig config;
+  RequestsPerSecondInputVariableSetter setter(config);
+  nighthawk::client::CommandLineOptions options;
+  setter.SetInputVariable(options, 5.0);
```
mum4k, line 45: Should we have a test where the double is set to some floating point value? E.g. 5.3?

Additionally maybe tests for negative and very large values?

source/adaptive_load/input_variable_setter_impl.h
eric846: Done

test/adaptive_load/input_variable_setter_test.cc
eric846: I realized SetInputVariable needs an opportunity to return errors. Updated the interface and added tests for these input cases.

include/nighthawk/adaptive_load/input_variable_setter.h
```diff
@@ -28,9 +30,12 @@ class InputVariableSetter {
    * @param command_line_options A Nighthawk Service CommandLineOptions load spec template with most
    * fields already filled.
    * @param input_value A value generated by a StepController to be applied to the next load spec.
+   *
+   * @return Status OK if the value could be applied, or an error if it could not be applied (e.g.
+   * out of range).
    */
-  virtual void SetInputVariable(nighthawk::client::CommandLineOptions& command_line_options,
-                                double input_value) PURE;
+  virtual absl::Status SetInputVariable(nighthawk::client::CommandLineOptions& command_line_options,
```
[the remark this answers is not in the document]
oschaaf: No reasons to avoid that I can think of. I think the concept fits well here.

PR #429, opened 2020-07-31
Adaptive Load builtin metrics plugin

eric846:
Adaptive Load "builtin" MetricsPlugin that extracts metrics from latency stats and counters in Nighthawk responses.


source/adaptive_load/metrics_plugin_impl.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Nighthawk {
+
+// Emulated MetricPlugin that wraps already collected Nighthawk Service counters and stats in a
+// MetricPlugin interface. This class is not registered with the Envoy registry mechanism. It will
+// be constructed on the fly from each Nighthawk Service result.
+class NighthawkStatsEmulatedMetricsPlugin : public MetricsPlugin {
+public:
+  explicit NighthawkStatsEmulatedMetricsPlugin(const nighthawk::client::Output& nighthawk_output);
+  double GetMetricByName(const std::string& metric_name) override;
```
oschaaf, line 19: I wonder would `s/override/const override/` make sense?

test/adaptive_load/metrics_plugin_test.cc
```diff
@@ -0,0 +1,103 @@
...
+}
+
+TEST(NighthawkStatsEmulatedMetricsPluginTest, ComputesCorrectMeanLatencyPlus3Stdev) {
+  NighthawkStatsEmulatedMetricsPlugin plugin =
+      NighthawkStatsEmulatedMetricsPlugin(MakeStandardNighthawkOutput());
+  double value = plugin.GetMetricByName("latency-ns-mean-plus-3stdev");
+  EXPECT_EQ(value, 533.0);
+}
+
+TEST(NighthawkStatsEmulatedMetricsPluginTest, ReturnsZeroForNonexistentMetric) {
```
oschaaf, line 83: Up to here the test pattern seems pretty monotonic. I wonder if this could be condensed by leveraging `TEST_P`.

test/adaptive_load/utility.cc
```diff
@@ -0,0 +1,45 @@
...
+  statistic->mutable_mean()->set_nanos(mean_ns % kOneBillion);
+  statistic->mutable_max()->set_seconds(max_ns / kOneBillion);
+  statistic->mutable_max()->set_nanos(max_ns % kOneBillion);
+  statistic->mutable_pstdev()->set_seconds(pstdev_ns / kOneBillion);
+  statistic->mutable_pstdev()->set_nanos(pstdev_ns % kOneBillion);
+}
+
+} // namespace
+
+nighthawk::client::Output MakeStandardNighthawkOutput() {
```
oschaaf, line 28: possible overkill, but this could take a (deterministic) random generator to add the potential to test with some more entropy. entirely optional, just floating an idea.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+        static_cast<double>(nighthawk_output.options().duration().seconds());
+    metric_from_name_["send-rate"] =
+        static_cast<double>(total_sent) / static_cast<double>(total_specified);
+    metric_from_name_["success-rate"] =
+        static_cast<double>(total_2xx) / static_cast<double>(total_sent);
+
+    for (const nighthawk::client::Statistic& statistic : result.statistics()) {
+      if (statistic.id() == "benchmark_http_client.request_to_response") {
+        double min = statistic.min().seconds() * 1.0e9 + statistic.min().nanos();
+        double mean = statistic.mean().seconds() * 1.0e9 + statistic.mean().nanos();
```
oschaaf, line 48: We have the `const long kOneBillion = 1000 * 1000 * 1000;`, which is nice. Consider moving that to a place where we can share it, and re-use that here for consistency? I think existing places in the code bases could benefit from this too; out of scope here, but might be worth filing an issue to suggest a pass to clean that up afterwards.

eric846: I ended up finding some Google time utilities for dealing with Durations -- filed #434 to apply it elsewhere.

test/adaptive_load/metrics_plugin_test.cc
eric846: Done

source/adaptive_load/metrics_plugin_impl.h
eric846: Yep, will fix this in #430.

test/adaptive_load/utility.cc
```diff
@@ -0,0 +1,45 @@
+#include "test/adaptive_load/utility.h"
+
+namespace Nighthawk {
+
+namespace {
+
+void SetCounterValue(nighthawk::client::Counter* counter, const std::string& name, int value) {
```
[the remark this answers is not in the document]
eric846: Done

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+        static_cast<double>(total_sent) /
+        static_cast<double>(nighthawk_output.options().duration().seconds());
+    metric_from_name_["send-rate"] =
+        static_cast<double>(total_sent) / static_cast<double>(total_specified);
+    metric_from_name_["success-rate"] =
+        static_cast<double>(total_2xx) / static_cast<double>(total_sent);
+
+    for (const nighthawk::client::Statistic& statistic : result.statistics()) {
+      if (statistic.id() == "benchmark_http_client.request_to_response") {
+        double min = statistic.min().seconds() * 1.0e9 + statistic.min().nanos();
```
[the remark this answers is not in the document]
eric846: Done

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+      if (counter.name() == "benchmark.http_2xx") {
+        total_2xx = counter.value();
+      } else if (counter.name() == "upstream_rq_total") {
+        total_sent = counter.value();
+      }
+    }
+    metric_from_name_["attempted-rps"] =
+        static_cast<double>(total_specified) /
+        static_cast<double>(nighthawk_output.options().duration().seconds());
+    metric_from_name_["achieved-rps"] =
```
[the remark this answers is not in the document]
eric846: Done

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Nighthawk {
+
+NighthawkStatsEmulatedMetricsPlugin::NighthawkStatsEmulatedMetricsPlugin(
+    const nighthawk::client::Output& nighthawk_output) {
+  for (const nighthawk::client::Result& result : nighthawk_output.results()) {
+    if (result.name() != "global") {
+      continue;
+    }
+    int concurrency;
```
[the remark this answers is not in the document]
eric846: Replaced with alternate code.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+        metric_from_name_["latency-ns-mean-plus-1stdev"] = mean + stdev;
+        metric_from_name_["latency-ns-mean-plus-2stdev"] = mean + 2 * stdev;
+        metric_from_name_["latency-ns-mean-plus-3stdev"] = mean + 3 * stdev;
+        break;
+      }
+    }
+  }
+}
+
+double NighthawkStatsEmulatedMetricsPlugin::GetMetricByName(const std::string& metric_name) {
```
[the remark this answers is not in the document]
eric846: Filed #435 to convert the headers.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,85 @@
...
+namespace Nighthawk {
+
+NighthawkStatsEmulatedMetricsPlugin::NighthawkStatsEmulatedMetricsPlugin(
+    const nighthawk::client::Output& nighthawk_output) {
+  for (const nighthawk::client::Result& result : nighthawk_output.results()) {
+    if (result.name() != "global") {
+      continue;
+    }
+    int concurrency;
+    // "auto" is not supported since we have no way to detect the number of cores Nighthawk Service
```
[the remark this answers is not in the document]
eric846: Thanks, this is a huge improvement for both the user and us if we simply support "auto" instead of making it an error.

test/adaptive_load/utility.cc
eric846: The random generator might be overkill for now, but I did end up expanding this a bit.

source/adaptive_load/metrics_plugin_impl.h
eric846: I realized this particular method is non-const on purpose because a real metrics plugin will have members that are RPC connections that can't be const.

oschaaf: Sure, makes sense to be non-const then.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#include "absl/strings/numbers.h"
+
+namespace Nighthawk {
+
+using ::Envoy::Protobuf::util::TimeUtil;
+
+NighthawkStatsEmulatedMetricsPlugin::NighthawkStatsEmulatedMetricsPlugin(
+    const nighthawk::client::Output& nighthawk_output) {
+  // Values that could not be determined for any reason are omitted from the map.
```
oschaaf, line 16: So as I understand it, incomplete results will be handled up higher in the stack?
I wonder, should we be able to handle different causes of that distinctly? Concretely,
should metrics plugins have an opportunity to have a say about that and interpret / indicate
failure (or status in general)?

test/adaptive_load/utility.cc
```diff
@@ -0,0 +1,46 @@
+#include "test/adaptive_load/utility.h"
+
+#include <google/protobuf/util/time_util.h>
```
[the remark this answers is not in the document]
oschaaf: fwiw, I'm surprised check-format didn't catch this, we might want to code in a check as I think we tend to avoid
depending on proto directly; the idea is that we do so by proxy via Envoy's facilities.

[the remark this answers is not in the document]
eric846: Oh yes, forgot about that. It turns out I had even already included protobuf.h.

source/adaptive_load/metrics_plugin_impl.cc
eric846: This 0.0 in the interface never really sat well with me. I brought in Envoy::StatusOr and introduced a bunch of error handling.

This does mean metrics_plugin.h needs to change, and it's just done in this PR.

oschaaf: Nice, I didn’t know about ‘StatusOr’ but that seems like a good solution 👍

mum4k:
@wjuan-AFK please review and assign to me once done.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+#include "external/envoy/source/common/protobuf/protobuf.h"
+
+#include "absl/status/status.h"
+#include "absl/strings/numbers.h"
+
+namespace Nighthawk {
+
+NighthawkStatsEmulatedMetricsPlugin::NighthawkStatsEmulatedMetricsPlugin(
+    const nighthawk::client::Output& nighthawk_output) {
+  errors_ = "'global' result not found in Nighthawk output.\n";
```
wjuan-AFK, line 15: Why do we start by setting this errors_ property?

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+        total_2xx = counter.value();
+      } else if (counter.name() == "upstream_rq_total") {
+        total_sent = counter.value();
+      }
+    }
+    if (total_2xx < 0) {
+      errors_ += "Counter 'benchmark.total_2xx' not found.\n";
+    }
+    if (total_sent < 0) {
+      errors_ += "Counter 'upstream_rq_total' not found.\n";
```
wjuan-AFK, line 41: I don't like having the possibility of having multiple errors stored in this one string. Could we use a string vector or some such thing, and then just combine the errors together when we need them?

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (actual_duration_seconds > 0.0) {
+      metric_from_name_["attempted-rps"] =
+          static_cast<double>(total_specified) / actual_duration_seconds;
```
wjuan-AFK, line 45: Maybe I'm missing something really obvious. Why do we have a static cast to double for something we just declared as long higher up? Why not just declare it as a double if we want a double?

test/adaptive_load/utility.cc
```diff
@@ -0,0 +1,42 @@
...
+  statistic->set_id(std::string(id));
+  *statistic->mutable_min() = Envoy::Protobuf::util::TimeUtil::NanosecondsToDuration(min_ns);
+  *statistic->mutable_mean() = Envoy::Protobuf::util::TimeUtil::NanosecondsToDuration(mean_ns);
+  *statistic->mutable_max() = Envoy::Protobuf::util::TimeUtil::NanosecondsToDuration(max_ns);
+  *statistic->mutable_pstdev() = Envoy::Protobuf::util::TimeUtil::NanosecondsToDuration(pstdev_ns);
+}
+
+} // namespace
+
+nighthawk::client::Output MakeSimpleNighthawkOutput(const SimpleNighthawkOutputSpec& spec) {
```
wjuan-AFK, line 27: Might be helpful to have some comment on this one. I didn't find this as clear as the other two.

test/adaptive_load/metrics_plugin_test.cc
```diff
@@ -0,0 +1,233 @@
...
+      /*max_ns=*/600,
+      /*pstdev_ns=*/11,
+  });
+  output.mutable_results(0)->clear_counters();
+  NighthawkStatsEmulatedMetricsPlugin plugin = NighthawkStatsEmulatedMetricsPlugin(output);
+  EXPECT_THAT(plugin.GetMetricByName("x").status().message(),
+              testing::HasSubstr("'benchmark.total_2xx' not found"));
+}
+
+TEST(NighthawkStatsEmulatedMetricsPluginTest, ReturnsErrorIfUpstreamRqTotalMissing) {
```
wjuan-AFK, line 75: Could these two not be the same test with both expectations? Or is that bad practice?

test/adaptive_load/metrics_plugin_test.cc
```diff
@@ -0,0 +1,233 @@
...
+  std::vector<std::string> supported_metrics = plugin.GetAllSupportedMetricNames();
+  EXPECT_THAT(supported_metrics,
+              ::testing::ElementsAre("attempted-rps", "achieved-rps", "send-rate", "success-rate",
+                                     "latency-ns-min", "latency-ns-mean", "latency-ns-max",
+                                     "latency-ns-mean-plus-1stdev", "latency-ns-mean-plus-2stdev",
+                                     "latency-ns-mean-plus-3stdev", "latency-ns-pstdev"));
+}
+
+} // namespace
+
```
wjuan-AFK, line 232: Would it make sense to also test that if you have a worker that has valid statistics and a worker that has erroneous statistics what would happen? Or multiple instances of global and such?

source/adaptive_load/metrics_plugin_impl.cc
eric846: I was thinking of the calculation being clearer if we compute the result as the type it really is, then cast it to double only when forced by the map that expects doubles.

test/adaptive_load/utility.cc
eric846: This one is commented over in utility.h.

source/adaptive_load/metrics_plugin_impl.cc
eric846: Changed to a vector.

test/adaptive_load/metrics_plugin_test.cc
eric846: I split this into two tests because it's technically two behaviors. Theoretically a future rewrite could break one behavior but not the other.

> Not sure my comments were sensible.

Absolutely, thanks for the review!

test/adaptive_load/metrics_plugin_test.cc
eric846: For workers with erroneous statistics, and for duplicated "global," we're trusting the unit tests of the Nighthawk code that computes the "global" statistics.

I don't know for sure whether "global" can ever be missing (e.g. in the event of a Nighthawk Service error), but the implementation checks for it just in case.

source/adaptive_load/metrics_plugin_impl.cc
wjuan-AFK: I guess so, but it feels a little weird when we only ever use it as a double.

eric846: I moved all the double casts to the time of initialization, and it does seem simpler now.

wjuan-AFK:
LGTM

test/adaptive_load/metrics_plugin_test.cc
oschaaf: re: "I don't know for sure whether "global" can ever be missing (e.g. in the event of a Nighthawk Service error), but the implementation checks for it just in case."

I can imagine situations where Nighthawk's could successfully terminate without any latency measurements. E.g. lots of random jitter, a very short execution duration and a very low qps: chances are no requests will be initiated before execution terminates based on exceeding duration.  An edge case, but not unthinkable.

include/nighthawk/adaptive_load/metrics_plugin.h
```diff
@@ -20,9 +22,10 @@ class MetricsPlugin {
    *
    * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
    *
-   * @return double The metric value, or 0.0 if the metric was unsupported or unavailable.
+   * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
+   * unavailable.
    */
-  virtual double GetMetricByName(const std::string& metric_name) PURE;
+  virtual Envoy::StatusOr<double> GetMetricByName(absl::string_view metric_name) PURE;
```
mum4k, line 28: Curious - why is this Envoy::StatusOr rather than absl::StatusOr?

source/adaptive_load/metrics_plugin_impl.h
```diff
@@ -0,0 +1,32 @@
...
+namespace Nighthawk {
+
+// Emulated MetricPlugin that wraps already collected Nighthawk Service counters and stats in a
+// MetricPlugin interface. This class is not registered with the Envoy registry mechanism. It will
+// be constructed on the fly from each Nighthawk Service result.
+class NighthawkStatsEmulatedMetricsPlugin
+    : public MetricsPlugin,
+      public Envoy::Logger::Loggable<Envoy::Logger::Id::main> {
+public:
+  explicit NighthawkStatsEmulatedMetricsPlugin(const nighthawk::client::Output& nighthawk_output);
```
mum4k, line 23: Let's document the constructor.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+
+#include "external/envoy/source/common/common/assert.h"
+#include "external/envoy/source/common/protobuf/protobuf.h"
+
+#include "absl/status/status.h"
+#include "absl/strings/numbers.h"
+
+namespace Nighthawk {
+
+NighthawkStatsEmulatedMetricsPlugin::NighthawkStatsEmulatedMetricsPlugin(
```
mum4k, line 15: This is a fairly long function, any way how we can extract some logical sub-portions from it into well named helper functions in an anonymous space? That would help the reader since they can follow the business logic first in this function and dive into the details later.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+    const nighthawk::client::Output& nighthawk_output) {
+  bool found_global_result = false;
+  for (const nighthawk::client::Result& result : nighthawk_output.results()) {
+    if (result.name() != "global") {
```
mum4k, line 19: To the above comment - e.g. this portion could be a helper function that retrieves the global result. And so on.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,123 @@
...
+        absl::StatusCode::kInternal,
+        absl::StrCat("Metric '", metric_name, "' was not computed by the 'builtin' plugin.")));
+  }
+  return metric_from_name_[metric_name];
+}
+
+const std::vector<std::string>
+NighthawkStatsEmulatedMetricsPlugin::GetAllSupportedMetricNames() const {
+  return {
+      "attempted-rps",
```
mum4k, line 107: How would you feel about sorting this list lexicographically?

test/adaptive_load/utility.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
mum4k, line 1: Can we give this library a name that is a bit less generic? File named like "utility" or "tools" will simply attract anything (since anything is a utility) and may become hard to manage.

Consider: minimal_output.(h|cc)

include/nighthawk/adaptive_load/metrics_plugin.h
wjuan-AFK: I had the same question but I don't think absl::StatusOr is open yet. I think that the reason it's in envoy is because it hasn't been made public.

test/adaptive_load/metrics_plugin_test.cc
eric846: Thanks, good to know.

include/nighthawk/adaptive_load/metrics_plugin.h
eric846: I found it's actually possible to use absl::StatusOr.

source/adaptive_load/metrics_plugin_impl.h
eric846: Done

source/adaptive_load/metrics_plugin_impl.cc
eric846: Done.

source/adaptive_load/metrics_plugin_impl.cc
eric846: Done.

source/adaptive_load/metrics_plugin_impl.cc
eric846: Done.

test/adaptive_load/utility.h
eric846: Done.

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,221 @@
...
+ */
+void ExtractCounters(const nighthawk::client::Output& nighthawk_output,
+                     absl::flat_hash_map<std::string, double>& metric_from_name,
+                     std::vector<std::string>& errors) {
+  nighthawk::client::Result global_result;
+  if (!ExtractValueOrRecordError(GetResult(nighthawk_output, "global"), global_result, errors)) {
+    return;
+  }
+  const int64_t actual_duration_seconds =
+      Envoy::Protobuf::util::TimeUtil::DurationToSeconds(global_result.execution_duration());
```
mum4k, line 118: Have you considered aliasing these long statements (this one and the one to Nanoseconds) via a using statement in the anonymous namespace?

source/adaptive_load/metrics_plugin_impl.cc
```diff
@@ -0,0 +1,221 @@
...
+ * @param status_or The StatusOr to extract. The value must be static_castable to the type of
+ * |value|.
+ * @param value The variable where the value should be extracted. Only written if the status is OK.
+ * @param errors A place to accumulate error messages. If the status is not OK, its message is
+ * appended here.
+ *
+ * @return bool True if the status was OK.
+ */
+template <typename T, typename U>
+bool ExtractValueOrRecordError(absl::StatusOr<T> status_or, U& value,
```
mum4k, line 91: I understand the motivation behind this template (removal of code duplication), however we should also consider the complexity it adds for the reader, it is both a redirection and a meta code just to process errors.

While the code without this template would contain a few more lines on each error check, what we would get in return is lower barrier of adoption, especially for potential contributors who aren't very versed with c++. Let's try to keep the code as simple as possible even if it means having a few repeated lines.

source/adaptive_load/metrics_plugin_impl.cc
eric846: Done

source/adaptive_load/metrics_plugin_impl.cc
eric846: Done

PR #437, opened 2020-08-04
Add the functionality to flush stats to sinks periodically in NH

qqustc:
Add a flush worker to periodically flush stats to sinks.

Add the functionality to configure sinks in process_impl.cc

Start the flush worker in process_impl.cc when there is stats sink configured.

Change the SinkableStatistic's tagExtractedName() to return `worker_id + "." + Nighthawk::Statistic::id()` which can be used in customized sinks.

Linked Issues: #344

Signed-off-by: Qin Qin <qqin@google.com>

### DIFF
--- a/include/nighthawk/common/factories.h
+++ b/include/nighthawk/common/factories.h
@@ -6,6 +6,7 @@
 #include "envoy/common/pure.h"
 #include "envoy/common/time.h"
 #include "envoy/event/dispatcher.h"
+#include "envoy/stats/symbol_table.h"
 #include "envoy/upstream/cluster_manager.h"
 
 #include "nighthawk/common/platform_util.h"
@@ -46,6 +47,28 @@ public:
   virtual TerminationPredicatePtr
   create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
          const Envoy::MonotonicTime scheduled_starting_time) const PURE;
+};
+
+/**
+ * Factory Interface to create Envoy::Stats::Sink in Nighthawk.
+ * Implemented for each Envoy::Stats::Sink and registered via
+ * Registry::registerFactory() or the convenience class RegisterFactory.
+ */
+class NighthawkStatsSinkFactory : public Envoy::Config::TypedFactory {
+public:
+  ~NighthawkStatsSinkFactory() override = default;
+
+  /**
+   * Create a particular Envoy::Stats::Sink implementation. If the
+   * implementation is unable to produce a Sink with the provided parameters, it
+   * should throw an EnvoyException. The returned pointer should always be
+   * valid.
+   * @param symbol_table supplies the symbol_table instance
+   */
+  virtual std::unique_ptr<Envoy::Stats::Sink>
+  createStatsSink(Envoy::Stats::SymbolTable& symbol_table) PURE;
+
+  std::string category() const override { return "nighthawk.stats_sinks"; }
 };
 
 } // namespace Nighthawk
--- a/include/nighthawk/common/factories.h
+++ b/include/nighthawk/common/factories.h
@@ -63,7 +63,8 @@ public:
    * implementation is unable to produce a Sink with the provided parameters, it
    * should throw an EnvoyException. The returned pointer should always be
    * valid.
-   * @param symbol_table supplies the symbol_table instance
+   * @param symbol_table supplies the symbol_table instance. For the definition
+   * of SymbolTable, see envoy/include/envoy/stats/symbol_table.h.
    */
   virtual std::unique_ptr<Envoy::Stats::Sink>
   createStatsSink(Envoy::Stats::SymbolTable& symbol_table) PURE;
--- a/source/client/flush_worker_impl.cc
+++ b/source/client/flush_worker_impl.cc
@@ -1,0 +1,51 @@
+#include "client/flush_worker_impl.h"
+
+#include "external/envoy/source/common/stats/symbol_table_impl.h"
+#include "external/envoy/source/server/server.h"
+
+#include "common/utility.h"
+
+namespace Nighthawk {
+namespace Client {
+
+FlushWorkerImpl::FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+                                 Envoy::Stats::Store& store,
+                                 std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                                 const std::chrono::milliseconds& stats_flush_interval)
+    : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
+  for (auto& sink : stats_sinks) {
+    stats_sinks_.emplace_back(std::move(sink));
+  }
+}
+
+void FlushWorkerImpl::work() {
+  stat_flush_timer_ = dispatcher_->createTimer([this]() -> void { flushStats(); });
+  stat_flush_timer_->enableTimer(stats_flush_interval_);
+
+  dispatcher_->run(Envoy::Event::Dispatcher::RunType::RunUntilExit);
+}
+
+void FlushWorkerImpl::shutdownThread() {
+  if (stat_flush_timer_ != nullptr) {
+    stat_flush_timer_->disableTimer();
+    stat_flush_timer_.reset();
+  }
+  // Make the final flush before the flush worker gets shutdown.
+  flushStats();
+}
+
+void FlushWorkerImpl::flushStats() {
+  // Create a snapshot and flush to all sinks. Even if there are no sinks,
+  // creating the snapshot has the important property that it latches all counters on a periodic
+  // basis.
+  Envoy::Server::MetricSnapshotImpl snapshot(store_);
+  for (auto& sink : stats_sinks_) {
+    sink->flush(snapshot);
+  }
+  if (stat_flush_timer_ != nullptr) {
+    stat_flush_timer_->enableTimer(stats_flush_interval_);
+  }
+}
+
+} // namespace Client
+} // namespace Nighthawk
--- a/source/client/flush_worker_impl.cc
+++ b/source/client/flush_worker_impl.cc
@@ -21,7 +21,6 @@ FlushWorkerImpl::FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Insta
 void FlushWorkerImpl::work() {
   stat_flush_timer_ = dispatcher_->createTimer([this]() -> void { flushStats(); });
   stat_flush_timer_->enableTimer(stats_flush_interval_);
-
   dispatcher_->run(Envoy::Event::Dispatcher::RunType::RunUntilExit);
 }
 
--- a/source/client/flush_worker_impl.cc
+++ b/source/client/flush_worker_impl.cc
@@ -8,10 +8,9 @@
 namespace Nighthawk {
 namespace Client {
 
-FlushWorkerImpl::FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+FlushWorkerImpl::FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval, Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
                                  Envoy::Stats::Store& store,
-                                 std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
-                                 const std::chrono::milliseconds& stats_flush_interval)
+                                 std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks)
     : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
   for (auto& sink : stats_sinks) {
     stats_sinks_.emplace_back(std::move(sink));
--- a/source/client/flush_worker_impl.cc
+++ b/source/client/flush_worker_impl.cc
@@ -38,7 +38,7 @@ void FlushWorkerImpl::flushStats() {
   // creating the snapshot has the important property that it latches all counters on a periodic
   // basis.
   Envoy::Server::MetricSnapshotImpl snapshot(store_);
-  for (auto& sink : stats_sinks_) {
+  for (std::unique_ptr<Envoy::Stats::Sink>& sink : stats_sinks_) {
     sink->flush(snapshot);
   }
   if (stat_flush_timer_ != nullptr) {
--- a/source/client/flush_worker_impl.cc
+++ b/source/client/flush_worker_impl.cc
@@ -8,7 +8,8 @@
 namespace Nighthawk {
 namespace Client {
 
-FlushWorkerImpl::FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval, Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+FlushWorkerImpl::FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval,
+                                 Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
                                  Envoy::Stats::Store& store,
                                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks)
     : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -1,0 +1,50 @@
+// Flush worker implementation. Flush worker periodically flushes metrics
+// snapshot to all configured stats sinks in Nighthawk.
+#pragma once
+
+#include <vector>
+
+#include "envoy/api/api.h"
+#include "envoy/event/dispatcher.h"
+#include "envoy/stats/sink.h"
+#include "envoy/stats/store.h"
+#include "envoy/thread_local/thread_local.h"
+
+#include "common/worker_impl.h"
+
+namespace Nighthawk {
+namespace Client {
+
+// Flush worker periodically flushes metrics snapshot to all configured stats
+// sinks in Nighthawk. A single flush worker is created in Nighthawk
+// process_impl.cc. It will keep running until exitDispatcher() gets called
+// after all client workers are completed in process_impl.cc. It will make the
+// last flush before shutdown in shutdownThread().
+class FlushWorkerImpl : public WorkerImpl {
+public:
+  FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+                  Envoy::Stats::Store& store,
+                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                  const std::chrono::milliseconds& stats_flush_interval);
+
+  void shutdownThread() override;
+
+  // Must be called in after all client workers are completed in
+  // process_impl.cc.
+  void exitDispatcher() { dispatcher_->exit(); }
+
+protected:
+  void work() override;
+
+private:
+  // Flush the stats sinks. Note: stats flushing may not be synchronous.
+  // Therefore, this function may return prior to flushing taking place.
+  void flushStats();
+
+  std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
+  std::chrono::milliseconds stats_flush_interval_;
+  Envoy::Event::TimerPtr stat_flush_timer_;
+};
+
+} // namespace Client
+} // namespace Nighthawk
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -32,8 +32,8 @@ public:
   // @param store supplies the stats store instance for WorkerImpl's constructor.
   // @param stats_sinks list of configured stats sinks where the stats will be
   // flushed to.
-  FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval, Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
-                  Envoy::Stats::Store& store,
+  FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval, Envoy::Api::Api& api,
+                  Envoy::ThreadLocal::Instance& tls, Envoy::Stats::Store& store,
                   std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks);
 
   void shutdownThread() override;
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -29,8 +29,9 @@ public:
 
   void shutdownThread() override;
 
-  // Must be called in after all client workers are completed in
-  // process_impl.cc.
+  // exitDispatcher() stops the dispatcher and the flush timer running in flush worker. It must be
+  // called after all client workers are completed in process_impl.cc to make sure all metrics will
+  // be flushed.
   void exitDispatcher() { dispatcher_->exit(); }
 
 protected:
@@ -42,7 +43,7 @@ private:
   void flushStats();
 
   std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
-  std::chrono::milliseconds stats_flush_interval_;
+  const std::chrono::milliseconds stats_flush_interval_;
   Envoy::Event::TimerPtr stat_flush_timer_;
 };
 
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -38,8 +38,8 @@ protected:
   void work() override;
 
 private:
-  // Flush the stats sinks. Note: stats flushing may not be synchronous.
-  // Therefore, this function may return prior to flushing taking place.
+  // Flush the stats sinks. Note: stats flushing may not be synchronous, depending on each stat
+  // sink's implementation. Therefore, this function may return prior to flushing taking place.
   void flushStats();
 
   std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -22,10 +22,19 @@ namespace Client {
 // process_impl.cc. It will make the last flush before shutdown in shutdownThread().
 class FlushWorkerImpl : public WorkerImpl {
 public:
-  FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+  // Constructor to call parent class's constructor and initialize member
+  // variables stats_sinks_ and stats_flush_interval_.
+  // @param stats_flush_interval time interval between each flush.
+  // @param api supplies the Api instance for WorkerImpl's constructor. See
+  // envoy/include/envoy/api/api.h for its definition.
+  // @param tls supplies the ThreadLocal::Instance for WorkerImpl's constructor.
+  // See envoy/include/envoy/thread_local/thread_local.h for its definition.
+  // @param store supplies the stats store instance for WorkerImpl's constructor.
+  // @param stats_sinks list of configured stats sinks where the stats will be
+  // flushed to.
+  FlushWorkerImpl(const std::chrono::milliseconds& stats_flush_interval, Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
                   Envoy::Stats::Store& store,
-                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
-                  const std::chrono::milliseconds& stats_flush_interval);
+                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks);
 
   void shutdownThread() override;
 
--- a/source/client/flush_worker_impl.h
+++ b/source/client/flush_worker_impl.h
@@ -15,11 +15,11 @@
 namespace Nighthawk {
 namespace Client {
 
-// Flush worker periodically flushes metrics snapshot to all configured stats
-// sinks in Nighthawk. A single flush worker is created in Nighthawk
-// process_impl.cc. It will keep running until exitDispatcher() gets called
-// after all client workers are completed in process_impl.cc. It will make the
-// last flush before shutdown in shutdownThread().
+// Only a single live flush worker instance can be created in Nighthawk at any
+// time.
+// Flush worker periodically flushes metrics snapshot to all configured stats sinks in Nighthawk. It
+// will keep running until exitDispatcher() gets called after all client workers are completed in
+// process_impl.cc. It will make the last flush before shutdown in shutdownThread().
 class FlushWorkerImpl : public WorkerImpl {
 public:
   FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -429,6 +429,23 @@ void ProcessImpl::addRequestSourceCluster(
   socket->set_port_value(uri.port());
 }
 
+void ProcessImpl::setupStatsSinks(
+    envoy::config::bootstrap::v3::Bootstrap& bootstrap,
+    std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks) {
+  for (const envoy::config::metrics::v3::StatsSink& stats_sink :
+         bootstrap.stats_sinks()) {
+      ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
+      auto& factory =
+          Envoy::Config::Utility::getAndCheckFactory<NighthawkStatsSinkFactory>(
+              stats_sink);
+      stats_sinks.emplace_back(
+          factory.createStatsSink(store_root_.symbolTable()));
+    }
+    for (std::unique_ptr<Envoy::Stats::Sink>& sink : stats_sinks) {
+      store_root_.addSink(*sink);
+    }
+}
+
 bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriPtr>& uris,
                               const UriPtr& request_source_uri, const UriPtr& tracing_uri) {
   {
@@ -478,24 +495,13 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
     Runtime::LoaderSingleton::get().initialize(*cluster_manager_);
 
     std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks;
-    for (const envoy::config::metrics::v3::StatsSink& stats_sink : bootstrap.stats_sinks()) {
-      ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
-      auto& factory =
-          Envoy::Config::Utility::getAndCheckFactory<NighthawkStatsSinkFactory>(stats_sink);
-      stats_sinks.emplace_back(factory.createStatsSink(store_root_.symbolTable()));
-    }
-    for (std::unique_ptr<Envoy::Stats::Sink>& sink : stats_sinks) {
-      store_root_.addSink(*sink);
-    }
-
-    // Default flush interval is 5s.
+    setupStatsSinks(bootstrap, stats_sinks);
     std::chrono::milliseconds stats_flush_interval = std::chrono::milliseconds(
         Envoy::DurationUtil::durationToMilliseconds(bootstrap.stats_flush_interval()));
 
     if (!options_.statsSinks().empty()) {
       // There should be only a single live flush worker instance at any time.
-      flush_worker_ = std::make_unique<FlushWorkerImpl>(*api_, tls_, store_root_, stats_sinks,
-                                                        stats_flush_interval);
+      flush_worker_ = std::make_unique<FlushWorkerImpl>(stats_flush_interval, *api_, tls_, store_root_, stats_sinks);
       flush_worker_->start();
     }
 
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -204,7 +204,7 @@ void ProcessImpl::configureComponentLogLevels(spdlog::level::level_enum level) {
 
 uint32_t ProcessImpl::determineConcurrency() const {
   // Spare 1 cpu for flush worker.
-  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount() - 1;
+  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount();
   bool autoscale = options_.concurrency() == "auto";
   // TODO(oschaaf): Maybe, in the case where the concurrency flag is left out, but
   // affinity is set / we don't have affinity with all cores, we should default to autoscale.
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -429,7 +429,7 @@ void ProcessImpl::addRequestSourceCluster(
   socket->set_port_value(uri.port());
 }
 
-void ProcessImpl::setupStatsSinks(envoy::config::bootstrap::v3::Bootstrap& bootstrap,
+void ProcessImpl::setupStatsSinks(const envoy::config::bootstrap::v3::Bootstrap& bootstrap,
                                   std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks) {
   for (const envoy::config::metrics::v3::StatsSink& stats_sink : bootstrap.stats_sinks()) {
     ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -9,9 +9,11 @@
 #include <random>
 
 #include "envoy/server/filter_config.h"
+#include "envoy/stats/sink.h"
 #include "envoy/stats/store.h"
 
 #include "nighthawk/client/output_collector.h"
+#include "nighthawk/common/factories.h"
 
 #include "external/envoy/source/common/api/api_impl.h"
 #include "external/envoy/source/common/common/cleanup.h"
@@ -24,6 +26,7 @@
 #include "external/envoy/source/common/runtime/runtime_impl.h"
 #include "external/envoy/source/common/singleton/manager_impl.h"
 #include "external/envoy/source/common/thread_local/thread_local_impl.h"
+#include "external/envoy/source/server/server.h"
 
 #include "absl/strings/str_replace.h"
 
@@ -133,6 +136,9 @@ void ProcessImpl::shutdown() {
       worker->shutdown();
     }
     workers_.clear();
+    if (flush_worker_) {
+      flush_worker_->shutdown();
+    }
   }
   if (cluster_manager_ != nullptr) {
     cluster_manager_->shutdown();
@@ -192,7 +198,8 @@ void ProcessImpl::configureComponentLogLevels(spdlog::level::level_enum level) {
 }
 
 uint32_t ProcessImpl::determineConcurrency() const {
-  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount();
+  // Spare 1 cpu for flush worker.
+  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount() - 1;
   bool autoscale = options_.concurrency() == "auto";
   // TODO(oschaaf): Maybe, in the case where the concurrency flag is left out, but
   // affinity is set / we don't have affinity with all cores, we should default to autoscale.
@@ -324,6 +331,11 @@ void ProcessImpl::createBootstrapConfiguration(envoy::config::bootstrap::v3::Boo
       addRequestSourceCluster(*request_source_uri, i, bootstrap);
     }
   }
+
+  for (const envoy::config::metrics::v3::StatsSink& stats_sink : options_.statsSinks()) {
+    *bootstrap.add_stats_sinks() = stats_sink;
+  }
+  bootstrap.mutable_stats_flush_interval()->set_seconds(options_.statsFlushInterval());
 }
 
 void ProcessImpl::addTracingCluster(envoy::config::bootstrap::v3::Bootstrap& bootstrap,
@@ -422,6 +434,12 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
     }
     int number_of_workers = determineConcurrency();
     shutdown_ = false;
+    envoy::config::bootstrap::v3::Bootstrap bootstrap;
+    createBootstrapConfiguration(bootstrap, uris, request_source_uri, number_of_workers);
+    // Needs to happen as early as possible (before createWorkers()) in the instantiation to preempt
+    // the objects that require stats.
+    store_root_.setTagProducer(Envoy::Config::Utility::createTagProducer(bootstrap));
+
     createWorkers(number_of_workers);
     tls_.registerThread(*dispatcher_, true);
     store_root_.initializeThreading(*dispatcher_, tls_);
@@ -441,8 +459,6 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
             ? Http1PoolImpl::ConnectionReuseStrategy::LRU
             : Http1PoolImpl::ConnectionReuseStrategy::MRU);
     cluster_manager_factory_->setPrefetchConnections(options_.prefetchConnections());
-    envoy::config::bootstrap::v3::Bootstrap bootstrap;
-    createBootstrapConfiguration(bootstrap, uris, request_source_uri, number_of_workers);
     if (tracing_uri != nullptr) {
       setupTracingImplementation(bootstrap, *tracing_uri);
       addTracingCluster(bootstrap, *tracing_uri);
@@ -455,6 +471,26 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
 
     Runtime::LoaderSingleton::get().initialize(*cluster_manager_);
 
+    std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks;
+    for (const envoy::config::metrics::v3::StatsSink& stats_sink : bootstrap.stats_sinks()) {
+      ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
+      auto& factory =
+          Envoy::Config::Utility::getAndCheckFactory<NighthawkStatsSinkFactory>(stats_sink);
+      stats_sinks.emplace_back(factory.createStatsSink(store_root_.symbolTable()));
+    }
+    for (auto& sink : stats_sinks) {
+      store_root_.addSink(*sink);
+    }
+
+    // Default flush interval is 5s.
+    std::chrono::milliseconds stats_flush_interval = std::chrono::milliseconds(
+        Envoy::DurationUtil::durationToMilliseconds(bootstrap.stats_flush_interval()));
+
+    flush_worker_ = std::make_unique<FlushWorkerImpl>(*api_, tls_, store_root_, stats_sinks,
+                                                      stats_flush_interval);
+
+    flush_worker_->start();
+
     for (auto& w : workers_) {
       w->start();
     }
@@ -462,6 +498,11 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
   for (auto& w : workers_) {
     w->waitForCompletion();
   }
+
+  // Stop the running dispatcher in flush_worker_. Need to be called after all
+  // client workers are complete so that all the metrics can be flushed.
+  flush_worker_->exitDispatcher();
+  flush_worker_->waitForCompletion();
 
   int i = 0;
   std::chrono::nanoseconds total_execution_duration = 0ns;
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -134,7 +134,7 @@ void ProcessImpl::shutdown() {
     // flush_worker_->shutdown() needs to happen before workers_.clear() so that
     // metrics defined in workers scope will be included in the final stats
     // flush which happens in FlushWorkerImpl::shutdownThread() after
-    // flush_worker_->shutdown(). For the order between worker shutdown() and
+    // flush_worker_->shutdown() is called. For the order between worker shutdown() and
     // shutdownThread(), see worker_impl.cc.
     if (flush_worker_) {
       flush_worker_->shutdown();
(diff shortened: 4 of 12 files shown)


source/client/process_impl.cc
```diff
@@ -192,7 +198,8 @@ void ProcessImpl::configureComponentLogLevels(spdlog::level::level_enum level) {
 }
 
 uint32_t ProcessImpl::determineConcurrency() const {
-  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount();
+  // Spare 1 cpu for flush worker.
+  uint32_t cpu_cores_with_affinity = Envoy::OptionsImplPlatform::getCpuCount() - 1;
```
qqustc, line 202: @oschaaf Do we need to reserve a cpu core for the flush worker?

oschaaf: I think this is a very good question. IMHO, the main thread and the flush worker would probably best be combined on a distinct dedicated core (if that luxury is an option): 
- this makes sure that the workers responsible for generating load won't be disturbed
- cpu consumption will be easily observable as spikes won't jump from core to core. chances of problematic behaviour slipping by unnoticed will be drastically lower.
Having said that, we don't do anything special here for the main thread either (but that's pretty much idle during load generation). Maybe we should discuss this on Slack for a bit with other devs to hear what they think about this?

include/nighthawk/common/factories.h
```diff
@@ -48,4 +49,26 @@ class TerminationPredicateFactory {
...
+   * Create a particular Envoy::Stats::Sink implementation. If the
+   * implementation is unable to produce a Sink with the provided parameters, it
+   * should throw an EnvoyException. The returned pointer should always be
+   * valid.
+   * @param symbol_table supplies the symbol_table instance
+   */
+  virtual std::unique_ptr<Envoy::Stats::Sink>
+  createStatsSink(Envoy::Stats::SymbolTable& symbol_table) PURE;
+
+  std::string category() const override { return "nighthawk.stats_sinks"; }
```
oschaaf, line 71: As this is in the interface, can we make this `PURE` and punt implementing this to a base implementation or the `FlushWorker` directly?

source/client/flush_worker_impl.h
```diff
@@ -0,0 +1,50 @@
...
+protected:
+  void work() override;
+
+private:
+  // Flush the stats sinks. Note: stats flushing may not be synchronous.
+  // Therefore, this function may return prior to flushing taking place.
+  void flushStats();
+
+  std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
+  std::chrono::milliseconds stats_flush_interval_;
```
oschaaf, line 45: const?

source/client/flush_worker_impl.h
```diff
@@ -0,0 +1,50 @@
...
+class FlushWorkerImpl : public WorkerImpl {
+public:
+  FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+                  Envoy::Stats::Store& store,
+                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                  const std::chrono::milliseconds& stats_flush_interval);
+
+  void shutdownThread() override;
+
+  // Must be called in after all client workers are completed in
```
oschaaf, line 32: Does this comment need updating to make it about FlushWorker? Also, tangentially related, do we want to ensure a single instance?

source/client/process_impl.cc
```diff
@@ -422,6 +434,12 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
     }
     int number_of_workers = determineConcurrency();
     shutdown_ = false;
+    envoy::config::bootstrap::v3::Bootstrap bootstrap;
+    createBootstrapConfiguration(bootstrap, uris, request_source_uri, number_of_workers);
+    // Needs to happen as early as possible (before createWorkers()) in the instantiation to preempt
+    // the objects that require stats.
+    store_root_.setTagProducer(Envoy::Config::Utility::createTagProducer(bootstrap));
```
oschaaf, line 441: Does this need to happen conditionally, when the sink is enabled through config? Or will this be a true no-op otherwise?

source/client/process_impl.cc
```diff
@@ -455,6 +471,26 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
...
+    }
+    for (auto& sink : stats_sinks) {
+      store_root_.addSink(*sink);
+    }
+
+    // Default flush interval is 5s.
+    std::chrono::milliseconds stats_flush_interval = std::chrono::milliseconds(
+        Envoy::DurationUtil::durationToMilliseconds(bootstrap.stats_flush_interval()));
+
+    flush_worker_ = std::make_unique<FlushWorkerImpl>(*api_, tls_, store_root_, stats_sinks,
```
oschaaf, line 489: We can skip this when we know there will be nothing to flush right? We could either opt to not do it, but we could also add a `FlushWorkerImpl` which just stubs the interface to keep the flow identical (might make testing easier too).

test/flush_worker_test.cc
```diff
@@ -0,0 +1,130 @@
...
+};
+
+TEST_F(FlushWorkerTest, WorkerFlushStatsPeriodically) {
+  setupDispatcherTimerEmulation();
+
+  auto worker = std::make_unique<FlushWorkerImpl>(mock_api_, tls_, store_, stats_sinks_,
+                                                  stats_flush_interval_);
+
+  std::thread thread = std::thread([&worker] {
+    sleep(1);
```
oschaaf, line 102: I feel it would be better to wait on some condition to be signalled here, instead of using `sleep()`, to try and get a deterministic order of events?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,130 @@
...
+  Envoy::Random::RandomGeneratorImpl rand_;
+  NiceMock<Envoy::Event::MockDispatcher>* dispatcher_ = nullptr;
+  std::unique_ptr<Envoy::Runtime::ScopedLoaderSingleton> loader_;
+  NiceMock<Envoy::LocalInfo::MockLocalInfo> local_info_;
+  NiceMock<Envoy::ProtobufMessage::MockValidationVisitor> validation_visitor_;
+
+  NiceMock<Envoy::Event::MockTimer>* timer_; // not owned
+  Envoy::Event::TimerCb timer_cb_;
+  bool timer_set_{};
+  bool stopped_{};
```
oschaaf, line 88: use `GUARDED_BY`

test/flush_worker_test.cc
```diff
@@ -0,0 +1,130 @@
...
+
+namespace Nighthawk {
+namespace Client {
+
+using namespace testing;
+
+class FlushWorkerTest : public Test {
+public:
+  FlushWorkerTest() : thread_factory_(Envoy::Thread::threadFactoryForTest()) {
+    dispatcher_ = new NiceMock<Envoy::Event::MockDispatcher>();
```
oschaaf, line 27: `std::make_unique`?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,130 @@
...
+    loader_ = std::make_unique<Envoy::Runtime::ScopedLoaderSingleton>(
+        Envoy::Runtime::LoaderPtr{new Envoy::Runtime::LoaderImpl(
+            *dispatcher_, tls_, {}, local_info_, store_, rand_, validation_visitor_, mock_api_)});
+    sink_ = new StrictMock<Envoy::Stats::MockSink>();
```
oschaaf, line 31: Maybe remove this field, and add a getter to obtain this pointer from the `stats_sinks_` vector?

test/process_test.cc
```diff
@@ -22,6 +23,37 @@ using namespace testing;
...
+class DummyStatsSink : public Envoy::Stats::Sink {
+public:
+  DummyStatsSink() : dummy_(1) { kNumFlushes = 0; }
+
+  // Envoy::Stats::Sink
+  void flush(Envoy::Stats::MetricSnapshot&) override { kNumFlushes++; }
+
+  void onHistogramComplete(const Envoy::Stats::Histogram&, uint64_t) override {}
+
+  int dummy_ = 0;
```
oschaaf, line 38: `private:`?

oschaaf: Also, it's not obvious to the reader what `dummy_` is for, maybe add a short comment?

test/process_test.cc
```diff
@@ -22,6 +23,37 @@ using namespace testing;
 namespace Nighthawk {
 namespace Client {
 
+constexpr absl::string_view kSinkName = "{name:\"nighthawk.dummy_stats_sink\"}";
+int kNumFlushes = 0;
+
+class DummyStatsSink : public Envoy::Stats::Sink {
```
oschaaf, line 29: Maybe add a short comment describing purpose of `DummyStatsSink`?

test/process_test.cc
```diff
@@ -22,6 +23,37 @@ using namespace testing;
 namespace Nighthawk {
 namespace Client {
 
+constexpr absl::string_view kSinkName = "{name:\"nighthawk.dummy_stats_sink\"}";
```
oschaaf, line 26: While this might technically be OK and the absl folks seem really happy with it, I wonder what's wrong with `constexpr const char kFoo[] = "";`. Maybe we should discuss this in the context of NH too.
/cc @jmarantz because we had a similar back and fort a little while ago.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,130 @@
...
+TEST_F(FlushWorkerTest, FinalFlush) {
+  auto worker = std::make_unique<FlushWorkerImpl>(mock_api_, tls_, store_, stats_sinks_,
+                                                  stats_flush_interval_);
+
+  // Final flush should be done in FlushWorkerImpl::shutdownThread().
+  EXPECT_CALL(*sink_, flush(_)).Times(1);
+
+  worker->start();
+  worker->waitForCompletion();
+  worker->shutdown();
```
oschaaf, line 126: does this guarantee that the flush we observed is called after `FlushWorkerImpl::shutdownThread()`? I think that's what the crown jewel is that we are after here right? Catching that particular even is important, because if that breaks, my concern is that this will slip by unnoticed during usage if it breaks in the future.

include/nighthawk/common/factories.h
qqustc: Here I follow the Envoy example https://github.com/envoyproxy/envoy/blob/2a6a4b8d758b4bf4238cb6d2babdd3d83fbf45bc/source/server/configuration_impl.h#L47

I think the category() should be here since different implementation of NighthawkStatsSinkFactory should share the same category.

oschaaf: I see; I think the difference between the Envoy code you point out and this file, is that the Envoy code is an implementation (`configuration_impl.h` in `source/...`), whereas NHs interfaces generally reside in `includes/`. 
Hence the suggestion to declare this PURE here, and create a base class to implement the common factor that should be shared across implementations under `source/..` (implementing only the getter for the `category()`).  That way, the interface is clean (which helps with that mocking the interface stays trivial), and the implementation can still be shared, which is what you're after, right?

source/client/flush_worker_impl.h
qqustc: Done

source/client/flush_worker_impl.h
qqustc: Updated the comment.

By single instance, do you mean to make sure it is only called once?

oschaaf: Yes, I think we don't ever want two instances to be alive at the same time, correct?

source/client/process_impl.cc
qqustc: Add a condition to only call setTagProducer when there is sink enabled.

It is no-op (it sets the tag producer in the store to control tags) when there is no sink enabled. As in the Envoy example, it is called even when there is no sink enabled.
https://github.com/envoyproxy/envoy/blob/2a6a4b8d758b4bf4238cb6d2babdd3d83fbf45bc/source/server/server.cc#L324

source/client/process_impl.cc
qqustc: Thanks Otto for notice this. Added a condition to skip creating flush_worker_ when there is no sink configured.

test/flush_worker_test.cc
qqustc: Thanks Otto. Done.

test/flush_worker_test.cc
qqustc: mock_api_ will take the ownership of dispatcher_ by 
`EXPECT_CALL(mock_api_, allocateDispatcher_(_, _)).WillOnce(Return(dispatcher_));`
https://github.com/envoyproxy/envoy/blob/e8aab1b4f14baf7d7cf89297e0422dac6c7aac8e/test/mocks/api/mocks.cc#L23

test/flush_worker_test.cc
qqustc: Removing sink_ and replace it with the pointer getting from the stats_sinks_ seems not work in `EXPECT_CALL(*sink_, flush(_))` since stats_sinks_ is a list of `Sink` not `MockSink`

test/process_test.cc
qqustc: The `dummy_ ` has no meaning and is not used, so I removed it.

test/process_test.cc
qqustc: Done

test/process_test.cc
qqustc: Changed to `constexpr const char kFoo[] = "";` to follow NH convention

test/flush_worker_test.cc
qqustc: Yes, this will verify the only flush done is the final flush in `FlushWorkerImpl::shutdownThread()` (note according to current implementation, the final flush is done inside `shutdownThread()` not after).

Added comment to clarify this.

test/process_test.cc
oschaaf: Thanks; somehow I find that's easier for the reader to see that this won't cause trouble, as it's just a plain old data type.

test/flush_worker_test.cc
```diff
@@ -47,7 +48,10 @@ class FlushWorkerTest : public Test {
         .WillRepeatedly(Invoke([&](const std::chrono::microseconds,
                                    const Envoy::ScopeTrackedObject*) { timer_set_ = true; }));
     EXPECT_CALL(*timer_, disableTimer()).WillOnce(Invoke([&]() { timer_set_ = false; }));
-    EXPECT_CALL(*dispatcher_, exit()).WillOnce(Invoke([&]() { stopped_ = true; }));
+    EXPECT_CALL(*dispatcher_, exit()).WillOnce(Invoke([&]() {
+      absl::MutexLock l(&lock_);
```
oschaaf, line 52: Sorry, I should have thought of this earlier, but thoughout the code base we've been using `Envoy::Thread::LockGuard` [1] in favor of `absl::MutexLock`. I bet one is implemented in terms of the other, but for consistency it might be good to stick to `Envoy::Thread::LockGuard`?

[1] https://github.com/envoyproxy/envoy/blob/7e88243610c09a8346226b49528bc7516fcfd7a7/source/common/common/lock_guard.h#L95

test/flush_worker_test.cc
oschaaf: Would there be anything against making stats_sinks_ a list of `MockSink`?

test/flush_worker_test.cc
oschaaf: I see, makes sense then, sorry about the noise.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+  }
+
+  NiceMock<Envoy::Api::MockApi> mock_api_;
+  Envoy::Thread::ThreadFactory& thread_factory_;
+  Envoy::Stats::IsolatedStoreImpl store_;
+  NiceMock<Envoy::ThreadLocal::MockInstance> tls_;
+  Envoy::Random::RandomGeneratorImpl rand_;
+  NiceMock<Envoy::Event::MockDispatcher>* dispatcher_ = nullptr; // not owned
```
oschaaf, line 87: Should the initialization be done in the constructor initializer list?

test/flush_worker_test.cc
oschaaf: I see, but what's not immediately clear to me is how it is guaranteed that the flush happens after the `shutdownThread()` call, and not before it? (I might be a little dense here, but I bet that someone else might have the same question pop up while reading this, so possibly a code-level comment explaining why this is guaranteed would be helpful here?).

test/flush_worker_test.cc
qqustc: If stats_sinks_ changes to a list of MockSink, then we got error `no matching constructor for initialization of 'Nighthawk::Client::FlushWorkerImpl'` when creating the FlushWorkerImpl 

`auto worker = std::make_unique<FlushWorkerImpl>(
      mock_api_, tls_, store_, stats_sinks_, stats_flush_interval_);`

oschaaf: I see this makes sense then. Sorry about the noise

test/flush_worker_test.cc
qqustc: Changed to use std::promise to wait for the while loop run at least 100 times in simulateTimerLoop().

test/flush_worker_test.cc
qqustc: Thanks Otto. Done

test/flush_worker_test.cc
qqustc: Done

test/process_test.cc
qqustc: Ah, after I changed to `constexpr const char kFoo[] = "";`, clang-tidy report error `do not declare C-style arrays, use std::array<> instead [modernize-avoid-c-arrays,-warnings-as-errors]`

So I will change back to `constexpr absl::string_view kSinkName =""` if that is ok

oschaaf: Well, if clang-tidy objects, I don't think we have much choice. Looking at this recommendation,
it should be allright: https://abseil.io/tips/168. Sorry about the fuzz.

test/flush_worker_test.cc
qqustc: Ah, the final flush call `flushStats()` happens inside `FlushWorkerImpl::shutdownThread()` as you can see in the flush_worker_impl.cc. And the `FlushWorkerImpl::shutdownThread()` is executed when `flush_worker_->shutdown()` is called in `ProcessImpl::shutdown()` in process_impl.cc.

Also added comment in the code.

include/nighthawk/common/factories.h
qqustc: ah, the `StatsSinkFactory` in the link above is an abstract class not an implementation.

Anyway, agree that it is better to make the category() PURE here.

source/client/flush_worker_impl.cc
```diff
@@ -0,0 +1,51 @@
...
+    : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
+  for (auto& sink : stats_sinks) {
+    stats_sinks_.emplace_back(std::move(sink));
+  }
+}
+
+void FlushWorkerImpl::work() {
+  stat_flush_timer_ = dispatcher_->createTimer([this]() -> void { flushStats(); });
+  stat_flush_timer_->enableTimer(stats_flush_interval_);
+
```
oschaaf, line 24: nit: unnecessary whitespace

source/client/process_impl.cc
oschaaf: I think that maybe the right thing to do with respect to the point you raised, is maybe punt this to an issue, and open a discussion on Slack or some such?

test/flush_worker_test.cc
oschaaf: Ah, so this works on the premise that if there's a single flush, it can only be the final one induced via `worker->shutdown();`, correct?
Would it then be better to move the expectation to right before `worker->shutdown();` to make sure nothing in the lines that executed before it triggered the flush?

include/nighthawk/common/factories.h
oschaaf: Fair enough

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+
+TEST_F(FlushWorkerTest, WorkerFlushStatsPeriodically) {
+  setupDispatcherTimerEmulation();
+
+  auto worker = std::make_unique<FlushWorkerImpl>(mock_api_, tls_, store_, stats_sinks_,
+                                                  stats_flush_interval_);
+
+  std::thread thread = std::thread([&worker, this] {
+    // Wait for the while loop run at least 100 times in simulateTimerLoop().
+    promise_.get_future().wait();
```
oschaaf, line 119: awesome

test/flush_worker_test.cc
qqustc: Good point. Moved to right before `worker->shutdown()`;

source/client/flush_worker_impl.cc
qqustc: Done.

source/client/process_impl.cc
qqustc: Thanks Otto. Opened https://github.com/envoyproxy/nighthawk/issues/441
Also reverted the change for now.

source/client/flush_worker_impl.h
qqustc: Discussed offline. Instead of making it a singleton, for simplicity, decided to add comments that there should be only a single live flush worker instance at any given time in NH.

qqustc:
> one whitespace nit, and my last feedback; thanks!

Thanks Otto, the PR is ready for review now

mum4k:

source/client/process_impl.cc
```diff
@@ -463,6 +507,13 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
     w->waitForCompletion();
   }
 
+  if (!options_.statsSinks().empty() && flush_worker_ != nullptr) {
+    // Stop the running dispatcher in flush_worker_. Need to be called after all
```
dubious90, line 511: grammar:  Needs to be called

source/common/statistic_impl.cc
```diff
@@ -323,6 +323,14 @@ void SinkableHdrStatistic::recordValue(uint64_t value) {
   scope_.deliverHistogramToSinks(*this, value);
 }
 
+std::string SinkableHdrStatistic::tagExtractedName() const {
+  if (worker_id().has_value()) {
+    return fmt::format("{}.", worker_id().value()) + id();
```
dubious90, line 328: This seems weird.  Why use fmt::Format if you're still going to concatenate the worker_id with the id?

Couldn't you do either:
```worker_id().value()) + id() + "."```
or 
or use fmt::format to replace bother worker_id().value() and id() as their own separate interpolations?

File: source/common/statistic_impl.cc [hunk 21]
```diff
@@ -334,4 +342,12 @@ void SinkableCircllhistStatistic::recordValue(uint64_t value) {
   scope_.deliverHistogramToSinks(*this, value);
 }
 
+std::string SinkableCircllhistStatistic::tagExtractedName() const {
+  if (worker_id().has_value()) {
+    return fmt::format("{}.", worker_id().value()) + id();
```
dubious90, line 347: Same fmt::format question as in SinkableHdrStatistic

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+
+class FlushWorkerTest : public Test {
+public:
+  FlushWorkerTest()
+      : thread_factory_(Envoy::Thread::threadFactoryForTest()),
+        dispatcher_(new NiceMock<Envoy::Event::MockDispatcher>()) {
+    loader_ = std::make_unique<Envoy::Runtime::ScopedLoaderSingleton>(
+        Envoy::Runtime::LoaderPtr{new Envoy::Runtime::LoaderImpl(
+            *dispatcher_, tls_, {}, local_info_, store_, rand_, validation_visitor_, mock_api_)});
+    sink_ = new StrictMock<Envoy::Stats::MockSink>();
```
dubious90, line 32: Two questions here:
1.  It appears that sink_ is a pointer that we are then passing into stats_sinks as a unique_ptr.  That doesn't seem like it should work.
2.  Why does sink_ need to be a member variable of FlushWorkerTest?  It is already exposed as part of stats_sinks_ being a unique pointer to it.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+#include "external/envoy/test/test_common/simulated_time_system.h"
+
+#include "client/flush_worker_impl.h"
+
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+namespace Client {
+
+using namespace testing;
```
dubious90, line 22: Style guide forbids using namespace declarations:
https://google.github.io/styleguide/cppguide.html#Namespaces

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+
```
dubious90, line 23: Tests should be inside an anonymous namespace.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+    sink_ = new StrictMock<Envoy::Stats::MockSink>();
+    stats_sinks_.emplace_back(sink_);
+
+    EXPECT_CALL(mock_api_, threadFactory()).WillRepeatedly(ReturnRef(thread_factory_));
+    // mock_api_ takes the ownership of dispatcher_.
+    EXPECT_CALL(mock_api_, allocateDispatcher_(_, _)).WillOnce(Return(dispatcher_));
+  }
+
+  // We set up emulating timer firing and moving simulated time forward in
+  // simulateTimerloop() below.
```
dubious90, line 41: nit: simulateTimerLoop

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+  Envoy::Event::TimerCb timer_cb_;
+  bool timer_set_{};
+  bool stopped_{false};
+  // Protect stopped_.
+  Envoy::Thread::MutexBasicLockable lock_;
+  std::promise<void> promise_;
+
+  Envoy::Stats::MockSink* sink_ = nullptr; // owned by stats_sinks_
+  std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
+  std::chrono::milliseconds stats_flush_interval_{10};
```
dubious90, line 109: This test seems really complicated, and is very hard to read.  I'm really having trouble following it.  One of the problems is you've declared 18 member variables.  Is that actually necessary here?  Can some of those be moved to local variables?  Local variables have the advantage of telling the user what they are when they are used.

Some more comments all around might make this easier as well.

Also, better naming will help.  Calling your promise promise_ tells us almost nothing.  What is being promised?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+        .WillOnce(Invoke([&](Envoy::Event::DispatcherImpl::RunType type) {
+          // The second dispatcher run is in FlushWorkerImpl::work().
+          ASSERT_EQ(Envoy::Event::DispatcherImpl::RunType::RunUntilExit, type);
+          simulateTimerLoop();
+        }));
+  }
+
+  void simulateTimerLoop() {
+    int i = 0;
+    while (true) {
```
dubious90, line 73: maybe i'm missing something, but could this be restructured to be while (!stopped_) and remove the break?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+          ASSERT_EQ(Envoy::Event::DispatcherImpl::RunType::NonBlock, type);
+        }))
+        .WillOnce(Invoke([&](Envoy::Event::DispatcherImpl::RunType type) {
+          // The second dispatcher run is in FlushWorkerImpl::work().
+          ASSERT_EQ(Envoy::Event::DispatcherImpl::RunType::RunUntilExit, type);
+          simulateTimerLoop();
+        }));
+  }
+
+  void simulateTimerLoop() {
```
dubious90, line 71: If we already have the existence of MockTimer, why do we need to simulate a timer?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+  worker->start();
+  worker->waitForCompletion();
+  thread.join();
+  worker->shutdown();
+}
+
+// Verify the final flush is always done in FlushWorkerImpl::shutdownThread()
+// even when the dispatcher and timer is not set up (expectDispatcherRun() is not called).
+TEST_F(FlushWorkerTest, FinalFlush) {
+  auto worker = std::make_unique<FlushWorkerImpl>(mock_api_, tls_, store_, stats_sinks_,
```
dubious90, line 137: Why does worker need to be a unique pointer.  Can't it just be a FlushWorkerImpl directly?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,149 @@
...
+
+  Envoy::Stats::MockSink* sink_ = nullptr; // owned by stats_sinks_
+  std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks_;
+  std::chrono::milliseconds stats_flush_interval_{10};
+};
+
+TEST_F(FlushWorkerTest, WorkerFlushStatsPeriodically) {
+  setupDispatcherTimerEmulation();
+
+  auto worker = std::make_unique<FlushWorkerImpl>(mock_api_, tls_, store_, stats_sinks_,
```
dubious90, line 115: Less confident than I am in the second test, but does worker here need to be a unique_ptr or can it just be a FlushWorkerImpl?

source/client/flush_worker_impl.cc
```diff
@@ -0,0 +1,50 @@
...
+                                 std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                                 const std::chrono::milliseconds& stats_flush_interval)
+    : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
+  for (auto& sink : stats_sinks) {
+    stats_sinks_.emplace_back(std::move(sink));
+  }
+}
+
+void FlushWorkerImpl::work() {
+  stat_flush_timer_ = dispatcher_->createTimer([this]() -> void { flushStats(); });
```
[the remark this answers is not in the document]
qqustc: It does not work with msg: `No viable conversion from 'void' to 'Envoy::Event::TimerCb' (aka 'function<void ()>')`

source/client/flush_worker_impl.h
```diff
@@ -0,0 +1,51 @@
...
+  // exitDispatcher() stops the dispatcher and the flush timer running in flush worker. It must be
+  // called after all client workers are completed in process_impl.cc to make sure all metrics will
+  // be flushed.
+  void exitDispatcher() { dispatcher_->exit(); }
+
+protected:
+  void work() override;
+
+private:
+  // Flush the stats sinks. Note: stats flushing may not be synchronous.
```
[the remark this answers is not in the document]
qqustc: Thanks Nate. Added  "may not be synchronous, depending on each stat sink's implementation." to clarify.

include/nighthawk/common/factories.h
```diff
@@ -48,4 +49,26 @@ class TerminationPredicateFactory {
...
+class NighthawkStatsSinkFactory : public Envoy::Config::TypedFactory {
+public:
+  ~NighthawkStatsSinkFactory() override = default;
+
+  /**
+   * Create a particular Envoy::Stats::Sink implementation. If the
+   * implementation is unable to produce a Sink with the provided parameters, it
+   * should throw an EnvoyException. The returned pointer should always be
+   * valid.
+   * @param symbol_table supplies the symbol_table instance
```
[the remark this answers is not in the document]
qqustc: Added a line point to the definition of SymbolTable in envoy

source/client/flush_worker_impl.cc
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Nighthawk {
+namespace Client {
+
+FlushWorkerImpl::FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
+                                 Envoy::Stats::Store& store,
+                                 std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                                 const std::chrono::milliseconds& stats_flush_interval)
+    : WorkerImpl(api, tls, store), stats_flush_interval_(stats_flush_interval) {
+  for (auto& sink : stats_sinks) {
```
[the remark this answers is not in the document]
qqustc: Ack

source/client/flush_worker_impl.cc
```diff
@@ -0,0 +1,50 @@
...
+  // Make the final flush before the flush worker gets shutdown.
+  flushStats();
+}
+
+void FlushWorkerImpl::flushStats() {
+  // Create a snapshot and flush to all sinks. Even if there are no sinks,
+  // creating the snapshot has the important property that it latches all counters on a periodic
+  // basis.
+  Envoy::Server::MetricSnapshotImpl snapshot(store_);
+  for (auto& sink : stats_sinks_) {
```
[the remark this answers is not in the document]
qqustc: Done.

source/client/process_impl.cc
```diff
@@ -455,6 +477,28 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
...
     Runtime::LoaderSingleton::get().initialize(*cluster_manager_);
 
+    std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks;
+    for (const envoy::config::metrics::v3::StatsSink& stats_sink : bootstrap.stats_sinks()) {
+      ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
+      auto& factory =
+          Envoy::Config::Utility::getAndCheckFactory<NighthawkStatsSinkFactory>(stats_sink);
+      stats_sinks.emplace_back(factory.createStatsSink(store_root_.symbolTable()));
+    }
+    for (auto& sink : stats_sinks) {
```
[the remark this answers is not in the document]
qqustc: Done

source/client/process_impl.cc
qqustc: Done

source/common/statistic_impl.cc
qqustc: Thanks Nate. Changed to 
`fmt::format("{}.{}", worker_id().value(), id())`

source/common/statistic_impl.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: 1. I think it can work because the unique_ptr has a constructor to take a raw pointer as argument. 
Same e.g. in Envoy https://github.com/envoyproxy/envoy/blob/cf850688a93b8dea5705170a211ea98b77beda61/test/server/server_test.cc#L66-L67

2. It needs to be  member variable since later in the test we check 
`EXPECT_CALL(*sink_, flush(_)).Times();`.
We can not replace it with stats_sinks_ in the `EXPECT_CALL()` since `EXPECT_CALL()` is expected to take `MockSink` instead of `Sink`

test/flush_worker_test.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: It is because the MockTimer does not act like a real timer, it can not periodically trigger a timer callback, so we need to have simulateTimerLoop(). This is following the example in
https://github.com/envoyproxy/nighthawk/blob/dba15a8a604fda23a78f32862ce349f892c4ac0a/test/sequencer_test.cc#L127

test/flush_worker_test.cc
qqustc: I don't think there is a easy way to do this since stopped_ needs to be guarded by the LockGuard below.

test/flush_worker_test.cc
qqustc: Thanks Nate. Changed rand_, local_info_, validation_visitor_, stats_flush_interval_ to local variables. The rest member variables are needed.

Also added some comments and renamed promise_ to signal_dispatcher_to_exit_ since it is used to guarantee `dispatcher->exit()` is called after `signal_dispatcher_to_exit_.set_value();` (the while loop runs at least 100 times).

Yes, agree that the test is difficult to read. Partly it is because we are trying to simulate that a periodical timer and a callback keeps firing until a stop signal from another thread.

test/flush_worker_test.cc
qqustc: Yes, it can be FlushWorkerImpl. Changed.

test/flush_worker_test.cc
qqustc: Done.

qqustc:
Thanks @dubious90 for the review. The PR is ready for review again.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+
+  std::thread thread = std::thread([&worker, this] {
+    // Wait for the while loop run at least 100 times in simulateTimerLoop().
+    signal_dispatcher_to_exit_.get_future().wait();
+    worker.exitDispatcher();
+  });
+
+  expectDispatcherRun();
+  // Check flush() is called at least 100 times in simulateTimerLoop().
+  EXPECT_CALL(*sink_, flush(_)).Times(testing::AtLeast(100));
```
dubious90, line 136: You know this better to me, so I'll defer to you.  Do you have any fears that this might be brittle or flaky?   Is this running 100 times a pretty safe assumption?  Or alternatively, if it doesn't run over 100 times, is that definitely indicative of a problem?

oschaaf: reading this, I wanted to float a suggestion: in another part of this source file there's:

```c++
      if (i == 100) {
        signal_dispatcher_to_exit_.set_value();
      }
```

I feel it would help readability if that was a nicely named constant. 
Also, while taking another look at this, removing the simulated_time_system header / dependency might be 
helpful, because I don't think simulated time is actually used, correct?

source/client/flush_worker_impl.h
```diff
@@ -0,0 +1,51 @@
...
+namespace Client {
+
+// Only a single live flush worker instance can be created in Nighthawk at any
+// time.
+// Flush worker periodically flushes metrics snapshot to all configured stats sinks in Nighthawk. It
+// will keep running until exitDispatcher() gets called after all client workers are completed in
+// process_impl.cc. It will make the last flush before shutdown in shutdownThread().
+class FlushWorkerImpl : public WorkerImpl {
+public:
+  FlushWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
```
mum4k, line 25: Can we document the constructor and its arguments?

source/client/flush_worker_impl.h
```diff
@@ -0,0 +1,51 @@
...
+                  Envoy::Stats::Store& store,
+                  std::list<std::unique_ptr<Envoy::Stats::Sink>>& stats_sinks,
+                  const std::chrono::milliseconds& stats_flush_interval);
```
mum4k, line 28: Per the style guide the input-only arguments (e.g. const references) should come before input/output and output only arguments.

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

test/statistic_test.cc
```diff
@@ -403,7 +403,7 @@ TYPED_TEST(SinkableStatisticTest, EmptySinkableStatistic) {
   EXPECT_EQ(Envoy::Stats::Histogram::Unit::Unspecified, stat.unit());
   EXPECT_FALSE(stat.used());
   EXPECT_EQ("", stat.name());
-  EXPECT_DEATH(stat.tagExtractedName(), ".*");
+  EXPECT_EQ("", stat.tagExtractedName());
```
mum4k, line 406: Are these changes intended to be in this PR? If yes, can we not the reason behind them in the PR description as it isn't obvious.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+  // Simulate the periodical timer which keeps running until dispatcher->exit()
+  // is called from another thread. The std::promise signal_dispatcher_to_exit_
+  // is used here to guarantee the while loop runs for at least 100 times before dispatcher->exit()
+  // is called.
+  void simulateTimerLoop() {
+    int i = 0;
+    while (true) {
+      i++;
+      // At least run the while loop for 100 times.
+      if (i == 100) {
```
mum4k, line 88: It certainly does raise a question of why is this 100, so it is a good idea to put a comment here explaining it. Unfortunately the current comment doesn't help and goes against the style guide as it just describes the code:

https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts

Can we try to improve this comment? Also I think @oschaaf had a suggestion to replace the literal 100 with a well named constant which could improve this further.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+        timer_set_ = false;
+        timer_cb_();
+      }
+    }
+  }
+
+  NiceMock<Envoy::Api::MockApi> mock_api_;
+  Envoy::Thread::ThreadFactory& thread_factory_;
+  Envoy::Stats::IsolatedStoreImpl store_;
+  NiceMock<Envoy::ThreadLocal::MockInstance> tls_;
```
mum4k, line 104: Can we try to consistently call all mock members mock_something_ (like mock_api_ above), or consistently not give them the mock_ prefix?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+  NiceMock<Envoy::Event::MockDispatcher>* dispatcher_ = nullptr; // not owned
```
mum4k, line 105: Comment nit: instead of saying "not owned" it might be more useful if we say who does own it.

Applies here and below.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+  std::unique_ptr<Envoy::Runtime::ScopedLoaderSingleton> loader_;
+
+  NiceMock<Envoy::Event::MockTimer>* timer_; // not owned
+  Envoy::Event::TimerCb timer_cb_;
+  bool timer_set_{};
```
mum4k, line 110: Can we add inline comments explaining the meaning of these two bool members?

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+// Verify stats are flushed periodically until dispatcher->exit() is called from
+// another thread.
+TEST_F(FlushWorkerTest, WorkerFlushStatsPeriodically) {
+  std::chrono::milliseconds stats_flush_interval{10};
+  setupDispatcherTimerEmulation();
+
+  FlushWorkerImpl worker(mock_api_, tls_, store_, stats_sinks_, stats_flush_interval);
+
+  std::thread thread = std::thread([&worker, this] {
+    // Wait for the while loop run at least 100 times in simulateTimerLoop().
```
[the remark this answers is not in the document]
mum4k: This comment doesn't seem addressed yet.

test/flush_worker_test.cc
mum4k: This comment doesn't seem addressed yet.

test/flush_worker_test.cc
```diff
@@ -0,0 +1,161 @@
...
+TEST_F(FlushWorkerTest, FinalFlush) {
+  std::chrono::milliseconds stats_flush_interval{10};
+
+  FlushWorkerImpl worker(mock_api_, tls_, store_, stats_sinks_, stats_flush_interval);
+
+  worker.start();
+  worker.waitForCompletion();
+  // Stats flush should happen exactly once as the final flush is done in
+  // FlushWorkerImpl::shutdownThread().
+  EXPECT_CALL(*sink_, flush(_)).Times(1);
```
mum4k, line 155: What guarantees that this was called one times via the final flush as compared to the periodic flush?

test/flush_worker_test.cc
mum4k, line 136: Can you help me understand the uncertainty around the number of calls? This is similar to what @dubious90 asked - why are we checking for "at least 100"? How is synchronisation working in this test? If we are controlling the timers and the dispatcher - cannot we write it in a way so that we know exactly how many times it gets called?

Tests without strong event based synchronisation tend to get flaky over time and are hard to troubleshoot.

source/client/process_impl.cc
```diff
@@ -128,6 +131,14 @@ void ProcessImpl::shutdown() {
 
   {
     auto guard = std::make_unique<Envoy::Thread::LockGuard>(workers_lock_);
+    // flush_worker_->shutdown() needs to happen before workers_.clear() so that
+    // metrics defined in workers scope will be included in the final stats
+    // flush which happens in FlushWorkerImpl::shutdownThread() after
+    // flush_worker_->shutdown() is called. For the order between worker shutdown() and
+    // shutdownThread(), see worker_impl.cc.
+    if (flush_worker_) {
```
mum4k, line 139: What guarantees that workers won't collect more statistics between the times when the flush worker was shut down and the time the workers get shut below?

source/client/process_impl.cc
```diff
@@ -455,6 +477,28 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
 
     Runtime::LoaderSingleton::get().initialize(*cluster_manager_);
 
+    std::list<std::unique_ptr<Envoy::Stats::Sink>> stats_sinks;
```
mum4k, line 480: runInternal is getting a bit too long to follow. Can we try to think of a different structure where the new (or some part of the old) code gets moved into a well named function (or functions) in the anonymous namespace?

That way this gets to keep the high level business logic of the run and the reader can absorb that before looking at the implementation details.

source/client/process_impl.cc
```diff
@@ -455,6 +477,28 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
...
+      ENVOY_LOG(info, "loading stats sink configuration in Nighthawk");
+      auto& factory =
+          Envoy::Config::Utility::getAndCheckFactory<NighthawkStatsSinkFactory>(stats_sink);
+      stats_sinks.emplace_back(factory.createStatsSink(store_root_.symbolTable()));
+    }
+    for (std::unique_ptr<Envoy::Stats::Sink>& sink : stats_sinks) {
+      store_root_.addSink(*sink);
+    }
+
+    // Default flush interval is 5s.
```
mum4k, line 491: Does this comment help, or is it likely to get out of sync with the real default value (if that ever gets changed) and confuse the reader?

test/process_test.cc
```diff
@@ -88,10 +126,17 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProcessTest,
                          Envoy::TestUtility::ipTestParamsToString);
 
 TEST_P(ProcessTest, TwoProcessInSequence) {
+  FakeStatsSinkFactory factory;
+  Envoy::Registry::InjectFactory<NighthawkStatsSinkFactory> registered(factory);
+  numFlushes = 0;
```
mum4k, line 131: If we need to manually reset it to zero at the beginning of a test case, we might be better off doing it in the SetUp method on the test fixture (ProcessTest).

https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests

test/process_test.cc
```diff
@@ -88,10 +126,17 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProcessTest,
                          Envoy::TestUtility::ipTestParamsToString);
 
 TEST_P(ProcessTest, TwoProcessInSequence) {
+  FakeStatsSinkFactory factory;
```
mum4k, line 129: Instead of modifying the existing test case by adding another tested behavior, we would be better off defining a new test case for this new behavior that includes the stats sink. The general advice is to only test one behavior per test case, since test failures are then easier to understand.

Applies here and below.

test/flush_worker_test.cc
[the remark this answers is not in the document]
qqustc: Done

test/flush_worker_test.cc
qqustc: Thanks Nate and Otto.

I think the test is not going to be flaky since the loop is guaranteed to run at least 100 times. (@oschaaf to correct me if I'm wrong.)
This is because 
 - `signal_dispatcher_to_exit_.get_future().wait()` will make the call `worker.exitDispatcher()` wait after the loop run at least 100 times
- the loop will keep running until `worker.exitDispatcher()` is called, so the loop is guaranteed to run >= 100 times.

Also defined a const int kNumTimerLoops to replace 100 and removed simulated_time_system header.

source/client/flush_worker_impl.h
qqustc: Done

source/client/flush_worker_impl.h
qqustc: Thanks Jakub. Done

test/statistic_test.cc
qqustc: Yes, it is intended. Added to PR description

test/flush_worker_test.cc
qqustc: Thanks Jakub. Replaced the 100 with a constant and updated the comment clarifying it does not to be 100.

test/flush_worker_test.cc
qqustc: Removed the mock_ prefix

test/flush_worker_test.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: Done

test/flush_worker_test.cc
qqustc: From here [`complete_.set_value()`](https://github.com/envoyproxy/nighthawk/blob/f57e8d5e4d7a66ef93a9399b2c584434c91cb803/source/common/worker_impl.cc#L32), we know that when `worker.waitForCompletion()` finished, `FlushWorkerImpl::work()` also finished, which means dispatcher is exited. So any flush after this must not be periodic flush.

Also we know that because of `signal_thread_to_exit_.get_future().wait()`, `FlushWorkerImpl::shutdownThread()` (where the final flush happens) will be called after `worker.shutdown()` is called. So the final flush is guaranteed to happen

test/flush_worker_test.cc
oschaaf: @qqustc I had some hands on with the test code, to see if this could be simplified, and ended up with a small change I'd like to float for consideration: https://github.com/qqustc/nighthawk/compare/pr1...oschaaf:pr1#diff-e0be675b84ffe6e6fd41947249a3bafe
(please ignore all other changes, only the changes to `flush_worker_test.cc` are relevant, the rest was induced by merging master).

Tangentially related, there might be an opportunity for another potential enhancement, perhaps best done in an follow up iteration if consensus is that this is desirable: Maybe we can refactor `SequencerTestWithTimerEmulation` to be more easily shareable, so we consolidate some of the complexity that comes with that into a single place. An aside to that would be that we will have simulated time flowing as well here, which can be used for something that might be useful: testing the number of periodic flushes based on the actual passage of (simulated) time.

qqustc: Thank you Otto for the simplification!

For refactoring the the complexity of SequencerTest and FlushWorkerTest into a single place, since the 2 tests are still quite different even though they both simulate timer, I'm not sure the benefit is worth the hassle of the refactoring work.

qqustc: Thanks to @oschaaf , updated the test to make the loop run exactly kNumTimerLoops(100) times.

Previously it is not deterministic because the 2 threads are trying to acquire the lock on `stopped_` and the loop will exit only 1 thread get the lock and set  `stopped_ = true`.

source/client/process_impl.cc
qqustc: In `runInternal()`, we first call `waitForCompletion()` on all workers and then call `waitForCompletion()` on flush worker.
```
...
for (auto& w : workers_) {
    w->waitForCompletion();
  }

  if (!options_.statsSinks().empty() && flush_worker_ != nullptr) {
    // Stop the running dispatcher in flush_worker_. Need to be called after all
    // client workers are complete so that all the metrics can be flushed.
    flush_worker_->exitDispatcher();
    flush_worker_->waitForCompletion();
  }
...
```
On one hand, since `waitForCompletion()` only finished after all workers `work()` finished (where sending requests happens), so after workers' `waitForCompletion()` finished, there will be no requests generated.

On the other hand,  flush worker is shut down only happens after `runInternal()` finished which is after all workers' and flush worker's `waitForCompletion()` finished.

source/client/process_impl.cc
qqustc: Yes, it could be out of sync in the future and the information about the default value is not very useful here, so removed this comment.

test/process_test.cc
qqustc: Done. Added new test cases for the new behavior

test/process_test.cc
qqustc: The original intention is to make it obvious that `numFlushes` will get updated after `runProcess`. Since now we added the new test case for testing stats sink, and not all test cases need to set `numFlushes = 0`, so I think it is better to leave it in the test case.

source/client/process_impl.cc
qqustc: Moved the new code of set up sinks to a member function `setupStatsSinks()`.

qqustc:
> For future reference - ideally this would have been two separate PRs. One that implements the flush worker and another that integrates it into the process layer. A good rule to follow is only include in a single PR what absolutely needs to be in it. This allows us to decrease the size and better load balance the review load.

Thanks @mum4k . The PR is ready for review again.

source/client/process_impl.h
```diff
@@ -114,6 +114,15 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
...
   mergeWorkerStatistics(const std::vector<ClientWorkerPtr>& workers) const;
   void setupForHRTimers();
+  /**
+   * If there are sinks configured in bootstrap, populate stats_sinks with sinks
+   * created through NighthawkStatsSinkFactory and add them to store_root_.
+   *
+   * @param bootstrap the bootstrap configuration which include the stats sink configuration.
+   * @param stats_sinks a Sink list to be populated.
+   */
+  void setupStatsSinks(envoy::config::bootstrap::v3::Bootstrap& bootstrap,
```
mum4k, line 124: Looks like `envoy::config::bootstrap::v3::Bootstrap&` may be a read-only argument, if it is so, can we make it a const reference?

```cc
void setupStatsSinks(const envoy::config::bootstrap::v3::Bootstrap& bootstrap, ... )
```

test/flush_worker_test.cc
```diff
@@ -26,6 +25,8 @@ using ::testing::ReturnRef;
 using ::testing::StrictMock;
 using ::testing::Test;
 
+const int kNumTimerLoops = 100;
```
mum4k, line 28: Since this is defined in a fairly large scope, can we add a comment with an explanation of what this is?

source/client/process_impl.h
qqustc: Done.

test/flush_worker_test.cc
qqustc: Done

qqustc:
> Looks good modulo last two comments.

Done. Thanks @mum4k for the review!

PR #440, opened 2020-08-06
Adaptive Load plugin loader and input validation

eric846:
Adds adaptive load plugin loader.

Also updates the header files to introduce plugin-specific input validation.


mum4k:
@pamorgan please review and assign back to me once done.

source/adaptive_load/plugin_util.cc
```diff
@@ -0,0 +1,75 @@
...
+  try {
+    auto& config_factory = Envoy::Config::Utility::getAndCheckFactoryByName<InputVariableSetterConfigFactory>(config.name());
+    absl::Status validation_status = config_factory.ValidateConfig(config.typed_config());
+    if (!validation_status.ok()) {
+      return validation_status;
+    }
+    return config_factory.createInputVariableSetter(config.typed_config());
+  } catch (const Envoy::EnvoyException& e) {
+    return absl::InvalidArgumentError(
+        absl::StrCat("Could not load plugin: ", config.DebugString(), ": ", e.what()));
```
oschaaf, line 24: could this log line accidentally bleed sensitive information into the logs? (may also apply below)

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,377 @@
...
+public:
+  std::string name() const override { return "nighthawk.test-input-variable-setter"; }
+  absl::Status ValidateConfig(const Envoy::Protobuf::Message& message) override {
+    return DoValidateConfig(message);
+  }
+
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    return std::make_unique<nighthawk::adaptive_load::LinearScoringFunctionConfig>();
+  }
+  InputVariableSetterPtr
```
oschaaf, line 73: Add a newline here?

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,377 @@
...
+    return std::make_unique<TestMetricsPlugin>(config);
+  }
+};
+
+REGISTER_FACTORY(TestMetricsPluginConfigFactory, MetricsPluginConfigFactory);
+
+/**
+ * StepController for testing.
+ */
+class TestStepController : public StepController {
```
oschaaf, line 160: this seems pretty close to mocking a step controller. It works for me as is, but out of curiosity, I do wonder if there was an explicit choice made here to not use the mocking and instead pull it off like this?

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,377 @@
...
+
+REGISTER_FACTORY(TestStepControllerConfigFactory, StepControllerConfigFactory);
+
+/**
+ * Creates an Any wrapping a TypedExtensionConfig for use in the |typed_config| of all test
+ * plugins in this file. The choice of the particular proto LinearScoringFunctionConfig is
+ * arbitrary. We don't leave the Any empty because we need to check that the plugin utils can
+ * correctly pass the proto through to the plugin.
+ */
+Envoy::ProtobufWkt::Any CreateTypedConfigAny(double threshold) {
```
oschaaf, line 211: const double threshold?

test/adaptive_load/plugin_util_test.cc
eric846: I couldn't find a way to cleanly test all the behavior using mocks (specifically, the plugin util must propagate its input args to the plugin constructor). But admittedly the fakes in this test are not very good. Filed #448 to create better fakes that can be used across the tests. I already have a much better fake StepController in one of the upcoming PRs, and I can turn it into a library and reuse it in this test. 

This PR is a prerequisite for another PR that will take some time to review, so can we merge this with these bad fakes for now, rather than changing to mocks which can't easily test the full behavior, then replace these fakes with the better fakes when they're ready?

oschaaf: Sounds great to me, thanks for sharing context.

include/nighthawk/adaptive_load/config_validator.h
```diff
@@ -0,0 +1,36 @@
...
+   * recursively validated at load time.
+   *
+   * @param message The Any config proto taken from the TypedExtensionConfig that activated this
+   * plugin, to be checked for validity in plugin-specific ways.
+   *
+   * @return Status OK for valid config, InvalidArgument with detailed error message otherwise.
+   */
+  virtual absl::Status ValidateConfig(__attribute__((unused))
+                                      const Envoy::Protobuf::Message& message) {
+    return absl::OkStatus();
```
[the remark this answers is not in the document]
eric846: Managed to make it pure (no mock for now).

include/nighthawk/adaptive_load/config_validator.h
```diff
@@ -0,0 +1,36 @@
...
+   * to call Load...Plugin() on each TypedExtensionConfig field value. See plugin_util.h. If all
+   * plugin config factories follow this convention, the entire adaptive load session spec will be
+   * recursively validated at load time.
+   *
+   * @param message The Any config proto taken from the TypedExtensionConfig that activated this
+   * plugin, to be checked for validity in plugin-specific ways.
+   *
+   * @return Status OK for valid config, InvalidArgument with detailed error message otherwise.
+   */
+  virtual absl::Status ValidateConfig(__attribute__((unused))
```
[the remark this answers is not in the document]
eric846: Filed #447 to consider noexcept or other options. Meanwhile added a comment.

source/adaptive_load/plugin_util.cc
eric846: Good point, this will end up in the log later -- changed to name().

test/adaptive_load/plugin_util_test.cc
eric846: Done

test/adaptive_load/plugin_util_test.cc
eric846: Done

include/nighthawk/adaptive_load/input_variable_setter.h
```diff
@@ -3,9 +3,10 @@
 #pragma once
 
 #include "envoy/common/pure.h"
-#include "envoy/config/core/v3/base.pb.h"
 #include "envoy/config/typed_config.h"
 
+#include "nighthawk/adaptive_load/config_validator.h"
+
 #include "external/envoy/source/common/protobuf/protobuf.h"
```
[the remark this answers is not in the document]
eric846: What do you mean about a "/"? 

This is what I'm seeing: 

image

source/adaptive_load/plugin_util.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
mum4k, line 1: Can we try to find a more specific name for this library? Using a generic name like *util* will eventually attract any code related to plugins. How about plugin_loader.h?

Feel free to choose any other name that is specific enough not to attract unrelated code.

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,378 @@
...
+
+#include "adaptive_load/plugin_util.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+namespace {
+
+const double kBadConfigThreshold = 98765.0;
```
mum4k, line 24: Considering the large scope for this constant - can we add a comment explaining what this is?

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,378 @@
...
+    return absl::OkStatus();
+  }
+  const double value_from_config_proto_;
+};
+
+/**
+ * A factory that creates a TestInputVariableSetter from a LinearScoringFunctionConfig (see
+ * TestInputVariableSetter constructor).
+ */
+class TestInputVariableSetterConfigFactory : public InputVariableSetterConfigFactory {
```
mum4k, line 63: Is this going to be the only unit test where we will need these "test only" elements? Or should we add them into their own testonly library so they are reusable?

Or is this something we want to address separately in #448 ?

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,378 @@
...
+/**
+ * ScoringFunction for testing.
+ */
+class TestScoringFunction : public ScoringFunction {
+public:
+  // Any plugin in the adaptive load system can freely choose an arbitrary single proto as its
+  // config type. We use LinearScoringFunctionConfig for all plugins in this test.
+  TestScoringFunction(const nighthawk::adaptive_load::LinearScoringFunctionConfig& config)
+      : value_from_config_proto_{config.threshold()} {}
+  double EvaluateMetric(double) const override { return 1.0; }
```
mum4k, line 94: (nit, optional) Can we add newlines between the constructor, the individual methods and the members so we have three distinct groups? It may improve readability.

If we choose to do it here, let's apply it to all the relevant objects.

test/adaptive_load/plugin_util_test.cc
```diff
@@ -0,0 +1,378 @@
...
+TEST(PluginUtilTest, ReturnsErrorFromInputVariableSetterConfigValidator) {
+  envoy::config::core::v3::TypedExtensionConfig config;
+  config.set_name("nighthawk.test-input-variable-setter");
+  *config.mutable_typed_config() = CreateTypedConfigAny(kBadConfigThreshold);
+  EXPECT_THAT(LoadInputVariableSetterPlugin(config).status().message(),
+              ::testing::HasSubstr("input validation failed"));
+}
+
+TEST(PluginUtilTest, PropagatesConfigProtoToInputVariableSetter) {
+  envoy::config::core::v3::TypedExtensionConfig config;
```
mum4k, line 238: (optional) Is this worth aliasing to make it shorter?

source/adaptive_load/plugin_util.h
eric846: Done

test/adaptive_load/plugin_util_test.cc
eric846: Done

test/adaptive_load/plugin_util_test.cc
eric846: I think all of these can be replaced by the fakes created in #448.

test/adaptive_load/plugin_util_test.cc
eric846: Done

test/adaptive_load/plugin_util_test.cc
eric846: Done

Issue #414, opened 2020-07-17
Code modernisation opportunities (c++17)


oschaaf:
In https://github.com/envoyproxy/nighthawk/pull/412 two clang-tidy checks have been disabled.
It's possible that Envoy may revert back to c++14, and until we are sure that won't happen, we 
should probably not rely on c++17isms.

This issue tracks following up on enabling of the following checks:
-modernize-use-nodiscard
-modernize-concat-nested-namespaces

PR #443, opened 2020-08-07
Update run_clang_tidy.sh

oschaaf:
- Update run_clang_tidy.sh from Envoy's version.

- doesn't enable `[[nodiscard]]` because that introduced problems
with gtest: google/googletest#2416
- doesn't collapse nested namespaces, as the auto-fix is
imperfect: it doesn't update the command at the closing '}'

With this clang_tidy's automatic fix feature can be applied
as follows:

```bash
ci/do_ci.sh clang_tidy
/opt/llvm/bin/clang-apply-replacements  .
```

Related: #414


oschaaf:
Force pushed an update and backed out the auto-applied changes, as both caused trouble:

- back out enable [[nodiscard]] because that introduced problems
with gtest: google/googletest#2416
- back out collapsing nested namespaces, as the auto-fix is
imperfect: it doesn't update the command at the closing '}'

mum4k:

eric846:
LGTM

PR #444, opened 2020-08-07
Use only Nighthawk as a namespace

oschaaf:
As discussed earlier, we'd like to move towards using a single
namespace, Nighthawk.

This strips the following namespaces in the c++ code.
- Nighthawk::Client
- Nighthawk::Server
- Nighthawk::Server::Configuration

The proto api's still follow the old convention of having namespaces
follow directory layout, as changing that would cause compatibility
issues.

While there's some large code churn, this change should be a no-op.


dubious90:
Is there any way we can also deal with the fact that our protos and code live in different namespaces?  that can be kinda confusing.  Example: https://github.com/envoyproxy/nighthawk/blob/master/api/client/options.proto

So CommandLineOptions lives in nighthawk (lowercase n), but our code lives in Nighthawk (uppercase N).  Changing the proto package would be backwards incompatible but is there a strong preference for the capital N in the rest of the code?

oschaaf:
> Is there any way we can also deal with the fact that our protos and code live in different namespaces? that can be kinda confusing. Example: https://github.com/envoyproxy/nighthawk/blob/master/api/client/options.proto
> 
> So CommandLineOptions lives in nighthawk (lowercase n), but our code lives in Nighthawk (uppercase N). Changing the proto package would be backwards incompatible but is there a strong preference for the capital N in the rest of the code?

No objections from me, I think it would be an enhancement. 
I pushed 61e5730 which lowercases the Nighthawk namespace.
I don't think there's a clear backstory on this other then that we copied Envoy's convention on this. I think the proposed lowercasing would better comply with the google c++ style guide and as you suggest resolve the divergence between proto generated code and our own.

oschaaf:
We discussed this on Slack, and it looks like the namespace change in here would inflict some unintended pain. That outweighs the consistency argument driving this change. Therefore closing this.

PR #449, opened 2020-08-12
Clean up do_ci.sh

oschaaf:
- Experimentally remove one-offs to throttle paralellism
- Eliminate some code duplication

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -72,28 +72,18 @@ function run_bazel() {
     fi
 }
 
-function do_asan() {
-    echo "bazel ASAN/UBSAN debug build with tests"
-    echo "Building and testing envoy tests..."
+function do_sanitizer() {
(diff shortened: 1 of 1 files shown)


oschaaf:
(this will be a draft for a little while, as it needs some tests to be re-run a few times to get a sense of stability)

oschaaf:
Respinning CI multiple times indicates that stability doesn't suffer from the increased parallelism. The sanitizer runs
didn't fail once, and while coverage did fail, it does that sometimes regardless of this change
(and OOM is no longer seems to happen, which was driving the reduced parallelism). 
So with this change we are able to make more effective use of CI resources by leveraging all vCPUs in all tasks, plus we clean up a little bit of duplicated code.

mum4k:

wjuan-AFK:
LGTM

PR #451, opened 2020-08-14
Update Envoy to 630511e4b83da462d4518c0652c2520382a0e79b

### LINKED ISSUE #291: Cleanup unused variables in service_main.cc
(Minor Tech Debt)
The code for starting up the gRPC service, after starting up said service, also creates a grpc channel and stub client for communicating with it.  These are not used by any process, and should be removed

oschaaf:
- Update Envoy to 630511e4b83da462d4518c0652c2520382a0e79b
- Clean up service main unused variables
- Remove a.out, which accidentally crept in. Modifies .gitignore
  to ignore *.out
- Exclude `.cache` directory in format checks (might get created via docker runs).
- Lower the number of concurrent tasks for the gcc test in CI from 8 to 6 to avoid OOM kills.
  This seems related to the Envoy update in here, the other suspect, switching to `-c dbg`,
  was moved out from here to #461 which seems to be fine.

Fixes #291


oschaaf:
Renamed this, as I backed out the Envoy update: the gcc build broke.
Also added a workaround for the CI failures which just started to occur
across all jobs. I'm not sure why we need to explicitly apt-get the things
added here, but this may buy some time to investigate while unblocking
practical usage.

source/client/stream_decoder.cc
```diff
@@ -156,7 +156,7 @@ void StreamDecoder::setupForTracing() {
   // We pass in a fake remote address; recently trace finalization mandates setting this, and will
   // segfault without it.
   const auto remote_address = Envoy::Network::Address::InstanceConstSharedPtr{
-      new Envoy::Network::Address::Ipv4Instance("127.0.0.1", 0)};
+      new Envoy::Network::Address::Ipv4Instance("127.0.0.1")};
```
oschaaf, line 159: Note to self/reviewers: this is a no-op, but will avoid breaking on the next Envoy
update.

oschaaf:
I am surprised that gcc is still broken

dubious90:
gcc test is failing.  do we know what's causing that?

oschaaf:
> gcc test is failing. do we know what's causing that?

It looks like an OOM kill. I should look into that, I thought
We got rid of those. Let’s see if happens structurally:
/retest

oschaaf:
Two sequential gcc test failures caused by running out of memory suggests that something significant changed with this. Looking into that.

oschaaf:
@dubious90 this is ready for review again. Changes:
- Moved the addition of `-c dbg` out of here and into https://github.com/envoyproxy/nighthawk/pull/461. This didn't resolve the gcc test problem.
- Lowered CI [parallelism for the gcc test task](
5fa7754). This did resolve the problem. I can only attribute having to do this to the Envoy update in here.
- Unrelated: added [a small tweak](2ca31a1) to avoid the git hooks from running format checks against a directory created via docker builds.

oschaaf:
(restarted all CI tasks to make sure things look allright on repeat in CI)

PR #453, opened 2020-08-17
Adding test for request source factory

wjuan-AFK:
Adding an additional unit test for request source factory that tests the rpc path. Important because we want to add to the proto to extend this logic.


mum4k:

source/client/factories_impl.cc
```diff
@@ -165,14 +179,16 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
...
   } else {
-    RELEASE_ASSERT(!service_cluster_name.empty(), "expected cluster name to be set");
     // We pass in options_.requestsPerSecond() as the header buffer length so the grpc client
     // will shoot for maintaining an amount of headers of at least one second.
-    return std::make_unique<RemoteRequestSourceImpl>(cluster_manager, dispatcher, scope,
-                                                     service_cluster_name, std::move(header),
-                                                     options_.requestsPerSecond());
+    return request_source_constructor.createRemoteRequestSource(std::move(header),
+                                                                options_.requestsPerSecond());
+    // return std::make_unique<RemoteRequestSourceImpl>(cluster_manager, dispatcher, scope,
```
oschaaf, line 189: Is this comment accidentally left in?

include/nighthawk/common/factories.h
```diff
@@ -33,12 +33,21 @@ class StatisticFactory {
   virtual StatisticPtr create() const PURE;
 };
 
+class RequestSourceConstructorInterface {
+public:
+  virtual ~RequestSourceConstructorInterface() = default;
+  virtual RequestSourcePtr
+  createStaticRequestSource(Envoy::Http::RequestHeaderMapPtr&&,
+                            const uint64_t max_yields = UINT64_MAX) const PURE;
```
oschaaf, line 41: Somehow specifying the default value in the interface feels off to me / implementation-like. But then this interface seems to target materializing concrete implementations of `RequestSourceContructor`. I'm not sure I fully understand the rationale here, could you help me there by sharing some context? My concern is that for each implementation we will have to add to this interface, which somehow feel counterintuitive.

source/client/factories_impl.cc
```diff
@@ -165,14 +179,16 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
   }
 
   if (options_.requestSource() == "") {
-    return std::make_unique<StaticRequestSourceImpl>(std::move(header));
+    return request_source_constructor.createStaticRequestSource(std::move(header));
+    //    return std::make_unique<StaticRequestSourceImpl>(std::move(header));
```
oschaaf, line 183: accidental comment left in here?

oschaaf (changes requested):
A few nits about what seem to be accidentally left in comments. That aside, requesting a little more context so I can provide better feedback.

include/nighthawk/common/factories.h
```diff
@@ -33,12 +33,21 @@ class StatisticFactory {
...
+};
+
 class RequestSourceFactory {
 public:
   virtual ~RequestSourceFactory() = default;
-  virtual RequestSourcePtr create(const Envoy::Upstream::ClusterManagerPtr& cluster_manager,
-                                  Envoy::Event::Dispatcher& dispatcher, Envoy::Stats::Scope& scope,
-                                  absl::string_view service_cluster_name) const PURE;
+  virtual RequestSourcePtr
+  create(const RequestSourceConstructorInterface& request_source_constructor) const PURE;
```
qqustc, line 50: Do we want to replace old `create()` method or add new method in addition to the old  method?

include/nighthawk/common/factories.h
```diff
@@ -33,12 +33,21 @@ class StatisticFactory {
   virtual StatisticPtr create() const PURE;
 };
 
+class RequestSourceConstructorInterface {
```
qqustc, line 36: Instead of create another interface, impl and mock for RequestSourceConstructor, would it be cleaner to modify current RequestSourceFactory to create both static and remote RequestSource?

source/client/factories_impl.cc
wjuan-AFK: Yes. Thank you for spotting this.

include/nighthawk/common/factories.h
wjuan-AFK: This is part of a wider set of changes I'm making for the request source plugin implementation. So ideally, once everything is in plugin form, the RequestSourceConstructor interface will be fairly stable. The reason I am splitting this off is to have better separation of logic between the decision of what kind of requestsource I am making, and how to make that request source. This also has the added benefit of making things more testable.

source/client/factories_impl.cc
wjuan-AFK: Yes. Thank you.

include/nighthawk/common/factories.h
wjuan-AFK: I want to replace the old create() method. This refactor is intended to separate out the logic for how to make a requestSource into the requestSourceConstructor.

include/nighthawk/common/factories.h
wjuan-AFK: Part of the intention of creating the separate interface is to make it easier to test, by allowing us to mock that portion of the code separately. As well as to separate out logic for creating the types of RequestSource so that I can add the RequestSourcePlugin to the RequestSourceConstructor in isolation.

wjuan-AFK:
General context for this PR is that I am adding an additional test for the logic of which request source am I given for this CLI. And separating out the logic for constructing requestSources from the requestSourceFactory so that it can be dependency injected. The reason for these changes is that I want to be able to make some CLI changes for the requestSourcePlugin and want that change to be isolated from changes I am making to the requestSourceConstructor to actually make those requestSourcePlugins.

include/nighthawk/common/factories.h
oschaaf: I wonder: What if the factories 'create' methode would be modified instead to take a proto config as an argument? And then its imementation instantiate  the right class  based that? (The proto config would use a envoy.config.core.v3.TypedExtensionConfig to indicate the desired request source)

qqustc:
@dubious90 
The only comment I have is that instead of create a `RequestSourceConstructorInterface`, `RequestSourceConstructorImpl` and the mock, we can just simply add a new method `createRemoteRequestSource` to the existing `RequestSourceFactory`.

Assign back to @dubious90 to leave the decision to Nate and unblock the review process.

include/nighthawk/common/factories.h
dubious90: I think I agree with Otto here.  Rather than having two different functions for createRemoteRequestSource and createStaticRequestSource, just have the interface contain one `create` function, that includes a proto for specific configuration, based on what people want to create.

Right now, the difference between these two methods appears to be the inclusion of max_yields.  I might not really understand what max_yields is doing here, but it seems likely that would be an implementation detail of specific implementations of this interface, rather than a part of the interface itself.  Otto's proto solution handles this by letting those implementations define max_yields or anything else they might not need.

include/nighthawk/common/factories.h
dubious90, line 36: I am a little concerned about the complexity of these changes.  Some of this might be naming but differentiating between a constructor and a factory here is pretty confusing, as both of them generally accomplish the same thing.

The more standard way to use a factory interface would be more like what Adaptive Load does, defining a factory that takes in something configurable like a proto.  I don't think it's necessary to ahve two separate methods here, createStaticRequestSource and createRemoteRequestSource, as those sound like implementation details.  And without that, I don't know what the ConstructorInterface here brings.

I'd like to schedule us a meeting to talk about this, if that's okay with you.  I'm sure there are things in the test code you're referring to that I'm not seeing for why this makes things less complex, and I'd like to get on the same page as you.

PR #454, opened 2020-08-17
Revert "Clean up do_ci.sh"

oschaaf:
This reverts commit 4bd0e30d7455efc237fcb87e72727ba6437fb776.
I can't relate this change to the current CI breakage, but let's see what happens
when backing it out to make sure I didn't miss anything here.


oschaaf:
This breaks in the same way as everything else, closing.

PR #456, opened 2020-08-17
CI: minimal fix to unbreak

oschaaf:
Python's psutil, which gets pulled in via pytest-xdist, needs gcc
to build. While doing it that it runs into that:
a) It can't find python.h, which we solve by installing python3-dev
b) It can't find /usr/bin/x86_64-linux-gnu-gcc which we resolve by
   symlinking in an existing version at the expected place.

Pinning the packages to specific versions doesn't solve the problem
by itself, at least not in a combination of versions that I could
find. Instead, I'll create a follow up to pin them to recent versions
in a follow-up.


oschaaf:
Right, so this actually fixes some CI tasks, but not all of them. Closing this, as we would then need #457 which includes this.

PR #457, opened 2020-08-17
Python requirements.txt - pin to (recent) versions 

oschaaf:
Pin python versions to recent versions. 
CI broke on pytest-xdist releasing `2.0.0` which got pulled in.

Signed-off-by: Otto van der Schaaf oschaaf@we-amp.com

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -159,6 +159,13 @@ if grep 'docker\|lxc' /proc/1/cgroup; then
     export USER=bazel
     export TEST_TMPDIR=/build/tmp
     export BAZEL="bazel"
+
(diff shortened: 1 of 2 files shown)


oschaaf:
As discussed on Slack, we hold back upgrading `pytest-xdist` to 2.0 here, and stay on 1.x. The changelist over at https://github.com/pytest-dev/pytest-xdist/blob/master/CHANGELOG.rst  shows nothing critical we should have in the short term, and this allows for a smaller fix. If we ever do want to move forward, e6bd6518f2ac8502e944ec72b5e4b6e47b8b909f has *a* way to make that work (maybe there's a better way).

PR #462, opened 2020-08-18
Update Envoy to a6f5d4b0d310d64a800a7592fa5182975efb0a0a 

oschaaf:
------

### Prerequisites
- [x] Land #451 & merge master into this
- [x] Needs envoyproxy/envoy#12706 to get merged first.


oschaaf:
Hmm, the test_gcc task failed because of getting OOM killed. Respinning. 
We might have to lower parallelism more for that task :-(

source/server/README.md
```diff
@@ -249,6 +249,9 @@ Path to logfile
 Prefix all occurrences of '%v' in log format with with '[%g:%#] '
 ('[path/to/file.cc:99] ').
 
+--enable-fine-grain-logging
+Logger mode: enable file level log control(Fancy Logger)or not
```
dubious90, line 253: Is this auto-generated from somewhere?  I thought our README was, but not seeing a corresponding change to TCLAP for this.

This is missing spaces around the parentheses. Also, I do not think that the "or not" is valuable.  Can we change to:
`enable file level log control (Fancy Logger)`

dubious90: Also, what is Fancy Logger?  Is that a library?  If so, we should provide a link to it.

oschaaf: Yup, this is auto-generated, these changes are brought in via the updated Envoy version, which propagates to our test server. So any changes should be made over at the Envoy repo.

oschaaf: And yeah, I think fancy logger is a library, this came with an accidental breaking change for us. We update Envoy's sha to the fix I proposed for that: https://github.com/envoyproxy/envoy/commit/74f4857726205f5de83ea1916aa4717f7f95d227

dubious90: Understood.  If it comes in directly from envoy, then there's nothing we can really do about the weirdness in the helptext.

Issue #448, opened 2020-08-12
Create full-fledged Adaptive Load fake plugins


eric846:
- Having their own unit tests
- For use in multiple unit tests

PR #463, opened 2020-08-18
Adaptive Load FakeStepController

eric846:
FakeStepController activated with a full TypedExtensionConfig to test plugin registration and config validation.
  - Regardless of state, returns a constant RPS (configurable via the config proto).
  - Reports doom if it has just received a benchmark result with an error from Nighthawk Service.
  - Reports converged if it has just received a benchmark result that received score >0 (manipulate the score using some ScoringFunction and some MetricsPlugin); if the latest result was <=0, reports non-converged. Initially reports non-converged.
  - Config proto supports setting an artificial config validation error.

Partially addresses issue #448.


dubious90:

include/nighthawk/adaptive_load/step_controller.h
```diff
@@ -66,7 +66,7 @@ using StepControllerPtr = std::unique_ptr<StepController>;
  * A factory that must be implemented for each StepController plugin. It instantiates the
  * specific StepController class after unpacking the plugin-specific config proto.
  */
-class StepControllerConfigFactory : public Envoy::Config::TypedFactory, public ConfigValidator {
+class StepControllerConfigFactory : public virtual Envoy::Config::TypedFactory, public virtual ConfigValidator {
```
htuch, line 69: Can we avoid multiple implementation inheritance here with some refactoring or changes to is-a vs. has-a relationships? Google/Envoy C++ style discourages multiple implementation inheritance.

eric846: Yes, I am removing the shared concrete NullConfigValidator that we were going to inherit from. Instead a plugin factory that doesn't need to validate anything can just explicitly override ValidateConfig() with a body that immediately returns absl::OkStatus(). No other ConfigValidator implementations will need to be shared because they will all be plugin-specific.

eric846: For now, I'm eliminating the multiple inheritance but not shaking up plugin_loader or all the tests written according to these interfaces. Filed #465 to consider whether a refactor will clean it up further, after all these PRs land.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+  Envoy::ProtobufWkt::Any config_any;
+  config_any.PackFrom(config);
+  CommandLineOptions options;
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<StepControllerConfigFactory>(
+          "nighthawk.fake-step-controller");
+  EXPECT_EQ(config_factory.name(), "nighthawk.fake-step-controller");
+}
+
+TEST(FakeStepControllerConfigFactory, CreateMetricsPluginCreatesCorrectPluginType) {
```
wjuan-AFK, line 41: What does this have to do with the metrics plugin?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+}
+
+TEST(FakeStepController, IsDoomedDoesNotWriteDoomedReasonAfterSuccessfulBenchmarkResult) {
+  FakeStepController step_controller(FakeStepControllerConfig{}, CommandLineOptions{});
+  BenchmarkResult benchmark_result;
+  benchmark_result.mutable_status()->set_code(::grpc::OK);
+  step_controller.UpdateAndRecompute(benchmark_result);
+  std::string doomed_reason = "untouched";
+  ASSERT_FALSE(step_controller.IsDoomed(doomed_reason));
+  EXPECT_EQ(doomed_reason, "untouched");
```
wjuan-AFK, line 143: Why do we expect these to be equal if it does not write doomed reason after successful benchmark result? Shouldn't it be empty or something?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Good catch, this was a typo.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: The local variable doomed_reason starts out with the value "untouched" (line 141) before we call step_controller.IsDoomed(doomed_reason). So the value "untouched" should still be intact instead of getting overwritten somewhere inside IsDoomed.

wjuan-AFK:
LGTM

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+using ::nighthawk::client::CommandLineOptions;
+using ::testing::HasSubstr;
+
+TEST(FakeStepControllerConfigFactory, CreateEmptyConfigProtoCreatesCorrectType) {
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<StepControllerConfigFactory>(
+          "nighthawk.fake-step-controller");
+  Envoy::ProtobufTypes::MessagePtr message = config_factory.createEmptyConfigProto();
+  FakeStepControllerConfig expected_config;
+  EXPECT_EQ(message->DebugString(), expected_config.DebugString());
```
oschaaf, line 27: Should we double down with the Message until based comparison to avoid missing edge cases?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<StepControllerConfigFactory>(
+          "nighthawk.fake-step-controller");
+  absl::Status status = config_factory.ValidateConfig(config_any);
+  EXPECT_TRUE(status.ok());
+}
+
+TEST(FakeStepController, GetCurrentCommandLineOptionsReturnsRpsFromConfig) {
+  FakeStepControllerConfig config;
+  config.set_fixed_rps_value(5678);
```
oschaaf, line 100: Maybe create a shared constant for the magic number?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+  EXPECT_TRUE(step_controller.IsConverged());
+}
+
+TEST(FakeStepController, IsDoomedReturnsFalseAfterSuccessfulBenchmarkResult) {
+  FakeStepController step_controller(FakeStepControllerConfig{}, CommandLineOptions{});
+  BenchmarkResult benchmark_result;
+  benchmark_result.mutable_status()->set_code(::grpc::OK);
+  step_controller.UpdateAndRecompute(benchmark_result);
+  std::string doomed_reason;
+  EXPECT_FALSE(step_controller.IsDoomed(doomed_reason));
```
oschaaf, line 133: Reading this, Ibwonder if a single call to obtain status (converged) or failure (doomed) would simplify things here and for the consing code. Just a thought after recently learning about that construct, please feel free to ignore

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+TEST(MakeFakeStepControllerPluginConfig, ActivatesFakeStepControllerPlugin) {
+  absl::StatusOr<StepControllerPtr> plugin_or =
+      LoadStepControllerPlugin(MakeFakeStepControllerPluginConfig(5), nighthawk::client::CommandLineOptions{});
+  ASSERT_TRUE(plugin_or.ok());
+  EXPECT_NE(dynamic_cast<FakeStepController*>(plugin_or.value().get()), nullptr);
+}
+
+TEST(MakeFakeStepControllerPluginConfig, ProducesFakeStepControllerPluginWithConfiguredValue) {
+  absl::StatusOr<StepControllerPtr> plugin_or =
+      LoadStepControllerPlugin(MakeFakeStepControllerPluginConfig(5), nighthawk::client::CommandLineOptions{});
```
oschaaf, line 176: Make 5 a named/shared constant?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,187 @@
...
+
+TEST(FakeStepController, IsDoomedSetsDoomedReasonToStatusMessageAfterFailedBenchmarkResult) {
+  FakeStepController step_controller(FakeStepControllerConfig{}, CommandLineOptions{});
+  BenchmarkResult benchmark_result;
+  benchmark_result.mutable_status()->set_code(::grpc::INTERNAL);
+  benchmark_result.mutable_status()->set_message("error from nighthawk");
+  step_controller.UpdateAndRecompute(benchmark_result);
+  std::string doomed_reason;
+  ASSERT_TRUE(step_controller.IsDoomed(doomed_reason));
+  EXPECT_EQ(doomed_reason, "error from nighthawk");
```
oschaaf, line 164: "error from nighthawk" might be reduplicated by using a constant?

oschaaf (changes requested):
Looks great, here is my feedback

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
wjuan-AFK: I agree with that. And even if you want to keep the IsDoomed call, you can still just have that refer to the status. Especially since Doomed and Converged are mutually exclusive states.

eric846: Opened #466

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Done

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Done -- keeping these constants local since they have no shared meaning across tests

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Done

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Done (made another 5 zero)

oschaaf (approved):
Looks good to me

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,208 @@
...
+  std::string doomed_reason;
+  EXPECT_FALSE(step_controller.IsDoomed(doomed_reason));
+}
+
+TEST(FakeStepController, IsDoomedDoesNotWriteDoomedReasonAfterSuccessfulBenchmarkResult) {
+  FakeStepController step_controller(FakeStepControllerConfig{}, CommandLineOptions{});
+  BenchmarkResult benchmark_result;
+  benchmark_result.mutable_status()->set_code(::grpc::OK);
+  step_controller.UpdateAndRecompute(benchmark_result);
+  std::string doomed_reason = "untouched";
```
dubious90, line 149: I got confused by the same thing as William.  I think this would be clearer, and more accurate to the API of the call if we set this to "", and check that it's still empty after IsDoomed is called.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -0,0 +1,208 @@
...
+
+TEST(FakeStepController, IsDoomedSetsDoomedReasonToStatusMessageAfterFailedBenchmarkResult) {
+  const std::string kErrorMessage = "error from nighthawk";
+  FakeStepController step_controller(FakeStepControllerConfig{}, CommandLineOptions{});
+  BenchmarkResult benchmark_result;
+  benchmark_result.mutable_status()->set_code(::grpc::INTERNAL);
+  benchmark_result.mutable_status()->set_message(kErrorMessage);
+  step_controller.UpdateAndRecompute(benchmark_result);
+  std::string doomed_reason;
+  ASSERT_TRUE(step_controller.IsDoomed(doomed_reason));
```
dubious90, line 172: Why ASSERT_TRUE rather than EXPECT_TRUE here?  Won't the code after still work?

Also, I honestly think that this and the previous test are essentially testing two very similar parts of the same thing, and we could combine them.  But that's an opinion, so I'll leave it up to you.

eric846: Changed to EXPECT_TRUE.

Deleted the previous test that was a subset of this one.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: One of the potential bugs this test is guarding against is IsDoomed accidentally writing "" to the variable instead of leaving it untouched, so we would want to fail if it started to do that.

What do you think of this clarified variable name and value?

eric846: Hmm actually I see what you mean: the caller isn't supposed to have anything important in the variable ahead of time, since it might get overwritten. 

The behavior of not writing the variable when IsDoomed returns false is really not very valuable at all. Maybe the behavior when returning false should be undefined, or IsDoomed should write the variable unconditionally, writing "" if it's returning false.

eric846: Opened #482 to rewrite IsDoomed to return absl::Status. The only reason for this weird contract is we had not brought in absl::Status at the time.

Meanwhile the test as it currently stands should explicitly test the ill-conceived spec of this function.

What do you think?

dubious90: Sounds good.  Thanks for taking a look at what could be done.  I like the Status solution.

PR #477, opened 2020-08-21
Echo back and track origin request-receipt timing deltas

oschaaf:
This adds a rough cut of the capability to the test server to allow tracking of the delta's between
request arrival times, and echo these back in a configurable response header name.

Subsequently, the nighthawk_client gets a feature to track numbers received via an arbitrary
header in a histogram. Right now it's hard coded to track `x-nh-do-not-use-origin-timings`. 
But the plan would be to allow for at least one configurable header name to track. 

Combined, these test server and client features enable obtaining insight into the origin-side point of
view regarding arrival timings [1]. The direct use case is manual inspection / sanity checking, but possibly
divergence of what would be expected based on client side configuration can be used as a way to quantify
"distortion" introduced by an intermediary proxy.

In a follow up, documentation, configuration of the client to specify the response header name that should
be used to track latency, and end to end tests will be added. 

When setting up a test between the Nighthawk client and test server using this feature, a new histogram
will be send to the output. On my machine, a simple 1000qps test yields the following perspective from
the test-server when it comes to deltas between inbound requests:

```
  0.5         2519        0s 000ms 999us 
  0.75        3767        0s 001ms 001us 
  0.8         4002        0s 001ms 002us 
  0.9         4508        0s 001ms 003us 
  0.95        4756        0s 001ms 004us 
  0.990625    4953        0s 001ms 005us 
  0.99902344  4995        0s 001ms 015us ```

[1] Concretely, if a straightforward 1000 qps is configured, the histogram obtained from tracking the server side should reasonably reflect that and show a histogram diverging very little from 1ms.

### CODE REVIEW

File: source/client/stream_decoder.cc [hunk 1]
```diff
@@ -19,6 +19,19 @@ void StreamDecoder::decodeHeaders(Envoy::Http::ResponseHeaderMapPtr&& headers, b
   response_header_sizes_statistic_.addValue(response_headers_->byteSize());
   const uint64_t response_code = Envoy::Http::Utility::getResponseStatus(*response_headers_);
   stream_info_.response_code_ = static_cast<uint32_t>(response_code);
+  const auto timing_header_name = Envoy::Http::LowerCaseString("x-nh-do-not-use-origin-timings");
```
oschaaf, line 22: Note to reviewers: the odd name here is to scare people away from using this. My plan is to make a new option for nighthawk_client to allow configuration of this header name.

mum4k:

source/server/http_test_server_filter.cc
```diff
@@ -79,6 +103,10 @@ HttpTestServerDecoderFilter::decodeTrailers(Envoy::Http::RequestTrailerMap&) {
 void HttpTestServerDecoderFilter::setDecoderFilterCallbacks(
     Envoy::Http::StreamDecoderFilterCallbacks& callbacks) {
   decoder_callbacks_ = &callbacks;
+  // TODO(oschaaf): this adds locking in the hot path. Consider moving this into a separate
+  // extension, which will also allow tracking multiple points via configuration.
+  last_request_delta_ns_ =
```
eric846, line 108: Could this stopwatch call be conditional on `!base_config_.emit_previous_request_delta_in_response_header().empty()` since we don't need it otherwise?

source/client/stream_decoder.cc
```diff
@@ -19,6 +19,19 @@ void StreamDecoder::decodeHeaders(Envoy::Http::ResponseHeaderMapPtr&& headers, b
...
+  const auto* timing_header = response_headers_->get(timing_header_name);
+  if (timing_header != nullptr) {
+    auto timing_value = timing_header->value().getStringView();
+    int64_t origin_delta;
+    if (absl::SimpleAtoi(timing_value, &origin_delta) && origin_delta >= 0) {
+      origin_latency_statistic_.addValue(origin_delta);
+    } else {
+      // TODO(XXX): Can we make sure we avoid high frequency logging for this somehow?
```
oschaaf, line 30: Punted to https://github.com/envoyproxy/nighthawk/issues/484, which has a candidate PoC solution. I remember there are other spots where this would be good to have in the existing code base as well.

source/server/http_test_server_filter.cc
```diff
@@ -16,6 +17,22 @@ HttpTestServerDecoderFilterConfig::HttpTestServerDecoderFilterConfig(
...
     : server_config_(std::move(proto_config)) {}
 
+uint64_t HttpTestServerDecoderFilterConfig::ThreadSafeMontonicTimeStopwatch::getElapsedNsAndReset(
+    Envoy::TimeSource& time_source) {
+  Envoy::Thread::LockGuard guard(lock_);
+  // Note that we obtain monotonic time under lock, to ensure that start_ will be updated
+  // monotonically.
+  const Envoy::MonotonicTime new_time = time_source.monotonicTime();
+  const uint64_t elapsed = start_ == Envoy::MonotonicTime::min() ? 0 : (new_time - start_).count();
+  start_ = new_time;
```
eric846, line 27: Warning: Attempting to reason about concurrency...

Do you see any weird effects running with concurrency > 1 on the test server?

I was imagining a situation with 2 test server threads operating almost in lockstep, but 1ns apart. Suppose the requests come in evenly every 1ms. If the test server got stalled for a few milliseconds and the OS ended up buffering several packets, and then the test server threads started consuming them round robin, wouldn't the timing reports end up alternating 0, 1ns, 999999ns, 1ns, 999999ns, 1ns, ...?

This would actually represent the true response timing coming out of the test server, so the code in this PR would be working as intended.

Extreme statistics like that could give the impression that there is extreme distortion from the intermediary proxy, but it's entirely an artifact of the Nighthawk test server having two threads and stalling for a few milliseconds.

(Also the intermediary might also introduce its own equally extreme irregularities, since it could also be a multithreaded Envoy.)

The obvious way to avoid this issue is to use concurrency=1 on the test server. Then we would only be measuring the distortions from intermediaries and the OS.

Do you think there's any future reporting that could diagnose this issue? For example, if we also had exactly this PR but with thread-local stopwatches, we would see perfect evenness rather than 1-999999 fluctuation. We would know that the 1-999999 was the fault of our own multithreading, and any timing deviations would be our OS or the intermediary.

api/server/response_options.proto
```diff
@@ -32,4 +32,6 @@ message ResponseOptions {
     // Concurrency based linear delay configuration.
     ConcurrencyBasedLinearDelay concurrency_based_linear_delay = 5;
   }
+  // If set, makes the extension echo timing data in the supplied response header name.
```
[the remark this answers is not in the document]
oschaaf: This is all correct. I'll add comments based on this to clarify.

source/server/http_test_server_filter.cc
eric846: Actually, maybe varying between 1ns and 1ms would be considered negligible distortion, especially if the intermediary added multiple milliseconds.

oschaaf: Your reasoning here makes sense to me. The intent of this is to be able to sanity check part of our timings, so if it would catch those extremes where we don't expect them, I think it would be providing value.

The way I see it is that this is going to be particularly useful in A/B tests, for example:
- across different clients
- with and without an intermediary
- with nighthawk client features that might perform relatively heavy lifting on cpu cores shared with the workers (e.g. stats flushing).

Some existing OSS clients can probably easily be tweaked to take these as inputs to their latency reports to figure out their request-release timing precisions. 

Also, something like the delta between the expected and observed curves could perhaps serve as a means to quantify distortion; Indeed I think the extremes in the scenario you described would be signal in this case, and not noise (looking at you, adaptive load controller :-)).

As for tracking thread-local values: I've been thinking about that, and I think it's pretty doable doing that on the server side by adding a new Envoy histogram in the extension. On the client side however, I think it will take more consideration and work, as I think there would be some feature gaps:
- be able to set multiple response headers for tracking, or create a means to read multiple values from the header.
- if we want to be able to correlate thread id's we need to be able to tag values too, but this may not be nessecary.
- having a means to dynamically create histograms or a histogram capable of grouping on a tag would be needed
  for the Nighthawk client to handle it

oschaaf: Afterthought: maybe, if we'd have the per worker histograms on the test server side, plus a means to read/sample these histograms on the NH client side and propagate these to the output, that might be a relatively low effort means to achieve this.

source/server/http_test_server_filter.cc
oschaaf: We can, but that would prevents this feature from working with per-request reconfiguration.

eric846: Probably should keep it unconditional, since I don't know of any actual performance impact, and it's supposed to cost only 100ns to take a lock.

eric846:
LGTM modulo adding the comment

source/server/http_test_server_filter.cc
oschaaf: Last thought: per-worker tracking may also yield data that's interesting in light of exposing unbalancedness.
I filed #487 to get this on the radar and track this / serve as a point for further discussion.

oschaaf:
> LGTM modulo adding the comment

Thanks @eric846. Changes according to feedback applied in cf1170c2

source/client/stream_decoder.cc
```diff
@@ -19,6 +19,19 @@ void StreamDecoder::decodeHeaders(Envoy::Http::ResponseHeaderMapPtr&& headers, b
   response_header_sizes_statistic_.addValue(response_headers_->byteSize());
   const uint64_t response_code = Envoy::Http::Utility::getResponseStatus(*response_headers_);
   stream_info_.response_code_ = static_cast<uint32_t>(response_code);
+  const auto timing_header_name = Envoy::Http::LowerCaseString("x-nh-do-not-use-origin-timings");
+  const auto* timing_header = response_headers_->get(timing_header_name);
```
mum4k, line 23: (nit) Since the type isn't obvious from the statement, it may be a good idea to replace auto with the actual type here to improve readability.

Applies to timing_value below as well.

source/server/http_test_server_filter.cc
```diff
@@ -79,6 +104,10 @@ HttpTestServerDecoderFilter::decodeTrailers(Envoy::Http::RequestTrailerMap&) {
 void HttpTestServerDecoderFilter::setDecoderFilterCallbacks(
     Envoy::Http::StreamDecoderFilterCallbacks& callbacks) {
   decoder_callbacks_ = &callbacks;
+  // TODO(#486): this adds locking in the hot path. Consider moving this into a separate
+  // extension, which will also allow tracking multiple points via configuration.
+  last_request_delta_ns_ =
```
mum4k, line 109: Did we (or can we) verify the performance implication of this? How likely is this going to cause significant regressions in performance tests that use the test server?

If we can't verify it, we may be better off keeping this behind a feature flag to begin with? Or is it already on the path only when we ask for it by setting `emit_previous_request_delta_in_response_header`?

source/server/http_test_server_filter.h
```diff
@@ -2,20 +2,68 @@
...
   HttpTestServerDecoderFilterConfig(nighthawk::server::ResponseOptions proto_config);
+
+  /**
+   * @return const nighthawk::server::ResponseOptions& read-only reference to the proto config
+   * object.
+   */
   const nighthawk::server::ResponseOptions& server_config() { return server_config_; }
 
+  /**
+   * Gets the number of elapsed nanoseconds since the last call (server wide).
```
mum4k, line 36: How useful is this statistic if it is server wide as compared to having it per each session? I likely don't fully understand the use case for this statistic, so this may as well be working as intended.

source/server/http_test_server_filter.h
```diff
@@ -2,20 +2,68 @@
...
+   * @return uint64_t 0 on the first call, else the number of elapsed nanoseconds since the last
+   * call.
+   */
+  static uint64_t getElapsedNanosSinceLastRequest(Envoy::TimeSource& time_source);
+
 private:
+  /**
+   *  Utility class for thread safe tracking of elapsed monotonic time.
+   */
+  class ThreadSafeMontonicTimeStopwatch {
```
mum4k, line 49: Why is this class part of the HttpTestServerDecoderFilterConfig? is there anything the binds them together other than the fact that one uses the other? Or can we (for clarity and reusability) define it as a standalone one?

source/server/http_test_server_filter.h
```diff
@@ -2,20 +2,68 @@
...
+  public:
+    ThreadSafeMontonicTimeStopwatch() : start_(Envoy::MonotonicTime::min()) {}
```
mum4k, line 51: Can we document the constructor?

source/server/http_test_server_filter.h
```diff
@@ -2,20 +2,68 @@
...
+     * last invocation otherwise.
+     */
+    uint64_t getElapsedNsAndReset(Envoy::TimeSource& time_source);
+
+  private:
+    Envoy::Thread::MutexBasicLockable lock_;
+    Envoy::MonotonicTime start_ GUARDED_BY(lock_);
+  };
+
+  static ThreadSafeMontonicTimeStopwatch& lastRequestStopwatch() {
```
mum4k, line 64: Can we document this?

Note the current name suggests that maybe there are multiple of these, new ones created for each new "last request". Unless I misunderstood something, I think this returns a global singleton. If so, can we mention the fact in documentation and optionally try to find another name, e.g. `getRequestStopwatch()` or similar.

source/server/http_test_server_filter.h
oschaaf: The way I see it is that the current approach is going to be particularly useful in A/B tests, for example:

with and without an intermediary
with nighthawk client features that might perform relatively heavy lifting on cpu cores shared with the workers (e.g. stats flushing).
across different clients

So the use case is sanity checking and exposing a low effort means to dig up a potential root cause of latency measurement divergence.

I think per-worker or per-session tracking could also serve as a purpose to further dig into things as that would allow one to get a sense of balancedness across threads and sessions.  Also see https://github.com/envoyproxy/nighthawk/pull/477/files#r476803233 for an earlier discussion related to this.

source/server/http_test_server_filter.h
oschaaf: I didn't want other things to rely on it yet, but if we feel this is good to go for that, then we can move it out. Maybe it should go into a separate file then and have direct test coverage. Let me know what you think.

source/server/http_test_server_filter.cc
oschaaf: I couldn't measure a significant difference in manual tests. Having said that, I'd love to move this into its own extension because intuitively I'd say that there's no way that this can have a positive impact.
The locking now is unconditional, and it is hard to make it conditional because the configuration for it can be changed on a per-request basis. 

But a separate extension would allow one to easily take this out of the equation altogether, and possibly also allow one to inject it into multiple places in the server side extension pipeline if desired so, emitting these timings from various places. 
If you don't mind me bloating this PR a bit, I'd be up to move this into a new extension.

oschaaf:
@mum4k feedback partially addressed in e813e7525c20d7df3692a8b7b5c2534b2711152a. Resolved the relevant comments. For the comments that are still unresolved I left some questions/remarks.

source/server/http_test_server_filter.cc
mum4k: Considering the unknown performance impact, I think it would be worthwhile to move this to a new extension in this PR. Thank you for looking into it.

source/server/http_test_server_filter.h
mum4k: I think the idea of having it in its own file with test coverage is a good one and as a bonus we will end up with a simpler interface.

oschaaf:
@mum4k assuming that this passes CI, ready for further review.
There's now a `Stopwatch` interface as well as a separate extension, so the files to watch are

`http_time_tracking_*` and `stopwatch_*`

@eric846 the `FakeIncrementingMonotonicTimeSource` came in handy in the  FakeTimeStopwatchTest/ThreadedStopwatchSpamming test in here, nice

source/common/thread_safe_monotonic_time_stopwatch.h
```diff
@@ -0,0 +1,32 @@
...
+#include "nighthawk/common/stopwatch.h"
+
+#include "external/envoy/source/common/common/lock_guard.h"
+#include "external/envoy/source/common/common/thread.h"
+
+namespace Nighthawk {
+
+/**
+ *  Utility class for thread safe tracking of elapsed monotonic time.
+ */
```
mum4k, line 12: (nit) Can we include an example use of this class? Either here, or where the interface is defined.

source/server/http_time_tracking_filter.h
```diff
@@ -0,0 +1,80 @@
...
+   */
+  static ThreadSafeMontonicTimeStopwatch& getRequestStopwatch() {
+    MUTABLE_CONSTRUCT_ON_FIRST_USE(ThreadSafeMontonicTimeStopwatch); // NOLINT
+  }
+  const nighthawk::server::ResponseOptions server_config_;
+};
+
+using HttpTimeTrackingFilterConfigSharedPtr = std::shared_ptr<HttpTimeTrackingFilterConfig>;
+
+class HttpTimeTrackingFilter : public Envoy::Http::PassThroughFilter {
```
mum4k, line 59: Can we document this class and its constructor also?

mum4k:
Handing over to @dubious90 due to my vacation.

oschaaf:
Addressed @mum4k his comments in  70502e1. Marking this as ready for another round!

With respect to user-facing documentation, I'm planning for the following as one or more followups to this, where I'm trying to keep PR size reasonable:
- Add configuration (cli arg/proto arg) of the client with respect to the response header name that should be treated as a latency input that should be tracked in a histogram.
- End to end tests
- User facing documentation

oschaaf:
Opened #500 as a WIP follow up to this

source/server/http_time_tracking_filter_config.cc
```diff
@@ -0,0 +1,55 @@
...
+namespace Configuration {
+
+class HttpTimeTrackingFilterConfig
+    : public Envoy::Server::Configuration::NamedHttpFilterConfigFactory {
+public:
+  Envoy::Http::FilterFactoryCb
+  createFilterFactoryFromProto(const Envoy::Protobuf::Message& proto_config, const std::string&,
+                               Envoy::Server::Configuration::FactoryContext& context) override {
+
+    auto& validation_visitor = Envoy::ProtobufMessage::getStrictValidationVisitor();
```
dubious90, line 23: can we be explicit here, instead of auto.  I'm not seeing a place nearby where that type becomes obvious

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+#include "api/server/response_options.pb.h"
+#include "api/server/response_options.pb.validate.h"
+
+#include "server/configuration.h"
+#include "server/http_time_tracking_filter.h"
+#include "server/well_known_headers.h"
+
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
```
dubious90, line 19: Can we put our test code inside an anonymous namespace?

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+
+  // We don't override SetUp(): tests in this file will call setup() instead to avoid having to
+  // create a fixture per filter configuration.
+  void setup(const std::string& config) {
+    config_helper_.addFilter(config);
+    HttpIntegrationTest::initialize();
+  }
+
+  // Fetches a response with request-level configuration set in the request header.
+  Envoy::IntegrationStreamDecoderPtr getResponse(absl::string_view request_level_config,
```
dubious90, line 49: There's a lot of overloading happening here that I think makes it a little hard to read.  Can we name these functions differently? Or alternatively, are all of them actually necessary?

In particular, the one with no arguments seems to counter the recommendation here: https://google.github.io/styleguide/cppguide.html#Function_Overloading

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+    return response;
+  }
+
+  const Envoy::Http::TestRequestHeaderMapImpl request_headers_;
+};
+
+class HttpTimeTrackingFilterConfigTest : public testing::Test,
+                                         public Envoy::Event::TestUsingSimulatedTime {};
+
+// WARNING: Don't move this test. Keep it on top.
```
dubious90, line 86: This comment indicates a design problem to me.  Could we instead abstract this to a function that is called before every test?  And then, if it's necessary for it not be called multiple times, verify that there instead?

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  EXPECT_EQ(Server::HttpTimeTrackingFilterConfig::getElapsedNanosSinceLastRequest(time_system), 0);
+  time_system.setMonotonicTime(1ns);
+  EXPECT_EQ(Server::HttpTimeTrackingFilterConfig::getElapsedNanosSinceLastRequest(time_system), 1);
+}
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, HttpTimeTrackingIntegrationTest,
+                         testing::ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()));
+
+// Verify expectations with static/file-based time-tracking configuration.
+TEST_P(HttpTimeTrackingIntegrationTest, StaticConfiguration) {
```
dubious90, line 100: Applies to all tests:  Prefer names that are affirmations of the behavior being tested (usually including a verb)
Something more like:
```
TEST_P(HttpTimeTrackingIntegrationTest, ReturnsPositiveLatencyForStaticConfiguration) {
```

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  const Envoy::Http::HeaderEntry* latency_header =
+      response->headers().get(Envoy::Http::LowerCaseString(kLatencyResponseHeaderName));
+  ASSERT_NE(latency_header, nullptr);
+  int64_t latency;
+  EXPECT_TRUE(absl::SimpleAtoi(latency_header->value().getStringView(), &latency));
+  // TODO(oschaaf): figure out if we can use simtime here, and verify actual timing matches
+  // what we'd expect using that.
+  EXPECT_GT(latency, 0);
+
+  // Send a malformed config request header. This ought to shortcut and directly reply,
```
dubious90, line 130: Can these two parts where we check invalid configuration be their own test cases?

test/stopwatch_test.cc
```diff
@@ -0,0 +1,68 @@
...
+#include "external/envoy/test/test_common/simulated_time_system.h"
+#include "external/envoy/test/test_common/utility.h"
+
+#include "common/thread_safe_monotonic_time_stopwatch.h"
+
+#include "test/common/fake_time_source.h"
+
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
```
dubious90, line 15: Can we put test code inside an anonymous namespace?

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -0,0 +1,147 @@
...
+const std::string kDefaultProtoFragment = fmt::format(
+    "emit_previous_request_delta_in_response_header: \"{}\"", kLatencyResponseHeaderName);
+const std::string kProtoConfigTemplate = R"EOF(
+name: time-tracking
+typed_config:
+  "@type": type.googleapis.com/nighthawk.server.ResponseOptions
+  {}
+)EOF";
+
+class HttpTimeTrackingIntegrationTest
```
dubious90, line 33: In the other test file, we define classes, then the tests they're for, which seems a little cleaner to me than defining all of the classes, then all of the tests.  Is it possible to do that here too?

test/stopwatch_test.cc
```diff
@@ -0,0 +1,68 @@
...
+  EXPECT_EQ(stopwatch.getElapsedNsAndReset(time_system), 1);
+  time_system.setMonotonicTime(3ns);
+  EXPECT_EQ(stopwatch.getElapsedNsAndReset(time_system), 1);
+  time_system.setMonotonicTime(5ns);
+  EXPECT_EQ(stopwatch.getElapsedNsAndReset(time_system), 2);
+}
+
+class FakeTimeStopwatchTest : public testing::Test {};
+
+TEST_F(FakeTimeStopwatchTest, ThreadedStopwatchSpamming) {
```
dubious90, line 38: Unless I'm missing something, FakeTimeStopWatchTest isn't accomplishing anything here.  Could we just use the gmock TEST macro instead, that doesn't use a fixture?

test/stopwatch_test.cc
```diff
@@ -0,0 +1,68 @@
...
+  constexpr uint32_t kNumThreads = 100;
+  ThreadSafeMontonicTimeStopwatch stopwatch;
+  FakeIncrementingMonotonicTimeSource time_system;
+  std::vector<std::thread> threads(kNumThreads);
+  std::promise<void> signal_all_threads_running;
+  std::shared_future<void> future(signal_all_threads_running.get_future());
+
+  // The first call should always return 0.
+  EXPECT_EQ(stopwatch.getElapsedNsAndReset(time_system), 0);
+  for (auto& thread : threads) {
```
dubious90, line 49: in my opinion, std::thread& isn't really long enough to justify the readability loss of using auto here.

source/client/stream_decoder.cc
oschaaf: re: separately, can you explain why having a configurable header name here is beneficial? just not sure why we wouldn't just dictate this

Well, in my opinion, there's certainly something to say for dictating the names: less code, less configuration. Here are the thoughts I had that made me lean towards allow the end user control of the header names involved. I think this mostly applies to both the client and test-server aspects of it: 

1. Being able to configure the header names both on the server and client side of Nighthawk decouples them. This flexibility may come in handy when mixing Nighthawk components with other OSS or homegrown clients/proxies/test servers with similar capabilities (especially when these dictate the naming involved). 
2. If both the proxy and test server emit timing data using different header names, one can switch which one the clients should track
3. If multiple proxies or test-servers are set up to emit latencies using different names in horizontal scaling, it could be used to track a single instance instead of all of them. 
4. If timing-filter gets extended to emit more timings, and we end up with multiple headers, clients can easily switch which one they track. 
5. In examples the header name could be a little bit longer for clarity, but in final testing, one might want to opt to make it as short as possible to minimise overhead. 

Considering the above, let me know what you think;

source/common/BUILD
```diff
@@ -70,6 +70,24 @@ envoy_cc_library(
...
+envoy_cc_library(
+    name = "thread_safe_monotonic_time_stopwatch_lib",
+    srcs = [
+        "thread_safe_monotonic_time_stopwatch.cc",
+    ],
+    hdrs = [
+        "thread_safe_monotonic_time_stopwatch.h",
+    ],
+    repository = "@envoy",
+    visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
oschaaf: I think this is a good point, I filed an issue tagged as tech-debt to track this: https://github.com/envoyproxy/nighthawk/issues/502

source/server/http_time_tracking_filter.cc
```diff
@@ -0,0 +1,68 @@
+#include "server/http_time_tracking_filter.h"
```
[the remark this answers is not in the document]
oschaaf: Yes I think that would make sense. Alternatively, we could keep this directory just for extensions and their configuration, and re-home everything else that's in there so far: configuration.h/cc (helpers), well_known_headers.h (http header name definitions). Should I file an issue to track this?

source/server/http_time_tracking_filter.cc
```diff
@@ -0,0 +1,68 @@
...
+uint64_t
+HttpTimeTrackingFilterConfig::getElapsedNanosSinceLastRequest(Envoy::TimeSource& time_source) {
+  return getRequestStopwatch().getElapsedNsAndReset(time_source);
+}
+
+HttpTimeTrackingFilter::HttpTimeTrackingFilter(HttpTimeTrackingFilterConfigSharedPtr config)
+    : config_(std::move(config)) {}
+
+Envoy::Http::FilterHeadersStatus
+HttpTimeTrackingFilter::decodeHeaders(Envoy::Http::RequestHeaderMap& headers, bool /*end_stream*/) {
```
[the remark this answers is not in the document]
oschaaf: My 2p:

1. I was thinking that this doesn't hurt, and may avoid a future reader having to look it up? 
2. For context: we derive from `PassThroughFilter` it as it conveniently stubs all the overrides precisely like we need, allowing us to override and focus on just the virtual methods of the interface that we are interested in. Maybe this deserves a code-level comment, wdyt?  By throwing an error, do you mean that it replies with a 500 when it doesn't understand the configuration? Could you elaborate on the nature of the potential problem you would like to rule out?

source/common/thread_safe_monotonic_time_stopwatch.h
```diff
@@ -0,0 +1,40 @@
...
+class ThreadSafeMontonicTimeStopwatch : public Stopwatch {
+public:
+  /**
+   * Construct a new ThreadSafe & MontonicTime-based Stopwatch object.
+   */
+  ThreadSafeMontonicTimeStopwatch() : start_(Envoy::MonotonicTime::min()) {}
+
+  /**
+   * @param time_source used to obtain a sample of the current monotonic time.
+   * @return uint64_t 0 on the first invocation, and the number of elapsed nanoseconds since the
```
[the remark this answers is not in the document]
oschaaf: Well, while I didn't consider `absl::Duration` (I didn't know it existed). But I did consider `std::chrono::duration`, and ended up leaning towards `uint64_t` because that:
- can't turn negative, which doesn't make sense here
- is less prone to some of the trickyness with conversions vs the underlying types I ran into over at https://github.com/envoyproxy/envoy/pull/9229#issuecomment-562498924. 

Having said that, I feel this is a subjective matter, so I'd be happy to change this if you lean towards `absl::Duration` after reading my considerations above.

source/server/http_time_tracking_filter.cc
```diff
@@ -0,0 +1,68 @@
...
+
+Envoy::Http::FilterHeadersStatus
+HttpTimeTrackingFilter::encodeHeaders(Envoy::Http::ResponseHeaderMap& response_headers, bool) {
+  if (!json_merge_error_) {
+    const std::string previous_request_delta_in_response_header =
+        base_config_.emit_previous_request_delta_in_response_header();
+    if (!previous_request_delta_in_response_header.empty() && last_request_delta_ns_ > 0) {
+      response_headers.appendCopy(
+          Envoy::Http::LowerCaseString(previous_request_delta_in_response_header),
+          absl::StrCat(last_request_delta_ns_));
```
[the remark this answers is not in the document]
oschaaf: Well, we're converting an `uint64_t` to a `string` here, and what made me think `absl::StrCat` was the right way to do it was [a comment over at absl's number.h](https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/refs/heads/lts_2018_06_20/absl/strings/numbers.h#20):

```c++
// This package contains functions for converting strings to numbers. For
--
  | // converting numbers to strings, use `StrCat()` or `StrAppend()` in str_cat.h,
  | // which automatically detect and convert most number values appropriately.
```

I do agree though, that I suspect that if I would be at the other end of this review, the same thing would have popped up in my mind. Maybe we should define a shorthand that does look more intuitive in the case where we're converting a single int to string, otoh I'm not sure it's worth it?

test/server/http_time_tracking_filter_integration_test.cc
oschaaf: The problem is that the static persists across tests. So only the first call from this compilation unit will truly excite the behaviour we are after consolidating in the test here, regardless of which test triggers it. I fully agree it's a code smell, but resolving it either requires  introducing test-specific code to reset the static to its initial value. I'll be in touch on Slack to discuss this some more -- I'm not sure I grasp what you propose here so I think I need some more context -- I'm surely open to ideas to improve upon this.

oschaaf: I think I resolved this code smell in 7774e2d; indeed design was flawed. With those changes, tests no longer need to be order-dependent, as we can control instantiation, the static/singleton wasn't needed. Verified that with gtest's `--gtest_shuffle` flag .

oschaaf:
@dubious90 Feedback partially addressed in 7774e2d

test/server/http_time_tracking_filter_integration_test.cc
oschaaf: I'd like to address this in a follow up; this is partially duplicated from other NH extensions, and I think this could be taken on as part of https://github.com/envoyproxy/nighthawk/issues/498.
Not having the code diverge makes it easier to refactor for now. Are you ok with that?

source/client/stream_decoder.cc
dubious90: This makes sense.  Thanks for the explanation.

source/common/thread_safe_monotonic_time_stopwatch.h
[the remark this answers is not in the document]
dubious90: That makes sense.  I don't know if it's worth bringing in absl::Duration just for this.  If we wanted to, we could try to bring in absl::Duration on a more global change later.

source/server/http_time_tracking_filter.cc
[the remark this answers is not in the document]
dubious90: An issue to track this would be great.  I'm happy to defer to you on what makes the most sense.  The `filters` subdirectory was just what made sense to me on a quick lookthrough.

source/server/http_time_tracking_filter.cc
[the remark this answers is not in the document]
dubious90: 1. Sounds good.  I was going back and forth on whether it was helpful.  Decided to raise the question to hear your opinion.
2. I did mean reply with a 500, sorry for being unclear.  I think all of that makes sense.  Just wasn't sure if we were breaking some sort of envoy contract.  I had looked up PassThroughFilter and it specifies that it always returns Continue.  I wanted to make sure that we were using this as a convenience and it was okay to break that rule, rather than it being a contract for the class that is relied upon

source/server/http_time_tracking_filter.cc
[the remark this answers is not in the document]
dubious90: Oops got that backward. My bad. This probably doesn't matter unless we are worrying about very high performance here, but I think StrCat might be a heavy hand for converting one int to a string.  It's optimized for concatenating strings and ints together efficiently.  Happy to leave it here, and agree that it's not worth creating a helper function.

test/server/http_time_tracking_filter_integration_test.cc
dubious90: Yep. That's completely reasonable.  Thanks!

PR #483, opened 2020-08-24
Adaptive Load main loop

eric846:
This is the main function of the Adaptive Load Controller library:

- Check the input proto for errors and apply default input values.
- For the Adjusting Stage: One big `while` loop:
  - Check for convergence deadline exceeded.
  - Get the latest dynamically generated CommandLineOptions from the StepController.
  - Run a short benchmark with the Nighthawk Service.
  - Obtain metric values for this benchmark from MetricsPlugins.
  - Score the metrics using ScoringFunction plugins.
  - Report scores back to the StepController, which recalculates the load for the next iteration.
  - Check for StepController convergence.
  - Check for StepController doom.
- For the Testing Stage: Run one long benchmark on the Nighthawk Service at the converged load.

The unit test has an example of mocking a gRPC stub, which is not easy.


mum4k:

tools/check_format.sh
```diff
@@ -8,7 +8,7 @@ TO_CHECK="${2:-$PWD}"
 bazel run @envoy//tools:code_format/check_format.py -- \
   --skip_envoy_build_rule_check  --namespace_check Nighthawk \
   --build_fixer_check_excluded_paths=$(realpath ".") \
-  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,test_common,test \
+  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,test_common,grpcpp,test \
```
eric846, line 11: Without separating grpcpp into its own section, the formatter arranges the include files in an order that is then detected as invalid.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+absl::Status CheckSessionSpec(const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec) {
+  std::vector<std::string> errors;
+  if (spec.nighthawk_traffic_template().has_duration()) {
+    errors.emplace_back(
+        "nighthawk_traffic_template should not have |duration| set. Set |measuring_period| "
+        "and |testing_stage_duration| in the AdaptiveLoadSessionSpec proto instead.");
+  }
+  if (spec.nighthawk_traffic_template().has_open_loop()) {
+    errors.emplace_back(
+        "nighthawk_traffic_template should not have |open_loop| set. Adaptive Load always "
```
oschaaf, line 272: I may misremember, but open loop needs to be *set* to avoid closed loop mode?
One tought I have here is that there are OSS load gens out there that don't have an open
loop mode, or have one with a partial feedback loop (e.g. when hitting connection limits). That might
be worth considering in the context of enforcing this, if encouraging other implementations of the 
adaptive load controllers is desired?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+      }
+    } else {
+      errors.emplace_back(absl::StrCat(
+          "MetricSpec referred to nonexistent metrics_plugin_name '",
+          metric_spec.metrics_plugin_name(),
+          "'. You must declare the plugin in metrics_plugin_configs or use plugin ",
+          "'nighthawk.builtin'. Available plugins: ", absl::StrJoin(plugin_names, ", "), "."));
+    }
+  }
+  if (errors.size() > 0) {
```
oschaaf, line 328: I'm not sure how large this can grow; but if it hypothetically can grow very large, it might be a nice
courtesy to bound this somehow, possibly both in the input to avoid OOM and on the output to avoid spam.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+          step_controller_or.status().message()));
+  StepControllerPtr step_controller = std::move(step_controller_or.value());
+
+  for (const nighthawk::adaptive_load::MetricSpecWithThreshold& threshold :
+       spec.metric_thresholds()) {
+    *output.mutable_metric_thresholds()->Add() = threshold;
+  }
+
+  Envoy::MonotonicTime start_time = time_source.monotonicTime();
+  while (!step_controller->IsConverged()) {
```
oschaaf, line 403: Assiuming we can't converge right away, I wonder if expressing this along the lines of:

```c++
do {
   ....
   terminated = step_controller->IsDoomed(doom_reason) || step_controller->IsConverged(); 
} while(!terminated);

if (!doom_reason.empty()) {
  // doom handling
}
```

would help clarity. entirely subjective -- please ignore if you feel otherwise.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+    ENVOY_LOG_MISC(info, message);
+    return false;
+  }
+  command_line_options = command_line_options_or.value();
+  return true;
+}
+
+} // namespace
+
+AdaptiveLoadSessionOutput PerformAdaptiveLoadSession(
```
dubious90, line 364: I realize that this API was already partially reviewed, but that was before we brought in absl::Status.  I question whether much of this would be easier to follow (and potentially easier to use) if you used StatusOr rather than having a proto that contains status fields.  Open to pushback here.

test/adaptive_load/fake_time_source.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "envoy/common/time.h"
+
+namespace Nighthawk {
+/**
+ * Fake time source that ticks 1 second on every query, starting from the Unix epoch. Supports only
+ * monotonicTime().
+ */
+class FakeIncrementingMonotonicTimeSource : public Envoy::TimeSource {
```
dubious90, line 10: I thought envoy already supported mock timers.  Is there a reason why we need to implement a fake?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+  nighthawk::client::ExecutionRequest request;
+  nighthawk::client::ExecutionResponse response;
+  *request.mutable_start_request()->mutable_options() = options;
+
+  ::grpc::ClientContext context;
+  std::shared_ptr<::grpc::ClientReaderWriterInterface<nighthawk::client::ExecutionRequest,
+                                                      nighthawk::client::ExecutionResponse>>
+      stream(nighthawk_service_stub->ExecutionStream(&context));
+
+  stream->Write(request);
```
oschaaf, line 71: I think `Write()` and `WritesDone()` return a bool indicating success, should we handle those?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+ * @param command_line_options Nighthawk Service benchmark request proto generated by the
+ * StepController, without the duration set.
+ * @param duration Duration to insert into the benchmark request.
+ *
+ * @return ExecutionResponse If we reached the Nighthawk Service, this is the raw ExecutionResponse
+ * proto, containing the benchmark data or possibly an error message from Nighthawk Service; if we
+ * had trouble communicating with the Nighthawk Service, we insert the details into the
+ * |error_detail| of this proto.
+ */
+nighthawk::client::ExecutionResponse PerformNighthawkBenchmark(
```
oschaaf, line 54: Reading this file again, and probably for after this PR has landed / a nice to have, I wonder how hard it would be to extract an interface from this that could (hypothetically) support both CLI and gRPC based flows.
Then this implementation would be the first one, and serve as the canonical example of how to make the rubber hit the road with respect to the adaptive load controller framework. 

If this is desireable, one thing that I think that would be good to consider is naming. For example, maybe this class could then better be called something like `AdaptiveNighthawkGrpcLoadController`?  Also, some of the class method names mention "Nighthawk". We'd want to be less specific, as some of those might end up being overrides of the interface spec.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
+    const AdaptiveLoadSessionSpec& spec,
+    const absl::flat_hash_map<std::string, MetricsPluginPtr>& name_to_custom_plugin_map,
+    const nighthawk::client::CommandLineOptions& command_line_options,
+    const Envoy::Protobuf::Duration& duration) {
+  nighthawk::client::ExecutionResponse response =
+      PerformNighthawkBenchmark(nighthawk_service_stub, command_line_options, duration);
+  BenchmarkResult result = AnalyzeNighthawkBenchmark(response, spec, name_to_custom_plugin_map);
+  for (const MetricEvaluation& evaluation : result.metric_evaluations()) {
+    ENVOY_LOG_MISC(info, "Evaluation: {}", evaluation.DebugString());
```
oschaaf, line 200: If `info` the right level?

test/adaptive_load/fake_time_source.h
```diff
@@ -0,0 +1,29 @@
...
+  Envoy::SystemTime systemTime() override;
+  /**
+   * Ticks forward 1 second on each call.
+   *
+   * @return Envoy::MonotonicTime Fake time value.
+   */
+  Envoy::MonotonicTime monotonicTime() override;
+
+private:
+  int unix_time_{0};
```
oschaaf, line 26: Is this specific to unix? Maybe rename it to something not tied to a platform?

test/adaptive_load/adaptive_load_controller_test.cc
```diff
@@ -0,0 +1,847 @@
...
+ * Creates a valid TypedExtensionConfig proto selecting the real BinaryScoringFunction plugin
+ * and configuring it with a threshold.
+ *
+ * @param lower_threshold Threshold value to set within the config proto.
+ *
+ * @return TypedExtensionConfig Full scoring function plugin spec that selects
+ * nighthawk.binary-scoring and provides a config.
+ */
+envoy::config::core::v3::TypedExtensionConfig
+MakeLowerThresholdBinaryScoringFunctionConfig(double lower_threshold) {
```
oschaaf, line 65: `const double lower_treshold`?

oschaaf (commented):
Another round of feedback; I've read most of the PR now, but still need to go through most of the tests

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+ * had trouble communicating with the Nighthawk Service, we insert the details into the
+ * |error_detail| of this proto.
+ */
+nighthawk::client::ExecutionResponse PerformNighthawkBenchmark(
+    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
+    const nighthawk::client::CommandLineOptions& command_line_options,
+    const Envoy::Protobuf::Duration& duration) {
+  nighthawk::client::CommandLineOptions options = command_line_options;
+  *options.mutable_duration() = duration;
+  options.mutable_open_loop()->set_value(false);
```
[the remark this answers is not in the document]
eric846: This was exactly the opposite of what I intended! We should always be using open loop mode for adaptive load.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+  *request.mutable_start_request()->mutable_options() = options;
+
+  ::grpc::ClientContext context;
+  std::shared_ptr<::grpc::ClientReaderWriterInterface<nighthawk::client::ExecutionRequest,
+                                                      nighthawk::client::ExecutionResponse>>
+      stream(nighthawk_service_stub->ExecutionStream(&context));
+
+  stream->Write(request);
+  stream->WritesDone();
+  if (!stream->Read(&response)) {
```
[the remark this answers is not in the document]
eric846: Done and done.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+}
+
+/**
+ * Returns a copy of the input spec with default values inserted.
+ *
+ * @param original_spec Valid adaptive load session spec.
+ *
+ * @return Adaptive load session spec with default values inserted.
+ */
+AdaptiveLoadSessionSpec SetDefaults(const AdaptiveLoadSessionSpec& original_spec) {
```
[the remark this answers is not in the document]
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: I definitely meant to *enable* open loop mode automatically and not allow mentioning it in the input. The intent here was to avoid confusion by not letting the open loop option even appear in example protos, whether true or false, since we force  open loop = true anyway.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: This should be bounded by the total number of fields in user-defined config proto. Each field can only result in at most one error line, as far as I know.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+
+  stream->Write(request);
+  stream->WritesDone();
+  if (!stream->Read(&response)) {
+    response.mutable_error_detail()->set_code(::grpc::UNKNOWN);
+    response.mutable_error_detail()->set_message("Nighthawk Service did not send a response.");
+  }
+  ::grpc::Status status = stream->Finish();
+  if (!status.ok()) {
+    response.mutable_error_detail()->set_code(status.error_code());
```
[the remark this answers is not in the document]
eric846: Done -- we are now all-in on StatusOr!

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+ * @param nighthawk_response Proto returned from Nighthawk Service describing a single benchmark
+ * session.
+ * @param spec Top-level proto defining the adaptive load session.
+ * @param name_to_custom_metrics_plugin_map Common map from plugin names to MetricsPlugins, loaded
+ * and initialized once at the beginning of the session and passed to all calls of this function.
+ *
+ * @return BenchmarkResult Proto containing metric scores for this Nighthawk Service benchmark
+ * session, or an error propagated from the Nighthawk Service or MetricsPlugins.
+ */
+BenchmarkResult AnalyzeNighthawkBenchmark(
```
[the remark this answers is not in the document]
eric846: Extracted EvaluateMetric function, should make it a bit simpler.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
+#include <chrono>
+
+#include "envoy/common/exception.h"
```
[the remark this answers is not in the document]
eric846: This main API is only a bare function, so there's no overridden interface or anything else to declare beyond include/adaptive_load/adaptive_load_controller.h

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+#include "absl/status/status.h"
+#include "absl/strings/str_join.h"
+#include "adaptive_load/metrics_plugin_impl.h"
+#include "adaptive_load/plugin_loader.h"
+
+namespace Nighthawk {
+
+namespace {
+
+using namespace Envoy; // for Envoy::Logger
```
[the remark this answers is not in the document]
eric846: This is needed because the Envoy logging macros assumes the code is in namespace Envoy. I don't see a way around it.

I think most Nighthawk code is already using namespace Envoy, for the same reason, via this include file: https://github.com/envoyproxy/nighthawk/blob/6a0d3a3c9c306bf4914ddc460b6fd5931e1685e4/source/client/benchmark_client_impl.h#L34

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,444 @@
...
+  if (nighthawk_response.error_detail().code() != ::grpc::OK) {
+    return benchmark_result;
+  }
+  // MetricSpecs in original order of definition.
+  std::vector<const nighthawk::adaptive_load::MetricSpec*> metric_specs;
+  // Pointer to the corresponding ThresholdSpec, or nullptr for informational metrics.
+  absl::flat_hash_map<const nighthawk::adaptive_load::MetricSpec*,
+                      const nighthawk::adaptive_load::ThresholdSpec*>
+      threshold_spec_from_metric_spec;
+  for (const MetricSpecWithThreshold& metric_threshold : spec.metric_thresholds()) {
```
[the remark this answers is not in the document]
eric846: Commented the complicated map and also extracted another function.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done, adopted StatusOr throughout, including the API.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Changed many of these to `error`

test/adaptive_load/fake_time_source.h
eric846: Renamed to `seconds_since_epoch_`

test/adaptive_load/adaptive_load_controller_test.cc
eric846: Shouldn't need that as it's passed by value

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: There's actually not a class here, only a bare public function and helper functions in the private namespace. This might all have to be refactored if we need to open it to overriding some of the parts, including rethinking the signatures of the functions that rely on Nighthawk-specific protos. Opened #485 to consider this area.

test/adaptive_load/fake_time_source.h
eric846: I see a mock TimeSource being used, and there's TestTimeSystem, but here we need something where:
- time advances between calls
- no intervention is needed between calls

because once we unleash the main loop we can't pause it to manipulate the clock.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
oschaaf: I see; but I still am curious as to why we are enforcing open loop mode? Is there a fundamental incompatibility with closed loop mode? I can see that encouraging best practices is a good thing, but enforcing them would bar users from experimenting with the different testing methodologies? E.g. "Let's see how far we can raise rps before the blocking histogram scores Y". Maybe a silly question to seek an answer to, but it's hard to see up front why someone would like to investigate that.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
oschaaf: Yeah, this is quite annoying. Maybe we should put this on the radar by filing an issue? It would be great to resolve, to avoid the broken window effect? :-) I wonder if we can make a modification to Envoy to avoid having to do this.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,481 @@
...
+
+  ::grpc::ClientContext context;
+  std::shared_ptr<::grpc::ClientReaderWriterInterface<nighthawk::client::ExecutionRequest,
+                                                      nighthawk::client::ExecutionResponse>>
+      stream(nighthawk_service_stub->ExecutionStream(&context));
+
+  if (!stream->Write(request)) {
+    return absl::UnknownError("Failed to write request to the Nighthawk Service gRPC channel.");
+  }
+  if (!stream->WritesDone()) {
```
oschaaf, line 75: Would using `else if` here be better? `WritesDone` probably will behave nicely, but to me that would feel cleaner.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
oschaaf: resolving this comment as the changes requested by me and @dubious90 look good to me (@dubious90 please check if you agree with that)

source/adaptive_load/adaptive_load_controller_impl.cc
oschaaf: Thanks for the context; still had a question about the rationale around this which I just asked in a comment higher up. Resolving this one to deduplicate / consolidate the discussion on a single thread.

source/adaptive_load/adaptive_load_controller_impl.cc
oschaaf: nice, thank you

test/adaptive_load/fake_time_source.h
oschaaf: So personally, I love the simplicity of this. It might be possible to use simulated time, but then we may need another thread to advance time or some other relatively complex construct. Also, as I think monotonic time promises us to advance each time we call it, this seems quite an accurate modelling of it. So personally, when weighing the potential duplication introduced against the simplicity we win, I'm rooting for the way this is right now.

source/adaptive_load/adaptive_load_controller_impl.cc
oschaaf: Thanks for providing context and filing that issue. Resolving this here.

source/adaptive_load/adaptive_load_controller_impl.cc
oschaaf: Thanks; resolving.

test/adaptive_load/adaptive_load_controller_test.cc
oschaaf: ah yes, sorry about that.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
eric846: There's certainly no problem with letting the user experiment, especially in the future if they are experimenting with something besides rps. I can just default open-loop to true when it's unset. I'll mention the override in the proto, but remove this enforcement.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
eric846: Agreed, filed #488.

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
oschaaf: Awesome

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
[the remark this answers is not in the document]
dubious90: Thanks for explaining!

eric846:
I'm in the process of deleting the status code fields from the protos, which serve no purpose now that we are using absl::StatusOr. This requires some minor updates to the test and fake StepController s since the fake behavior was partly based on passing around protos containing error states.

oschaaf (commented):
LGTM -- but one last ask: 
Coverage [1] shows that some of the non-happy flow isn't covered. Would it be reasonably doable to cover these cases?

[1] https://18620-180498819-gh.circle-artifacts.com/0/coverage/source/adaptive_load/adaptive_load_controller_impl.cc.gcov.html

eric846:
Added several more commits with nontrivial content, starting from "delete proto error status fields." I had to introduce new functionality to the FakeStepController to be able to cover the last bit of the main function.

mum4k:
I am wondering what we can do about the fact that this PR is both fairly long and hard to navigate / review. The main issue may be that there are multiple effort tracks combined in this PR which makes it hard to mentally map the code changes to individual tracks. As it stands, I don't feel confident enough that I won't miss anything important while reviewing this.

The best solution might be to shelve this PR, while keeping it as a reference for the reviewers who already posted comments in here. We can then try to split the individual effort tracks into multiple separate PRs, each of them ideally focused on just one thing. This can also allow us to better track the reasoning for individual changes in the description of each PR.

I may not know enough to prescribe the split exactly, but here is a rough list of ideas. Please feel free to adapt this to reality as you see fit.

- a separate PR performing the API changes related to StatusOr.
- a separate PR fixing any proto errors.
- a separate PR implementing the fake time source.
- a separate PR performing changes to the fake step controller.

Next we should address how long the adaptive load controller implementation is itself. Skimming over the helper functions in the anonymous namespace - some of them look like they contain enough logic to warrant their own tests. I will leave this to your best judgment. Having private functions makes the code agile, but there is a balance between keeping everything private and having separate test coverage for large enough behaviors. Maybe we can find some such candidates and their move to standalone libraries with test coverage will both simplify the PRs and make it faster to find errors when their individual unit tests break. E.g maybe the function that evaluates metrics deserves its own library? This should leave us in a situation where the main library of the adaptive load controller only contains its main business logic.

Any such libraries can then be sent as separate PRs. We can then send one last PR to add the main logic of the adaptive load controller.

eric846:
Commencing this split now

PR #491, opened 2020-08-27
Adaptive Load fake time source

eric846:
A fake `Envoy::TimeSource` that ticks forward every time it is called, starting from 1970-01-01 00:00:00. Only `monotonicTime()` is implemented.

This simple autonomous behavior is needed because the adaptive load controller checks the clock many times automatically and can't be interrupted in the middle of the run to update the fake time.

Part 2 of splitting PR #483.


oschaaf (approved):
Nice, LGTM

test/adaptive_load/fake_time_source.cc
```diff
@@ -0,0 +1,17 @@
+#include "test/adaptive_load/fake_time_source.h"
```
oschaaf, line 1: Oh sorry, I was looking at this in the context of `ThreadSafeMonoticTimeSource` which might also be coming up, to see where this ended up in terms of physical location. I wonder if instead of placing this under `test/adaptive_load/`, moving it into a shared directory like `test/common/` would be possible?

oschaaf: ('common' might not be the best name actually, but imho this ends up in a place where in terms of directory layout this and a new non-test time source would both end up with a consistent location, `test/foo` and `source/foo`)

dubious90: This makes sense.  Holding off on merging until Eric weighs in on where this should live.

eric846: Moved to `test/common`.

oschaaf (approved):

test/adaptive_load/fake_time_source.cc
eric846: What do you think of `test/common`?

- It's parallel to `source/common`
- It definitely makes more sense than `test/adaptive_load` since a fake time source has nothing to do with adaptive load.
- I don't want to add something directly in `test/`.
- `test/test_common` looks like something we are forced to have by Envoy precedent but didn't make sense for other Nighthawk stuff?
- It would be alongside tests for `ThreadSafeMonotonicTimeSource` and could be used from them if it made sense.

If we rearrange the whole `test/` directory later, we can easily eliminate `test/common` if there's an alternative.

Issue #466, opened 2020-08-19
Consider consolidating IsConverged and IsDoomed in Adaptive Load StepController


eric846:
Both are const methods that can change value from the same events.

- We might think of other predicates.
- It could also allow us to encode constraints like doomed implying not converged. 
- It may simplify testing.

oschaaf:
I was looking at https://github.com/envoyproxy/nighthawk/pull/492, another thought related to this topic
So please hear me out ... currently on master StepController looks like:

```c++  
  virtual absl::StatusOr<nighthawk::client::CommandLineOptions> GetCurrentCommandLineOptions() const PURE;
  virtual bool IsConverged() const PURE;
  virtual bool IsDoomed(std::string& doom_reason) const PURE;
  virtual void UpdateAndRecompute(const nighthawk::adaptive_load::BenchmarkResult& result) PURE;
```

I am now wondering, what if StepController would condense these 4 methods into a single one. 
Thinking it through, I end up with something like:

```c++
  //  Return value imposes three possible transitions for consuming code:  
  //    - StatusOr indicated success AND the inner unique_ptr == nullptr: converged / terminate succesfully
  //    - StatusOr indicated success AND  the inner unique_ptr != nullptr: more work
  //    - StatusOr indicated failure: doomed / terminate with an error
  // result parameter can equate to nullptr, implying the first call. This might be ASSERTED on to avoid mistakes in consuming code.
  virtual absl::StatusOr<std::unique_ptr<nighthawk::client::CommandLineOptions>> 
       NextCommandLine(const std::unique_ptr<const nighthawk::adaptive_load::BenchmarkResult>& result) const PURE;
```

This thought wouldn't make sense if it is valid/desirable that consuming (prod) code is able to call `UpdateAndRecompute` multiple times without interleaving it with `GetCurrentCommandLineOptions()` as it stands on master. But if that's not the case, then doing this could be considered api abuse, and collapsing the interface to a single method might rule that out?

PR #492, opened 2020-08-27
Adaptive Load FakeStepController doom update

eric846:
Updates FakeStepController:

- Set doomed state when `FakeStepController::UpdateAndRecompute()` receives a `BenchmarkResult` with a negative metric score. Previously `BenchmarkResult` could contain an error status, and we used the error status to trigger simulated doom. Now errors do not get passed to  `StepController::UpdateAndRecompute()`, so we need a new mechanism.
- Adds the ability to simulate an `InputVariableSetter` failure in `GetCurrentCommandLineOptions()`. The artificial failure is specified in the config proto.
- Adds a countdown mechanism so that `GetCurrentCommandLineOptions()` returns successfully until `UpdateAndRecompute()` has been called a configured number of times, then starts returning the simulated input variable setter failure mentioned above. This simple autonomous behavior is needed to test error handling in the testing stage without triggering an early exit in the adjusting stage; the adaptive load controller can't be paused in the middle of the test to reconfigure the `FakeStepController` to change its behavior.

The behavior of `FakeStepController` after an `UpdateAndRecompute()` with a `BenchmarkResult` containing the given metric scores:
- Any positive score: converged
- All scores zero: neither converged nor doomed
- Any negative score: doomed

Part 3 of splitting PR #483.


test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.proto
```diff
@@ -5,12 +5,23 @@ package nighthawk.adaptive_load;
...
+// from input value setting every time, or return an error after some number of UpdateAndRecompute()
+// iterations.
 message FakeStepControllerConfig {
-  // RPS that should always be returned. Optional, default 0.
+  // RPS that should always be returned, except when artificial errors are configured. Optional,
+  // default 0.
   int32 fixed_rps_value = 1;
   // Artificial error that the plugin factory should return during validation. Optional.
   google.rpc.Status artificial_validation_failure = 2;
+  // Artificial error that should be returned from GetCurrentCommandLineOptions(). Optional.
```
dubious90, line 19: Can we enhance this comment to indicate the relation to the countdown?

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.proto
```diff
@@ -5,12 +5,23 @@ package nighthawk.adaptive_load;
...
+  google.rpc.Status artificial_input_setting_failure = 3;
```
dubious90, line 20: I think I'm having trouble understanding the full use case here.  If this is 3, then we are going to succeed 3 times, then fail on the 4th attempt. Makes sense, but I'm not sure I understand why that's useful.  What is the test that you're supporting by creating it in this way?

dubious90:
@mum4k Can you give this a review and assign back to me when done?

mum4k:
@eric846 please address / respond to the comments from @dubious90 as that can help me by providing additional context for the review.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.cc
```diff
@@ -15,9 +15,10 @@ absl::Status StatusFromProtoRpcStatus(const google::rpc::Status& status_proto) {
 } // namespace
 
 FakeStepController::FakeStepController(
-    const nighthawk::adaptive_load::FakeStepControllerConfig& config,
+    nighthawk::adaptive_load::FakeStepControllerConfig config,
     nighthawk::client::CommandLineOptions command_line_options_template)
-    : is_converged_{false}, is_doomed_{false}, fixed_rps_value_{config.fixed_rps_value()},
+    : input_setting_failure_countdown_{config.artificial_input_setting_failure_countdown()},
+      config_{std::move(config)}, is_converged_{false}, is_doomed_{false},
```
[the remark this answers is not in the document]
eric846: Funny story, this was actually required by clang-tidy. If the function implementation is making a copy of a const reference parameter, clang-tidy tells you to change the parameter to be pass by value so it's obvious to the caller that a copy is happening. The argument gets copied at the call site into an unnamed temporary value. Then in the constructor we can actually move this temporary value into the field without incurring the cost of a second copy.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.proto
eric846: Done

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.proto
eric846: `GetCurrentCommandLine()` can return an error status that should be handled cleanly by the main controller loop. The controller calls `GetCurrentCommandLine()` repeatedly during the adjusting stage until the step controller says it's converged, and then in the testing stage we call `GetCurrentCommandLine()` one last time, reusing the last converged value. 

In order to test handling of these errors, we need the `FakeStepController` to return successfully from `GetCurrentCommandLine()` during the adjusting stage, but then start returning errors just in time for the testing stage. We don't have any way to update the `FakeStepController` during the run, so it has to somehow be programmed up front to behave differently at different times.

An alternative would be for magic values in `UpdateAndRecompute()` to trigger `GetCurrentCommandLine()` error behavior. We already use magic values to control convergence and doom. But there's only so much information we can encode in metric score `double`s without having it get out of hand. Currently the `UpdateAndRecompute()` behavior is: zero scores=non-converged non-doomed, any positive score=converged, any negative score=doomed.

This trick wouldn't be necessary if the step controller were aware of what stage it was operating in.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
```diff
@@ -163,5 +239,29 @@ TEST(MakeFakeStepControllerPluginConfigWithValidationError,
   EXPECT_EQ(plugin_or.status().message(), kValidationErrorMessage);
 }
 
+TEST(MakeFakeStepControllerPluginConfigWithInputSettingError,
+     ProducesFakeStepControllerPluginWithConfiguredErrorAndCountdown) {
+  const int kExpectedRpsValue = 123;
+  std::string kInputSettingErrorMessage = "artificial input setting error";
```
oschaaf, line 245: const

oschaaf (approved):
Just a const nit, LGTM otherwise. Left a question that popped up over at https://github.com/envoyproxy/nighthawk/issues/466#issuecomment-682442611 as that's out of scope here.

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.h
```diff
@@ -22,7 +22,7 @@ class FakeStepController : public StepController {
    * @param config FakeStepControllerConfig proto for setting the fixed RPS value.
```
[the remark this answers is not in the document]
eric846: Done

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller.h
```diff
@@ -100,4 +103,21 @@ MakeFakeStepControllerPluginConfig(int fixed_rps_value);
 envoy::config::core::v3::TypedExtensionConfig MakeFakeStepControllerPluginConfigWithValidationError(
     const absl::Status& artificial_validation_error);
 
+/**
+ * Creates a valid TypedExtensionConfig proto that activates a FakeStepController with a
+ * FakeInputVariableSetterConfig that returns an error from GetCurrentCommandLineOptions().
+ *
+ * @param fixed_rps_value Value for RPS to set in the FakeStepControllerConfig proto until the
+ * countdown reaches zero.
+ * @param artificial_input_setting_failure An error status.
```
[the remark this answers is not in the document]
eric846: Done

test/adaptive_load/fake_plugins/fake_step_controller/fake_step_controller_test.cc
eric846: Done

oschaaf (approved):
Thanks for resolving the const nit. LGTM for my part of the review.

PR #493, opened 2020-08-27
Adaptive Load library for calling Nighthawk Service

eric846:
A library that calls a Nighthawk Service gRPC stub with the given `CommandLineOptions`, translating all possible gRPC errors into `absl::StatusOr`.

Will need to be updated when Nighthawk Service starts returning more than one message over the stream.

Part 4 of splitting PR #483.


tools/check_format.sh
```diff
@@ -8,7 +8,7 @@ TO_CHECK="${2:-$PWD}"
 bazel run @envoy//tools:code_format/check_format.py -- \
   --skip_envoy_build_rule_check  --namespace_check Nighthawk \
   --build_fixer_check_excluded_paths=$(realpath ".") \
-  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,test_common,test \
+  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,grpcpp,test_common,test \
```
eric846, line 11: This `grpcpp` section is a simple way to prevent the script from rearranging the grpcpp, gtest, gmock headers into an order that is then considered invalid.

source/adaptive_load/nighthawk_service_client.cc
```diff
@@ -0,0 +1,45 @@
+#include "adaptive_load/nighthawk_service_client.h"
+
+#include "external/envoy/source/common/common/assert.h"
+
+namespace Nighthawk {
+
+absl::StatusOr<nighthawk::client::ExecutionResponse> PerformNighthawkBenchmark(
```
oschaaf, line 7: Looking at this again, I wonder if we could unify this somehow with `RemoteProcessImpl::run(OutputCollector& collector)` https://github.com/envoyproxy/nighthawk/blob/53fa6132757c1a4f40868c94ee058c4d4ffa44cb/source/client/remote_process_impl.cc#L23

They seem similar in some regards, but this has a cleaner interface, and behaves more nicely when it receives unexpected intermediate updates. Maybe it is viable to implement one in terms of the other? (probably out of scope here, but I thought it was worth calling out the similarity.

tools/check_format.sh
oschaaf: Oh, that might have been a hassle to figure out, sorry about that. I wonder if we can improve there, ideally the "fix" doesn't produce something that the "check" objects to :-(

oschaaf (approved):
One remark with respect to overlapping functionality with existing code; but this LGTM

source/adaptive_load/nighthawk_service_client.cc
eric846: We should definitely unify them. Currently they have slightly different inputs, but we can probably find a common subset.

eric846:
Moved this bare function into an interface so we can mock it when testing future layers. Hopefully these are the right places to put the files for internal libraries.

mum4k:
@dubious90 please review and merge once ready.

source/adaptive_load/nighthawk_service_client.cc
oschaaf: The `OutputCollector` has a method which I think would accept `nighthawk::client::ExecutionResponse::Output`. Hopefully a small tweak would suffice to make this happen. Filed #496 to track, resolving this comment here.

oschaaf (approved):

include/nighthawk/adaptive_load/nighthawk_service_client.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "api/client/options.pb.h"
+#include "api/client/service.grpc.pb.h"
+
+namespace Nighthawk {
+
+/**
+ * An interface for interacting with a Nighthawk Service gRPC stub.
+ */
+class NighthawkServiceClient {
```
dubious90, line 15: Curious what you think of this train of thought:

At first I was going to ask if you could rename this to be Adaptive Load specific, since NighthawkServiceClient is the kind of name that could be used by a future class in Nighthawk, for creating a generic client for the NighthawkService.

Then I looked at the code, and I don't think there's anything adaptive-load specific about this class or the impl

include/nighthawk/adaptive_load/nighthawk_service_client.h
```diff
@@ -0,0 +1,38 @@
...
+#include "external/envoy/source/common/common/statusor.h"
+#include "external/envoy/source/common/protobuf/protobuf.h"
+
+#include "api/client/options.pb.h"
+#include "api/client/service.grpc.pb.h"
+
+namespace Nighthawk {
+
+/**
+ * An interface for interacting with a Nighthawk Service gRPC stub.
```
dubious90, line 13: Can you add a line here indicating whether or not this class is thread-safe?

include/nighthawk/adaptive_load/nighthawk_service_client.h
```diff
@@ -0,0 +1,38 @@
...
+   * StepController, without the duration set.
+   * @param duration Duration to insert into the benchmark request.
+   *
+   * @return StatusOr<ExecutionResponse> If we reached the Nighthawk Service, this is the raw
+   * ExecutionResponse proto, containing the benchmark data or possibly an error message from
+   * Nighthawk Service; if we had trouble communicating with the Nighthawk Service, we return an
+   * error status.
+   */
+  virtual absl::StatusOr<nighthawk::client::ExecutionResponse> PerformNighthawkBenchmark(
+      nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
```
dubious90, line 33: Not a request for change, but I wanted to hear your rationale: why have this as an argument on the function, rather than an argument to the class constructor?

It's interesting to me that neither this nor the impl version seem to hold any state at all, which leads to a question of why they need to be classes at all.

source/adaptive_load/nighthawk_service_client_impl.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include "external/envoy/source/common/common/assert.h"
+
+namespace Nighthawk {
+
+absl::StatusOr<nighthawk::client::ExecutionResponse>
+NighthawkServiceClientImpl::PerformNighthawkBenchmark(
+    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
+    const nighthawk::client::CommandLineOptions& command_line_options,
+    const Envoy::Protobuf::Duration& duration) {
```
dubious90, line 11: This seems like an odd api to me. Accepting a const command_line_options reference and then immediately copying it into a new object feels counterproductive.

There also doesn't seem to me to be a good reason to have duration as its own parameter as opposed to set on the command_line_options here, since the first and only thing you do with duration is set it in options.  Could you accept a const reference to command line options, with the duration properly included, and then use that in ExecutionRequest?

eric846: This API has always bothered me a bit. It made more sense when it was a private helper and I was trying to save lines at the call sites. Changing this as you suggest.

include/nighthawk/adaptive_load/nighthawk_service_client.h
eric846: This used to be a bare function because it had no need to track state across calls. But in order to mock out this functionality for testing, with gtest it's much simpler to make a library export an interface rather than plain functions.

Leaving it stateless seems simpler than storing the stub and taking on responsibility for its lifetime. The intent here is definitely not to introduce a new layer into the system architecture, only to wrap a stateless helper function in a class for unit test purposes.

include/nighthawk/adaptive_load/nighthawk_service_client.h
eric846: You're right, moved it to `common/`.

include/nighthawk/adaptive_load/nighthawk_service_client.h
eric846: Added a comment to the impl.

tools/check_format.sh
eric846: This is definitely a quick fix. It seems to pass all the CI checks. I don't know how it's arriving at the order or how it's checking it. If we figure out how it works, this would no longer be necessary.

oschaaf: I think it's the right tweak to address this. But I do find it's disappointing that check and fix don't line up when the order isn't explicitly specified. I'll file an issue to see if Envoy has progressed here, and if there are changes worth including in our version of this.

include/nighthawk/adaptive_load/nighthawk_service_client.h
dubious90: Makes sense.  Thanks for explaining.

source/common/nighthawk_service_client_impl.cc
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  bool got_response = false;
+  while (stream->Read(&response)) {
+    RELEASE_ASSERT(!got_response,
+                   "Nighthawk Service has started responding with more than one message.");
+    got_response = true;
+  }
+  if (!got_response) {
+    return absl::UnknownError("Nighthawk Service did not send a gRPC response.");
```
dubious90, line 33: This one doesn't seem like an Unknown Error.  Can we say this is an internal error?  As currently, this is something that shouldn't happen and is entirely internal to the codebase

test/common/nighthawk_service_client_test.cc
```diff
@@ -0,0 +1,154 @@
...
+using ::testing::HasSubstr;
+using ::testing::Return;
+using ::testing::SaveArg;
+using ::testing::SetArgPointee;
+
+TEST(PerformNighthawkBenchmark, UsesSpecifiedCommandLineOptions) {
+  const int kExpectedRps = 456;
+  ExecutionRequest request;
+  nighthawk::client::MockNighthawkServiceStub mock_nighthawk_service_stub;
+  EXPECT_CALL(mock_nighthawk_service_stub, ExecutionStreamRaw)
```
dubious90, line 33: Two small asks, but on each of these tests.

1. Can you add a linebreak between this EXPECT_CALL and the CommandLineOptions below (or in other tests, whatever is following it)?  Just helps the eyes parse this a little as one block.

2. Can we comment at the top of the EXPECT_CALL statement roughly what this block is mocking?  Or if that seems like a bad way to document this, something else that helps clarify for future readers what is happening.

dubious90 (changes requested):
Hi Eric, mostly looks good.  Just a couple comments.

source/common/nighthawk_service_client_impl.cc
eric846: After looking at the docs, I found there is Unavailable, meaning you to retry the request as-is. This might apply if the original Write fails since the Nighthawk Service might just have been down. But if you managed to Write but then got an error from WritesDone, Read, or Finish, those could all be Internal since they shouldn't happen. Does that make sense?

test/common/nighthawk_service_client_test.cc
eric846: Done

oschaaf (approved):

dubious90 (approved):

Issue #290, opened 2020-01-17
Dependency Injection for Logging


dubious90:
Our systems that we will be running nighthawk in have their own logging framework that does not seem to inherently work with the envoy logging that nighthawk uses.  It would be great if it were possible to have a dependency injection or binary configuration of some sort where we could configure what happens with the Nighthawk logging messages.

oschaaf:
By chance I just came across some possibly relevant enhancements in Envoy, there seem to be more sink types now. 

For example see `LogRecordingSink`:
https://github.com/envoyproxy/envoy/blob/d25fb83ddb85d2de376df3fc91e15e6a3df02091/test/test_common/logging.cc

(This doesn't seem to cover any sort of extension model for this, but this might offer a better alternative to the file log scraping going on today in some of the tests)

PR #494, opened 2020-08-27
Allow ENVOY_LOG to log messages in user-specified logging level and format

jiajunye:
This is requested by https://github.com/envoyproxy/nighthawk/issues/290.

I added another constructor in ServiceImpl to accept an Envoy Logger Context pointer which contains user-specified log format and log level.


oschaaf (approved):
Nice; it is a shame we don't have an easy means to unit test this :-(
LGTM

oschaaf:
The current change LGTM and makes sense, but I'm not sure I can fully match it to the PR description?

source/client/service_impl.h
```diff
@@ -35,9 +35,15 @@ class ServiceImpl final : public nighthawk::client::NighthawkService::Service,
   /**
    * Constructs a new ServiceImpl instance
    */
-  ServiceImpl() : process_wide_(std::make_shared<Envoy::ProcessWide>()) {
-    logging_context_ = std::make_unique<Envoy::Logger::Context>(
-        spdlog::level::from_str("info"), "[%T.%f][%t][%L] %v", log_lock_, false);
+  ServiceImpl(std::unique_ptr<Envoy::OptionsImpl> options = nullptr)
```
oschaaf, line 38: Wait .. one ask, does this then scan/parse the CLI for arguments that match Envoy's CLI in nighthawk_service? If so, wouldn't it be better to replicate these options in Nighthawk's own?

oschaaf (changes requested):
Sorry -- one question, I think I need a little more context to understand this change.

jiajunye:
> The current change LGTM and makes sense, but I'm not sure I can fully match it to the PR description?

Sorry for the confusion Otto. My previous PR title "Allow ENVOY_LOG to log messages in user specified log delegate" actually contains two parts: 1) in users' own logging framework, they need add some code to use their own log delegate. 2) in order to correctly display the messages in users' own log delegate, in the open source, we should enable the ability for users to choose the log level and format. This PR is just responsible for the part 2).

I've updated the title and description. Please let me know if this look better to you.

oschaaf:
@jiajunye thanks for updating and sharing context, that is helpful, and now it makes sense to me.

source/client/service_impl.h
```diff
@@ -17,7 +17,6 @@
 #include "external/envoy/source/common/common/thread.h"
 #include "external/envoy/source/common/event/real_time_system.h"
 #include "external/envoy/source/exe/process_wide.h"
-
```
oschaaf, line 20: Is this PR updated to the latest on master? I’m surprised the formatting check doesn’t object to this?
Regardless this change doesn’t seem nessecary?

source/client/service_impl.h
```diff
@@ -39,6 +38,12 @@ class ServiceImpl final : public nighthawk::client::NighthawkService::Service,
     logging_context_ = std::make_unique<Envoy::Logger::Context>(
         spdlog::level::from_str("info"), "[%T.%f][%t][%L] %v", log_lock_, false);
   }
+
+  ServiceImpl(std::unique_ptr<Envoy::Logger::Context> logging_context)
```
oschaaf, line 42: Optional: using a && arg here might save a copy?

oschaaf (commented):
A nit, a suggestion, and a question. Of it's not merges to recent master, could you do that? Otherwise LGTM

source/client/service_impl.h
jiajunye: What's the benefit to choose && comparing to &?

source/client/service_impl.h
jiajunye: Agreed. I will leave it unchanged.

jiajunye: PR is updated to the latest on master.

source/client/service_impl.h
oschaaf: Well, my take on it would be:
- With `&param` the smart pointer would be passed by reference, but ownership may or may not be taken by the constructor of `ServiceImpl`. The caller should make no assumptions about that and live with both scenarios after the call. 
- With a `const & std:unique_ptr<...>` parameter it would be clear that no ownership will be taken, but it might be better to unwrap and just pass a reference to the contained object itself as that avoids having to check for nullptr -- or at least delegates that responsibility to the caller.
- With `&&param` it is clear to consuming code that it will no longer own whatever it passes in to such a parameter (irrespective of wether there is a smart pointer involved). And `&&` saves a copy, which isn't a deciding factor in the code line we're talking about here (as it's not frequently called) -- but minding that might be a good habit.

Having said that, it may be good to add a test to cover this constructor, which will then probably also give our friend clang-tidy an opportunity to have a say about it when it inspects how this is going to be used :-) Next to adding actual coverage, which won't hurt either.

jiajunye: Thanks so much Otto! This helps a lot!

test/service_test.cc
```diff
@@ -111,6 +113,43 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
   int grpc_server_port_{0};
 };
 
+class ServiceTestWithParameterizedConstructor : public ServiceTest {
+public:
+  void SetUp() override {
+    Envoy::Thread::MutexBasicLockable log_lock_;
```
oschaaf, line 119: Might want to make this a member instead of local variable to ensure the proper lifetime,
as the Logging::Context constructor takes a ref?:

```
  Context(spdlog::level::level_enum log_level, const std::string& log_format,
          Thread::BasicLockable& lock, bool should_escape, bool enable_fine_grain_logging = false);
```

jiajunye: Agreed. I'm updating this PR to retest, since my local test can succeed somehow. Thanks Otto!

jiajunye:
@oschaaf Is this integration test flaky? I had couple failed builds but one successful build.

oschaaf:
> @oschaaf Is this integration test flaky? I had couple failed builds but one successful build.
I'm not sure if this is a flake, it is trending just below the threshold. Did you check the coverage report? Maybe there's some code that can be easily covered in a unit test, if not we need to lower the threshold. (the ratio may also change because of code churn).

jiajunye:
> > @oschaaf Is this integration test flaky? I had couple failed builds but one successful build.
> 
> I'm not sure if this is a flake, it is trending just below the threshold. Did you check the coverage report? Maybe there's some code that can be easily covered in a unit test, if not we need to lower the threshold. (the ratio may also change because of code churn).

oschaaf (dismissed):

oschaaf:
(looks great from my perspective; thanks for addressing the feedback)

jiajunye:
> (looks great from my perspective; thanks for addressing the feedback)

Thanks for reviewing the code. I really appreciate your help! Let's wait for dubious90's response on the threshold.

test/service_test.cc
```diff
@@ -111,6 +113,44 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+    setBasicRequestOptions();
+  }
+
+  Envoy::Thread::MutexBasicLockable log_lock_;
+};
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, ServiceTestWithParameterizedConstructor,
+                         ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()),
+                         Envoy::TestUtility::ipTestParamsToString);
+
```
dubious90, line 139: Recognize that this is copy-paste of initial code, but I'm going to review the new code to match better style guide conventions.

First: this comment should be superfluous with a good test name.. We can remove that, and then change the test name.

Test names should ideally read like a sentence explaining what the test case is.  I'll defer to you on what makes sense for a name but typically it's something like:
ConstructorWithLoggingParameterCanRespondToRequests

test/service_test.cc
```diff
@@ -111,6 +113,44 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+// Test single NH run with parameterized constructor
+TEST_P(ServiceTestWithParameterizedConstructor, BasicWithUserSpecifiedLoggingContext) {
+  auto r = stub_->ExecutionStream(&context_);
```
dubious90, line 142: Two things here:
1. We try to avoid using auto for readability reasons, unless it is incredibly obvious what the returning type is (such as when the constructor already indicates the type or we're iterating through an array where the type is immediately present there).  See the "Decision" at https://google.github.io/styleguide/cppguide.html#Type_deduction

2. We also try to avoid abbreviations in general.  If this is a request, then we'd usually call the variable request instead of r.  See https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

test/service_test.cc
```diff
@@ -111,6 +113,44 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+  r->Write(request_, {});
+  r->WritesDone();
+  EXPECT_TRUE(r->Read(&response_));
+  ASSERT_TRUE(response_.has_error_detail());
+  EXPECT_THAT(response_.error_detail().message(), HasSubstr(std::string("Unknown failure")));
+  EXPECT_TRUE(response_.has_output());
+  EXPECT_GE(response_.output().results(0).counters().size(), 8);
+  auto status = r->Finish();
```
dubious90, line 150: auto critique applies here too, especially since there are two different statuses - rpc status and absl status, so it's important to distinguish which one is here.

dubious90 (changes requested):
Hi Jason, mostly looks good with a couple style comments.  Please reach out if you have any questions/concerns.

test/service_test.cc
```diff
@@ -111,6 +113,44 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+
+    builder.AddListeningPort(fmt::format("{}:0", loopback_address_),
+                             grpc::InsecureServerCredentials(), &grpc_server_port_);
+    builder.RegisterService(service_.get());
+    server_ = builder.BuildAndStart();
+    setupGrpcClient();
+    setBasicRequestOptions();
+  }
+
+  Envoy::Thread::MutexBasicLockable log_lock_;
```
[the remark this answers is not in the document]
jiajunye: I agree that this should be private. The reason it needs to be a member variable is that the logging_context constructor takes log_lock_ by ref.

dubious90:
> You're right. We split the integration tests and unit tests, and while the unit test covers the new constructor, the integration tests do not. I'm think that lowering the threshold is what we should do in this case, if @dubious90 agrees. That threshold can be tweaked in `ci/do_ci.sh`.

I agree with Otto here.  I think I might have further questions about how the coverage works now, but that is separate from this PR.

test/service_test.cc
jiajunye: It looks like we typically match test suite name to the class name. ConstructorWithLoggingParameterCanRespondToRequests sounds like a good test name in TEST_P:
http://google3/third_party/googletest/googletest/include/gtest/gtest-param-test.h?l=416&rcl=317696457

jiajunye: Oh, I guess you mean the line 141. Now I get it.

test/service_test.cc
```diff
@@ -130,24 +130,27 @@ class ServiceTestWithParameterizedConstructor : public ServiceTest {
...
-TEST_P(ServiceTestWithParameterizedConstructor, BasicWithUserSpecifiedLoggingContext) {
-  auto r = stub_->ExecutionStream(&context_);
-  r->Write(request_, {});
-  r->WritesDone();
-  EXPECT_TRUE(r->Read(&response_));
+TEST_P(ServiceTestWithParameterizedConstructor,
+       ConstructorWithLoggingContextParameterCanRespondToRequests) {
+  std::unique_ptr<::grpc::ClientReaderWriter<::nighthawk::client::ExecutionRequest,
+                                             ::nighthawk::client::ExecutionResponse>>
+      request = stub_->ExecutionStream(&context_);
```
dubious90, line 145: Ah, I actually gave you a bad name for this. This isn't actually a request.  Sorry about that!  Easier to tell what it is now that it's not an `auto`.

Whatever makes sense to you as a name here.  reader_writer is an option.  Or if you don't like that, something that describes this accurately, like stream_client or stream_handler.  Whatever you feel good about.

test/service_test.cc
```diff
@@ -111,6 +113,47 @@ class ServiceTest : public TestWithParam<Envoy::Network::Address::IpVersion> {
...
+  Envoy::Thread::MutexBasicLockable log_lock_;
+};
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, ServiceTestWithParameterizedConstructor,
+                         ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()),
+                         Envoy::TestUtility::ipTestParamsToString);
+
+TEST_P(ServiceTestWithParameterizedConstructor,
+       ConstructorWithLoggingContextParameterCanRespondToRequests) {
+  std::unique_ptr<::grpc::ClientReaderWriter<::nighthawk::client::ExecutionRequest,
```
dubious90, line 143: Note that we are actually within namespace ::nighthawk::client, so you can just say ExecutionRequest and ExecutionResponse with no namespace declarations.  Should make this more concise.

dubious90 (commented):
Probably a last round of requests.  Thanks for your work!

test/service_test.cc
jiajunye: I would go with stream. Thanks for the advice!

test/service_test.cc
jiajunye: It actually complains about _Unknown type name 'ExecutionRequest'; did you mean 'nighthawk::client::ExecutionRequest'?_ after removing ::nighthawk::client. Maybe because I added an empty namespace wrapper?

jiajunye: Talked in person with @dubious90. The error was caused by the namespace discrepancy between the code file and proto buffer. In code, our namespaces start with uppercase, while in proto buffer they start with lowercase.

dubious90 (approved):
LGTM.  Thanks for responding to all these changes!

jiajunye:
@dubious90 Do you mind merging this PR please? Thanks!

PR #495, opened 2020-08-27
Adaptive Load metrics evaluator library

eric846:
A library that combines the latest Nighthawk Service response, metric and threshold spec configuration, and results from MetricsPlugins to produce metric scores.

Part 5 of splitting PR #483.


mum4k:
@wjuan-AFK please review and assign to @dubious90 once done.

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -0,0 +1,399 @@
+#include "envoy/config/core/v3/base.pb.h"
+#include "envoy/registry/registry.h"
+
+#include "adaptive_load/metrics_evaluator_impl.h"
+
+// #include "nighthawk/adaptive_load/adaptive_load_controller.h"
```
oschaaf, line 6: nit: accidentally left in comments?

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,87 @@
...
+   * metrics.
+   *
+   * @param spec The adaptive load session spec.
+   * @param metric_specs A list to store extracted MetricSpecs in order of definition.
+   * @param threshold_spec_from_metric_spec A map to store each MetricSpec and its threshold if it
+   * had one, or nullptr if it was an informational metric.
+   */
+  virtual void
+  ExtractMetricSpecs(const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec,
+                     std::vector<const nighthawk::adaptive_load::MetricSpec*>& metric_specs,
```
oschaaf, line 60: I wonder if instead of having two output args, a single one consisting of a vector of `std::pair<>`s with a mandatory metric + optional threshold could simplify this interface?

oschaaf (changes requested):
generally looks good to me, Leaving a nit about accidental comments left in here, plus a question

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -0,0 +1,399 @@
...
+      /*min_ns=*/400,
+      /*mean_ns=*/500,
+      /*max_ns=*/600,
+      /*pstdev_ns=*/11,
+  });
+  *response.mutable_output() = output;
+  return response;
+}
+
+TEST(EvaluateMetric, SetsMetricId) {
```
wjuan-AFK, line 94: I think we prefer the ConditionAWillResultinB format for naming these.

test/adaptive_load/metrics_evaluator_test.cc
eric846: Cleaned up these includes.

include/nighthawk/adaptive_load/metrics_evaluator.h
eric846: Done

test/adaptive_load/metrics_evaluator_test.cc
eric846: Added some more detail throughout.

include/nighthawk/adaptive_load/metrics_evaluator.h
eric846: The reason for the two separate structures is that I couldn't find a standard map that preserves the order of insertion.

eric846: Made both data structures conts so there is no way for them to get out of sync.

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -0,0 +1,122 @@
...
+            absl::flat_hash_map<const MetricSpec*, const ThresholdSpec*>>
+      metric_spec_list_map = ExtractMetricSpecs(spec);
+  // MetricSpecs in original order of definition.
+  std::vector<const MetricSpec*> metric_specs = metric_spec_list_map.first;
+  // Pointer to the corresponding ThresholdSpec, or nullptr for informational metrics.
+  absl::flat_hash_map<const MetricSpec*, const ThresholdSpec*> threshold_spec_from_metric_spec =
+      metric_spec_list_map.second;
+
+  std::vector<std::string> errors;
+  for (const MetricSpec* metric_spec : metric_specs) {
```
oschaaf, line 105: Reading this, I'm still wondering; it seems to me that having `ExtractMetricSpecs` return a `vector<std::pair<MetricSpec*, ThresholdSpec*>>` could simplify that interface as well as the code here; wouldn't that avoid the map lookups? The map usage would be in implementation detail in ExtractMetricSpecs, and not propagate outside of it. Of course, that might thwart re-use for other purposes perhaps.

oschaaf (commented):
looks great to me, one last question regarding the return type of ExtractMetricSpecs

source/adaptive_load/metrics_evaluator_impl.cc
eric846: Oh that's a good idea, it preserves the order and is simpler. I definitely don't need the map. If someone did need a map later, I think they could pass the vector of pairs into a map constructor.

eric846: Updated now to return a vector of pairs.

oschaaf (approved):
LGTM

wjuan-AFK:
LGTM

wjuan-AFK (approved):

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -0,0 +1,114 @@
...
+  return spec_threshold_pairs;
+}
+
+absl::StatusOr<nighthawk::adaptive_load::BenchmarkResult>
+MetricsEvaluatorImpl::AnalyzeNighthawkBenchmark(
+    const nighthawk::client::ExecutionResponse& nighthawk_response,
+    const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec,
+    const absl::flat_hash_map<std::string, MetricsPluginPtr>& name_to_custom_metrics_plugin_map)
+    const {
+  if (nighthawk_response.error_detail().code() != static_cast<int>(absl::StatusCode::kOk)) {
```
dubious90, line 73: I'm sure this is tested so I'm probably wrong, but isn't error_detail().code() not guaranteed to exist here?  What happens if it doesn't?

If there's any concern of that creating weird behavior, we could wrap it with a has_error_detail / has_code

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -0,0 +1,368 @@
...
+}
+
+TEST(AnalyzeNighthawkBenchmark, PropagatesErrorFromFailedMetricEvaluation) {
+  nighthawk::adaptive_load::AdaptiveLoadSessionSpec spec;
+
+  const std::string kMetricName = "bad-metric";
+  const std::string kExpectedStatusMessage = "artificial metric error";
+
+  FakeMetricsPluginConfig metrics_plugin_config;
+  FakeMetricsPluginConfig::FakeMetric* fake_metric =
```
dubious90, line 317: Optional:  It seems like we're doing this same pattern a lot, and it takes up a lot of code (lack of textproto parsing).  Any way to create helper functions to reduce some of the lines of boilerplate?

dubious90 (changes requested):

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+ * An interface with utilities for translating between metrics definitions, thresholds, scores,
+ * MetricsPlugins, and Nighthawk Service results.
+ */
+class MetricsEvaluator {
+public:
+  virtual ~MetricsEvaluator() = default;
+
+  /**
+   * Given a MetricSpec, obtains a single metric value from the MetricPlugin and optionally scores
+   * it according to a threshold and scoring function.
```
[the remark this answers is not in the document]
eric846: Done

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+#include "api/client/service.pb.h"
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/status/status.h"
+#include "absl/strings/str_join.h"
+
+namespace Nighthawk {
+
+/**
+ * An interface with utilities for translating between metrics definitions, thresholds, scores,
```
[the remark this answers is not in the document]
eric846: I added a better explanation of what this library does. Only one of the methods is expected to be called directly.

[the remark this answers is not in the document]
eric846: This should be clearer now.

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+class MetricsEvaluator {
+public:
+  virtual ~MetricsEvaluator() = default;
+
+  /**
+   * Given a MetricSpec, obtains a single metric value from the MetricPlugin and optionally scores
+   * it according to a threshold and scoring function.
+   *
+   * @param metric_spec The metric spec identifying the metric by name and plugin name.
+   * @param metrics_plugin An already activated MetricsPlugin used by the metric_spec.
```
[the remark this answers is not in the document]
eric846: Dropped that wording, it should be clearer now.

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+   * @param threshold_spec A proto describing the threshold and scoring function. Nullptr if the
+   * metric is informational only.
```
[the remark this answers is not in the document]
eric846: We will add a user guide later, but this is mentioned in the proto comment: https://github.com/envoyproxy/nighthawk/blob/365b97a8061f4ede707d1ea398200b3b4cdd0348/api/adaptive_load/adaptive_load.proto#L32

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+   * @return Vector of pairs of pointers to MetricSpec and ThresholdSpec within |spec|. For
+   * informational metrics, the ThresholdSpec pointer is nullptr.
+   */
+  virtual const std::vector<std::pair<const nighthawk::adaptive_load::MetricSpec*,
+                                      const nighthawk::adaptive_load::ThresholdSpec*>>
+  ExtractMetricSpecs(const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec) const PURE;
+
+  /**
+   * Analyzes a Nighthawk Service benchmark against configured MetricThresholds. Queries
+   * outside MetricsPlugins for current metric values, and/or uses "nighthawk.builtin" plugin to
```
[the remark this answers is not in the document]
eric846: Dropping this discussion of the builtin plugin. Mentioned the nighthawk.builtin in some of the parameter descriptions.

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+   * extract stats and counters from the latest Nighthawk Service output. The Nighthawk benchmark is
+   * assumed to have finished recently so values from MetricsPlugins will be relevant.
```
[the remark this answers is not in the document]
eric846: Done.

include/nighthawk/adaptive_load/metrics_evaluator.h
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param nighthawk_response Proto returned from Nighthawk Service describing the latest single
+   * benchmark session.
+   * @param spec Top-level proto defining the adaptive load session.
+   * @param name_to_custom_metrics_plugin_map Common map from plugin names to MetricsPlugins, loaded
```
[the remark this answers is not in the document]
eric846: I see, dropped the information about how the caller operates and and added a description of the true constraints.

source/adaptive_load/metrics_evaluator_impl.cc
eric846: We will get the default instance of the `error_detail` field if it's not set, and that should have `code` 0.

test/adaptive_load/metrics_evaluator_test.cc
eric846: There are two occurrences here, with one having a value and one having an error, so we wouldn't save anything in this test by adding helpers that would only get called once each. The reason we don't have helpers in `FakeMetricsPlugin` is that it seemed like too much work after `FakeMetricsPlugin` became more flexible. It would be great if we could get textproto parsing working.  Opened #520.

dubious90 (approved):

PR #499, opened 2020-08-29
Fix (enable) integration test coverage in CI

oschaaf:
Whoops, I think I found why the integration test coverage task in not
showing in CircleCI. Next to defining it, it should be separately enabled too.


oschaaf:
Yep, I can see it showing now.

PR #500, opened 2020-08-31
Finalize emission and tracking of latencies in response headers

oschaaf:
Adds a client option and wires it through in TCLAP. Amends tests and code to work with that
instead of the hard coded value. Adds end to end test, and enables test-server extension build.

Follow up to https://github.com/envoyproxy/nighthawk/pull/477

Fixes https://github.com/envoyproxy/nighthawk/issues/360: with this the feature is ready to use.

-------

TODO: 
- [x] Land #477 and merge it in here 
- [x] Wire up TCLAP
- [x] Tests for the new option parsing/handling
- [x] Enable build of the new time-tracking extension into nighthawk_test_server
- [x] End-to-end tests to prove the new client-side feature can be used to track latencies delivered by response header emitted by the test-server's new feature. 
- [x] CLI & proto description / comments, regen docs. 
- [x] Replace Stopwatch shared_ptr with unique_ptr


dubious90:
@qqustc please review and assign to me once done.  Thanks!

qqustc:
Thanks Otto!
LGTM. Assign back to @dubious90 .

qqustc (approved):

README.md
```diff
@@ -80,6 +81,12 @@ format>
 
 Where:
 
+--response-header-with-latency-input <string>
```
dubious90, line 84: I think the meaning here is a little hard to parse from the name / comment combo.

How do you feel about response_latency_header_name?

README.md
```diff
@@ -80,6 +81,12 @@ format>
...
+Set an optional response header name, whose values will be tracked in
```
dubious90, line 85: For the description, can I suggest this edit to the first sentence:
Set an optional header name that will be returned in responses, whose values will be tracked in a latency histogram if set.

README.md
```diff
@@ -80,6 +81,12 @@ format>
...
+a latency histogram if set. Can be used in tandem with the test server
```
dubious90, line 86: For the second sentence, can we edit to something like:
"... if set.  e.g. Can be used in tandem with the test server's response option "emit_previous_request_delta_in_response_header" to record elapsed time between request arrivals."

(Adding in the word response option clarifies where the reader should look for the definition of that)

source/client/stream_decoder.h
```diff
@@ -46,7 +46,8 @@ class StreamDecoder : public Envoy::Http::ResponseDecoder,
                 Statistic& response_body_sizes_statistic, Statistic& origin_latency_statistic,
                 HeaderMapPtr request_headers, bool measure_latencies, uint32_t request_body_size,
                 Envoy::Random::RandomGenerator& random_generator,
-                Envoy::Tracing::HttpTracerSharedPtr& http_tracer)
+                Envoy::Tracing::HttpTracerSharedPtr& http_tracer,
+                std::string response_header_with_latency_input)
```
dubious90, line 50: I'm a little surprised by your use of a std::string copy with std::move here, as opposed to having this be an absl::string_view (or const std::string&, whicever is more common within our code).

And then having doing response_header_with_latency_input_(response_header_with_latency_input), which I'm pretty sure just works.

This suggestion is more in line with the style I'm familiar with.  Is there a strong reason not to do that?

test/benchmark_http_client_test.cc
```diff
@@ -179,7 +179,7 @@ class BenchmarkClientHttpTest : public Test {
   void setupBenchmarkClient(const RequestGenerator& request_generator) {
     client_ = std::make_unique<Client::BenchmarkClientHttpImpl>(
         *api_, *dispatcher_, store_, statistic_, false, cluster_manager_, http_tracer_, "benchmark",
-        request_generator, true);
+        request_generator, true, "");
```
dubious90, line 182: This is a lot of arguments here.  Would it be okay if we added comments helping users track the corresponding parameter under use?  For the new one, it'd be
```/*response_header_with_latency_input=*/""```
And then, the two boolean parameters could also benefit from this.

File: test/integration/configurations/nighthawk_track_timings.yaml [hunk 6]
```diff
@@ -0,0 +1,34 @@
+admin:
```
dubious90, line 1: Can we add a comment at the top of this, explaining its purpose in just a sentence, and draw attention to the relevant/unique part of the configuration (which I think is the time-tracking filter)

test/integration/test_integration_basics.py
```diff
@@ -700,6 +700,24 @@ def test_cancellation_with_infinite_duration(http_test_server_fixture):
   asserts.assertCounterGreaterEqual(counters, "benchmark.http_2xx", 1)
 
 
+@pytest.mark.parametrize('server_config',
+                         ["nighthawk/test/integration/configurations/nighthawk_track_timings.yaml"])
+def test_http_h1_response_header_latency_tracking(http_test_server_fixture):
+  """Test emission and tracking of response header latencies.
+
+  Run the CLI configured to track latencies delivered by response header from the test-server
+  which is set up emit those. Ensure the expected histogram is observed.
```
dubious90, line 709: nit: set up to emit those

dubious90, line 709: nit: Can we instead say ensure the histogram receives the correct number of inputs or something?

test/integration/test_integration_basics.py
```diff
@@ -700,6 +700,24 @@ def test_cancellation_with_infinite_duration(http_test_server_fixture):
...
+  """
+  parsed_json, _ = http_test_server_fixture.runNighthawkClient([
+      http_test_server_fixture.getTestServerRootUri(), "--connections", "1", "--rps", "100",
+      "--duration", "100", "--termination-predicate", "benchmark.http_2xx:99",
+      "--response-header-with-latency-input", "x-origin-request-receipt-delta"
+  ])
+  global_histograms = http_test_server_fixture.getNighthawkGlobalHistogramsbyIdFromJson(parsed_json)
+  asserts.assertEqual(
+      int(global_histograms["benchmark_http_client.origin_latency_statistic"]["count"]), 99)
```
dubious90, line 718: Can we add a second integration test that works as a crash test of sorts, for adding in the header but receiving no responses that include that header?

Not sure of the exact correct assertions.  We would expect it to work, but for origin_latency_statistic to equal 0, I think.  Working could be defined as the responses_xxx counter increasing?

dubious90 (changes requested):
This feature looks cool, Otto.  Thanks for your efforts.

README.md
```diff
@@ -43,8 +43,9 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--stats-flush-interval <uint32_t>]
-[--stats-sinks <string>] ...
+bazel-bin/nighthawk_client  [--response-header-with-latency-input
```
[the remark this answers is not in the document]
oschaaf: Yeah, I agree.

source/client/stream_decoder.h
oschaaf: Good catch. I'm surprised too, I must have grown tired of the tried and proven path in this repo :-) Using absl::string_view now.

test/integration/test_integration_basics.py
oschaaf: Done; we now test using two server side configurations, one with the new time-tracking extension enabled, and one that doesn't have it set up. We check expectations based on which one we're testing. Let me know if that looks good to you. (68ed58af)

oschaaf:
@dubious90 comments addressed in 68ed58af.

oschaaf:
> @dubious90 comments addressed in [68ed58a](https://github.com/envoyproxy/nighthawk/commit/68ed58affe29eaddd11b5d2e5486a1d9fad746c7).

Also, merged master in here and resolved conflicts. Thanks for the review!

README.md
```diff
@@ -43,7 +43,8 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--stats-flush-interval <uint32_t>]
+bazel-bin/nighthawk_client  [--response-latency-header-name <string>]
```
dubious90, line 46: Sorry for extra churn here.  Realized I gave you a misleading name (implies this would be response latency, which, while possible to do with this, is not the only usecase).

How do you feel about `--latency-response-header-name` (just inverting those two words).

test/integration/test_integration_basics.py
```diff
@@ -700,6 +700,30 @@ def test_cancellation_with_infinite_duration(http_test_server_fixture):
...
+  Ensure that the origin_latency_statistic histogram receives the correct number of inputs.
+  """
+  parsed_json, _ = http_test_server_fixture.runNighthawkClient([
+      http_test_server_fixture.getTestServerRootUri(), "--connections", "1", "--rps", "100",
+      "--duration", "100", "--termination-predicate", "benchmark.http_2xx:99",
+      "--response-latency-header-name", "x-origin-request-receipt-delta"
+  ])
+  global_histograms = http_test_server_fixture.getNighthawkGlobalHistogramsbyIdFromJson(parsed_json)
+  asserts.assertEqual(int(global_histograms["benchmark_http_client.latency_2xx"]["count"]), 100)
+  # Verify behavior is correct both with and without the timing filter enabled.
```
dubious90, line 720: This is a very helpful comment here.  Thanks

dubious90 (commented):
LGTM modulo the one naming comment. Thanks!

README.md
oschaaf: Sounds good, done in 594bbbec

dubious90 (approved):

PR #503, opened 2020-09-02
Fix uri parsing bug & add some unit test coverage for client.cc

oschaaf:
- Avoid std::stoi exception during uri parsing
- Add unit tests for client.cc to improve coverage

### DIFF
--- a/test/client_test.cc
+++ b/test/client_test.cc
@@ -80,5 +80,19 @@ TEST_F(ClientTest, BadRun) {
   EXPECT_FALSE(program.run());
 }
 
+TEST_F(ClientTest, BadRemoteRun) {
+  Main program(Nighthawk::Client::TestUtility::createOptionsImpl(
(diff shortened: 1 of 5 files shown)


test/client_test.cc
```diff
@@ -80,5 +80,19 @@ TEST_F(ClientTest, BadRun) {
   EXPECT_FALSE(program.run());
 }
 
+TEST_F(ClientTest, BadRemoteRun) {
```
dubious90, line 83: It's not very clear to me what is actually being tested here.  If we're always expecting false from these tests, are we actually verifying anything?

oschaaf: I agree it's  not much, but we verify it isn't crashing and that the return value indicates failure. Functional testing resides in the integration tests. That verifies a little more. (Also see the todo above in this file). Below, failure is expected too, because there's no actual gRPC service here for the test to talk to.

dubious90: We've decided offline to keep these tests for now, and discuss later on what our philosophy is around tests/coverage.

PR #506, opened 2020-09-03
Set explicit --jobs value in ci/do_ci.sh only when running on CircleCI

eric846:
For local `ci/do_ci.sh` runs, this change allows the explicit `--jobs` setting from `bazelrc` files to take effect, or falls back on the default `--jobs=auto`.

This change stops passing ` --jobs $NUM_CPUS` for non-CircleCI Bazel runs, but any other uses of `NUM_CPUS` are assumed to be critical and are all left unchanged. For example, `NUM_CPUS` must be in the environment for the call out to `ci/run_clang_tidy.sh` to work. For clang-tidy and anything else that doesn't use a remote Bazel build, we should always allow `NUM_CPUS` to take effect, both locally and on CircleCI.

Motivation for this change:
When using a `user.bazelrc` that uses Remote Build Execution with a high `--jobs` value, we discovered that `NUM_CPUS` in `ci/do_ci.sh` is artificially limiting the parallelism of the remote builds based on the local CPU count.

(`--jobs` is only useful for limiting the amount of parallelism, not increasing it: https://docs.bazel.build/versions/master/user-manual.html#flag--jobs.)

Also fixing the printing of the help message when the user forgot the arg.


oschaaf:
this change makes sense, the limiting we want to perform is specific to circleci as its VMs provide more vCPUs then their available memory allows us to use while building, inducing OOM if parallelism is not tuned down.
I don't grasp the clang-tidy flake that popped up all of a sudden.
/retest

oschaaf:
The clang-tidy breakage seems real; it depends on `NUM_CPUS` being set: https://github.com/envoyproxy/nighthawk/blob/master/ci/run_clang_tidy.sh#L60

eric846:
> The clang-tidy breakage seems real; it depends on `NUM_CPUS` being set: https://github.com/envoyproxy/nighthawk/blob/master/ci/run_clang_tidy.sh#L60

ohh I see, will try to handle that...

dubious90:
Re: https://github.com/envoyproxy/nighthawk/pull/506#issuecomment-686596956 (which I can't reply to for some reason), please also update the description to match this new knowledge.  If I understand correctly, we now have an assumption listed there that is found to be false.

eric846:
> Re: [#506 (comment)](https://github.com/envoyproxy/nighthawk/pull/506#issuecomment-686596956) (which I can't reply to for some reason), please also update the description to match this new knowledge. If I understand correctly, we now have an assumption listed there that is found to be false.

Done.

oschaaf (approved):

dubious90 (approved):
LGTM

PR #508, opened 2020-09-03
Adaptive load session spec proto helpers

eric846:
- `SetSessionSpecDefaults()`: Returns a copy of the `AdaptiveLoadSessionSpec` with default values added.
- `CheckSessionSpec()`: Checks an `AdaptiveLoadSessionSpec` for illegal values, invalid plugin references, and invalid plugin configs.

Part 6 of splitting PR #483.


dubious90:
@wjuan-AFK could please review and assign back to me once complete?

source/adaptive_load/session_spec_proto_helper_impl.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "api/adaptive_load/adaptive_load.pb.h"
+#include "api/adaptive_load/metric_spec.pb.h"
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/strings/str_join.h"
+#include "absl/strings/string_view.h"
+#include "adaptive_load/metrics_plugin_impl.h"
+#include "adaptive_load/plugin_loader.h"
+
+// #include "api/adaptive_load/step_controller_impl.pb.h"
```
oschaaf, line 15: nit: accidental comment?

oschaaf (approved):
modulo the (probably) accidental comment, LGTM.

source/adaptive_load/session_spec_proto_helper_impl.cc
```diff
@@ -0,0 +1,119 @@
...
+namespace Nighthawk {
+
+nighthawk::adaptive_load::AdaptiveLoadSessionSpec
+AdaptiveLoadSessionSpecProtoHelperImpl::SetSessionSpecDefaults(
+    nighthawk::adaptive_load::AdaptiveLoadSessionSpec spec) {
+  if (!spec.nighthawk_traffic_template().has_open_loop()) {
+    spec.mutable_nighthawk_traffic_template()->mutable_open_loop()->set_value(true);
+  }
+  if (!spec.has_measuring_period()) {
+    spec.mutable_measuring_period()->set_seconds(10);
```
wjuan-AFK, line 24: Should we have these as consts at the top so that they're easier to modify, rather than being numbers inside the method itself? E.g. set_seconds(DEFAULT_MEASURING_PERIOD)

dubious90: Just to chime in, I think this is a matter of opinion, but in the past, I think I've seen mum4k argue against constants for this purpose, because they're only used once, and potentially create misdirection when trying to figure out what the defaults are.

This function is already properly abstracted to be a SetDefaults method, so it's relatively separated from the rest of the code, and easy to find how it works.

wjuan-AFK: Ok. LGTM.

dubious90 (approved):

PR #510, opened 2020-09-03
Build an image containing the benchmark infrastructure

abaptiste:
This script and Dockerfile creates the "benchmark-dev" container used by the fully dockerized flow here:

https://github.com/envoyproxy/nighthawk/tree/master/benchmarks#example-fully-dockerized-flow

I was not able to find a script that creates this image in the nighthawk repo.  The steps here were extracted from the `oschaaf/benchmark-dev:latest` container.

I am able to use this to run the steps in benchmarks doc above to execute tests.


ci/docker/Dockerfile-benchmark
```diff
@@ -0,0 +1,22 @@
...
+
+CMD ["./benchmarks" "--help"]
+ENV RUNFILES_DIR=/usr/local/bin/benchmarks/benchmarks.runfiles
+ENV ENVOY_PATH=envoy
+ENV ENVOY_IP_TEST_VERSIONS=v4only
+WORKDIR /usr/local/bin/benchmarks
+
+COPY benchmarks /usr/local/bin/benchmarks/
+
+RUN apk add docker openrc
```
oschaaf, line 11: should this point to a specific versions to pin them?

ci/docker/benchmark_build.sh
```diff
@@ -0,0 +1,28 @@
...
+# Builds a docker image benchmark-dev:latest containing the benchmark scripts
+
+# Stop on errors.
+set -e
+
+# NOTE: explicit no -x for verbose commands. Because this is run in CI, doing so may result in
+# publishing sensitive information into public CI logs if someone makes a change in a
+# consuming script that is off guard.
+
+DOCKER_NAME="benchmark"
```
oschaaf, line 12: I wonder, should we leave a trace of the source in this name? E.g. 'nighthawk-benchmark'?

ci/docker/Dockerfile-benchmark
```diff
@@ -0,0 +1,22 @@
...
+RUN rc-update add docker boot
+RUN echo "**** install Python ****" && \
+  apk add --no-cache python3 && \
+  if [ ! -e /usr/bin/python ]; then ln -sf python3 /usr/bin/python; fi && \
+  echo "**** install pip ****" && \
+  python3 -m ensurepip && \
+  rm -r /usr/lib/python*/ensurepip && \
+  pip3 install --no-cache --upgrade pip setuptools wheel && \
```
oschaaf, line 19: I'm also wondering if we should consider pinning versions here?

oschaaf (commented):
Thanks! Generally looks great to me, a few questions

ci/docker/benchmark_build.sh
abaptiste: Certainly can update this.

ci/docker/Dockerfile-benchmark
abaptiste: I have an idea of how we can do this.  It does make sense.  I'll test and update.

ci/docker/Dockerfile-benchmark
abaptiste: Yep.  Let me spend some more time on this and I will update.

oschaaf (commented):
LGTM, deferring to @dubious90 for final review.

Cross linking https://github.com/envoyproxy/envoy/issues/11266#issuecomment-687212684 for context, which in turn points the relevant design-doc

oschaaf (approved):

ci/docker/Dockerfile-nighthawk-benchmark
```diff
@@ -0,0 +1,24 @@
+FROM frolvlad/alpine-glibc
+
+CMD ["./benchmarks" "--help"]
```
dubious90, line 3: Forgive me, I've been out of docker for a while.  Why is it necessary for us to run --help at the start of this?

ci/docker/benchmark_build.sh
```diff
@@ -0,0 +1,28 @@
...
+
+[ -d "${TMP_DIR}"  ] && rm -rf "${TMP_DIR}"
+
+echo "Preparing docker build context in ${TMP_DIR}"
+cp -r "${WORKSPACE}/ci/docker/" "${TMP_DIR}/"
+cp -rL "${BAZEL_BIN}/benchmarks" "${TMP_DIR}"
+
+cd "${TMP_DIR}"
+echo "running docker build ... "
+docker build -f "${TMP_DIR}/Dockerfile-${DOCKER_NAME}" -t "${DOCKER_IMAGE_PREFIX}-dev:latest" .
```
dubious90, line 26: I could very well be missing your intent here, but would it make sense for us to parameterize some of these arguments?  The one that seems to have the most utility to me is `latest` here.  Could we default to latest but let others specify other docker tags?

The other parameter I'm wondering about is TMP_DIR.  I've found in the past that requiring certain paths sometimes results in people being unable to proceed due to permissions errors on a directory.  Do you think that's a possible concern here, justifying a parameter?  Or is that just needless complexity

dubious90 (commented):
Looks good.  I just have a couple questions.

ci/docker/Dockerfile-nighthawk-benchmark
abaptiste: I took this verbatim from the `docker history` of the original container.  I want to think that this is a quick way to ensure that all the python dependencies for `benchmarks` exist.  If not, the container build would fail immediately.

ci/docker/benchmark_build.sh
abaptiste: We can certainly parameterize the image tag.

I can also use `mktemp` here for TMP_DIR.  I don't envision that this container will be built repeatedly, but yes, having a static name for a temporary path could present an issue.

I will update both of these.

abaptiste:
I didn't add the sign-off to one of my commits, hence the git adventure.

dubious90 (dismissed):

dubious90:
> I didn't add the sign-off to one of my commits, hence the git adventure.

Changes look good, but it appears that this is still blocking the pull request.  Is further rebasing or anything required, or does this need to be refreshed somehow?

oschaaf:
Here's a (hopefully helpful) readup on fixing DCO: https://github.com/envoyproxy/envoy/blob/master/CONTRIBUTING.md#fixing-dco

For the future, the repo has [pre-push and pre-commit git hooks](https://github.com/envoyproxy/nighthawk/tree/master/support) that may assist with this. Nearly everyone forgets at some point :-)

abaptiste:
Thanks Otto for the link.   It's been on the back of my mind to set this up for this repo.  No time like the present!

oschaaf (approved):

oschaaf:
@dubious90 DCO has been fixed, any objections to merging this?

dubious90 (approved):

abaptiste:
Thank you for the merge!

PR #511, opened 2020-09-04
Avoid high frequency logging

### LINKED ISSUE #484: Avoiding high frequency logging
Sometimes there's error flow control in the hot path that deserves logging, but plain logging there
risks generating a large amount of log spam.
It would be great to have a low-barrier and easy-to-consume way to emit a single or bounded number of log lines in these cases.

oschaaf:
Throttle log emission in a few potentially high frequency paths.

Fixes #484


dubious90 (approved):
Looks great!

PR #516, opened 2020-09-08
Allow 0s as a value to the jitter-uniform option.

### LINKED ISSUE #513: Accept 0s as an argument to --jitter-uniform
In the options proto we enforce that the uniform jitter option can't be set to 0:
https://github.com/envoyproxy/nighthawk/blob/8667f7f7adf9c4fbed5872de5585e75e1c93e5b7/api/client/options.proto#L183

There are times when being able to do so is convenient, and this should be allowed but result in a no-op.
AFAICT this would just work when we tweak the validation to allow it.

Context: https://github.com/istio/tools/pull/1190#discussion_r484168703

oschaaf:
Fixes #513

### DIFF
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -273,7 +273,9 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
   *fortio_output.mutable_requestedduration() = output.options().duration();
   auto actual_duration = getAverageExecutionDuration(output);
   fortio_output.set_actualduration(actual_duration.count());
-  fortio_output.set_jitter(output.options().has_jitter_uniform());
+  fortio_output.set_jitter(output.options().has_jitter_uniform() &&
+                           (output.options().jitter_uniform().nanos() > 0 ||
+                            output.options().jitter_uniform().seconds() > 0));
   fortio_output.set_runtype("HTTP");
 
   // The stock Envoy h2 pool doesn't offer support for multiple connections here. So we must ignore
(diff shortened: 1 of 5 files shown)


source/client/output_formatter_impl.cc
```diff
@@ -273,7 +273,9 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
   *fortio_output.mutable_requestedduration() = output.options().duration();
   auto actual_duration = getAverageExecutionDuration(output);
   fortio_output.set_actualduration(actual_duration.count());
-  fortio_output.set_jitter(output.options().has_jitter_uniform());
+  fortio_output.set_jitter(output.options().has_jitter_uniform() &&
```
dubious90, line 276: There's something I'm missing here. As far as I can tell, the jitter in options is basically just being used as a boolean.  If it's there (and now, if it's greater than 0), then we use jitter.  But where does the int value actually come into play?

dubious90 (commented):
Looks good. Just one question I think because I'm confused as to how jitter actually works.

source/client/output_formatter_impl.cc
oschaaf: In Fortio jittering is a bool flag. On our side it's quite different, but we can emulate Fortio's version of it by using a certain configuration (10% of the qps from the top of my head).
So here we are more or less doing the best we can. Setting the right value here will make the Fortio UI display
wether jitter was set.

oschaaf: Oh, and what might be confusing here, is that on our side, next to it being set, it must also have a value.

dubious90 (approved):

PR #517, opened 2020-09-09
Add HttpFilterIntegrationTestBase

oschaaf:
Prelude to #512, which includes it and provides a means to
see the end goal.

- Extracts shared needs between tested extensions.
- Sanitized api, better naming.
- Doc comments.
- Support for testing POST request methods and entity bodies
  to check the alternative flow that will trigger in extensions
  when using that.


dubious90:
@eric846 please review and assign to me when done.

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,122 @@
...
+   * "{{response_body_size:1024}".
+   * @param setup_for_upstream_request Set to true iff the filter extension under test is expected
+   * to short-circuit and supply a response directly. For example because it couldn't parse the
+   * supplied request-level configuration. Otherwise this should be set to true, and a stock
+   * response will be yielded by the integration test server through an upstream request.
+   * @return Envoy::IntegrationStreamDecoderPtr Pointer to the integration stream decoder, which can
+   * be used to inspect the response.
+   */
+  Envoy::IntegrationStreamDecoderPtr getResponse(absl::string_view request_level_config,
+                                                 bool setup_for_upstream_request);
```
eric846, line 101: Could we reduce the number of public methods by half and eliminate the need to figure out the meaning of `true` or `false` in the implementation by creating a mode enum with values `kUpstream` and `kExtension`? Or if it's more convenient for the test authors to have the separate public methods, a private enum could still make the impl easier to read.

test/server/http_filter_integration_test_base.cc
```diff
@@ -0,0 +1,72 @@
...
+
+Envoy::IntegrationStreamDecoderPtr HttpFilterIntegrationTestBase::getResponseFromExtension(
+    const Envoy::Http::TestRequestHeaderMapImpl& request_headers) {
+  return getResponse(request_headers, false);
+}
+
+Envoy::IntegrationStreamDecoderPtr
+HttpFilterIntegrationTestBase::getResponse(absl::string_view request_level_config,
+                                           bool setup_for_upstream_request) {
+  const Envoy::Http::LowerCaseString key("x-nighthawk-test-server-config");
```
eric846, line 44: It would be good to mention in all doc comments that `request_level_config` is sent as the `x-nighthawk-test-server-config` header -- I didn't realize how it worked.

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,122 @@
...
+  void setup(const std::string& config);
+
+  /**
+   * @brief Fetch a response with the default request headers, and set up a fake upstream to supply
+   * the response.
+   *
+   * @return Envoy::IntegrationStreamDecoderPtr Pointer to the integration stream decoder, which can
+   * be used to inspect the response.
+   */
+  Envoy::IntegrationStreamDecoderPtr getResponseFromUpstream();
```
eric846, line 37: Are we missing `getResponseFromExtension()` with no args, or does that one not make sense?

eric846 (changes requested):

oschaaf:
Thanks for the review @eric846. Following some of your suggestions, I refactored this some more to further simplify in 
8d30820. I think it captures most of your comments modulo the one about default_request_headers_, which no longer makes sense as it is now private and used in a stateful manner. If you want, it's possible to take a peek at an example that consumes the state as it is now proposed here [over here](https://github.com/envoyproxy/nighthawk/blob/8b3db6c12e437934feca1eddf221a8ea28043470/test/server/http_time_tracking_filter_integration_test.cc).

eric846:
LGTM, thanks!

oschaaf:
@eric846 Sorry -- added a small new helper here in 7eeb7e9 which I turned out to need, PTAL.

test/server/http_filter_integration_test_base.cc
```diff
@@ -0,0 +1,56 @@
...
+
+Envoy::IntegrationStreamDecoderPtr
+HttpFilterIntegrationTestBase::getResponse(ResponseOrigin expected_origin) {
+  cleanupUpstreamAndDownstream();
+  codec_client_ = makeHttpConnection(lookupPort("http"));
+  Envoy::IntegrationStreamDecoderPtr response;
+  const bool is_post = request_headers_.Method()->value().getStringView() ==
+                       Envoy::Http::Headers::get().MethodValues.Post;
+  const uint64_t request_body_size = is_post ? 1024 : 0;
+  if (expected_origin == ResponseOrigin::UPSTREAM) {
```
dubious90, line 42: I'm having trouble understanding this function's full intent.  There are a lot of conditionals that seem to make the behavior very different.  Is this actually common behavior across filters?

Maybe some comments in this section about why things are happening would be helpful.

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+   */
+  HttpFilterIntegrationTestBase(Envoy::Network::Address::IpVersion ip_version);
+
+  /**
+   * We don't override SetUp(): tests using this fixture must call setup() instead.
+   * This is to avoid imposing the need to create a fixture per filter configuration.
+   *
+   * @param config configuration to pass to Envoy::HttpIntegrationTest::config_helper_.addFilter.
+   */
+  void setup(const std::string& config);
```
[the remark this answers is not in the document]
oschaaf: Done

test/server/http_filter_integration_test_base.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+HttpFilterIntegrationTestBase::HttpFilterIntegrationTestBase(
+    Envoy::Network::Address::IpVersion ip_version)
+    : HttpIntegrationTest(Envoy::Http::CodecClient::Type::HTTP1, ip_version),
+      request_headers_({{":method", "GET"}, {":path", "/"}, {":authority", "host"}}) {}
+
+void HttpFilterIntegrationTestBase::setup(const std::string& config) {
```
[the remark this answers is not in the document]
oschaaf: Done -- this was slightly more convenient in the implementation as we need to copy, but I agree consistency outweighs that. Thanks for catching

test/server/http_filter_integration_test_base.cc
oschaaf: I added some comments in ca17f41. Does that provide the type of context you'd consider helpful when reading this code? 
Basically, this support call is an extract of the what we need to cover the stream codec code for all our extensions. We'll later see that this allows us deprecate much more complex code (which was the way to do it a while back), as wel as deduplicate some similar code.

oschaaf:
@dubious90 feedback address in ca17f41, thanks for the review!

test/server/http_filter_integration_test_base.cc
oschaaf: Additionally, for some more concrete context, https://github.com/envoyproxy/nighthawk/pull/533 is the direct follow up to this (currently in draft, as it depends on this PR, and will lean up once we merge this one and merge the result into that). It brings in a [new test file](https://github.com/envoyproxy/nighthawk/blob/97322d867ca1ebeb35bed0412e7b603bff4ba8a6/test/server/http_filter_base_test.cc) which is using this extensively. 
That test file adds tests for ensuring basic behavioural guarantees across all of our extensions: https://github.com/envoyproxy/nighthawk/blob/97322d867ca1ebeb35bed0412e7b603bff4ba8a6/test/server/http_filter_base_test.cc

The behaviours we are looking for involve certain configuration (error) handling, as well as handling default/empty configuration, both static from disk as well as dynamically delivered via response headers. These tests are parameterised over GET and POST requests.

Once we land that, we can in turn start purging / enhancing the other test files, and make them more about just testing the functionality that the extension brings in. Where applicable, it will also use the functionality we add in this PR, and we can purge their own variant(s).

oschaaf: (NOTE: the changes from the last paragraph above, are already staged over at #512)

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
+#include <string>
+
+#include "external/envoy/test/integration/http_integration.h"
+
+namespace Nighthawk {
+
+/**
+ * Base class with shared functionality for testing Nighthawk test server http filter extensions.
```
dubious90, line 10: Can we add two notes/clarifications here: 1. This class is stateful.  2. Unless I'm mistaken I don't think this class is threadsafe

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Construct a new HttpFilterIntegrationTestBase instance.
+   *
+   * @param ip_version Specify the ip version that the integration test server will use to listen
+   * for connections.
+   */
+  HttpFilterIntegrationTestBase(Envoy::Network::Address::IpVersion ip_version);
+
+  /**
+   * We don't override SetUp(): tests using this fixture must call setup() instead.
```
dubious90, line 27: nit: reference to setup() here is no longer accurate

test/server/http_filter_integration_test_base.cc
```diff
@@ -0,0 +1,63 @@
...
+  cleanupUpstreamAndDownstream();
+  codec_client_ = makeHttpConnection(lookupPort("http"));
+  Envoy::IntegrationStreamDecoderPtr response;
+  const bool is_post = request_headers_.Method()->value().getStringView() ==
+                       Envoy::Http::Headers::get().MethodValues.Post;
+  // Upon observing a POST request method, we inject a content body, as promised in the header file.
+  // This is useful, because emitting an entity body will hit distinct code in extensions. Hence we
+  // facilitate that.
+  const uint64_t request_body_size = is_post ? 1024 : 0;
+  // An extension can either act as an origin and synthesize a reply, or delegate that
```
dubious90, line 45: Looking through again, these comments are very helpful.  Thank you for adding them

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+   * Set a request header value. Overwrites any existing value.
+   *
+   * @param header_name Name of the request header to set.
+   * @param header_value Value to set for the request header.
+   */
+  void setRequestHeader(const Envoy::Http::LowerCaseString& header_name,
+                        absl::string_view header_value);
+
+  /**
+   * Fetch a response. The request headers default to a minimal GET, but this may be changed
```
dubious90, line 60: Can we rephrase this comment a little (correct anything here that is inaccurate):
Fetch a response, according to the options specified by the class methods. By default, simulates a GET request with minimal headers.

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+   * @param ip_version Specify the ip version that the integration test server will use to listen
+   * for connections.
+   */
+  HttpFilterIntegrationTestBase(Envoy::Network::Address::IpVersion ip_version);
+
+  /**
+   * We don't override SetUp(): tests using this fixture must call setup() instead.
+   * This is to avoid imposing the need to create a fixture per filter configuration.
+   *
+   * @param config configuration to pass to Envoy::HttpIntegrationTest::config_helper_.addFilter.
```
dubious90, line 30: is this configuration for the filter being tested?  If so, can we clarify "filter configuration"

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Base class with shared functionality for testing Nighthawk test server http filter extensions.
+ */
+class HttpFilterIntegrationTestBase : public Envoy::HttpIntegrationTest {
+protected:
+  /**
+   * Indicate the expected response origin.
+   */
+  enum class ResponseOrigin { UPSTREAM, EXTENSION };
```
dubious90, line 17: Can we add a one line summary here for each type of ResponseOrigin?  This may also allow you to remove a comment elsewhere, if it's duplication at this point.

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Make getResponse send request-level configuration. Test server extensions read that
+   * configuration and merge it with their static configuration to determine a final effective
+   * configuration. See TestServerConfig in well_known_headers.h for the up to date header name.
+   *
+   * @param request_level_config Configuration to be delivered by request-header in future calls to
+   * getResponse(). For example: "{response_body_size:1024}".
+   */
+  void updateRequestLevelConfiguration(absl::string_view request_level_config);
```
dubious90, line 42: nit/optional:  How do you feel about changing this to setRequestLevelConfiguration? I think it'd then be more consistent with setRequestHeader

test/server/http_filter_integration_test_base.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Switch getResponse() to use the POST request method with an entity body.
+   * Doing so will make tests hit a different code paths in extensions.
+   */
+  void switchToPostWithEntityBody();
```
dubious90, line 48: Just a thought, curious what you think - it might be more self-documenting here, if we take in a parameter for the post body size.  Or even then, rename this to setPostBodySize, with an indication that unless this is called, the requests into the filter will be GET requests.

dubious90 (changes requested):
Thank you for adding all of the comments.  I understand what is happening here a lot more now.  Have some suggestions on naming and documentation, but as always, open to pushback.

oschaaf:
@dubious90 feedback addressed in dubious90 modulo one thing, I will reply to that specific comment to explain.

test/server/http_filter_integration_test_base.h
oschaaf: I would be more up for doing this, but rather do it once a functional need for it arises (which is very possible). Right now, in the context of this PR and the changes coming after this, I'm on a mission to minimize and contain complexity that has organically grown while we added the test server and the two new extensions. Part of effort is making everything go through this hopefully well defined but minimal functionality. It should be trivial to refactor this later on according to what you suggest, once the need arises in new tests?

dubious90 (approved):
LGTM.  Thank you for addressing my concerns.

test/server/http_filter_integration_test_base.h
dubious90: That makes sense to me.  I'm happy to push it through as is, and if it comes up, do the refactor as you suggest.

PR #518, opened 2020-09-09
Add FilterConfigurationBase

oschaaf:
Base class for test server http filter extension configuration.
Provides support for shared configuration and consolidates
shared code.

Potentially useful in that it provides a path to cache parsed
and interpreted configuration accross extensions as it flows
through the server pipeline. This isn't a goal for now however,
this is a prelude to #512, which includes it and provides a means
to see the end goal.


oschaaf:
note to reviewers: to avoid a dip in coverage, this doesn't wire up in BUILD yet. There's no separate testing on this around yet, though it will get tested once we're done via extensions as the extensions rely on this (see #512).

source/server/http_filter_config_base.cc
```diff
@@ -0,0 +1,42 @@
...
+      effective_config_ =
+          std::make_shared<const nighthawk::server::ResponseOptions>(std::move(response_options));
+    } else {
+      effective_config_ = absl::InvalidArgumentError(error_message);
+    }
+  }
+}
+
+bool FilterConfigurationBase::maybeSendErrorReply(
+    Envoy::Http::StreamDecoderFilterCallbacks& decoder_callbacks) const {
```
dubious90, line 29: Can this be const?  Or is sendLocalReply a non-const function?

dubious90 (commented):
Looks good.  Have a couple comments.

Just to make sure I follow, there is a PR coming soon that will integrate these changes into our existing filters, and that's when we'll bring in both the BUILD changes and the testing?

oschaaf (commented):
@dubious90 review feedback addressed in 562a1d9, including the updated canonical representation which uses absl::StatusOr, so we can open up the discussion on that.

oschaaf:
> Looks good. Have a couple comments.
> 
> Just to make sure I follow, there is a PR coming soon that will integrate these changes into our existing filters, and that's when we'll bring in both the BUILD changes and the testing?

that is correct. this is a subset of #512 which has the BUILD file and everything else. The BUILD file is excluded to avoid coverage from failing because of dipping below the threshold.

oschaaf:
@dubious90 as discussed, backed out the absl::StatusOr typedef in b894e9606518c7131b9558300009dff90046a4ac

dubious90 (approved):

PR #522, opened 2020-09-10
Adding config factory

wjuan-AFK:
Adding config factories for RequestSourcePlugins. Specifically starting with FileBasedRequestSourcePlugin. These will be provided as options to be used inside RequestSourceFactory. This is part of a series of PRs for providing the ability to use statically linked requestSources.


api/client/options.proto
```diff
@@ -18,6 +18,10 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+message RequestOptionses {
```
oschaaf, line 21: Also, maybe add a comment about the purpose of this message?
Also, `RequestOptionses` reads a bit unnatural to me. After reading some more, there seems to be a naming conflict with two things actually competing for the name RequestOptions... I wonder: can we figure out a better name? I'm not 100% sure if this is on target, but to float a suggestion something like `TrafficProfile` instead?

include/nighthawk/common/factories.h
```diff
@@ -72,4 +72,4 @@ class NighthawkStatsSinkFactory : public Envoy::Config::TypedFactory {
   std::string category() const override { return "nighthawk.stats_sinks"; }
 };
 
-} // namespace Nighthawk
+} // namespace Nighthawk
```
oschaaf, line 75: This file could be left unchanged?

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+class RequestSourcePluginConfigFactory : public Envoy::Config::TypedFactory {
+public:
+  ~RequestSourcePluginConfigFactory() override = default;
+  std::string category() const override { return "nighthawk.request_source_plugin"; }
+  /**
+   * Instantiates the specific RequestSourcePlugin class. Casts |message| to Any, unpacks it to the
+   * plugin-specific proto, and passes the strongly typed proto to the plugin constructor.
+   *
+   * @param message Any typed_config proto taken from the TypedExtensionConfig.
```
oschaaf, line 37: Missing a description of the api argument?

source/client/client_worker_impl.cc
```diff
@@ -106,4 +106,4 @@ StatisticPtrMap ClientWorkerImpl::statistics() const {
 }
 
 } // namespace Client
-} // namespace Nighthawk
+} // namespace Nighthawk
```
oschaaf, line 109: Some changes that seem unnecessary here and below

source/common/request_source_plugin_impl.h
```diff
@@ -0,0 +1,87 @@
...
+
+// This factory is activated through ???.
+DECLARE_FACTORY(DummyRequestSourceConfigFactory);
+
+using RequestOptionsIterator =
+    Envoy::ProtobufWkt::internal::RepeatedPtrIterator<const nighthawk::client::RequestOptions>;
+
+/**
+ */
+class FileBasedRequestSourcePlugin : public RequestSourcePlugin {
```
oschaaf, line 53: If I understand correctly, this plugin will replay request specifications from a file. This file gets read into memory by each worker, and each worker will replay it. Is that correct? Maybe it's worth adding a doc comment at the class level for this?

oschaaf (commented):
Looks great to me overall, a couple of nits and doc/naming suggestions to get started.

api/client/options.proto
wjuan-AFK: Renaming it to RequestOptionsList. I hope that works better? I know calling it a list might be misleading.

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Nighthawk {
+
+/**
+ * An interface for different RequestSource plugins.
+ * Uses a plugin-specific config proto.
+ */
+class RequestSourcePlugin : RequestSource {
+public:
+  RequestSourcePlugin(Envoy::Api::Api& api) : api_(api) {}
```
oschaaf, line 17: Mocking might be easier if the api argument is added via a setter or separate init call. It may even void the need to be stateful with respect to keeping a reference to the api.

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,50 @@
...
+
+protected:
+  Envoy::Api::Api& api_;
```
oschaaf, line 20: This seems like implementation in an interface. Consider creating a common base class which derives from the interface for this?

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,106 @@
...
+                                                 : std::numeric_limits<double>::infinity()} {}
+RequestGenerator DummyRequestSourcePlugin::get() {
+  RequestGenerator request_generator = []() {
+    Envoy::Http::RequestHeaderMapPtr header = Envoy::Http::RequestHeaderMapImpl::create();
+    auto returned_request_impl = std::make_unique<RequestImpl>(std::move(header));
+    return returned_request_impl;
+  };
+  return request_generator;
+}
+void DummyRequestSourcePlugin::initOnThread() {}
```
oschaaf, line 47: this could perhaps fit nicely into a common base implementation as the default implementation, to save plugins that don't need it some bootstrapping.

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,106 @@
...
+}
+
+REGISTER_FACTORY(FileBasedRequestSourceConfigFactory, RequestSourcePluginConfigFactory);
+
+FileBasedRequestSourcePlugin::FileBasedRequestSourcePlugin(
+    const nighthawk::request_source::FileBasedPluginRequestSourceConfig& config,
+    Envoy::Api::Api& api)
+    : RequestSourcePlugin{api}, uri_(config.uri()), file_path_(config.file_path()) {
+  Envoy::MessageUtil util;
+  util.loadFromFile(file_path_, options_list_, Envoy::ProtobufMessage::getStrictValidationVisitor(),
```
oschaaf, line 72: Looking at this, I wonder what would happen here when reading super large files.  If the use case is reading files that fit into random access memory (also when considering running with lots of workers), then maybe renaming this class to something to reflect that would be good?

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,106 @@
...
+                                               : Envoy::Http::Headers::get().SchemeValues.Http);
+    header->setMethod(envoy::config::core::v3::RequestMethod_Name(request_option.request_method()));
+    const uint32_t content_length = request_option.request_body_size().value();
+    if (content_length > 0) {
+      header->setContentLength(content_length);
+    }
+    for (const auto& option_header : request_option.request_headers()) {
+      auto lower_case_key = Envoy::Http::LowerCaseString(std::string(option_header.header().key()));
+      header->remove(lower_case_key);
+      header->addCopy(lower_case_key, std::string(option_header.header().value()));
```
oschaaf, line 96: I wonder, Isn't there a `setCopy()` around instead of the remove/add? Also -- should this just add? Or ... maybe what happens should be part of the options spec (replace/remove/append)?

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,106 @@
...
+    }
+    auto returned_request_impl = std::make_unique<RequestImpl>(std::move(header));
```
oschaaf, line 98: maybe return the value directly instead of via this temporary variable?

source/common/request_source_plugin_impl.cc
oschaaf: Also -- I don't remember, but this doesn't happen concurrently right? If it does, should we be concerned about loadFromFile using exclusive file-modes? 

Also, independently -- I am curious, when multiple workers are involved, should they all start at 0 in the file / have a full copy? Or can they share that memory? 
Or could/should they offsets, each taking an equal chunk of the file for reply purposes? In that case we might save some memory by storing only what a worker needs.

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,114 @@
...
+using nighthawk::request_source::FileBasedPluginRequestSourceConfig;
+using ::testing::Test;
+
+class DummyRequestSourcePluginTest : public Test {
+public:
+  DummyRequestSourcePluginTest() : api_(Envoy::Api::createApiForTest(stats_store_)) {}
+  Envoy::Api::ApiPtr api_;
+  Envoy::Stats::MockIsolatedStatsStore stats_store_;
+}; // RequestSourcePluginTest
+class FileBasedRequestSourcePluginTest : public Test {
```
oschaaf, line 28: please add a newline here.

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,114 @@
...
+using nighthawk::request_source::DummyPluginRequestSourceConfig;
+using nighthawk::request_source::FileBasedPluginRequestSourceConfig;
+using ::testing::Test;
+
+class DummyRequestSourcePluginTest : public Test {
+public:
+  DummyRequestSourcePluginTest() : api_(Envoy::Api::createApiForTest(stats_store_)) {}
+  Envoy::Api::ApiPtr api_;
+  Envoy::Stats::MockIsolatedStatsStore stats_store_;
+}; // RequestSourcePluginTest
```
oschaaf, line 27: This comment seems unusual, this is what we usually do with namespaces. The rest of the code doesn't do this, so for consistency please consider removing it.

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,114 @@
...
+class FileBasedRequestSourcePluginTest : public Test {
+public:
+  FileBasedRequestSourcePluginTest() : api_(Envoy::Api::createApiForTest(stats_store_)) {}
+  Envoy::Api::ApiPtr api_;
+  Envoy::Stats::MockIsolatedStatsStore stats_store_;
+  nighthawk::request_source::FileBasedPluginRequestSourceConfig
+  MakeFileBasedPluginConfigWithTestYaml(const std::string& request_file) {
```
oschaaf, line 34: Instead of `const std::string&` we usually go for `absl::string_view` in this code base. I there a strong reason to use `const std::string&` here?

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,114 @@
...
+    nighthawk::request_source::FileBasedPluginRequestSourceConfig config;
+    config.mutable_uri()->assign("http://foo/");
+    config.mutable_file_path()->assign(request_file);
+    return config;
+  }
+}; // RequestSourcePluginTest
```
oschaaf, line 40: See my other comment, we don't usually do this at the class level

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,114 @@
...
+TEST_F(DummyRequestSourcePluginTest, CreateEmptyConfigProtoCreatesCorrectType) {
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<RequestSourcePluginConfigFactory>(
+          "nighthawk.dummy-request-source-plugin");
+  const Envoy::ProtobufTypes::MessagePtr empty_config = config_factory.createEmptyConfigProto();
+  const nighthawk::request_source::DummyPluginRequestSourceConfig expected_config;
+  EXPECT_EQ(empty_config->DebugString(), expected_config.DebugString());
+  EXPECT_TRUE(Envoy::MessageUtil()(*empty_config, expected_config));
+}
+TEST_F(DummyRequestSourcePluginTest, FactoryRegistrationUsesCorrectPluginName) {
```
oschaaf, line 51: nit: missing newline here and below in a couple of other spots.

test/request_source/test_data/test-config.yaml
```diff
@@ -0,0 +1,13 @@
+options:
+  - request_body_size: 10
+    request_headers:
+      - { header: { key: ":path", value: "/a" } }
+      - { header: { key: "foo", value: "bar" } }
+      - { header: { key: "foo", value: "bar2" }, append: true }
```
oschaaf, line 6: Ah, I see this append here, but it's ignored in the file backed implementation. That may be pre-existing, not sure if the gRPC request source variant that exists today minds this field.

oschaaf (changes requested):
Flushing out some more comments. Looks great overall, but I had one higher level question about the file backed implementation and its purpose; is there a cap to the file size we want to impose? If there is, we may want to pre-check before loading up the file and optimize memory usage. If there's not, we may want to look into streaming the input somehow to avoid OOM.

include/nighthawk/common/request_source_plugin.h
qqustc, line 17: Add `const` into `RequestSourcePlugin(const Envoy::Api::Api& api)`?

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "nighthawk/common/request_source.h"
+
+namespace Nighthawk {
+
+/**
+ * An interface for different RequestSource plugins.
+ * Uses a plugin-specific config proto.
+ */
+class RequestSourcePlugin : RequestSource {
```
qqustc, line 15: If RequestSourcePlugin inherit from RequestSource just to add the constructor, then we don't need to create  RequestSourcePlugin. We can directly create an implementation of RequestSource instead.

source/common/request_source_plugin_impl.h
```diff
@@ -0,0 +1,87 @@
...
+
+#include "api/client/options.pb.h"
+#include "api/request_source/request_source_plugin_impl.pb.h"
+
+#include "common/uri_impl.h"
+
+namespace Nighthawk {
+
+/**
+ */
```
qqustc, line 15: Plz add comment.

source/common/request_source_plugin_impl.h
```diff
@@ -0,0 +1,87 @@
...
+class DummyRequestSourcePlugin : public RequestSourcePlugin {
```
qqustc, line 16: Plz correct me if I'm wrong, it seems DummyRequestSourcePlugin can directly inherit from RequestSource?

source/common/request_source_plugin_impl.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+class DummyRequestSourceConfigFactory : public virtual RequestSourcePluginConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  RequestSourcePluginPtr createRequestSourcePlugin(const Envoy::Protobuf::Message& message,
+                                                   Envoy::Api::Api& api) override;
+};
+
+// This factory is activated through ???.
```
qqustc, line 45: plz update comment.

api/request_source/request_source_plugin_impl.proto
```diff
@@ -0,0 +1,23 @@
...
+
+// Configuration for DummyPluginRequestSource (plugin name: "nighthawk.dummy-request-source-plugin")
+// that does nothing
+message DummyPluginRequestSourceConfig {
+  // Dummy input value.
+  google.protobuf.DoubleValue dummy_value = 1;
+}
+// Configuration for RPCPluginRequestSource (plugin name: "nighthawk.rpc-request-source-plugin")
+// that uses rpc
+message RPCPluginRequestSourceConfig {
```
qqustc, line 15: It seems this RPCPluginRequestSourceConfig proto is not used in this PR?

qqustc (changes requested):

include/nighthawk/common/factories.h
wjuan-AFK: Fixed.

include/nighthawk/common/request_source_plugin.h
wjuan-AFK: Done.

source/client/client_worker_impl.cc
wjuan-AFK: Fixed.

api/request_source/BUILD
```diff
@@ -16,6 +16,18 @@ api_cc_py_proto_library(
     ],
 )
 
+api_cc_py_proto_library(
+    name = "request_source_plugin",
+    srcs = [
+        "request_source_plugin_impl.proto",
```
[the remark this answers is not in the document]
wjuan-AFK: I went with that because we used a similar naming scheme for the adaptive_load protos. I agree that it feels weird though. Fixed.

source/common/request_source_plugin_impl.cc
wjuan-AFK: The load from file concern is something I am addressing, but the use case for this is indeed for smaller files, since this is mostly a small plugin to act as a test case for the change.

source/common/request_source_plugin_impl.h
wjuan-AFK: Good point.

source/common/request_source_plugin_impl.cc
wjuan-AFK: There is a setCopy(). I'm not sure about making it an option, but I'll spin off a separate issue for adding that so I can get this PR in.

source/common/request_source_plugin_impl.cc
wjuan-AFK: Makes sense.

test/request_source/request_source_plugin_test.cc
wjuan-AFK: Sure.

test/request_source/request_source_plugin_test.cc
wjuan-AFK: Understood.

test/request_source/request_source_plugin_test.cc
wjuan-AFK: No strong reason to avoid the string_view. I'll swap it.

test/request_source/request_source_plugin_test.cc
wjuan-AFK: Ack.

test/request_source/request_source_plugin_test.cc
wjuan-AFK: Cleaned those up. Thanks.

test/request_source/test_data/test-config.yaml
wjuan-AFK: I'll remove this.

api/client/options.proto
```diff
@@ -18,6 +18,10 @@ message RequestOptions {
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
 }
 
+message RequestOptionsList {
```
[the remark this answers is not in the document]
wjuan-AFK: Can do.

api/request_source/request_source_plugin_impl.proto
```diff
@@ -0,0 +1,23 @@
...
+package nighthawk.request_source;
+
+import "google/protobuf/wrappers.proto";
+
+// Configuration for DummyPluginRequestSource (plugin name: "nighthawk.dummy-request-source-plugin")
+// that does nothing
+message DummyPluginRequestSourceConfig {
+  // Dummy input value.
+  google.protobuf.DoubleValue dummy_value = 1;
+}
```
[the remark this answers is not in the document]
wjuan-AFK: Done.

include/nighthawk/common/request_source_plugin.h
wjuan-AFK: I don't think I can do that, because I need to use the api for "loadFromFile" which doesn't take a const api.

source/common/request_source_plugin_impl.cc
oschaaf: Sounds good

include/nighthawk/common/request_source_plugin.h
wjuan-AFK: I need this interface for the register factory mechanism. And the plugins I'm making need to be request sources to be used with the RequestSourceFactory. I might be misunderstanding your comment.

source/common/request_source_plugin_impl.h
wjuan-AFK: Done.

source/common/request_source_plugin_impl.h
wjuan-AFK: So, generally speaking, you're right. However, in this case it's made awkward by the fact that I want the create method for the generic case to take in an api parameter. This means that unless I make the constructor for all RequestSourcePlugins take the api parameter, the compiler complains about an unused parameter. I agree that this seems inelegant, but I wasn't sure if there was a better solution for this.

api/request_source/request_source_plugin_impl.proto
wjuan-AFK: Correct. I'll get rid of it.

source/common/request_source_plugin_impl.h
wjuan-AFK: Done.

include/nighthawk/common/request_source_plugin.h
oschaaf: Also, aren't the things we needed to construct an implementation specific to that implementation?
Maybe consider passing a RequestSourcePluginContext(Ptr) to abstract that away here? 
Or if we don't want to impose an interface for that, just a (void*)? Otherwise, we can expect to have to update the interface as implementations are added.
Then at the implementation layer plugins could use (dynamic_)casts to get to a specific re-implementation of it to satisfy their needs?

source/common/request_source_plugin_impl.h
wjuan-AFK: nevermind, I was dumb.

include/nighthawk/common/request_source_plugin.h
wjuan-AFK: I just got rid of this.

source/common/request_source_plugin_impl.cc
wjuan-AFK: I'm not sure I entirely understand this comment.

oschaaf: An example:

```c++

// Interface
class RequestSourcePlugin {
    virtual initOnThread() PURE;
};

// BaseImpl
class RequestSourcePluginBaseImpl {
    // Add default uninteresting behavior.
    void initOnThread() override {};
};

// Actual concrete implementations can derive from RequestSourcePluginBase above, e.g.
class RequestSourcePluginFooImpl : RequestSourcePluginBaseImpl {
    // no need to override initOnThread() here
};

```

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,28 @@
...
+// Configuration for RPCPluginRequestSource (plugin name: "nighthawk.rpc-request-source-plugin")
+// that uses rpc
+message RPCPluginRequestSourceConfig {
+  string uri = 1;
+}
+
+// Configuration for FileBasedPluginRequestSource (plugin name:
+// "nighthawk.file-based-request-source-plugin") that uses rpc
+message FileBasedPluginRequestSourceConfig {
+  string uri = 1;
```
oschaaf, line 24: Silly perhaps, but at the first glance I thought this `uri` field might be used to point out a to-be-downloaded file for replay.  Maybe a rename and/or a helpful comment is good here? For example "request_header_base_uri" or some such. 

I'm not sure if it's intentional, but it is also worth explicitly calling out that with the way this is now implemented, the result of applying `uri` to a request header can be overridden by setting related request headers (":path, :authority") on a per-request basis in "records" from the file contents.

That leads to the thought: if this behavior is intended/desired, should we allow this uri to be empty? If the file is expected to fully supply this for every request, that would make sense. Maybe, in that case, this field should be optional (via `google.protobuf.StringValue`?), so we can make a distinction between empty and not set. 

Alternatively, it is also possible to make a change to ignore or error out when ambiguity arises based on `uri` and file-based content. But either way, I think we need to define / document the desired behaviour. 

Or, possibly even generalize this accross plugins, and enforce consistent behavior.

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,28 @@
...
+  string file_path = 2;
+  google.protobuf.UInt32Value num_requests = 3 [(validate.rules).uint32 = {gte: 0, lte: 1000000}];
+  google.protobuf.UInt32Value max_file_size = 4;
```
oschaaf, line 27: Naive question: we'll be reading a proto stream from disk, right? I have no idea how that works, but that won't be subject to things along the lines of [zip boms](https://en.wikipedia.org/wiki/Zip_bomb) right?

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,38 @@
...
+   * @param message Any typed_config proto taken from the TypedExtensionConfig.
+   *
+   * @param api Api parameter that contains timesystem, filesystem, and threadfactory.
+   *
+   * @return RequestSourcePtr Pointer to the new plugin instance.
+   *
+   * @throw Envoy::EnvoyException If the Any proto cannot be unpacked as the type expected by the
+   * plugin.
+   */
+  virtual RequestSourcePtr createRequestSourcePlugin(const Envoy::Protobuf::Message& message,
```
oschaaf, line 34: It just might, but are we sure `api` always suffices as an input for every implementation to construct it?

source/client/factories_impl.h
```diff
@@ -80,4 +80,4 @@ class TerminationPredicateFactoryImpl : public OptionBasedFactoryImpl,
 };
 
 } // namespace Client
-} // namespace Nighthawk
+} // namespace Nighthawk
```
oschaaf, line 83: accidental change?

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,118 @@
...
+    const Envoy::Protobuf::Message& message, Envoy::Api::Api& api) {
+  const auto& any = dynamic_cast<const Envoy::ProtobufWkt::Any&>(message);
+  nighthawk::request_source::FileBasedPluginRequestSourceConfig config;
+  Envoy::MessageUtil util;
+  util.unpackTo(any, config);
+  RELEASE_ASSERT(api.fileSystem().fileSize(config.file_path()) < config.max_file_size().value(),
+                 "file size must be less than max_file_size");
+  auto temp_list = std::make_unique<nighthawk::client::RequestOptionsList>();
+  {
+    Envoy::Thread::LockGuard lock_guard(file_lock_);
```
oschaaf, line 68: Why do we need to lock here? Can we spend a code-level comment on that?  Also, should we then also have a simple spam test for this, calling it from multiple threads? We might not even want to have any specific expectations in it, but just so we give tsan a good opportunity to have a say about this?
(sorry, I don't want to bloat this PR, but given that it's targetting production and not test code, this might be worth considering?)

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,118 @@
...
+FileBasedRequestSourcePlugin::FileBasedRequestSourcePlugin(
+    const nighthawk::request_source::FileBasedPluginRequestSourceConfig& config,
+    std::unique_ptr<nighthawk::client::RequestOptionsList> options_list)
+    : uri_(config.uri()), file_path_(config.file_path()),
+      options_list_(std::move(options_list)), request_max_(config.num_requests().value()) {}
+
+RequestGenerator FileBasedRequestSourcePlugin::get() {
+  uint32_t counter = 0;
+  request_count_.push_back(counter);
+  uint32_t* lambda_counter = &request_count_.back();
```
oschaaf, line 89: instead of a raw pointer, I wonder if we could instead do:
```c++
uint32_t& lambda_counter = request_count_.back();
```

that way, we eliminate the possibility of nulling it, which simplifies things slightly for future readers of this code?

source/common/request_source_plugin_impl.cc
wjuan-AFK: Understood.

oschaaf (changes requested):
Thanks for following up; I had another pass over this, and as a result here are a few more thoughts and asks.

api/request_source/request_source_plugin.proto
oschaaf: Oh -- afterthought here; I think the gRPC plugin also behaves like this; so we're consistent, but we're not enforcing that consistency. This could perhaps also be facilitated by a base implementation. I don't want to bloat this PR though.

include/nighthawk/common/request_source_plugin.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "nighthawk/common/request_source.h"
+
+namespace Nighthawk {
+
+/**
+ * A factory that must be implemented for each RequestSourcePlugin. It instantiates the specific
+ * RequestSourcePlugin class after unpacking the plugin-specific config proto.
+ */
+class RequestSourcePluginConfigFactory : public Envoy::Config::TypedFactory {
```
qqustc, line 15: Rename the file name to something like `request_source_plugin_config_factory.h`?

source/common/request_source_plugin_impl.cc
```diff
@@ -0,0 +1,116 @@
...
+
+REGISTER_FACTORY(FileBasedRequestSourceConfigFactory, RequestSourcePluginConfigFactory);
+
+FileBasedRequestSourcePlugin::FileBasedRequestSourcePlugin(
+    const nighthawk::request_source::FileBasedPluginRequestSourceConfig& config,
+    std::unique_ptr<nighthawk::client::RequestOptionsList> options_list)
+    : uri_(config.uri()), file_path_(config.file_path()), options_list_(std::move(options_list)),
+      request_max_(config.num_requests().value()) {}
+
+RequestGenerator FileBasedRequestSourcePlugin::get() {
```
qqustc, line 84: nit: this fuction seems quite complicated, plz try to add some comments in the function body to explain the logic.

qqustc (changes requested):

source/common/request_source_plugin_impl.cc
wjuan-AFK: I added a test with multiple calls to create, though I'm not sure that this tests for what you want to test for. I added a comment about the lock also.

oschaaf: Thanks for that comment, that is helpful. So now that the intent is clear, looking at the test you added, how about adding something like [1] to ensure multiple *concurrent* calls will only ever result in a single config file being read from disk? Such an amendment would also give the CI TSAN job a chance to verify.

```c++
void spamCall(std::function<void()>&& call_to_spam, const uint32_t num_threads) {
  std::vector<std::thread> threads(num_threads);
  auto barrier = std::make_unique<absl::Barrier>(num_threads);

  for (auto& thread : threads) {
    thread = std::thread([&call_to_spam, &barrier] {
      // Allow threads to accrue, to maximize concurrency on the call we are testing.
      if (barrier->Block()) {
        barrier.reset();
      }
      call_to_spam();
    });
  }
  for (std::thread& thread : threads) {
    thread.join();
  }
}

....

TEST_F(FileBasedRequestSourcePluginTest, CreateMultipleRequestSourcePluginReadsFileOnce) {
   constexpr uint32_t kNumThreads = 100;
   spamCall([kNumThreads]() { 
       .. insert the current test body which ensures a single config file is read that you have here...
    }, 
    kNumThreads);
}

```

include/nighthawk/common/request_source_plugin_config_factory.h
```diff
@@ -0,0 +1,46 @@
...
+
+struct RequestSourceContext {
+  RequestSourceContext(RequestSourceContext& context)
+      : api(std::move(context.api)), header(std::move(context.header)) {}
+  RequestSourceContext(Envoy::Api::ApiPtr api_input, Envoy::Http::RequestHeaderMapPtr header_input)
+      : api(std::move(api_input)), header(std::move(header_input)) {}
+  Envoy::Api::ApiPtr api;
+  Envoy::Http::RequestHeaderMapPtr header;
+};
+using RequestSourceContextPtr = std::unique_ptr<RequestSourceContext>;
```
oschaaf, line 19: nit: missing space. I wonder, if you run `ci/do_ci.sh fix_format`, does it inject this space for you?

source/common/request_source_plugin_impl.cc
oschaaf: Also, it seems to me that there might be value to put every known implementation of `::initOnThread` through a similar test to help ensure concurrent calls of these stay well behaved during future maintenance. I don't think this is an issue now or with the code here, but if you agree on the principle, maybe file an issue on it and tag it as tech-debt?

oschaaf (commented):
Thanks for addressing feedback. This looks great. I'm posting a last round of suggestions, based on the last set of changes.

include/nighthawk/common/request_source_plugin_config_factory.h
wjuan-AFK: Calling ci/do_ci.sh fix_format just now didn't seem to do anything on this file.

source/common/request_source_plugin_impl.cc
wjuan-AFK: After further thought, this doesn't seem appropriate to have as a unit test. The primary problem being that if it were to be an appropriate unit test for a threading issue, then it is inherently trying to be non deterministic, which is not what a unit test should be. It seems like what we want is maybe closer to an integration test, but that's probably best done in a separate pr.

include/nighthawk/common/request_source_plugin_config_factory.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "envoy/api/api.h"
+#include "envoy/common/pure.h"
+#include "envoy/config/typed_config.h"
+
+#include "nighthawk/common/request_source.h"
+
+namespace Nighthawk {
+
+struct RequestSourceContext {
```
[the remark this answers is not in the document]
wjuan-AFK: Getting rid of this context because it does not make sense with the RegisterFactory mechanism we are using, since the actual arguments that are being passed are chosen before we decide which factory is being used.

qqustc (dismissed):
LGTM. Assign back to @mum4k

include/nighthawk/common/request_source_plugin_config_factory.h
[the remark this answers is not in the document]
oschaaf: Yes, thanks for explaining, makes sense.

source/common/request_source_plugin_impl.cc
oschaaf: A separate PR works for me; maybe file an issue to track?

oschaaf (dismissed):
lgtm

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package nighthawk.request_source;
+
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+import "api/client/options.proto";
+
+// Configuration for DummyPluginRequestSource (plugin name: "nighthawk.dummy-request-source-plugin")
```
mum4k, line 9: I think we could use better documentation outlining what these individual plugin types do. This being the public API might be a good place to add a few sentences next to each plugin. That documentation should help the reader understand what the plugin is for. The current documentation is fairly short and requires the reader to already have some context.

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
...
+// that does nothing
+message DummyPluginRequestSourceConfig {
```
mum4k, line 11: What is this one used for? It it only used for testing? If so, we could move it down in the file, so this isn't the first thing the reader encounters.

Assuming it is just for testing, we should use better terminology than "dummy". A test only implementation that do just one thing are usually called "Stubs".

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
...
+  // Dummy input value.
+  google.protobuf.DoubleValue dummy_value = 1;
```
mum4k, line 13: Why do we need a value on a plugin that doesn't do anything? I suspect it is for testing, but just making sure I understand it correctly.

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Configuration for RPCPluginRequestSource (plugin name: "nighthawk.rpc-request-source-plugin")
+// that uses rpc
+message RPCPluginRequestSourceConfig {
+  string uri = 1;
```
mum4k, line 19: Can we add a comment explaining this field? All fields in proto messages should have good comments since they essentially represent the public API. Please apply throughout.

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Configuration for FileBasedPluginRequestSource (plugin name:
+// "nighthawk.file-based-request-source-plugin") that uses rpc
+message FileBasedPluginRequestSourceConfig {
+  string file_path = 1;
+  google.protobuf.UInt32Value num_requests = 2 [(validate.rules).uint32 = {gte: 0, lte: 1000000}];
+  google.protobuf.UInt32Value max_file_size = 3;
```
mum4k, line 27: Why is the max_file_size a configuration knob? Do we have a good use case where the user needs to tweak this value?

api/request_source/request_source_plugin.proto
mum4k, line 11: These names are fairly repetitive. They all currently live in the request_source namespace, but also contain the words "RequestSource". Can we remove one or the other?

This especially shows in the name RequestOptionsListPluginRequestSourceConfig which reached a length that makes it hard to parse meaning from it.

include/nighthawk/common/request_source_plugin_config_factory.h
```diff
@@ -0,0 +1,37 @@
...
+public:
+  ~RequestSourcePluginConfigFactory() override = default;
+  std::string category() const override { return "nighthawk.request_source_plugin"; }
+  /**
+   * Instantiates the specific RequestSourcePlugin class. Casts |message| to Any, unpacks it to the
+   * plugin-specific proto, and passes the strongly typed proto to the plugin constructor.
+   *
+   * @param message Any typed_config proto taken from the TypedExtensionConfig.
+   *
+   * @param api Api parameter that contains timesystem, filesystem, and threadfactory.
```
mum4k, line 25: Looks like we may be missing comment related to the header param.

source/common/BUILD
```diff
@@ -81,6 +81,30 @@ envoy_cc_library(
     ],
 )
 
+envoy_cc_library(
+    name = "request_source_plugin_impl",
```
mum4k, line 85: This is a good point to start thinking about slightly different file / directory structure here. The common directory is growing which makes it harder to navigate. How about we create a new request_source directory instead at the same level where common is?

In addition to this, we should split out the individual request source plugin implementations into their own files. Placing them all into one will quickly outgrow "good file size". There isn't a hard rule, but as a recommendation I usually start thinking about a refactor when a file reaches about 300 lines. Let's try to refactor these as suggested and see how it looks.

source/common/request_source_plugin_impl.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include "api/client/options.pb.h"
+#include "api/request_source/request_source_plugin.pb.h"
+
+#include "common/uri_impl.h"
+
+namespace Nighthawk {
+
+/**
+ * Sample Request Source implementation for comparison.
```
mum4k, line 18: We should improve these comments. There should be a single location (either here or in the proto file) that clearly explains what each plugin does.

In addition to that, each class comment ideally has all of the following:

- a one line summary of what the class does
- a longer context for the users of the class.
- an example use of the class.
- a statement of whether the class is or isn't thread safe.

Please update comments on all classes accordingly. The current wording on their comments isn't very informative for someone without context on request source.

mum4k (changes requested):
First round of comments, only high level at this point.

api/request_source/request_source_plugin.proto
wjuan-AFK: Understood. I hope that I've added some context for this.

api/request_source/request_source_plugin.proto
wjuan-AFK: Understood.

source/common/request_source_plugin_impl.h
wjuan-AFK: Understood. I have attempted to improve my class comments. Please let me know if they remain substandard.

source/common/BUILD
wjuan-AFK: Understood.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+// memory and replays them.
+// @param request_max The number of requests the requestGenerator produced by get() will generate. 0
+// means it is unlimited.
+// @param header the default header that will be overridden by values taken from the options_list,
+// any values not overridden will be used.
+// @param options_list A copy of the options_list will be loaded in memory. The RequestGenerator
+// produced by get() will use options from the options_list to overwrite values in the header, and
+// create new requests. if request_max is greater than the length of options_list, it will loop.
+// This is not thread safe.
+class RequestOptionsListRequestSource : public RequestSource {
```
mum4k, line 69: What is the practical use case for this request source implementation? Looking back, our plan was to implement the file based request source to serve as an example and a gRPC request source for backward compatibility.

Can you help me understand what do we need this one for?

EDIT: After looking at the .cc file, it looks like this request source implementation is used internally by the FileBased request source to do its job. If that it the plan, that is fine, but then this shouldn't be part of the API. The API (the header) should only contain the FileBased request source. The .cc file can have a section with anonymous namespace where we define the RequestOptionsListRequestSource internally.

Alternatively, if we plan to reuse this by multiple request sources, we should move it into its own library and say (in the header file) that this is for internal purposes only.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+  void initOnThread() override;
+
+private:
+  Envoy::Http::RequestHeaderMapPtr header_;
+  const std::unique_ptr<nighthawk::client::RequestOptionsList> options_list_;
+  std::vector<uint32_t> request_count_;
+  const uint32_t request_max_;
+};
+
+// Factory that creates a RequestOptionsListRequestSource from a FileBasedPluginConfig proto.
```
mum4k, line 88: Possible typo, looks like this factory creates a FileBasedRequestSourcePlugin rather than a RequestOptionsListRequestSource.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+// Registered as an Envoy plugin.
+// Implementation of RequestSourceConfigFactory which produces a RequestSource that keeps an
```
mum4k, line 90: This comment contains a mix of good class comments and implementation comments. We shouldn't comment or describe implementation here as it is bound to change and isn't part of the API contract. An example of implementation comments here are:

- the fact that the class keeps a RequestOptionsList in memory
- the fact that file_lock_ is used to protect the list.

On the other hand saying that the class is not thread safe is a very good comment here, but it doesn't need more details.

Please use your best judgment to adjust the comments so they talk about *what* a class does and how it can *be used*, but never about *how* it does its job.

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+
+  // Locking to avoid issues with multiple threads reading the same file.
+  {
+    Envoy::Thread::LockGuard lock_guard(file_lock_);
+    // Reading the file only the first time.
+    if (options_list_.options_size() == 0) {
+      util.loadFromFile(config.file_path(), options_list_,
+                        Envoy::ProtobufMessage::getStrictValidationVisitor(), *api, true);
+    }
+    temp_list->CopyFrom(options_list_);
```
mum4k, line 43: Since we load options_list_ only once, why do we need to make a copy here? Can't we give a const reference to all the created plugins instead?

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+    Envoy::Http::RequestHeaderMapPtr header) {
+  const auto& any = dynamic_cast<const Envoy::ProtobufWkt::Any&>(message);
+  nighthawk::request_source::FileBasedPluginConfig config;
+  Envoy::MessageUtil util;
+
+  util.unpackTo(any, config);
+  if (api->fileSystem().fileSize(config.file_path()) > config.max_file_size().value()) {
+    throw NighthawkException("file size must be less than max_file_size");
+  }
+  auto temp_list = std::make_unique<nighthawk::client::RequestOptionsList>();
```
mum4k, line 33: We should choose a better variable name here, "temp_list" doesn't actually communicate anything useful to the reader. All variables are essentially temporary and the fact that this is a list can be seen from the object type.

If we do keep this (see my comment below), let's try to come up with a more descriptive name. I can help by suggesting such a name once I understand why do we need this variable.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+  std::string name() const override;
+  // This returns an empty version of the expected FileBasedPluginConfig from
+  // request_source_plugin.proto
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+  // This is the primary method that is used to get RequestSources.
+  // This implementation is not thread safe. Only the first call to createRequestSourcePlugin will
+  // load the file from memory and subsequent calls just make a copy of the options_list that was
+  // already loaded. The FileBasedRequestSourcePluginConfigFactory will not work with multiple
+  // different files for this reason.
+  RequestSourcePtr createRequestSourcePlugin(const Envoy::Protobuf::Message& message,
```
mum4k, line 112: Looks like this method would throw if the file cannot be parsed correctly. Can we mention that in the comment?

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+                                             Envoy::Api::ApiPtr api,
+                                             Envoy::Http::RequestHeaderMapPtr header) override;
+};
+
+// This factory will be activated through RequestSourceFactory in factories.h
+DECLARE_FACTORY(DummyRequestSourcePluginConfigFactory);
+
+// Sample Request Source for small RequestOptionsLists. Loads a copy of the RequestOptionsList in
+// memory and replays them.
+// @param request_max The number of requests the requestGenerator produced by get() will generate. 0
```
mum4k, line 61: If I understand correctly what "request_max" specifies, a name like "total_requests" would probably be more descriptive.

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+    const uint32_t request_max, Envoy::Http::RequestHeaderMapPtr header,
+    std::unique_ptr<nighthawk::client::RequestOptionsList> options_list)
+    : header_(std::move(header)), options_list_(std::move(options_list)),
+      request_max_(request_max) {}
+
+RequestGenerator RequestOptionsListRequestSource::get() {
+  uint32_t counter = 0;
+  request_count_.push_back(counter);
+  uint32_t& lambda_counter = request_count_.back();
+  RequestGenerator request_generator = [this, lambda_counter]() mutable -> RequestPtr {
```
mum4k, line 61: Can you help me understand the need for the "mutable" keyword here?

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+REGISTER_FACTORY(FileBasedRequestSourcePluginConfigFactory, RequestSourcePluginConfigFactory);
+
+RequestOptionsListRequestSource::RequestOptionsListRequestSource(
+    const uint32_t request_max, Envoy::Http::RequestHeaderMapPtr header,
+    std::unique_ptr<nighthawk::client::RequestOptionsList> options_list)
+    : header_(std::move(header)), options_list_(std::move(options_list)),
+      request_max_(request_max) {}
+
+RequestGenerator RequestOptionsListRequestSource::get() {
+  uint32_t counter = 0;
```
mum4k, line 58: Do we need to define the counter locally as a variable? Or can we just push 0 on and take its address?

```cc
request_count_.push_back(0);
uint32_t& lambda_counter = request_count_.back();
```

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+  request_count_.push_back(counter);
+  uint32_t& lambda_counter = request_count_.back();
+  RequestGenerator request_generator = [this, lambda_counter]() mutable -> RequestPtr {
+    // if request_max is 0, then we never stop generating requests.
+    if (lambda_counter >= request_max_ && request_max_ != 0) {
+      return nullptr;
+    }
+
+    // Increment the counter and get the request_option from the list for the current iteration.
+    auto index = lambda_counter % options_list_->options_size();
```
mum4k, line 68: The "auto" keyword here actually hides a type that would be useful to see. Ideally we only use auto when the type is too long or redundant in the statement.

https://google.github.io/styleguide/cppguide.html#Type_deduction

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+    nighthawk::client::RequestOptions request_option = options_list_->options().at(index);
+    lambda_counter++;
```
mum4k, line 70: In c++, we should prefer pre-increment and pre-decrement:

https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+
+    // Initialize the header with the values from the default header.
+    Envoy::Http::RequestHeaderMapPtr header = Envoy::Http::RequestHeaderMapImpl::create();
+    Envoy::Http::HeaderMapImpl::copyFrom(*header, *header_);
```
mum4k, line 74: Since we are making a copy of the original header map anyway, why do we need to pass it in as a pointer to the plugin? Can't the plugin take the header as a const reference to make it clear on the API layer that the header isn't owned or modified?

Or is there a reason that forces us to give ownership?

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+
+    // Override the default values with the values from the request_option
+    header->setMethod(envoy::config::core::v3::RequestMethod_Name(request_option.request_method()));
+    const uint32_t content_length = request_option.request_body_size().value();
+    if (content_length > 0) {
+      header->setContentLength(content_length);
+    }
+    for (const auto& option_header : request_option.request_headers()) {
```
mum4k, line 82: This is another place where auto might be hiding useful type information.

mum4k (changes requested):
Thank you for improving the documentation. This is second review round, mostly on APIs and some implementation. I will complete the review of the implementation in the next round.

source/request_source/request_options_list_plugin_impl.cc
wjuan-AFK: I need this keyword because if I do not have mutable, the compiler will complain about the incrementing of the lambda_counter later in this lambda.

source/request_source/request_options_list_plugin_impl.cc
wjuan-AFK: I think that it is more consistent with the way that the other plugins do it, and it is possible that the plugin would want to modify the header. Generally it also seems like something that the RequestSource should own. In the factories_impl.cc where this RequestSource will be created, I believe the header passed in does not have a lifetime greater than that of the RequestSource, and that might matter also, though I'm not sure if that is correct.

source/request_source/request_source_plugin_impl.h
wjuan-AFK: Agreed.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+
+// Stub Request Source implementation for comparison.
+class DummyRequestSource : public RequestSource {
+public:
+  explicit DummyRequestSource(const nighthawk::request_source::StubPluginConfig& config);
+  // The generator function will only return empty headers.
+  // The function is threadsafe.
+  RequestGenerator get() override;
+
+  // default implementation
```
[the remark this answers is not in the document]
wjuan-AFK: Understood.

source/request_source/request_source_plugin_impl.h
```diff
@@ -0,0 +1,124 @@
...
+// means it is unlimited.
+// @param header the default header that will be overridden by values taken from the options_list,
+// any values not overridden will be used.
+// @param options_list A copy of the options_list will be loaded in memory. The RequestGenerator
+// produced by get() will use options from the options_list to overwrite values in the header, and
+// create new requests. if request_max is greater than the length of options_list, it will loop.
+// This is not thread safe.
+class RequestOptionsListRequestSource : public RequestSource {
+public:
+  explicit RequestOptionsListRequestSource(
```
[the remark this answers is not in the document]
wjuan-AFK: That makes sense. Thanks.

source/request_source/request_source_plugin_impl.h
wjuan-AFK: Not a typo. I wanted this to be a more generic version of the RequestOptionsListRequestSource since it seemed the more versatile option.

source/request_source/request_source_plugin_impl.h
wjuan-AFK: I felt like it made sense for both the RequestSource and the factory that makes the requestSource to live in the same file together, I can move it into its own library if you prefer.

source/request_source/request_source_plugin_impl.h
wjuan-AFK: Sure.

test/adaptive_load/adaptive_load_controller_test.cc
```diff
@@ -27,6 +27,10 @@
 #include "api/client/service.pb.h"
 #include "api/client/service_mock.grpc.pb.h"
 
+#include "gtest/gtest.h"
```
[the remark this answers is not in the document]
wjuan-AFK: Hopefully I've reverted the change that introduced these.

source/request_source/request_source_plugin_impl.h
wjuan-AFK: I'm not sure I'm doing a very good job at this, but I'll do my best.

test/request_source/stub_plugin_impl.h
```diff
@@ -0,0 +1,55 @@
+// Implementations of RequestSourceConfigFactory and the RequestSources that those factories make.
```
mum4k, line 1: We probably want to say something else here, i.e. we should talk about the stub plugin which is implemented in this file.

api/request_source/request_source_plugin.proto
```diff
@@ -0,0 +1,35 @@
...
+  // if it's too large it will throw an error.
+  google.protobuf.UInt32Value max_file_size = 3 [(validate.rules).uint32 = {lte: 1000000}];
+}
+
+// Configuration for StubPluginRequestSource (plugin name: "nighthawk.stub-request-source-plugin")
+// The plugin does nothing. This is for testing and comparison of the Request Source Plugin Factory
+// mechanism using a minimal version of plugin that does not require a more complicated proto or
+// file reading.
+message StubPluginConfig {
+  // test input value.
```
mum4k, line 33: We had a comment thread about improving this comment, but looks like the it wasn't actually done. Repeating the comment:


Can we improve this comment? We could say what the input argument is for / what does the plugin do with it. It raises a question as to why a plugin that does nothing need an input argument.

test/request_source/request_source_plugin_test.cc
```diff
@@ -0,0 +1,179 @@
...
+  }
+};
+
+TEST_F(StubRequestSourcePluginTest, CreateEmptyConfigProtoCreatesCorrectType) {
+  auto& config_factory =
+      Envoy::Config::Utility::getAndCheckFactoryByName<RequestSourcePluginConfigFactory>(
+          "nighthawk.stub-request-source-plugin");
+  const Envoy::ProtobufTypes::MessagePtr empty_config = config_factory.createEmptyConfigProto();
+  const nighthawk::request_source::StubPluginConfig expected_config;
+  EXPECT_EQ(empty_config->DebugString(), expected_config.DebugString());
```
mum4k, line 52: We shouldn't compare protocol buffers using their string representation, since that isn't their API, just a debug string for humans to read. It isn't guaranteed to capture all the important details, nor it is guaranteed to be stable.

The gmock test suite comes with ::testing::EqualsProto helper we should be able to utilize in these cases.

Something like:

EXPECT_THAT(empty_config, ::testing::EqualsProto(expected_config))

mum4k (changes requested):
Looks good, thank you for iterating on this. This is now a complete review and we only have few minor comments left to address.

test/request_source/request_source_plugin_test.cc
mum4k: Please ignore this comment, looks like `::testing::EqualsProto` isn't as of yet available publically:

google/googletest#1761

api/request_source/request_source_plugin.proto
wjuan-AFK: Oh I thought I fixed it by making it do something.

mum4k (approved):

PR #526, opened 2020-09-11
Adaptive load controller main loop

eric846:
Adaptive load controller object which contains the main loop:

- Validates the adaptive load session spec
- Performs adjusting stage:
  - Gets latest load value from StepController
  - Calls Nighthawk Service
  - Scores metrics
  - Checks StepController for convergence and doom
  - Checks for timeout
- Calls Nighthawk Service for one last testing stage benchmark with the latest load value
- Returns an output proto

Part 7 of splitting PR #483.


dubious90:
Before I send this out to review, can you break it up into smaller pieces if it's not too much work?  This PR is currently 25 files
You know the code better obviously, but in an initial scan, it looks to me like there are three things happening here that could be separate PRs:
1. The main Adaptive Load Loop
2. Mock Metrics Evaluator
3. Some cleanup work around declaring const/PURE and such
  a. This appears to be most of the files, but few of the lines, and splitting it out will likely make that review very easy, while reducing the overhead on the other reviews.

eric846:
> Before I send this out to review, can you break it up into smaller pieces if it's not too much work? This PR is currently 25 files
> You know the code better obviously, but in an initial scan, it looks to me like there are three things happening here that could be separate PRs:
> 
> 1. The main Adaptive Load Loop
> 2. Mock Metrics Evaluator
> 3. Some cleanup work around declaring const/PURE and such
>    a. This appears to be most of the files, but few of the lines, and splitting it out will likely make that review very easy, while reducing the overhead on the other reviews.

Will do, closing this one

PR #532, opened 2020-09-14
Enhance end-to-end test for POST requests with an entity body.

oschaaf:
Make the end-to-end test for POST handling cover all extensions.

Split out from #512 

### DIFF
--- a/test/integration/test_integration_basics.py
+++ b/test/integration/test_integration_basics.py
@@ -418,7 +418,7 @@ def test_cli_output_format(http_test_server_fixture):
     'filter_configs',
     ["{}", "{static_delay: \"0.01s\"}", "{emit_previous_request_delta_in_response_header: \"aa\"}"])
 def test_request_body_gets_transmitted(http_test_server_fixture, filter_configs):
-  """Test request body transmission with.
+  """Test request body transmission handling code for our extensions.
 
   Ensure that the number of bytes we request for the request body gets reflected in the upstream
   connection transmitted bytes counter for h1 and h2.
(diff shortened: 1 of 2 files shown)


test/integration/test_integration_basics.py
```diff
@@ -414,8 +414,11 @@ def test_cli_output_format(http_test_server_fixture):
   asserts.assertIn("Percentile", output)
 
 
-def test_request_body_gets_transmitted(http_test_server_fixture):
-  """Test request body transmission.
+@pytest.mark.parametrize(
+    'filter_configs',
+    ["{}", "{static_delay: \"0.01s\"}", "{emit_previous_request_delta_in_response_header: \"aa\"}"])
+def test_request_body_gets_transmitted(http_test_server_fixture, filter_configs):
+  """Test request body transmission with.
```
dubious90, line 421: I think this comment is incomplete?

test/integration/test_integration_basics.py
```diff
@@ -434,13 +437,17 @@ def check_upload_expectations(fixture, parsed_json, expected_transmitted_bytes,
         expected_received_bytes)
 
   upload_bytes = 1024 * 1024 * 3
+  # TODO(#531): The dynamic-delay extension hangs unless we lower the request entity body size.
+  if "static_delay" in filter_configs:
```
dubious90, line 441: How do you feel about combining this with the above line and making it a ternary, rather than an if condition?

My 2 cents is that as is, it's possible to scan this and not notice that upload_bytes has changed from its original value.

dubious90 (commented):
Two really small comments.

oschaaf:
@dubious90 thanks for the review; I applied corresponding changes in 467dd6b97d325153fd35fa7fd46b539a30e22b06

dubious90 (approved):
LGTM

PR #533, opened 2020-09-14
Add generic c++ integration test for uniform extension behavior. 

oschaaf:
Test basic behavioural properties of all test-server extensions. Any new extensions
may piggy-back on this by enlisting themselves. In a follow up we'll purge code
from the pre-existing tests that can now be deprecated.

Split out from #512

------

Prerequisite:

- [x] Land #517


mum4k:

oschaaf:
Can't reproduce the clang-tidy failure locally, and in CI it doesn't state a specific complaint.
/retest

oschaaf:
Yup, clang-tidy flaked in CI for this PR, which is unusual :-( Filed #546 to put in on the radar.

test/server/http_filter_base_test.cc
```diff
@@ -0,0 +1,89 @@
...
+                       testing::ValuesIn({true, false})));
+
+// Verify extensions are well-behaved when it comes to:
+// - GET requests.
+// - POST requests with an entity body (takes a slightly different code path).
+// - Valid but empty configuration.
+// - Bad request-level json configuration.
+// We will be low on functional expectations for the extensions, but this will catch hangs and
+// ensure that bad configuration handling is in-place.
+TEST_P(HttpFilterBaseIntegrationTest, BasicExtensionFlows) {
```
wjuan-AFK, line 46: Should we split these into separate tests? I'm not sure what the preference in the code base is, but it seems like each chunk is testing a different behaviour and might deserve a separate test?

test/server/http_filter_base_test.cc
```diff
@@ -0,0 +1,89 @@
...
+  EXPECT_TRUE(response->body().empty());
+
+  // Test with a valid but empty request-level configuration.
+  setRequestLevelConfiguration("{}");
+  response = getResponse(happy_flow_response_origin);
+  ASSERT_TRUE(response->complete());
+  EXPECT_EQ("200", response->headers().Status()->value().getStringView());
+  EXPECT_TRUE(response->body().empty());
+
+  const std::string kBadConfigErrorSentinel =
```
wjuan-AFK, line 72: Why not use status instead of doing this substring check?

oschaaf: We do check the http response code, but then also look for this substring in the response body because we know it should look like that if the extension under test is using the shared configuration handling code. Does that sound good?

test/server/http_filter_base_test.cc
oschaaf: I can see your point, and in general I agree that the clearer test names that arise out of more finely grained separate tests are better. 
But in this case, when splitting this, up we would have to duplicate the parameter unpacking for the TEST_P params across the newly split up tests and/or add facilities to help with that. 
This was also sort of pre-existing (as we can eliminate three different instances of similar code in the test code for three extensions).

So personally I lean towards the compact version as is is now.  With the context above, are you comfortable with leaving this as-is?

wjuan-AFK: Yeah, makes sense to me.

test/server/http_filter_base_test.cc
wjuan-AFK: Yes. That also makes sense.

wjuan-AFK (approved):

test/server/http_filter_base_test.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "server/http_test_server_filter.h"
+#include "server/http_time_tracking_filter.h"
+
+#include "test/server/http_filter_integration_test_base.h"
+
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+using namespace testing;
```
mum4k, line 11: Do we need this, since most symbols from the namespace are qualified? E.g. ::testing::Combine.

If you do prefer to shorten them, can we list them explicitly instead of subscribing to the entire testing namespace which *could* cause a conflict and confuse someone in the future?

E.g. for symbols that repeat a lot we can:

using ::testing::Combine;
using ::testing::ValuesIn;

test/server/http_filter_base_test.cc
mum4k, line 46: Adding to what @wjuan-AFK commented on - it would really help if we can separate these into individual test cases - one per behavior. Such approach simplifies future code changes, because it ensures that the tests won't be affected when a new behavior is added or that less tests are affected when a behavior is changed. Additionally when a behavior breaks, we get a clear signal which behavior broke, rather than a generic failure saying "the BasicExtensionFlows no longer works".

I see your point about duplicating the parameter unpacking. I wonder if we could place that into a separate helper function. If we figure out we can't, I think we should prefer more instances of smaller independent tests for individual behavior to shorter code that uses parametrised tests. In other words, if we need to drop the parametrised test approach and write few more test cases, I think that is a worthwhile price to pay for shorter and easier to maintain test cases.

Can we burn another cycle on this and see if we can find a win-win solution?

mum4k (changes requested):

test/server/http_filter_base_test.cc
oschaaf: I will take a look. Hopefully a convenient helper will suffice to support this, and then I can split these out into different tests. I'd like to avoid the need to drop parameterized testing here if we can: we would not only have to duplicate the test(s) for our three current extensions, but also require that for future extensions that would want to piggy back on this.

test/server/http_filter_base_test.cc
oschaaf: Done; added an anonymous namespace as well based on that it can only benefit us.

test/server/http_filter_base_test.cc
oschaaf: Done in 5ff6d569f61e314e4307aaf5285088d0ef014869.
@wjuan-AFK sorry, this was actually straight forward, and looking at the results, I rather like it better compared to what it was.

mum4k (approved):
Looks great, thank you for splitting them up.

Issue #422, opened 2020-07-26
Fortio output slightly different than the real fortio


howardjohn:
```
# cat /tmp/results/nighthawk |jq '.RetCodes'
{
  "200": "49988"
}
# cat /tmp/results/fortio  |jq '.RetCodes'
{
  "200": 50000
}
```

You can see in fortio the value is a number but in nighthawk a string

Versions:
```
FROM envoyproxy/nighthawk-dev:c3d7b475ddaafbe93b8223b84281463e3d3834a1 as nighthawk
FROM fortio/fortio:1.6.3 as fortio
```

Fortio command with `-json <path>`, nighthawk with `--output-format fortio`

oschaaf:
Nice catch; 
So that field was defined as follows in 68e59719fe754401907c16e53156bf18cd0e4e65:

```proto3
 map<string, uint64> RetCodes = 11;
```

That looks allright to me; so to figure out why the integer typed map value gets serialized as a string, 
I think we need to dig into what happens in `Envoy::MessageUtil::getJsonStringFromMessage()`:

```c++
std::string MessageUtil::getJsonStringFromMessage(const Protobuf::Message& message,
                                                  const bool pretty_print,
                                                  const bool always_print_primitive_fields) {
  Protobuf::util::JsonPrintOptions json_options;
  ....
  std::string json;
  const auto status = Protobuf::util::MessageToJsonString(message, &json, json_options);
   ....
  return json;
}
```

This in turn would lead up to investigating `Protobuf::util::MessageToJsonString()` / if we need
to set a JsonPrintOption to make this work.

oschaaf:
I wonder, there's `bool JsonPrintOptions::always_print_enums_as_ints` for `Protobuf::util::MessageToJsonString()`
By default they are rendered as strings.

I wonder if setting that would apply to values in maps too (or if it does exactly says what it does on the tin,
and stick to enums).
If it does turn out to fix our problem, we'd be lucky. If not, I think we need to propose an amendment to Protobuf to make
this happen.

howardjohn:
Interesting, I am surprised it behaves that way...

By the way, this isn't actually breaking me in anyway, I just happened to notice this. Should have mentioned that in the original issue.

oschaaf:
I was surprised too. It's probably tidier to serialize these as ints, though I think a lot of parser
might coerce the value into an int iff possible. But as far as I know that is not mandatory,  a strict parser would be right in
[...]

PR #534, opened 2020-09-14
Add experimental "pedantic" fortio output formatter.

### LINKED ISSUE #514: Nighthawk fortio's output is discrepant then real fortio's output
**Description**
Nighthawk's fortio transform output is different than real fortio output for a few fields.

#### Nighthawk's Output
```json
{
 "Labels": "",
 "RequestedQPS": 5,
 "ActualQPS": 4.9999995930000329,
 "ActualDuration": 5000000407,
 "NumThreads": 100,
 "DurationHistogram": {
  "Count": "25",
  "Data": [
   {
    "Start": 0.024088575,
    "End": 0.024088575,
    "Percent": 0,
    "Count": "1"
   },
```

#### Fortio's Output
```json
{
  "RunType": "HTTP",
  "Labels": "test",
  "StartTime": "2020-09-08T16:29:19.35507+05:30",
  "RequestedQPS": "100",
  "RequestedDuration": "5s",
  "ActualQPS": 59.9799880886016,
  "ActualDuration": 5101701580,
  "NumThreads": 8,
  "Version": "1.6.8",
  "DurationHistogram": {
    "Count": 306,
    "Min": 0.114543168,
    "Max": 0.436206912,
    "Sum": 40.422087638,
    "Avg": 0.1320983256143791,
    "StdDev": 0.03333137761133907,
    "Data": [
      {
        "Start": 0.114543168,
        "End": 0.2,
        "Percent": 97.38562091503267,
        "Count": 298
      },
```

As can be seen 
- RequestedQPS is an integer but should be a string.
- Dataentry.count, Sizes.count, HeaderSizes.count is a string but should be an integer.

**Expected Behaviour**

Nighthawk Output should be similar to Fortio's Output.

oschaaf:
Adds a post-processing step to massage the fortio output formatting
to more exactly reproduce Fortio's output, while leaving the
original output formatter intact for backwards compatibility purposes.

Fixes known issues #422 and #514. Experimental until we have
some confirmation that this can be marked as final / all issues have
been resolved.

### DIFF
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -402,13 +402,13 @@ FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& 
   // This should be OK as the regular expression we use can be trusted.
   // 1. We misdefined RequestRPS as an int, whereas Fortio outputs that as a string.
   res = std::regex_replace(res, std::regex(R"EOF("RequestedQPS"\: ([0-9]*))EOF"),
-                           "\"RequestedQPS\": \"$1\"");
+                           R"EOF("RequestedQPS": "$1")EOF");
   // 2. Our uint64's get serialized as json strings. Fortio outputs them as json integers.
   // NOTE: [0-9][0-9][0-9] looks for string fields referring to http status codes, which get
   // counted.
   res = std::regex_replace(
       res, std::regex(R"EOF("([0-9][0-9][0-9]|Count|BytesSent|BytesReceived)"\: "([0-9]*)")EOF"),
-      "\"$1\": $2");
+      R"EOF("$1": $2)EOF");
   // clang-format on
   return res;
 }
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -3,6 +3,7 @@
 #include <google/protobuf/util/time_util.h>
 
 #include <chrono>
+#include <regex>
 #include <sstream>
 
 #include "nighthawk/common/exception.h"
@@ -391,6 +392,26 @@ const nighthawk::client::DurationHistogram FortioOutputFormatterImpl::renderFort
                        }
                      });
   return fortio_histogram;
+}
+
+std::string
+FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
+  std::string res = FortioOutputFormatterImpl::formatProto(output);
+  // Fix two types of quirks. We disable linting because we use std::regex directly.
+  // This should be OK as the regular expression we use can be trusted.
+
+  // clang-format off
+  // 1. We misdefined RequestRPS as an int, whereas Fortio outputs that as a string.
+  res = std::regex_replace(res, std::regex(R"EOF("RequestedQPS"\: ([0-9]*))EOF"),
+                           "\"RequestedQPS\": \"$1\"");
+  // 2. Our uint64's get serialized as json strings. Fortio outputs them as json integers.
+  // NOTE: [0-9][0-9][0-9] looks for string fields referring to http status codes, which get
+  // counted.
+  res = std::regex_replace(
+      res, std::regex(R"EOF("([0-9][0-9][0-9]|Count|BytesSent|BytesReceived)"\: "([0-9]*)")EOF"),
+      "\"$1\": $2");
+  // clang-format on
+  return res;
 }
 
 } // namespace Client
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -400,7 +400,7 @@ FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& 
   // clang-format off
   // Fix two types of quirks. We disable linting because we use std::regex directly.
   // This should be OK as the regular expression we use can be trusted.
-  // 1. We misdefined RequestRPS as an int, whereas Fortio outputs that as a string.
+  // 1. We misdefined RequestedRPS as an int, whereas Fortio outputs that as a string.
   res = std::regex_replace(res, std::regex(R"EOF("RequestedQPS"\: ([0-9]*))EOF"),
                            R"EOF("RequestedQPS": "$1")EOF");
   // 2. Our uint64's get serialized as json strings. Fortio outputs them as json integers.
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -397,10 +397,9 @@ const nighthawk::client::DurationHistogram FortioOutputFormatterImpl::renderFort
 std::string
 FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
   std::string res = FortioOutputFormatterImpl::formatProto(output);
+  // clang-format off
   // Fix two types of quirks. We disable linting because we use std::regex directly.
   // This should be OK as the regular expression we use can be trusted.
-
-  // clang-format off
   // 1. We misdefined RequestRPS as an int, whereas Fortio outputs that as a string.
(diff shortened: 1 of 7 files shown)


dubious90:
@eric846 Please review and assign back to me once done.

kushthedude (dismissed):
This works for us!
/cc @leecalcote

oschaaf:
Also /cc @carolynhu 
this is pulled of in a backwards compatible manner, but when we upgrade, there'll
be a way to get rid of the workarounds for the fortio output more quirks in NH once this lands.

source/client/output_formatter_impl.cc
```diff
@@ -393,5 +394,24 @@ const nighthawk::client::DurationHistogram FortioOutputFormatterImpl::renderFort
   return fortio_histogram;
 }
 
+std::string
+FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
+  std::string res = FortioOutputFormatterImpl::formatProto(output);
+  // clang-format off
+  // Fix two types of quirks. We disable linting because we use std::regex directly.
+  // This should be OK as the regular expression we use can be trusted.
+  // 1. We misdefined RequestRPS as an int, whereas Fortio outputs that as a string.
```
carolynhu, line 403: typo here? `RequestedQPS`

test/test_data/output_formatter.medium.fortio-noquirks.gold
```diff
@@ -0,0 +1,483 @@
...
+    "End": 0.310886399,
+    "Percent": 100,
+    "Count": 0
+   }
+  ],
+  "Min": 0.053798911,
+  "Max": 0.310886399,
+  "Sum": 4.156954618,
+  "Avg": 0.078433106,
+  "StdDev": 0.05052302,
```
carolynhu, line 201: Could we move these fields (line 197 to line 201) up to under line 8 "Count" field as what fortio has?

test/test_data/output_formatter.medium.fortio-noquirks.gold
```diff
@@ -0,0 +1,483 @@
...
+    "Percentile": 80,
+    "Value": 47331
+   },
+   {
+    "Percentile": 90,
+    "Value": 47339
+   },
+   {
+    "Percentile": 95,
+    "Value": 47353
```
carolynhu, line 475: are these value fake or real? as this is in seconds...

leecalcote:
@oschaaf thank you for today's call. Very much enjoyed it.

We can assist in verifying compatibility. Are the build artifacts of this PR (the container image) available?

test/test_data/output_formatter.medium.fortio-noquirks.gold
oschaaf: That is quite hard to pull off in the current approach. Looking at it, the proto json serialiser we use here alphabetically sorts this. I think it can do that, because after parsing the difference usually doesn't matter? Do you see a way that the re-ordering could cause trouble?

test/test_data/output_formatter.medium.fortio-noquirks.gold
oschaaf: I didn't come up with this test data, it's a copy of a pre-existing one modified to reflect the changes in here. If I interpret the file correctly, we're not looking at seconds, but rather a histogram of byte counts for the header?

source/client/output_formatter_impl.cc
oschaaf: Fix in 5ab48900

oschaaf:
> @oschaaf thank you for today's call. Very much enjoyed it.
> 
> We can assist in verifying compatibility. Are the build artifacts of this PR (the container image) available?

@leecalcote 
Unfortunately, I think fresh images will only be pushed post-merge of this. 
So to try this out one needs to manually build the docker image. 
Having said that, it is ok to iterate on this while we have the feature marked as --experimental, as we don't
need to be backwards compatible while that is the case. So waiting until after the merge is viable too.

test/test_data/output_formatter.medium.fortio-noquirks.gold
carolynhu: Not yet. Just noticed the order is mismatch with fortio. Don't think it will cause any trouble.

eric846 (dismissed):
LGTM

source/client/output_formatter_impl.cc
```diff
@@ -393,5 +394,24 @@ const nighthawk::client::DurationHistogram FortioOutputFormatterImpl::renderFort
...
 }
 
+std::string
+FortioPedanticOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
+  std::string res = FortioOutputFormatterImpl::formatProto(output);
+  // clang-format off
+  // Fix two types of quirks. We disable linting because we use std::regex directly.
+  // This should be OK as the regular expression we use can be trusted.
+  // 1. We misdefined RequestedRPS as an int, whereas Fortio outputs that as a string.
+  res = std::regex_replace(res, std::regex(R"EOF("RequestedQPS"\: ([0-9]*))EOF"),
```
mum4k, line 404: (optional) Would you mind including (in the comment) and example of the matched string? That usually helps when trying to understand regexes in code. This one is fairly trivial, so optional, but we could use one for the second regex.

source/client/output_formatter_impl.h
```diff
@@ -108,5 +108,10 @@ class FortioOutputFormatterImpl : public OutputFormatterImpl {
   double durationToSeconds(const Envoy::ProtobufWkt::Duration& duration) const;
 };
 
+class FortioPedanticOutputFormatterImpl : public FortioOutputFormatterImpl {
```
mum4k, line 111: Can we add a comment on the class roughly explaining what it is for?

test/output_formatter_test.cc
```diff
@@ -224,5 +225,12 @@ TEST_F(StatidToNameTest, TestTranslations) {
   }
 }
 
+TEST_F(MediumOutputCollectorTest, FortioPedanticFormatter) {
+  const auto input_proto = loadProtoFromFile("test/test_data/output_formatter.medium.proto.gold");
```
mum4k, line 229: The "auto" here doesn't seem to hide redundant type information, so it might be better for the reader to spell it out.

test/output_formatter_test.cc
```diff
@@ -224,5 +225,12 @@ TEST_F(StatidToNameTest, TestTranslations) {
...
+  FortioPedanticOutputFormatterImpl formatter;
+  expectEqualToGoldFile(formatter.formatProto(input_proto),
```
mum4k, line 231: This golden file is fairly long which makes it hard for the reader to figure out which portions actually are relevant to this test and which are just noise. Any way we could cut it down to the minimum required to test this new behavior?

mum4k (changes requested):

source/client/output_formatter_impl.cc
oschaaf: Yeah, I always felt that regular expressions are kind of a write-only concept :-)
Done in 2df8c2be

source/client/output_formatter_impl.h
oschaaf: Done in 2df8c2be

test/output_formatter_test.cc
oschaaf: Done here and in other spots in 2df8c2be

test/output_formatter_test.cc
oschaaf: This makes sense, but is kind of pre-existing; I think the initial work for this just grabbed a raw arbitrary snapshot and included it for testing the origin fortio output formatter code. 
We are now more or less forced to propagate that, as our new gold file is an amended version of the original one. 
I punted this to #542 to track doing this, do that work for you?

mum4k: That works, thank you.

mum4k (approved):

PR #535, opened 2020-09-14
Adaptive Load Controller main loop

### LINKED ISSUE #485: Consider making Adaptive Load Controller main API a class interface
Currently the API is a single bare function.

There might be opportunities to use most of the logic but swap out Nighthawk Service for another option.

eric846:
This is the main function of the Adaptive Load Controller library:

- Check the input proto for errors
- Apply default input values
- Adjusting Stage loop:
  - Get the latest dynamically generated CommandLineOptions from the StepController
  - Run a short benchmark with the Nighthawk Service
  - Obtain values from MetricsPlugins
  - Score metrics using ScoringFunction plugins
  - Report scores back to the StepController, which recalculates the load for the next iteration
  - Check for convergence deadline exceeded
  - Check for StepController convergence
  - Check for StepController doom
- Testing Stage: Run one long benchmark on the Nighthawk Service at the converged load

Fixes #485.

Part 9 of splitting PR #483.


source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,195 @@
...
+#include "absl/status/status.h"
+#include "absl/strings/str_join.h"
+#include "adaptive_load/metrics_plugin_impl.h"
+#include "adaptive_load/plugin_loader.h"
+
+namespace Nighthawk {
+
+namespace {
+
+using namespace Envoy; // for Envoy::Logger
```
oschaaf, line 34: Do we still have to do this? I was hoping that with https://github.com/envoyproxy/envoy/pull/12827 and https://github.com/envoyproxy/envoy/pull/12706 the need for this is gone.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,195 @@
...
+ * @return Map from MetricsPlugin names to initialized plugins, to be used in the course of a single
+ * adaptive load session based on the session spec.
+ */
+absl::flat_hash_map<std::string, MetricsPluginPtr>
+LoadMetricsPlugins(const AdaptiveLoadSessionSpec& spec) {
+  absl::flat_hash_map<std::string, MetricsPluginPtr> name_to_custom_metrics_plugin_map;
+  for (const envoy::config::core::v3::TypedExtensionConfig& config :
+       spec.metrics_plugin_configs()) {
+    absl::StatusOr<MetricsPluginPtr> metrics_plugin_or = LoadMetricsPlugin(config);
+    RELEASE_ASSERT(
```
oschaaf, line 59: unactionable, but I love these assertations. they're very helpful to the reader, as they serve to document the layering and what's safe to assume at this point, as well as help ensuring this stays water tight.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,195 @@
...
+  absl::StatusOr<nighthawk::client::CommandLineOptions> command_line_options_or =
+      step_controller.GetCurrentCommandLineOptions();
+  if (!command_line_options_or.ok()) {
+    ENVOY_LOG_MISC(error, "Error constructing Nighthawk input: {}: {}",
+                   command_line_options_or.status().code(),
+                   command_line_options_or.status().message());
+    return command_line_options_or.status();
+  }
+  nighthawk::client::CommandLineOptions command_line_options = command_line_options_or.value();
+  *command_line_options.mutable_duration() = std::move(duration);
```
oschaaf, line 112: it still feels a little bit off to me that we're specializing the way we treat the test duration. maybe this is a good place to add a comment about why we need that?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,195 @@
...
+        nighthawk_service_stub, spec, name_to_custom_metrics_plugin_map, *step_controller,
+        spec.measuring_period());
+    if (!result_or.ok()) {
+      return result_or.status();
+    }
+    BenchmarkResult result = result_or.value();
+    *output.mutable_adjusting_stage_results()->Add() = result;
+    step_controller->UpdateAndRecompute(result);
+
+    if (std::chrono::duration_cast<std::chrono::seconds>(time_source_.monotonicTime() - start_time)
```
oschaaf, line 172: is there a reason to enforce 1 second granularity?

oschaaf (changes requested):
Generally LGTM, it's quite cool to see everything come together. I feel it reads like poetry :-) except for the special way we treat duration, which still feels a tiny bit off; maybe a code level comment can help here. Somehow I'm also wondering if duration isn't something we should allow implementing as a parameter, like qps. E.g. to allow one to research a question like "how does test duration influence convergence speed at a steady rps". 
Left a few small other remarks.

dubious90:
@pamorgan Please review and assign back to me once done.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Ahh I had not been following that. Removed `using namespace Envoy;` and it still works.

> Generally LGTM, it's quite cool to see everything come together. I feel it reads like poetry :-) except for the special way we treat duration, which still feels a tiny bit off; maybe a code level comment can help here. Somehow I'm also wondering if duration isn't something we should allow implementing as a parameter, like qps. E.g. to allow one to research a question like "how does test duration influence convergence speed at a steady rps".
> Left a few small other remarks.

Shall I compare thee to a summer's *INSERT DURATION HERE*? ;-)

Opened #539 to discuss this. I agree duration sticks out like a sore thumb in this API, and it may be possible to generalize it with minimal cost.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Added some comments for now -- see issue #539 for a potential refactoring.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Nope -- changed this now to extract the duration in the standard way.

source/adaptive_load/adaptive_load_controller_impl.cc
oschaaf: Haha nice one :-) Thanks for adding the comment & filing that issue. Resolving.

oschaaf (dismissed):
lgtm

include/nighthawk/adaptive_load/adaptive_load_controller.h
```diff
@@ -10,26 +10,32 @@
...
-absl::StatusOr<nighthawk::adaptive_load::AdaptiveLoadSessionOutput> PerformAdaptiveLoadSession(
-    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
-    const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec, Envoy::TimeSource& time_source);
+class AdaptiveLoadController {
+public:
+  virtual ~AdaptiveLoadController() = default;
+  /**
+   * Performs an adaptive load session consisting of the Adjusting Stage and the
+   * Testing Stage. Adjusting Stage: Runs a series of short benchmarks, checks metrics according to
+   * MetricSpecs, and adjusts load up or down based on the result; returns an error if convergence
```
pamorgan, line 22: nit: extra quote.

include/nighthawk/adaptive_load/adaptive_load_controller.h
```diff
@@ -10,26 +10,32 @@
...
+   * is not detected before the deadline in the spec. Load adjustments and convergence detection are
+   * computed by a StepController plugin. Metric values are obtained through MetricsPlugins. Testing
+   * Stage: When the optimal load is found, runs one long benchmark to validate it.
+   *
+   * @param nighthawk_service_stub A Nighthawk Service gRPC stub.
```
pamorgan, line 27: nit: Maybe stub is more appropriate as a member in the class than an argument to the method. I don't think we expect it to change on a multiple call scenario.

include/nighthawk/adaptive_load/adaptive_load_controller.h
```diff
@@ -10,26 +10,32 @@
...
- *
- * @return StatusOr<AdaptiveLoadSessionOutput> A proto logging the result of all traffic attempted
- * and all corresponding metric values and scores, or an overall error status if the session failed.
+ * Contains the main loop of the adaptive load controller. Consults a StepController for load
+ * decisions, interacts with Nighthawk Service and MetricsPlugins.
  */
-absl::StatusOr<nighthawk::adaptive_load::AdaptiveLoadSessionOutput> PerformAdaptiveLoadSession(
-    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
-    const nighthawk::adaptive_load::AdaptiveLoadSessionSpec& spec, Envoy::TimeSource& time_source);
+class AdaptiveLoadController {
```
pamorgan, line 16: should this be renamed to interface now, since we now a pure methods.

source/adaptive_load/adaptive_load_controller_impl.h
```diff
@@ -0,0 +1,80 @@
...
+   *       MetricsEvaluatorImpl(),
+   *       AdaptiveLoadSessionSpecProtoHelperImpl(),
+   *       Envoy::Event::RealTimeSystem()); // NO_CHECK_FORMAT(real_time))
+   *   AdaptiveLoadSessionSpec spec;
+   *   // (set spec fields here)
+   *   StatusOr<AdaptiveLoadSessionOutput> output =
+   *       controller.PerformAdaptiveLoadSession(&nighthawk_service_stub, spec);
+   *
+   * @param nighthawk_service_client A helper that executes Nighthawk Service benchmarks given a
+   * gRPC stub.
```
pamorgan, line 33: What is the reason behing seperating the client from the stub?

source/adaptive_load/adaptive_load_controller_impl.h
```diff
@@ -0,0 +1,80 @@
...
+
+class AdaptiveLoadControllerImpl : public AdaptiveLoadController {
+public:
+  /**
+   * Constructs an implementation of the adaptive load controller main loop that relies on logic in
+   * several helper objects. Through helpers, it performs Nighthawk Service benchmarks, obtains
+   * metrics from MetricsPlugins, scores the results, and consults a StepController plugin to
+   * determine the next load and detect convergence. All plugins are specified through the
+   * AdaptiveLoadSessionSpec proto.
+   *
```
pamorgan, line 19: can you add a comment on the thread safety? Can you run concurrent sessions?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,198 @@
...
+  return benchmark_result;
+}
+
+absl::StatusOr<AdaptiveLoadSessionOutput> AdaptiveLoadControllerImpl::PerformAdaptiveLoadSession(
+    nighthawk::client::NighthawkService::StubInterface* nighthawk_service_stub,
+    const AdaptiveLoadSessionSpec& input_spec) {
+  AdaptiveLoadSessionSpec spec = session_spec_proto_helper_.SetSessionSpecDefaults(input_spec);
+  absl::Status validation_status = session_spec_proto_helper_.CheckSessionSpec(spec);
+  if (!validation_status.ok()) {
+    return validation_status;
```
pamorgan, line 147: Should we log reasons of failures.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,198 @@
...
+          "Failed to converge before deadline of %.2f seconds.", time_limit_ns.count() / 1e9));
+    }
+  } while (!step_controller->IsConverged() && !step_controller->IsDoomed(doom_reason));
+
+  if (step_controller->IsDoomed(doom_reason)) {
+    return absl::AbortedError(
+        absl::StrCat("Step controller determined that it can never converge: ", doom_reason));
+  }
+
+  absl::StatusOr<BenchmarkResult> result_or = PerformAndAnalyzeNighthawkBenchmark(
```
pamorgan, line 188: Can you comment on why calling this again after complete.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,198 @@
...
+
+    const std::chrono::nanoseconds time_limit_ns(
+        Envoy::Protobuf::util::TimeUtil::DurationToNanoseconds(spec.convergence_deadline()));
+    const auto elapsed_time_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+        time_source_.monotonicTime() - start_time);
+    if (elapsed_time_ns > time_limit_ns) {
+      return absl::DeadlineExceededError(absl::StrFormat(
+          "Failed to converge before deadline of %.2f seconds.", time_limit_ns.count() / 1e9));
+    }
+  } while (!step_controller->IsConverged() && !step_controller->IsDoomed(doom_reason));
```
pamorgan, line 181: Usually while loops are more readable than do/while. Is there a reason not to use a simple while loop here?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,198 @@
...
+  do {
+    absl::StatusOr<BenchmarkResult> result_or = PerformAndAnalyzeNighthawkBenchmark(
+        nighthawk_service_stub, spec, name_to_custom_metrics_plugin_map, *step_controller,
+        spec.measuring_period());
+    if (!result_or.ok()) {
+      return result_or.status();
+    }
+    BenchmarkResult result = result_or.value();
+    *output.mutable_adjusting_stage_results()->Add() = result;
+    step_controller->UpdateAndRecompute(result);
```
pamorgan, line 171: It took me a second to follow the logic because step_controller->GetCurrentCommandLineOptions and step_controller->UpdateAndRecompute(result); are called in 2 different methods. 
I think it would be easier to follow if you moved step_controller->GetCurrentCommandLineOptions to this method.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * @return Map from MetricsPlugin names to initialized plugins, to be used in the course of a single
+ * adaptive load session based on the session spec.
+ */
+absl::flat_hash_map<std::string, MetricsPluginPtr>
+LoadMetricsPlugins(const AdaptiveLoadSessionSpec& spec) {
+  absl::flat_hash_map<std::string, MetricsPluginPtr> name_to_custom_metrics_plugin_map;
+  for (const envoy::config::core::v3::TypedExtensionConfig& config :
+       spec.metrics_plugin_configs()) {
+    absl::StatusOr<MetricsPluginPtr> metrics_plugin_or = LoadMetricsPlugin(config);
```
pamorgan, line 57: nit: usually it's better to perform the load right after validations and pass in the metric plugins to avoid the assert. No action needed if it's too much of code refactoring.

pamorgan (approved):
some nits but LGTM.

include/nighthawk/adaptive_load/adaptive_load_controller.h
eric846: The convention for Nighthawk seems to be that everything under `include` should be pure but no existing interfaces have `Interface` in their name.

include/nighthawk/adaptive_load/adaptive_load_controller.h
eric846: Reformatted this

include/nighthawk/adaptive_load/adaptive_load_controller.h
eric846: This class is intended to be stateless helpers and it's only a class to make mocking easier. We will have a chance to design a convenient client when we merge `RemoteProcessImpl` and `NighthawkServiceClientImpl`.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Agreed, it's a bit awkward for historical reasons -- filed #547

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Just moved UpdateAndRecompute into the helper, which hopefully makes the same amount of sense. I had moved GetCurrentCommandLineOptions to the helper to reduce duplication in the main loop.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: We had discussed it, I think the motivation for do/while is that `IsConverged` and `IsDoomed` don't have meaningful answers until the loop body has executed.

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: Done (this is the testing stage)

source/adaptive_load/adaptive_load_controller_impl.h
eric846: This is for historical reasons: We had a helper that made it easier to call the stub, then moved the helper into its own class for testing purposes.

I'm not sure what will happen with this, but we may add a more convenient Nighthawk Service client wrapper object that contains the stub, and then take this wrapper instead. Or the stub might be the best thing for the public API to take.

source/adaptive_load/adaptive_load_controller_impl.h
eric846: Done

source/adaptive_load/adaptive_load_controller_impl.cc
eric846: BTW we have #539 to consider making the testing stage an optional thing, decided by the StepController, so this special call might go away.

mum4k (approved):
Looks great, it is obvious that a lot of work was done here to improve readability. This PR only contains the business logic of the adaptive controller. The main functions and test cases are easy to follow. Thank you for taking all the additional steps.

PR #537, opened 2020-09-15
Test benchmark docker image build in CI

oschaaf:
Now that we have a benchmark docker image build, it would be
good to make sure that keeps working in CI.

This also makes it easy to start pushing this image in CI in a
follow-up.


dubious90 (approved):
LGTM

abaptiste:
LGTM.  Thank you!

PR #540, opened 2020-09-17
Switch dynamic delay and time-tracking to use FilterConfigurationBase.

oschaaf:
Second step in unifying configuration handling across extensions:
make these two extensions inherit from and use FilterConfigurationBase.
Fix POST handling.

Split out from #512

### DIFF
--- a/source/server/BUILD
+++ b/source/server/BUILD
@@ -20,11 +20,20 @@ envoy_cc_library(
 
 envoy_cc_library(
     name = "configuration_lib",
-    srcs = ["configuration.cc"],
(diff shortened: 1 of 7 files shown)


mum4k:

qqustc (approved):
LGTM, want to make sure that there is no unit test need to be added/modified?

Assigned back to @mum4k .

oschaaf:
> LGTM, want to make sure that there is no unit test need to be added/modified?

There is pre-existing testing in place, this should be ok in that regard

mum4k (approved):

PR #544, opened 2020-09-17
Avoid crash with the experimental fortio pedantic formatter

### LINKED ISSUE #543: Fortio Output Transform: panic not reached error when using experimental_fortio_pedantic format
#### Description

After updating the nighthawk image, when I tried to run test with experimental_fortio_pedantic format, I encounter following error
```sh
[21:45:38.894151][1][C] panic: not reached
```
**CLI Input**
```sh
docker run envoyproxy/nighthawk-dev:latest nighthawk_client --rps 10 --concurrency 1 --duration 5 https://www.github.com --output-format experimental_fortio_pedantic
```

The following input runs well while using other output formats.

oschaaf:
I forgot the last step, which is to actually teach the output
formatter factory when to construct the new formatter. Sorry!

- Amends a pre-existing test to avoid regression when adding new
output formats in the future
- Fix the issue, add the new proto enum option to the switch
statement.

Fixes #543


mum4k (approved):
Thank you for adding the test.

PR #550, opened 2020-09-23
Update overview doc to include user-specified logging

jiajunye:


docs/root/overview.md
```diff
@@ -189,3 +189,9 @@ other formats (e.g. human, fortio). It can be very useful to always store the
 json output format, yet be able to easily get to one of the other output
 formats. It’s like having the cake and eating it too!
 
+## User-specified Nighthawk logging
+
+Users can use their own Nighthawk logging options when calling **ENVOY_LOG**.
+Basically it would require two steps: 1) Create a logging sink delegate
+inherited from [Envoy SinkDelegate](https://github.com/envoyproxy/envoy/blob/master/source/common/common/logger.h). 2) Constructor a ServiceImpl object with
```
oschaaf, line 196: Looking at [how this gets rendered](https://github.com/envoyproxy/nighthawk/blob/f8dfdd790655e6da01146b005ad7d5989d8dd34f/docs/root/overview.md), I feel specifying a markdown list here may enhance readability. E.g.

```
1. Create a logging sink delegate
2. ....
3. ....
```

oschaaf (approved):
one nit, otherwise LGTM

docs/root/overview.md
jiajunye: Thanks Otto! Please see this [updated version](https://github.com/envoyproxy/nighthawk/blob/701ab40f99e70c2e6443f6eafef4fc4bb0b507b9/docs/root/overview.md).

oschaaf (dismissed):

docs/root/overview.md
```diff
@@ -189,3 +189,9 @@ other formats (e.g. human, fortio). It can be very useful to always store the
 json output format, yet be able to easily get to one of the other output
 formats. It’s like having the cake and eating it too!
 
+## User-specified Nighthawk logging
+
+Users can use their own Nighthawk logging options when calling **ENVOY_LOG**.
```
mum4k, line 194: We may want to update the first sentence, since users don't really call ENVOY_LOG directly. It also isn't clear what we mean by "setting their own options". Let's try to write this from the user's perspective, how about something like this:

Users of Nighthawk can specify custom format and destination for all messages logged by Nighthawk. Nighthawk utilizes the Envoy's logging mechanism by performing all logging via the **ENVOY_LOG** macro. To customize this mechanism, users need to perform two steps:
...

mum4k (changes requested):

docs/root/overview.md
jiajunye: This sounds better to me indeed. Thanks for the suggestion!

docs/root/overview.md
```diff
@@ -189,3 +189,11 @@ other formats (e.g. human, fortio). It can be very useful to always store the
...
 formats. It’s like having the cake and eating it too!
 
+## User-specified Nighthawk logging
+
+Users of Nighthawk can specify custom format and destination (logging sink
+delegate) for all Nighthawk logging messages. Nighthawk utilizes the Envoy's
+logging mechanism by performing all logging via the **ENVOY_LOG** macro. To
+customize this mechanism, users need to perform two steps:
+1. Create a logging sink delegate inherited from [Envoy SinkDelegate](https://github.com/envoyproxy/envoy/blob/master/source/common/common/logger.h).
+2. Constructor a ServiceImpl object with an [Envoy Logger Context](https://github.com/envoyproxy/envoy/blob/master/source/common/common/logger.h) which contains user-specified log level and format.
```
mum4k, line 199: Possible typo: "Construct" ? instead of "Constructor".

mum4k (changes requested):
Looks good, just one nit.

docs/root/overview.md
jiajunye: Oops.. amended.

mum4k (approved):

PR #552, opened 2020-09-25
Fix jitter behavior for large values.

oschaaf:
By chance weird behaviour was observed for when specifying large jitter
values. This warranted analysis, and upon investigation the root problem seems
to be that when configuring `--uniform-jitter` to a value that would allow the
adjusted value(s) to overlap with the pacing of the underlying rate limiter,
the new overlapping value could overwrite the old one that was scheduled.

This modifies behaviour and tests to fix and consolidate, by modifying 
`DistributionSamplingRateLimiter` so that it can queue adjusted release
timings instead of storing a single one. This will allow it to accumulate multiple
release timings that are adjusted for jitter. 
This ensures that the pacing of the rate limiter that modifies timings to add jitter
gets properly disconnected from the pacing of the underlying rate limiter.

NOTE: There are no known consumers running into problems / inaccuracies
because of this. To tease the problem out, either one needs to:
- configure `--jitter-uniform` in a way that it would adjust timings to overlap
with the frequency imposed by `--rps`, or
- configure any combination of `--jitter-uniform > 0s` and `--burst-size > 0`.

### DIFF
--- a/source/common/rate_limiter_impl.cc
+++ b/source/common/rate_limiter_impl.cc
@@ -156,14 +156,17 @@ DelegatingRateLimiterImpl::DelegatingRateLimiterImpl(
       random_distribution_generator_(std::move(random_distribution_generator)) {}
 
 bool DelegatingRateLimiterImpl::tryAcquireOne() {
-  if (distributed_start_ == absl::nullopt) {
-    if (rate_limiter_->tryAcquireOne()) {
-      distributed_start_ = timeSource().monotonicTime() + random_distribution_generator_();
-    }
+  const Envoy::MonotonicTime now = timeSource().monotonicTime();
+  sanity_check_pending_release_ = false;
+  if (rate_limiter_->tryAcquireOne()) {
+    const Envoy::MonotonicTime adjusted = now + random_distribution_generator_();
+    distributed_timings_.insert(
+        std::lower_bound(distributed_timings_.begin(), distributed_timings_.end(), adjusted),
+        adjusted);
   }
 
-  if (distributed_start_ != absl::nullopt && distributed_start_ <= timeSource().monotonicTime()) {
-    distributed_start_ = absl::nullopt;
+  if (!distributed_timings_.empty() && distributed_timings_.front() <= now) {
+    distributed_timings_.erase(distributed_timings_.begin());
     return true;
   }
 
@@ -171,7 +174,9 @@ bool DelegatingRateLimiterImpl::tryAcquireOne() {
 }
 
 void DelegatingRateLimiterImpl::releaseOne() {
-  distributed_start_ = absl::nullopt;
+  RELEASE_ASSERT(!sanity_check_pending_release_,
+                 "unexpected call to DelegatingRateLimiterImpl::releaseOne()");
+  sanity_check_pending_release_ = true;
   rate_limiter_->releaseOne();
 }
 
--- a/source/common/rate_limiter_impl.cc
+++ b/source/common/rate_limiter_impl.cc
@@ -157,16 +157,18 @@ DelegatingRateLimiterImpl::DelegatingRateLimiterImpl(
 
 bool DelegatingRateLimiterImpl::tryAcquireOne() {
   const Envoy::MonotonicTime now = timeSource().monotonicTime();
-  sanity_check_pending_release_ = false;
   if (rate_limiter_->tryAcquireOne()) {
     const Envoy::MonotonicTime adjusted = now + random_distribution_generator_();
+    // We track a sorted list of timings, where the one at the front is the one that should
+    // be applied the soonest.
     distributed_timings_.insert(
-        std::lower_bound(distributed_timings_.begin(), distributed_timings_.end(), adjusted),
+        std::upper_bound(distributed_timings_.begin(), distributed_timings_.end(), adjusted),
         adjusted);
   }
 
   if (!distributed_timings_.empty() && distributed_timings_.front() <= now) {
-    distributed_timings_.erase(distributed_timings_.begin());
+    distributed_timings_.pop_front();
+    sanity_check_pending_release_ = false;
     return true;
   }
 
--- a/source/common/rate_limiter_impl.h
+++ b/source/common/rate_limiter_impl.h
@@ -1,7 +1,7 @@
 #pragma once
 
+#include <list>
 #include <random>
-#include <vector>
 
 #include "envoy/common/time.h"
 
@@ -157,7 +157,8 @@ protected:
   const RateLimiterDelegate random_distribution_generator_;
 
 private:
-  std::vector<Envoy::MonotonicTime> distributed_timings_;
+  std::list<Envoy::MonotonicTime> distributed_timings_;
+  // Used to enforce that releaseOne() is always paired with a successfull tryAcquireOne().
   bool sanity_check_pending_release_{true};
 };
 
--- a/source/common/rate_limiter_impl.h
+++ b/source/common/rate_limiter_impl.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <random>
+#include <vector>
 
 #include "envoy/common/time.h"
 
@@ -156,7 +157,8 @@ protected:
   const RateLimiterDelegate random_distribution_generator_;
 
 private:
-  absl::optional<Envoy::MonotonicTime> distributed_start_;
+  std::vector<Envoy::MonotonicTime> distributed_timings_;
+  bool sanity_check_pending_release_{true};
 };
 
 class UniformRandomDistributionSamplerImpl : public DiscreteNumericDistributionSampler {
--- a/test/rate_limiter_test.cc
+++ b/test/rate_limiter_test.cc
@@ -301,8 +301,8 @@ TEST_F(DistributionSamplingRateLimiterTest, QueuedAcquisitionCorrectReleaseOrder
   // will buffer those that indicate an offset > 0. Zero-valued offsets ought to be released
   // immediately.
   std::vector<uint64_t> acquisition_timings;
-  for (uint64_t k = 0; k < input_acquisition_timings_ms.size(); k++) {
-    if (input_acquisition_timings_ms[k] == 0) {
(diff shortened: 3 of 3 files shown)


mum4k:

source/common/rate_limiter_impl.cc
```diff
@@ -156,22 +156,27 @@ DelegatingRateLimiterImpl::DelegatingRateLimiterImpl(
...
 bool DelegatingRateLimiterImpl::tryAcquireOne() {
-  if (distributed_start_ == absl::nullopt) {
-    if (rate_limiter_->tryAcquireOne()) {
-      distributed_start_ = timeSource().monotonicTime() + random_distribution_generator_();
-    }
+  const Envoy::MonotonicTime now = timeSource().monotonicTime();
+  sanity_check_pending_release_ = false;
+  if (rate_limiter_->tryAcquireOne()) {
+    const Envoy::MonotonicTime adjusted = now + random_distribution_generator_();
+    distributed_timings_.insert(
```
eric846, line 163: Could you add a comment about the invariant we maintain on `distributed_timings_`? It looks like it's simply ascending order. Also, is it easy to add unit tests that insert multiple distribution values other than 1ns in sorted and nonsorted order to make sure each acquisition always happens at the expected nanosecond?

Also, is `distributed_timings_` ever anticipated to become large? If so, would a tree be better than a vector?

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
-  EXPECT_CALL(unsafe_mock_rate_limiter, tryAcquireOne)
-      .Times(AtLeast(1))
-      .WillRepeatedly(Return(true));
-  EXPECT_CALL(unsafe_mock_rate_limiter, releaseOne).Times(1);
-  EXPECT_CALL(*unsafe_discrete_numeric_distribution_sampler, getValue)
-      .Times(3)
+  std::unique_ptr<NiceMock<MockRateLimiter>> mock_rate_limiter_;
+  MockRateLimiter& unsafe_mock_rate_limiter_;
+  Envoy::Event::SimulatedTimeSystem time_system_;
+  MockDiscreteNumericDistributionSampler* unsafe_discrete_numeric_distribution_sampler_;
```
eric846, line 215: Could you comment what operations are unsafe on this field?

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
   // The distribution first yields a 1 ns offset. So we don't expect to be green lighted.
-  EXPECT_FALSE(rate_limiter->tryAcquireOne());
-  time_system.advanceTimeWait(1ns);
-  EXPECT_TRUE(rate_limiter->tryAcquireOne());
-  // We expect releaseOne to be propagated.
-  rate_limiter->releaseOne();
-  // The distribution will yield an offset of 0ns, we expect success.
-  EXPECT_TRUE(rate_limiter->tryAcquireOne());
+  EXPECT_FALSE(rate_limiter_->tryAcquireOne());
+  EXPECT_FALSE(rate_limiter_->tryAcquireOne());
```
eric846, line 244: I think the test makes sense, but could you add multiple comments to clarify the scenario, which seems to be:
1. The inner rate limiter releases 2 requests only (the true, true, false, false, false)
2. The 2 requests are delayed until 1ns has elapsed (the 1, 1)
3. After the 2 delayed requests are released, no further requests are released because the underlying rate limiter released nothing further

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
-      .WillRepeatedly(ReturnRef(time_system));
+// may arbitrarily delay that. We test that principle with tests that use this fixture, which
+// sets up a distribution sampling rate limiter instance to encapsulate a mock rate limiter,
+// relying on simulated time and a mock discrete numberic distribution sampler.
+class DistributionSamplingRateLimiterTest : public RateLimiterTest {
+public:
+  DistributionSamplingRateLimiterTest()
+      : mock_rate_limiter_(std::make_unique<NiceMock<MockRateLimiter>>()),
+        unsafe_mock_rate_limiter_(*mock_rate_limiter_),
+        unsafe_discrete_numeric_distribution_sampler_(new MockDiscreteNumericDistributionSampler()),
```
eric846, line 202: Is there a special need for bare `new`, or would making the field `std::unique_ptr<MockDiscreteNumericDistributionSampler> discrete_numeric_distribution_sampler_` and referring to `*discrete_numeric_distribution_sampler_` in the tests work just the same? This seems like it like it would get flagged by asan.

source/common/rate_limiter_impl.h
```diff
@@ -156,7 +157,8 @@ class DelegatingRateLimiterImpl : public ForwardingRateLimiterImpl,
   const RateLimiterDelegate random_distribution_generator_;
 
 private:
-  absl::optional<Envoy::MonotonicTime> distributed_start_;
+  std::vector<Envoy::MonotonicTime> distributed_timings_;
+  bool sanity_check_pending_release_{true};
```
eric846, line 161: Could you add a comment that explains what this protects us from?

eric846 (changes requested):

source/common/rate_limiter_impl.cc
oschaaf: > Also, is `distributed_timings_` ever anticipated to become large? If so, would a tree be better than a vector?

Yes I think you are right, but I feel it would become more prudent to optimise this if we find use cases that need to support specifying large ranges of values to distribute timings over. In that case, this could list could grow large. Then we should optimise, but for now I'm happy with correctness, and I'd rather add a documentation note recommending to not configure large values at this point in time.

(For the known use case where people add white noise for desynchronisation purposes, I wouldn't expect this to be a factor, as the distribution range should usually be smaller then the request-release interval).

test/rate_limiter_test.cc
oschaaf: This ends up being owned by the rate limiter, so there shouldn't be an ASAN report about it. 
Annoyingly, we need temporary member variables to grab references to the mock before we transfer ownership, as the constructor requires a unique_ptr&&. This works nice in prod code by making the path of least resistance a good one, but it is annoying here as we need to go through this hoop. I renamed some things, and made this code use std::move like we do with the rate limiter mocks, to hopefully clean this up and make it more easy to grok.

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
+        rate_limiter_(std::make_unique<DistributionSamplingRateLimiterImpl>(
+            std::unique_ptr<DiscreteNumericDistributionSampler>(
+                unsafe_discrete_numeric_distribution_sampler_),
+            std::move(mock_rate_limiter_))) {
+    EXPECT_CALL(unsafe_mock_rate_limiter_, timeSource)
+        .Times(AtLeast(0))
```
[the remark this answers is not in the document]
oschaaf: done

source/common/rate_limiter_impl.cc
oschaaf: > Also, is it easy to add unit tests that insert multiple distribution values other than 1ns in sorted and nonsorted order to make sure each acquisition always happens at the expected nanosecond?

Done.

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
-  EXPECT_CALL(unsafe_mock_rate_limiter, timeSource)
-      .Times(AtLeast(1))
-      .WillRepeatedly(ReturnRef(time_system));
+// may arbitrarily delay that. We test that principle with tests that use this fixture, which
+// sets up a distribution sampling rate limiter instance to encapsulate a mock rate limiter,
+// relying on simulated time and a mock discrete numberic distribution sampler.
+class DistributionSamplingRateLimiterTest : public RateLimiterTest {
+public:
+  DistributionSamplingRateLimiterTest()
+      : mock_rate_limiter_(std::make_unique<NiceMock<MockRateLimiter>>()),
```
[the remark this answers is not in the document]
oschaaf: Done

test/rate_limiter_test.cc
```diff
@@ -191,43 +191,80 @@ TEST_F(RateLimiterTest, DistributionSamplingRateLimiterImplTest) {
...
+  MockDiscreteNumericDistributionSampler* unsafe_discrete_numeric_distribution_sampler_;
+  RateLimiterPtr rate_limiter_;
+};
+
+TEST_F(DistributionSamplingRateLimiterTest, SingleAcquisition) {
+  EXPECT_CALL(unsafe_mock_rate_limiter_, tryAcquireOne)
+      .WillOnce(Return(true))
+      .WillOnce(Return(false))
+      .WillOnce(Return(false));
+  EXPECT_CALL(*unsafe_discrete_numeric_distribution_sampler_, getValue).WillOnce(Return(1));
```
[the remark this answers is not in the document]
oschaaf: Done

test/rate_limiter_test.cc
oschaaf: Done

oschaaf:
@eric846 thanks for the review, I pushed 131d2e0 to address everything modulo optimisation of the release-timings buffering via a tree.  I file https://github.com/envoyproxy/nighthawk/issues/555 to track that.

oschaaf:
@eric846 one last thing -- while I didn't implement a tree, the buffered timings are now tracked via a `std::list` instead of the `std::vector`. I haven't measured it yet, but I think that the list may be a better match here, with the sorting/insert and pop from the front that is being done.

eric846:
LGTM, thanks!

eric846 (approved):

mum4k (approved):

PR #560, opened 2020-10-07
Adding proto only Request Source Config factory

wjuan-AFK:
This is a more usable version of the Request Source Config Factory similar to the FIle Based Config Factory which is easier to pass through the CLI.


eric846 (changes requested):

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -45,6 +45,36 @@ RequestSourcePtr OptionsListFromFileRequestSourceFactory::createRequestSourcePlu
...
+    Envoy::Http::RequestHeaderMapPtr header) {
+  const auto& any = dynamic_cast<const Envoy::ProtobufWkt::Any&>(message);
+  nighthawk::request_source::InLinePluginConfig config;
+  Envoy::MessageUtil::unpackTo(any, config);
+  // Locking to avoid issues with multiple threads calling this at the same time and trying to set
+  // the options_list_
+  {
+    Envoy::Thread::LockGuard lock_guard(options_list_lock_);
+    // Only loading the config into memory the first time.
+    if (options_list_.options_size() == 0) {
```
oschaaf, line 68: Suggestion: I think that relying on `absl::optional` here to express if options_list_ was populated slightly simplifies things when considering what happens in the edge case where the passed message actually is empty (I think right now this will repeatedly set options_list_. I see no harm in it, but it seems different from the regular flow).

test/request_source/request_source_plugin_test.cc
```diff
@@ -175,5 +188,96 @@ TEST_F(FileBasedRequestSourcePluginTest,
...
+      Envoy::Config::Utility::getAndCheckFactoryByName<RequestSourcePluginConfigFactory>(
+          "nighthawk.in-line-options-list-request-source-plugin");
+  auto header = Envoy::Http::RequestHeaderMapImpl::create();
+  RequestSourcePtr plugin =
+      config_factory.createRequestSourcePlugin(config_any, *api_, std::move(header));
+  Nighthawk::RequestGenerator generator = plugin->get();
+  Nighthawk::RequestPtr request = generator();
+  Nighthawk::RequestPtr request2 = generator();
+  Nighthawk::RequestPtr request3 = generator();
+  Nighthawk::HeaderMapPtr header1 = request->header();
```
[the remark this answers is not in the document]
wjuan-AFK: Makes sense.

api/request_source/request_source_plugin.proto
```diff
@@ -7,7 +7,7 @@ import "google/protobuf/wrappers.proto";
 import "validate/validate.proto";
 import "api/client/options.proto";
 
-// Configuration for FileBasedPluginRequestSource (plugin name:
+// Configuration for OptionsListFromFileRequestSourceFactory (plugin name:
```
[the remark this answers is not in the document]
wjuan-AFK: I think that this naming better expresses that they're both making an optionsListRequestSource.

source/request_source/request_options_list_plugin_impl.cc
```diff
@@ -45,6 +45,36 @@ RequestSourcePtr OptionsListFromFileRequestSourceFactory::createRequestSourcePlu
 
 REGISTER_FACTORY(OptionsListFromFileRequestSourceFactory, RequestSourcePluginConfigFactory);
 
+std::string OptionsListFromProtoRequestSourceFactory::name() const {
```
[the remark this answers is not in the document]
wjuan-AFK: I think that this naming better expresses that they're both making an optionsListRequestSource. I would be open to renaming the configProto, but I'm not sure what a better name would be.

test/request_source/request_source_plugin_test.cc
```diff
@@ -175,5 +188,96 @@ TEST_F(FileBasedRequestSourcePluginTest,
...
+  EXPECT_EQ(config_factory.name(), "nighthawk.in-line-options-list-request-source-plugin");
+}
+
+TEST_F(InLineRequestSourcePluginTest, CreateRequestSourcePluginCreatesCorrectPluginType) {
+  Envoy::MessageUtil util;
+  nighthawk::client::RequestOptionsList options_list;
+  util.loadFromFile(TestEnvironment::runfilesPath("test/request_source/test_data/test-config.yaml"),
+                    options_list, Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_,
+                    true);
+  nighthawk::request_source::InLinePluginConfig config = MakeInLinePluginConfig(options_list, 2);
```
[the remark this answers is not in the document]
wjuan-AFK: Understood.

test/request_source/request_source_plugin_test.cc
```diff
@@ -175,5 +188,96 @@ TEST_F(FileBasedRequestSourcePluginTest,
...
+      config_factory.createRequestSourcePlugin(config_any, *api_, std::move(header));
+  EXPECT_NE(dynamic_cast<RequestOptionsListRequestSource*>(plugin.get()), nullptr);
+}
+TEST_F(InLineRequestSourcePluginTest,
+       CreateRequestSourcePluginGetsWorkingRequestGeneratorThatEndsAtNumRequest) {
+  Envoy::MessageUtil util;
+  nighthawk::client::RequestOptionsList options_list;
+  util.loadFromFile(TestEnvironment::runfilesPath("test/request_source/test_data/test-config.yaml"),
+                    options_list, Envoy::ProtobufMessage::getStrictValidationVisitor(), *api_,
+                    true);
```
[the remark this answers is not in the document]
wjuan-AFK: Sure thing. Added clarifying comments throughout.

api/request_source/request_source_plugin.proto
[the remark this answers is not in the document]
eric846: I mean isn't this a typo? This appears to be the comment above `FileBasedPluginConfig`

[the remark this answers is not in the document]
eric846: OHHH I didn't see `File` in the middle of the name, never mind!

eric846 (changes requested):

source/request_source/request_options_list_plugin_impl.cc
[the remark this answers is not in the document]
wjuan-AFK: I agree with this. I think this is sufficiently concise while also adding clarification.

eric846 (dismissed):
LGTM

api/request_source/request_source_plugin.proto
```diff
@@ -7,30 +7,47 @@ import "google/protobuf/wrappers.proto";
...
+}
+
 // Configuration for StubPluginRequestSource (plugin name: "nighthawk.stub-request-source-plugin")
 // The plugin does nothing. This is for testing and comparison of the Request Source Plugin Factory
 // mechanism using a minimal version of plugin that does not require a more complicated proto or
 // file reading.
 message StubPluginConfig {
   // test input value which is the only output value in the headers produced from the
-  // requestGenerator for the StubRequestSource.
+  // requestGenerator for the StubRequestSource. This field is optional with a default of 0.
```
mum4k, line 51: (nit) Consider removing the comment addition here, since it falls into the a comment category that just "describes the language". It is obvious from the definition below that the field is optional and what its default value is.

The style guide gives some advice on how to write good comments:
https://google.github.io/styleguide/cppguide.html#Comments

A good advice to follow is to avoid writing comments about obvious things, or comments that just describe the language syntax. We should assume that the reader of the comments is familiar with the syntax and just provide comments on tricky parts or logic.

source/request_source/request_options_list_plugin_impl.h
```diff
@@ -45,40 +44,69 @@ class RequestOptionsListRequestSource : public RequestSource {
...
+//         "nighthawk.in-line-options-list-request-source-plugin");
+// RequestSourcePtr plugin =
+//     config_factory.createRequestSourcePlugin(config, std::move(api), std::move(header));
+
+class InLineOptionsListRequestSourceFactory : public virtual RequestSourcePluginConfigFactory {
+public:
+  std::string name() const override;
+  Envoy::ProtobufTypes::MessagePtr createEmptyConfigProto() override;
+
+  // This implementation is not thread safe. There is a race condition because only the first call
```
mum4k, line 97: I believe we have discussed this in one of the previous PRs. We shouldn't explain why it isn't thread safe as that is an implementation comment. Also the comment whether a class is or isn't thread safe belongs at class level, not method level.

Let's remove all this and instead add a single sentence at the end of the class docstring:

```cc
// This class is not thread safe.
class InLineOptionsListRequestSourceFactory : public virtual  {...}
```

Edit after looking at the current implementation - the class actually may be thread safe, because it uses a lock around this call, so that doesn't seem to be a race. However let's first discuss what we need as noted in the other comment.

source/request_source/request_options_list_plugin_impl.h
```diff
@@ -45,40 +44,69 @@ class RequestOptionsListRequestSource : public RequestSource {
...
+  // to createRequestSourcePlugin will load the options list into memory and subsequent calls just
+  // make a copy of the options_list that was already loaded.
+  RequestSourcePtr createRequestSourcePlugin(const Envoy::Protobuf::Message& message,
+                                             Envoy::Api::Api& api,
+                                             Envoy::Http::RequestHeaderMapPtr header) override;
+
+private:
+  Envoy::Thread::MutexBasicLockable options_list_lock_;
+  absl::optional<nighthawk::client::RequestOptionsList> options_list_;
```
mum4k, line 106: Does this field need to be optional? When would we create an instance of InLineOptionsListRequestSourceFactory without an options_list_?

source/request_source/request_options_list_plugin_impl.h
```diff
@@ -45,40 +44,69 @@ class RequestOptionsListRequestSource : public RequestSource {
...
+
+  // This implementation is not thread safe. There is a race condition because only the first call
+  // to createRequestSourcePlugin will load the options list into memory and subsequent calls just
+  // make a copy of the options_list that was already loaded.
+  RequestSourcePtr createRequestSourcePlugin(const Envoy::Protobuf::Message& message,
+                                             Envoy::Api::Api& api,
+                                             Envoy::Http::RequestHeaderMapPtr header) override;
+
+private:
+  Envoy::Thread::MutexBasicLockable options_list_lock_;
```
mum4k, line 105: If we are saying the class is not thread safe, why do we need a lock? Once we proclaim it thread-unsafe, nobody should be calling it from multiple threads.

If we do include a lock, we should use it to its intended purpose which is to make the class thread safe. Maybe I am misunderstanding it, but looks like we are in a hybrid state where we do pay the cost of locking for no benefit.

Taking a step back, we should probably first agree whether we need this class to be thread safe. If you think we do, can you help by describing a scenario where this needs to be accessed by multiple threads? Once we understand that, we can discuss how to design it to be compatible with the intended use.

mum4k (changes requested):
Partial review.

api/request_source/request_source_plugin.proto
```diff
@@ -7,30 +7,47 @@ import "google/protobuf/wrappers.proto";
...
-  // RequestOptionList located in the file at file_path, it will loop. num_requests = 0 means no
-  // limit on the number of requests to be produced.
-  google.protobuf.UInt32Value num_requests = 2 [(validate.rules).uint32 = {gte: 0, lte: 1000000}];
+  // RequestOptionList located in the file at file_path, it will loop. num_requests = 0 means it
+  // will loop indefinitely, though it will still terminate by normal mechanisms. This field is
+  // optional with a default of 0.
+  uint32 num_requests = 2;
   // The pluginfactory will load the file located in file_path as long as it is below max_file_size,
-  // if it's too large it will throw an error.
+  // if it's too large it will throw an error. This field is optional with a default of 1000000.
```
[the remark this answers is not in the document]
wjuan-AFK: done.

api/request_source/request_source_plugin.proto
wjuan-AFK: Done.

source/request_source/request_options_list_plugin_impl.h
wjuan-AFK: I was trying to add the clarification because of a discussion I had with Eric on this CL. I think the issue may be that I shouldn't be storing the value in the factory itself. I'm a little uncertain on this one. Essentially the problem that has resulted is that by storing the value in the factory itself, although in practice we will always get the same proto passed in from the CLI to each of the workers and thus to each call of the RequestSourcePluginConfigFactory, that technically creates a condition wherein, if multiple calls are made to the same factory with different protos, only the first one is acknowledged. In practice, as I say, this doesn't matter, but in theory this behaviour seems like some kind of race condition.

source/request_source/request_options_list_plugin_impl.h
wjuan-AFK: I made this optional to indicate the state before and after having saved the options_list_.

source/request_source/request_options_list_plugin_impl.h
mum4k: Thank you for the clarification, sounds like all we need to do is adjust the terminology. This class actually is thread-safe. Definition of race condition is clear - a code is racy when there are multiple threads accessing the same memory and at least one of them writes. Such situation isn't possible in this code, since we are using a lock. To address this - let's update the comment at the class level and say "This class is thread-safe".

As for the interesting behavior you are describing, for now we should just document it. Can you update the documentation so that it doesn't refer to a race condition and instead mentions that this function has a specific behavior on the first call? We can choose to remove the premature optimization in a later PR.

wjuan-AFK: Understood.

source/request_source/request_options_list_plugin_impl.h
wjuan-AFK: It should be thread safe. Sorry for the confusion.

mum4k (commented):
Looks good, can you please make sure the automated checks pass?

mum4k (approved):

Issue #546, opened 2020-09-18
clang-tidy flake


oschaaf:
In https://github.com/envoyproxy/nighthawk/pull/533 I observed a first occurrency of a non-reproducible clang-tidy flake.
Since this CI task has proven to be quite stable over time, it's worth tracking further occurrences of this here.

If this turns out problematic, the first potential suspect I would investigate is the last Envoy update, which at this time is 8fcf416d1e31bf1f6f2e8688c86f24495aef87b7

oschaaf:
So personally, I'm observing about 30% of these clang-tidy CI jobs to fail. One case of a timeout observed, where clang-tidy seemed to hang, and a few other cases where clang-tidy wrote a diff, but there was no trace of an actual complaint in the logs. I'll ask around on Slack if this is a known issue for envoy proxy in general, as I suspect that this crept in via an Envoy update, or an envoy build image update (which we usually update at the same time).

oschaaf:
The last merge to master also ran into this, the logs ended with:
```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

oschaaf:
The latest merge on master (61996df8b68d71bade3fb1ac8b94ee4f48bc0e27) needed 4 tries before succeeding the clang-tidy CI check.

oschaaf:
In https://github.com/envoyproxy/nighthawk/pull/566 this seems to happen more or less structurally. I can't spot an actual problem, and neither does clang_tidy report an actual problem. But somehow the CI task keeps exiting with an exit code that indicates a problem.

PR #561, opened 2020-10-12
Update Envoy & HdrHistogram_c

oschaaf:
- Update Envoy to 2097fe908f2abb718757dbd4087d793c861d7c5a
- Update HdrHistogram_c to 0.11.2

Adds `--define tcmalloc=gperftools` to the opt build we use to push images,
to (continue to) allow cpu profiling. We need this because of https://github.com/envoyproxy/envoy/commit/4c0d2d2a36996f2ec7d00b519ea90068629a2c55

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -27,7 +27,8 @@ build --host_javabase=@bazel_tools//tools/jdk:remote_jdk11
 build --javabase=@bazel_tools//tools/jdk:remote_jdk11
 build --enable_platform_specific_config
 
-# Enable position independent code, this option is not supported on Windows and default on on macOS.
+# Enable position independent code (this is the default on macOS and Windows)
+# (Workaround for https://github.com/bazelbuild/rules_foreign_cc/issues/421)
 build:linux --copt=-fPIC
 build:linux --cxxopt=-std=c++17
 build:linux --conlyopt=-fexceptions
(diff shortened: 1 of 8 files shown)


dubious90 (approved):

dubious90:
Can you resolve the clang tidy failure, and then re-apply the waiting-for-review label?

oschaaf:
@dubious90 a128bfe addresses the clang tidy issue here. In this PR, that task started to structurally fail without showing an actual complaint (it started flaking earlier, which is tracked in #546).
I copied over changes from Envoy's version of the clang tidy check script to our own, and after doing so an actual complaint emerged that I addressed. As a side effect, I suspect #546 will also be resolved upon merging this.

dubious90 (approved):

oschaaf:
@dubious90 sorry - with dd0e19834dd117b2769f5420794adbf761f1e580, which resolves a last quotation issue this now passes. I'd like to retry a few times to see if this also deflakes the clang-tidy task in CI, I'll ping you on Slack once I know more about that.

oschaaf:
@dubious90 unfortunately, this doesn't resolve the clang-tidy flakes, though it is an enhancement and now passes again some of the time instead of structurally failing. resolving #546 needs more looking into.

dubious90 (approved):

PR #562, opened 2020-10-16
Update Envoy to f95f5391b0b8683081ec786ea946026594955fc6

oschaaf:
Notable changes:

- Handle the new `Envoy::Http::StreamResetReason:: ConnectError` & amend related test
- Handle that `Envoy::Http::HeaderMap::get()` now returns a `Envoy::Http::HeaderMap::GetResult`.
- Squelch Envoy's format check error that scans for urls in `bazel/repositories.bzl`

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -1,7 +1,7 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-ENVOY_COMMIT = "2097fe908f2abb718757dbd4087d793c861d7c5a"  # October 12th, 2020
-ENVOY_SHA = "323360544ee355f0eddea742b31a80a94899090db1d64029cd22880083b311c0"
+ENVOY_COMMIT = "9dce187eca9aa10e98c0ae81368c078384528801"  # October 17th, 2020
+ENVOY_SHA = "fa5a452d52475cfc64c0557c4258e1a91fdceb725131aaa02956e83523b23269"
 
 HDR_HISTOGRAM_C_VERSION = "0.11.2"  # October 12th, 2020
 HDR_HISTOGRAM_C_SHA = "637f28b5f64de2e268131e4e34e6eef0b91cf5ff99167db447d9b2825eae6bad"
@@ -11,7 +11,9 @@ def nighthawk_dependencies():
         name = "envoy",
         sha256 = ENVOY_SHA,
         strip_prefix = "envoy-%s" % ENVOY_COMMIT,
+        # // clang-format off: Envoy's format check: Only repository_locations.bzl may contains URL references
         url = "https://github.com/envoyproxy/envoy/archive/%s.tar.gz" % ENVOY_COMMIT,
+        # // clang-format on
     )
     http_archive(
         name = "dep_hdrhistogram_c",
@@ -50,5 +52,7 @@ cc_library(
   """,
         sha256 = HDR_HISTOGRAM_C_SHA,
         strip_prefix = "HdrHistogram_c-%s" % HDR_HISTOGRAM_C_VERSION,
+        # // clang-format off
         url = "https://github.com/HdrHistogram/HdrHistogram_c/archive/%s.tar.gz" % HDR_HISTOGRAM_C_VERSION,
+        # // clang-format on
     )
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -1,7 +1,7 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-ENVOY_COMMIT = "9dce187eca9aa10e98c0ae81368c078384528801"  # October 17th, 2020
-ENVOY_SHA = "fa5a452d52475cfc64c0557c4258e1a91fdceb725131aaa02956e83523b23269"
+ENVOY_COMMIT = "f95f5391b0b8683081ec786ea946026594955fc6"  # October 21st, 2020
+ENVOY_SHA = "1129dcb0e18ec79ab56f59cbe5150f564e32d9221edd7ba1bd84b9b5377cbe35"
 
 HDR_HISTOGRAM_C_VERSION = "0.11.2"  # October 12th, 2020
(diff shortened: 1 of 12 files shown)


dubious90:
@jiajunye can you please review this and assign back to me once done?  Also let me know if you have any questions.

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -44,16 +44,15 @@ TEST_P(HttpTimeTrackingIntegrationTest, ReturnsPositiveLatencyForStaticConfigura
...
+      response->headers().get(Envoy::Http::LowerCaseString(kLatencyResponseHeaderName)).size(), 0);
 
   // On the second request we should observe a delta.
   response = getResponse(ResponseOrigin::UPSTREAM);
-  const Envoy::Http::HeaderEntry* latency_header_2 =
+  const Envoy::Http::HeaderMap::GetResult& latency_header =
       response->headers().get(Envoy::Http::LowerCaseString(kLatencyResponseHeaderName));
-  ASSERT_NE(latency_header_2, nullptr);
-  EXPECT_TRUE(absl::SimpleAtoi(latency_header_2->value().getStringView(), &latency));
+  EXPECT_EQ(latency_header.size(), 1);
```
jiajunye, line 54: Is ASSERT_EQ better?

jiajunye (dismissed):
Just a nit.

test/server/http_time_tracking_filter_integration_test.cc
oschaaf: Sure is, thanks. Done here and one other spot in 49106b690747dae7111bbac611aa623ce1e75299

jiajunye:
LGTM

bazel/repositories.bzl
```diff
@@ -11,7 +11,9 @@ def nighthawk_dependencies():
         name = "envoy",
         sha256 = ENVOY_SHA,
         strip_prefix = "envoy-%s" % ENVOY_COMMIT,
+        # // clang-format off: Envoy's format check: Only repository_locations.bzl may contains URL references
```
dubious90, line 14: This seems to be the first instance where us just using envoy's .bazelrc isn't great for us.  No action required here on this PR, but we should keep an eye on it.

source/client/stream_decoder.cc
```diff
@@ -21,9 +21,10 @@ void StreamDecoder::decodeHeaders(Envoy::Http::ResponseHeaderMapPtr&& headers, b
   stream_info_.response_code_ = static_cast<uint32_t>(response_code);
   if (!latency_response_header_name_.empty()) {
     const auto timing_header_name = Envoy::Http::LowerCaseString(latency_response_header_name_);
-    const Envoy::Http::HeaderEntry* timing_header = response_headers_->get(timing_header_name);
-    if (timing_header != nullptr) {
-      absl::string_view timing_value = timing_header->value().getStringView();
+    const Envoy::Http::HeaderMap::GetResult& timing_header =
+        response_headers_->get(timing_header_name);
+    if (!timing_header.empty()) {
+      absl::string_view timing_value = timing_header[0]->value().getStringView();
```
dubious90, line 27: This could possibly use a comment or maybe even an assertion of sorts.  It looks like the envoy function was changed to now allow multiple HeaderEntries to be returned rather than one?

I gather that our use case would always expect only one - would it be reasonable to add a failure here if we received more than one?

source/server/http_filter_config_base.cc
```diff
@@ -13,11 +13,11 @@ FilterConfigurationBase::FilterConfigurationBase(
 
 void FilterConfigurationBase::computeEffectiveConfiguration(
     const Envoy::Http::RequestHeaderMap& headers) {
-  const auto* request_config_header = headers.get(TestServer::HeaderNames::get().TestServerConfig);
-  if (request_config_header) {
+  const auto& request_config_header = headers.get(TestServer::HeaderNames::get().TestServerConfig);
+  if (request_config_header.size() == 1) {
```
dubious90, line 17: should we have an else condition here now?  again, taking into consideration receiving more than 1

dubious90 (commented):
Mostly looks good.  Just two nits.

source/client/stream_decoder.cc
oschaaf: I added a sparse warning log line in dc0b618951bde50cc9eed7143fe39d0f7d433521 and also testing for that case to make sure we don't track timings when there's ambiguity because of multiple header/value pairs.

source/server/http_filter_config_base.cc
oschaaf: Yes, you're right. Done in dc0b618951bde50cc9eed7143fe39d0f7d433521 + explicit tests for the case.

test/server/http_filter_base_test.cc
```diff
@@ -90,5 +90,29 @@ TEST_P(HttpFilterBaseIntegrationTest, EmptyRequestLevelConfigurationShouldFail)
...
+  appendRequestLevelConfiguration("{}");
+  Envoy::IntegrationStreamDecoderPtr response = getResponse(ResponseOrigin::EXTENSION);
+  ASSERT_TRUE(response->complete());
+  EXPECT_THAT(response->body(),
+              HasSubstr("Received multiple configuration headers in the request"));
+}
+
+TEST_P(HttpFilterBaseIntegrationTest, SingleValidPlusEmptyConfigurationHeadersFails) {
+  // Make sure we fail when both a valid configuration plus an empty configuration header is send.
+  // Note that we could be more flexible and look for the first request header that has a value,
```
dubious90, line 105: This note seems like it belongs more in the function itself rather than in this test.

Also, I might encourage a rephrase of why our approach is reasonable.  Without understanding of a real use case for them, we are assuming that any existence of duplicate headers here is in error.

oschaaf: Amended this; also updated the Envoy dep to f95f5391b0b8683081ec786ea946026594955fc6 in here, as that didn't need require any changes on our side modulo updating the revision/sha in `.bazelrc`.

oschaaf: (changes associated to your review comment applied in 6136e81, the Envoy sha change is in 30d7f20)

dubious90 (approved):

Issue #433, opened 2020-08-03
Add safe proto diff helper


eric846:
Need to do this separately after test/adaptive_load/utility.h goes in.

The helper should do:
  EXPECT_EQ(message->DebugString(), expected_config.DebugString());
  EXPECT_TRUE(Envoy::MessageUtil()(*message, expected_config));

oschaaf:
I learned that with proto3 default values may not get serialised. I'm not 100% sure that equivalence here guarantees actual equivalence with respect to set/not-set/default-value. Therefore, to make sure no edge cases slip past, maybe it makes sense to also double down with MessageUtil()(message1, message1). This won't produce a helpful test message, so I'd put it after comparing the debug strings (which should produce a nice diff), maybe in a helper method.

eric846:
Actually this isn't specific to adaptive_load, could go in an existing file.

eric846:
Actually I don't see

eric846:
This may be clearer than the construct with MessageUtil():

```
  EXPECT_TRUE(::Envoy::Protobuf::util::MessageDifferencer::Equivalent(*empty_config, expected_config));
```

oschaaf:
Oh that's much nicer indeed

oschaaf:
In the context of this issue, also making a note of a helpful review comment by @mum4k [1].
TIL there's ` ::testing::EqualsProto`
Sounds like we may want to use that as a replacement where applicable?

[1] https://github.com/envoyproxy/nighthawk/pull/522#discussion_r498507310

mum4k:
Interestingly enough, `testing::EqualsProto` looks to be internal only:

https://github.com/google/googletest/issues/1761

While it is in the public documentation of gtest and gmock, it isn't actually present in the code. So back to plan B.

mum4k:
Fyi, we now have a proto matcher:

https://github.com/envoyproxy/nighthawk/blob/9a6d7181ad886d9ad00a769c7d667d07ca2dc5dc/test/test_common/proto_matchers.h#L40-L48

eric846:
\o/

PR #563, opened 2020-10-22
Cli change for file based request source

wjuan-AFK:
Commit to add RequestSourcePlugin option to the CLI. base PR: https://github.com/envoyproxy/nighthawk/pull/560.
This is necessary to make use of the previous PRs adding the requestsourcepluginfactory via the CLI.


mum4k:

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  // the unexpected (yaml) diff.
+
+  // The predicates are defined as proto maps, and these seem to re-serialize into a different
+  // order. Hence we trim the maps to contain a single entry so they don't thwart our textual
+  // comparison below.
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("benchmark.http_4xx"));
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("benchmark.http_5xx"));
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("requestsource.upstream_rq_5xx"));
+
+  OptionsImpl options_from_proto(*cmd);
```
oschaaf, line 362: Maybe leave `// TODO(#433)` here and below to make sure we don't forget to clean this up once we have a good test helper for diffing protos.

source/client/options_impl.cc
```diff
@@ -265,7 +265,14 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "Remote gRPC source that will deliver to-be-replayed traffic. Each worker will separately "
       "connect to this source. For example grpc://127.0.0.1:8443/.",
       false, "", "uri format", cmd);
-
+  TCLAP::ValueArg<std::string> request_source_plugin_config(
+      "", "request-source-plugin-config",
+      "Request Source plugin configuration in json or compact yaml. "
+      "Mutually exclusive with --request-source. Example (json): "
```
oschaaf, line 271: Is the mutual exclusivity enforced? Or will one option be prioritized over the other when both are configured?

ci/do_ci.sh
```diff
@@ -47,7 +47,8 @@ function do_unit_test_coverage() {
 
 function do_integration_test_coverage() {
     export TEST_TARGETS="//test:python_test"
-    export COVERAGE_THRESHOLD=78.6
+    #TODO(#564): Revert this to 78.6
+    export COVERAGE_THRESHOLD=75
```
oschaaf, line 51: This is fine, but looking at this, maybe we should leave a note here that the comparison is limited to one tenths of a percent (as that is the precision at which is gets determined by the coverage flow).

oschaaf (changes requested):
Looks great to me overall, a few small code level questions/asks in review comments.

Those aside, looking at the coverage report I wonder if it would be hard to:
- test the too-large-of-a-file edge case
- cover  `OptionsListRequestSource::initOnThread()`, perhaps only to ensure it doesn't crash. If it's impossible to hit that maybe add `NOT_IMPLEMENTED_GCOVR_EXCL_LINE` to make it assert/document that.
- test bad json input as a request source config input

api/client/options.proto
```diff
@@ -104,6 +105,7 @@ message H1ConnectionReuseStrategy {
 
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
+// highest unused number is 38
```
dubious90, line 108: I don't think it's conventional to add in a helper comment like this, partially because it is likely to get out of sync with future changes.  I'd suggest removing it.

README.md
```diff
@@ -109,6 +110,13 @@ Perform a simple single warmup request (per worker) before starting
 execution. Note that this will be reflected in the counters that
 Nighthawk writes to the output. Default is false.
 
+--request-source-plugin-config <string>
+Request Source plugin configuration in json or compact yaml. Mutually
+exclusive with --request-source. Example (json):
```
dubious90, line 115: I believe it's also mutually exclusive with --request-options.  If we are going to document this, we should do so completely/consistently.  Should we also be updating the documentation of the old fields to indicate their mutual exclusivity?

api/client/options.proto
```diff
@@ -148,6 +150,8 @@ message CommandLineOptions {
     // Remote gRPC source that will deliver to-be-replayed traffic. Each worker will separately
     // connect to this source.
     RequestSource request_source = 26;
+    // Selects and configures a RequestSource plugin.
```
dubious90, line 153: This comment doesn't provide any extra information to the user about why and how they would use this field.

source/client/factories_impl.cc
```diff
@@ -169,15 +169,15 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
     setRequestHeader(*header, option_header.header().key(), option_header.header().value());
   }
 
-  if (options_.requestSource() == "") {
-    return std::make_unique<StaticRequestSourceImpl>(std::move(header));
-  } else {
+  if (options_.requestSource() != "") {
```
dubious90, line 172: pre-existing, but since we edited this code:  Can we change this to be `.empty()` instead of doing empty string equality?

dubious90, line 172: I'd also wonder why we edited this code.  It seems that we took a statement that was:
```
If something is true{}
else{}
```
and made it
```
if something is not true{}
else{}
```
In general, I think not true, else statements tend to be a tad more confusing.

source/client/options_impl.cc
```diff
@@ -496,6 +503,17 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       throw MalformedArgvException(e.what());
     }
   }
+  if (!request_source_plugin_config.getValue().empty()) {
+    try {
+      request_source_plugin_config_.emplace(envoy::config::core::v3::TypedExtensionConfig());
```
dubious90, line 508: This is basically just a copy of the above code, which leads me to ask:
Would it be possible / help readability if we abstracted this recipe into a private helper?  SetFieldFromJSON or something?

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
   EXPECT_TRUE(util(*(options_from_proto.toCommandLineOptions()), *cmd));
 }
 
+// This test covers --RequestSourcePlugin, which can't be tested at the same time as --RequestSource
+// and some other options. This is the test for the inlineoptionslistplugin.
+TEST_F(OptionsImplTest, InLineOptionsListRequestSourcePlugin) {
```
dubious90, line 282: We try to name tests with the action you are testing/verifying.  It should be clear what a test is accomplishing before reading the code.

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+      "},"
+      "}"
+      "}";
+
+  std::unique_ptr<OptionsImpl> options = TestUtility::createOptionsImpl(
+      fmt::format("{} --request-source-plugin-config {} {}", client_name_, request_source_config,
+                  good_test_uri_));
+
+  // Check that our conversion to CommandLineOptionsPtr makes sense.
+  CommandLineOptionsPtr cmd = options->toCommandLineOptions();
```
dubious90, line 301: Stylistically, please avoid abbreviations in variable names.

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
   EXPECT_TRUE(util(*(options_from_proto.toCommandLineOptions()), *cmd));
 }
 
+// This test covers --RequestSourcePlugin, which can't be tested at the same time as --RequestSource
+// and some other options. This is the test for the inlineoptionslistplugin.
```
dubious90, line 281: nit:  If I understand correctly, this isn't a test of inlineoptionslistplugin.  It is a test of --RequestSourcePlugin that happens to use the inlineoptionslistplugin as its test requeest source plugin.

This is important because it would be weird to have a test of inlineoptionslistplugin in this file.

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+
+  std::unique_ptr<OptionsImpl> options = TestUtility::createOptionsImpl(
+      fmt::format("{} --request-source-plugin-config {} {}", client_name_, request_source_config,
+                  good_test_uri_));
+
+  // Check that our conversion to CommandLineOptionsPtr makes sense.
+  CommandLineOptionsPtr cmd = options->toCommandLineOptions();
+  EXPECT_TRUE(
+      util(cmd->request_source_plugin_config(), options->requestSourcePluginConfig().value()));
+  // Now we construct a new options from the proto we created above. This should result in an
```
dubious90, line 304: I think you're overdescribing here.  It's also particularly confusing because the code you're describing also doesn't appear to be the code that it immediately follows this comment.

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("benchmark.http_4xx"));
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("benchmark.http_5xx"));
+  EXPECT_EQ(1, cmd->mutable_failure_predicates()->erase("requestsource.upstream_rq_5xx"));
+
+  OptionsImpl options_from_proto(*cmd);
+  std::string s1 = Envoy::MessageUtil::getYamlStringFromMessage(
+      *(options_from_proto.toCommandLineOptions()), true, true);
+  std::string s2 = Envoy::MessageUtil::getYamlStringFromMessage(*cmd, true, true);
+  EXPECT_EQ(s1, s2);
+  // For good measure, also directly test for proto equivalence, though this should be
```
dubious90, line 321: If it is superfluous, it is likely (by definition) not necessary.

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  // superfluous.
+  EXPECT_TRUE(util(*(options_from_proto.toCommandLineOptions()), *cmd));
+}
+
+// This test covers --RequestSourcePlugin, which can't be tested at the same time as --RequestSource
+// and some other options. This is the test for the filebasedplugin.
+TEST_F(OptionsImplTest, FileBasedRequestSourcePlugin) {
```
dubious90, line 328: Having now read beyond the first test, I am a little confused.  What are these tests accomplishing?  If we are testing the functionality of the new CLI option, then one test suffices.  If we are testing functionality of our different plugins, those tests should be in files related to the plugins.  (I am assuming those plugin-specific tests already exist, which makes these tests unnecessary entirely).

test/options_test.cc
```diff
@@ -276,6 +277,141 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+      "options:[{request_method:\"1\",request_headers:[{header:{key:\"key\",value:\"value\"}}]}]"
+      "},"
+      "}"
+      "}";
+
+  std::unique_ptr<OptionsImpl> options = TestUtility::createOptionsImpl(
+      fmt::format("{} --request-source-plugin-config {} {}", client_name_, request_source_config,
+                  good_test_uri_));
+
+  // Check that our conversion to CommandLineOptionsPtr makes sense.
```
dubious90, line 300: This comment doesn't seem to be helpful, as it is obvious by reading the code directly below it.

README.md
```diff
@@ -109,6 +110,13 @@ Perform a simple single warmup request (per worker) before starting
 execution. Note that this will be reflected in the counters that
 Nighthawk writes to the output. Default is false.
 
+--request-source-plugin-config <string>
+Request Source plugin configuration in json or compact yaml. Mutually
```
dubious90, line 114: Request Source is not meaningful to the average user of nighthawk.  Should we point them to request source documentation?

In other words - If someone doesn't know what request source is, how would they know to use this feature?

dubious90 (changes requested):

source/client/factories_impl.cc
wjuan-AFK: I wanted it this way because the default option is the static one, so since in a later PR I am adding an additional condition, I thought it made sense to make the non functional change here to set that up.

dubious90: In the future, I think it's easier cognitively on the reviewers if you include the change in the PR where it is relevant, rather than in an unrelated PR.

If it's substantial, you could split it into its own PR, with the intent of making the future PR easier.  I don't think that would be warranted on this particular change

api/client/options.proto
wjuan-AFK: This seems to be relatively standard practice. It provides the convenience of not having to look through the proto to find what the highest number is. I think it is unlikely to get out of sync because if you're editing the proto, the comment is readily available where the proto is, and it only needs to be updated when you're trying to add a new field to the proto.

README.md
wjuan-AFK: That would make sense, but --request-options doesn't actually seem to be a command line options field. I am updating --request-source to indicate their mutual exclusivity though. Currently, the only way we can get request-options seems to be via StartRequest being passed to HandleExecutionRequest.

source/client/options_impl.cc
wjuan-AFK: The mutually exclusivity should be enforced. I added a validation in the constructor.

ci/do_ci.sh
wjuan-AFK: That makes sense thanks.

wjuan-AFK:
> * request source config

Added these tests. The bad json test is inside options_test. The others are inside request_source_plugin_test.

test/options_test.cc
wjuan-AFK: Makes sense.

test/options_test.cc
wjuan-AFK: That makes sense. Thanks dubious90. I see what you mean.

test/options_test.cc
wjuan-AFK: That makes sense.

test/options_test.cc
wjuan-AFK: Sure thing.

test/options_test.cc
wjuan-AFK: I found this comment helpful to me, so I thought it made sense to keep it. I think I would have found these lines odd if I didn't have them.

test/options_test.cc
wjuan-AFK: Sure.

test/options_test.cc
wjuan-AFK: I'm just testing that these different parameters pass in as options successfully.

test/options_test.cc
wjuan-AFK: Understood.

test/options_test.cc
```diff
@@ -272,10 +273,113 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  // comparison below.
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("benchmark.http_4xx"));
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("benchmark.http_5xx"));
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("requestsource.upstream_rq_5xx"));
+  // TODO(#433)
+  OptionsImpl options_from_proto(*command);
+  std::string s1 = Envoy::MessageUtil::getYamlStringFromMessage(
+      *(options_from_proto.toCommandLineOptions()), true, true);
+  std::string s2 = Envoy::MessageUtil::getYamlStringFromMessage(*command, true, true);
+  EXPECT_EQ(s1, s2);
```
oschaaf, line 310: Consider also adding the following comparison, as I suspect the yaml/text based comparison may miss edge cases.

```c++
  EXPECT_TRUE(util(*(options_from_proto.toCommandLineOptions()), *command));
```

oschaaf (approved):
LGTM, thanks!
Leaving one last suggestion with respect to a test that now uses a textual yaml diff comparison as the sole means to test for equivalence, I suspect that doubling down with a true equivalence tests helps catch edge cases. (We really need https://github.com/envoyproxy/nighthawk/issues/433)

test/options_test.cc
wjuan-AFK: Ok. Understood. That makes sense.

wjuan-AFK: @dubious90 Wanted to let you know that I was adding this back in since this reasoning makes sense to me.

test/options_test.cc
```diff
@@ -249,7 +250,7 @@ TEST_F(OptionsImplTest, AlmostAll) {
   EXPECT_TRUE(util(cmd->stats_sinks(0), options->statsSinks()[0]));
   EXPECT_TRUE(util(cmd->stats_sinks(1), options->statsSinks()[1]));
   EXPECT_EQ(cmd->latency_response_header_name().value(), options->responseHeaderWithLatencyInput());
-
+  // TODO(#433)
```
dubious90, line 253: Two nit asks on this set of TODOs:
1.  Can we add a oneliner, explaining what the action to do here specifically is
2.  Rather than repeating it on every test here, we could have in the comment something like "Here and below".  Then we only need to include the TODO once per relevant file

test/options_test.cc
dubious90: Sorry -- my previous code comment wasn't particularly helpful/directive.  I agree that comments here would be helpful, but I'm not finding this comment to clarify the code for me, so let me re-explain better:
In general, comments are most helpful when we explain the why, rather than the what.  This comment is actually doing a little bit of both.  One of the reasons why I think you're finding the need to explain "what" is that the comments aren't actually positionally with the code that they're explaining.

Rather than explaining a whole series of actions in one comment, can we split up the comments to be directly tied to their code sections?

test/options_test.cc
```diff
@@ -272,10 +273,115 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  // order. Hence we trim the maps to contain a single entry so they don't thwart our textual
+  // comparison below.
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("benchmark.http_4xx"));
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("benchmark.http_5xx"));
+  EXPECT_EQ(1, command->mutable_failure_predicates()->erase("requestsource.upstream_rq_5xx"));
+  // TODO(#433)
+  OptionsImpl options_from_proto(*command);
+  std::string s1 = Envoy::MessageUtil::getYamlStringFromMessage(
+      *(options_from_proto.toCommandLineOptions()), true, true);
+  std::string s2 = Envoy::MessageUtil::getYamlStringFromMessage(*command, true, true);
```
dubious90, line 309: One thing that might make this easier to parse is if we use more descriptive names, rather than s1 and s2.  Maybe expected_options_yaml and actual_options_yaml?  Or whatever names you think would be accurate/helpful.

test/options_test.cc
```diff
@@ -272,10 +273,115 @@ TEST_F(OptionsImplTest, RequestSource) {
...
+  EXPECT_EQ(s1, s2);
+  // Additional comparison to avoid edge cases missed.
+  EXPECT_TRUE(util(*(options_from_proto.toCommandLineOptions()), *command));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    HappyPathRequestSourceConfigJsonSuccessfullyTranslatesIntoOptions,
+    RequestSourcePluginTestFixture,
+    ::testing::ValuesIn(std::vector<std::string>{
```
dubious90, line 318: I'm having a really hard time parsing this.  There are two things that might help here:
1. C++ has a formatted string option that will make this much easier to read, rather than using \" so much. You can do R"( string that contains " with no problems)". https://abseil.io/tips/64
2. (optional) I _think_ this is just one string, but it's difficult to tell.  Can we abstract the actual textprotos into variables?

dubious90 (commented):
LGTM modulo a few commenting comments.  Will continue tracking the two conversations there, but otherwise reassigning to mumak, who will approve.

test/options_test.cc
wjuan-AFK: 1. sounds good to me. I feel like tagging the comparisons which need to be replaced is helpful, but I can remove the other ones if you insist.

test/options_test.cc
wjuan-AFK: For sure.

test/options_test.cc
wjuan-AFK: I'm not sure that 1 works for me here. I might be missing something, but it seems like there are issues with the newlines. At least it seems like using R"( makes you keep the newlines if you space them into separate lines. I can make each line a separate raw string though, I'm not sure that this makes it more readable, but maybe you'll find this preferable?
2. I've kind of tried to do this. Hope that this approach looks ok.

mum4k (approved):
Looks good, no actionable comments from my end.

mum4k:
@dubious90 this is good to go.

dubious90 (approved):

Issue #564, opened 2020-10-26
Revert Integration Test Coverage percentage. Temporarily Lowered it to get a PR in separately from coverage in follow up PR.


wjuan-AFK:
PR #563 Was blocked on Test_integration_coverage and I was working on the python tests in a separate PR. So I lowered the threshold in #563 in order to get the process started on reviewing it.

dubious90:
Can you rephrase this bug for me?  The issue appears to be that we are missing integration test coverage.  The lowered coverage percentage is really a side effect of that.  So then we can have one issue (fixed by one PR) where we add in the missing coverage, and raise the coverage percentage back appropriately.  Thanks!

oschaaf:
Is this issue still relevant?

mum4k:
Looks like this isn't relevant anymore. Filed #830 to run a fixit and improve our integration test coverage.

PR #566, opened 2020-10-28
Factory plugin implementation

wjuan-AFK:
Adding implementation in the factory_impl class for loading request source plugins. This also introduces python tests for the previous PR and undoes the temporary hack to reduce test_integration_coverage threshold in issue #564.


dubious90:
@mum4k Could you please review and merge when done?

source/client/factories_impl.h
```diff
@@ -58,14 +61,30 @@ class OutputFormatterFactoryImpl : public OutputFormatterFactory {
...
 class RequestSourceFactoryImpl : public OptionBasedFactoryImpl, public RequestSourceFactory {
 public:
-  RequestSourceFactoryImpl(const Options& options);
+  RequestSourceFactoryImpl(const Options& options, Envoy::Api::Api& api);
   RequestSourcePtr create(const Envoy::Upstream::ClusterManagerPtr& cluster_manager,
                           Envoy::Event::Dispatcher& dispatcher, Envoy::Stats::Scope& scope,
                           absl::string_view service_cluster_name) const override;
 
+protected:
+  Envoy::Api::Api& api_;
```
mum4k, line 70: Can this be private instead or do we have a good reason to make it protected?

test/factories_test.cc
```diff
@@ -52,17 +52,61 @@ TEST_F(FactoriesTest, CreateBenchmarkClient) {
...
+                                   request_source_plugin_config.value(),
+                                   Envoy::ProtobufMessage::getStrictValidationVisitor());
+  EXPECT_CALL(options_, requestMethod()).Times(1);
+  EXPECT_CALL(options_, requestBodySize()).Times(1);
+  EXPECT_CALL(options_, uri()).Times(2).WillRepeatedly(Return("http://foo/"));
+  EXPECT_CALL(options_, requestSource()).Times(1);
+  EXPECT_CALL(options_, requestSourcePluginConfig())
+      .Times(2)
+      .WillRepeatedly(ReturnRef(request_source_plugin_config));
+  auto cmd = std::make_unique<nighthawk::client::CommandLineOptions>();
```
mum4k, line 80: (nit) Can we replace some of these auto statements with the actual type? Only needed in places where the type isn't obvious from the expression.

mum4k (changes requested):
Looks good just one question, let's wait for review from @oschaaf.

source/client/factories_impl.h
wjuan-AFK: I believe that this can be private.

source/client/factories_impl.cc
```diff
@@ -168,18 +170,39 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
...
     RELEASE_ASSERT(!service_cluster_name.empty(), "expected cluster name to be set");
     // We pass in options_.requestsPerSecond() as the header buffer length so the grpc client
     // will shoot for maintaining an amount of headers of at least one second.
     return std::make_unique<RemoteRequestSourceImpl>(cluster_manager, dispatcher, scope,
                                                      service_cluster_name, std::move(header),
                                                      options_.requestsPerSecond());
+  } else if (options_.requestSourcePluginConfig().has_value()) {
+    absl::StatusOr<RequestSourcePtr> plugin_or = LoadRequestSourcePlugin(
+        options_.requestSourcePluginConfig().value(), api_, std::move(header));
+    RELEASE_ASSERT(plugin_or.ok(),
```
oschaaf, line 183: Maybe annotate with a `XXX_GCOVR_EXCL_XXX`. We don't care not every branch gets hit in tests, putting that in documents an explicit choice that we're ok with that. If it would ever start working, it would also help with correcting the reported coverage % (it seems to be ignored today).

source/client/factories_impl.cc
```diff
@@ -168,18 +170,39 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
...
+absl::StatusOr<RequestSourcePtr> RequestSourceFactoryImpl::LoadRequestSourcePlugin(
+    const envoy::config::core::v3::TypedExtensionConfig& config, Envoy::Api::Api& api,
+    Envoy::Http::RequestHeaderMapPtr header) const {
+  try {
+    auto& config_factory =
+        Envoy::Config::Utility::getAndCheckFactoryByName<RequestSourcePluginConfigFactory>(
+            config.name());
+    return config_factory.createRequestSourcePlugin(config.typed_config(), api, std::move(header));
+  } catch (const Envoy::EnvoyException& e) {
+    return absl::InvalidArgumentError(
```
oschaaf, line 202: Consider covering this branch in a unit test, or annotating with `XXX_GCOVR_EXCL_XXX` if we don't care.

source/client/factories_impl.h
```diff
@@ -58,14 +61,30 @@ class OutputFormatterFactoryImpl : public OutputFormatterFactory {
...
+
 private:
   void setRequestHeader(Envoy::Http::RequestHeaderMap& header, absl::string_view key,
                         absl::string_view value) const;
+  /**
+   * Instantiates a RequestSource using a RequestSourcePluginFactory based on the plugin name in
+   * |config|, unpacking the plugin-specific config proto within |config|. Validates the config
+   * proto.
+   *
+   * @param config Proto containing plugin name and plugin-specific config proto.
```
oschaaf, line 80: Not all parameters are documented, consider adding doc comments for `api` & `header`?

test/integration/test_request_source_plugin.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pytest
+import os
+
+from test.integration.integration_test_fixtures import (http_test_server_fixture, server_config)
+from test.integration import utility
+from test.integration import asserts
+
+
+def test_request_source_plugin_happy_flow(http_test_server_fixture):
```
oschaaf, line 11: Just a thought, I wonder if it would make sense to parameterize this test over plugin setup. So we iterate the same test over all known plugins. That way, future plugins could add themselves to this hypothetical parameterized list, and have basic integration test coverage through that. (This would imply that all plugin types would be set up with a configuration that yields the same traffic pattern). Wouldn't object to leaving this as is, but if you see sufficient value it might be worth doing.

oschaaf (commented):
Looks great, leaving a few small asks, and a few thoughts.

test/integration/test_request_source_plugin.py
wjuan-AFK: Sure thing.

source/client/factories_impl.cc
oschaaf: As discussed on slack, we could probably just append `// NOT_REACHED_GCOVR_EXCL_LINE` to the line itself, or use 
```c++.
// GCOVR_EXCL_START
... code to be ignored in coverage measurements ...
// GCOVR_EXCL_STOP
```

source/client/factories_impl.cc
```diff
@@ -168,18 +170,43 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
...
     RELEASE_ASSERT(!service_cluster_name.empty(), "expected cluster name to be set");
     // We pass in options_.requestsPerSecond() as the header buffer length so the grpc client
     // will shoot for maintaining an amount of headers of at least one second.
     return std::make_unique<RemoteRequestSourceImpl>(cluster_manager, dispatcher, scope,
                                                      service_cluster_name, std::move(header),
                                                      options_.requestsPerSecond());
+  } else if (options_.requestSourcePluginConfig().has_value()) {
+    absl::StatusOr<RequestSourcePtr> plugin_or = LoadRequestSourcePlugin(
+        options_.requestSourcePluginConfig().value(), api_, std::move(header));
+    // GCOVR_EXCL_START
```
oschaaf, line 183: I didn't expect the GCOVR exclusion here, but rather at/around the RELEASE_ASSERT line above -- so, sorry could you remove it here? Looking at the [coverage report](https://22476-180498819-gh.circle-artifacts.com/0/coverage/source/client/factories_impl.cc.gcov.html) everything is fine now, including the RELEASE_ASSERT line, so it seems we don't need it there either now (probably because there's no conditional argument being passed).

source/client/factories_impl.cc
```diff
@@ -168,18 +170,39 @@ RequestSourceFactoryImpl::create(const Envoy::Upstream::ClusterManagerPtr& clust
   for (const auto& option_header : request_options.request_headers()) {
     setRequestHeader(*header, option_header.header().key(), option_header.header().value());
   }
-
   if (!options_.requestSource().empty()) {
     RELEASE_ASSERT(!service_cluster_name.empty(), "expected cluster name to be set");
     // We pass in options_.requestsPerSecond() as the header buffer length so the grpc client
     // will shoot for maintaining an amount of headers of at least one second.
     return std::make_unique<RemoteRequestSourceImpl>(cluster_manager, dispatcher, scope,
```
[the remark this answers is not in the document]
oschaaf: Thanks for covering this in a test. Resolving.

source/client/factories_impl.cc
oschaaf: Resolving this, thanks for covering it in a test.

oschaaf (changes requested):
One last nit about removing `// GCOVR_EXCL_START` and `// GCOVR_EXCL_END`. It seems some of the changes outdated that comment, and these annotations no longer have value. Sorry about the back and forth on that. Other then that, this looks great to me.

mum4k (approved):
Ready for merging, blocked on the clang_tidy flake which we are planning to remove temporarily in the near future (#570).

mum4k:
@oschaaf, this is still marked as "change requested" from your last review, can you please indicate if there is anything outstanding or approve if otherwise?

oschaaf:
@mum4k  Sorry, I forgot to update the `waiting-for-...` label. There's one last (small) [change](https://github.com/envoyproxy/nighthawk/pull/566/files#r520878294) I requested.

mum4k:
Please merge from master to remove the hard block on clang_tidy.

source/client/factories_impl.cc
wjuan-AFK: Sounds good.

oschaaf (approved):
Looks great, thanks

mum4k (approved):

PR #570, opened 2020-11-10
CI: make clang-tidy always pass.

oschaaf:
As discussed yesterday, make clang-tidy ignore failures to stop
the excessive flaking from being a blocker to merge.
This is meant as a temporary stop gap to buy us some time to dig
up the root cause.

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -34,6 +34,7 @@ function do_test() {
 }
 
 function do_clang_tidy() {
+    # TODO(#546): deflake clang tidy runs, and remove '|| true' here.
     ci/run_clang_tidy.sh || true
 }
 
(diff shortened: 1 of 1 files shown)


ci/do_ci.sh
```diff
@@ -34,7 +34,7 @@ function do_test() {
 }
 
 function do_clang_tidy() {
-    ci/run_clang_tidy.sh
+    ci/run_clang_tidy.sh || true
```
mum4k, line 37: Can we add a comment here pointing to the issue about a long term fix?

mum4k (changes requested):

ci/do_ci.sh
oschaaf: Done in 0790741

mum4k (approved):

PR #573, opened 2020-11-16
Schedule option

### LINKED ISSUE #569: Building with clang's libc++ fails
PR #549 modified Nighthawk's use of `MonotonicTime` to `SystemTime`, which causes a failure to build with clang's/llvm's libc++ (--config=libc++), and gives the following error:

```
ERROR: /home/.../git/nighthawk/source/client/BUILD:11:17: C++ compilation of rule '//source/client:nighthawk_client_lib' failed (Exit 1) clang failed: error executing command /home/.../clang10/bin/clang -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -fcolor-diagnostics -fno-omit-frame-pointer '-stdlib=libc++' -MD -MF ... (remaining 765 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from source/client/process_impl.cc:1:
In file included from bazel-out/k8-fastbuild/bin/source/client/_virtual_includes/nighthawk_client_lib/client/process_impl.h:3:
In file included from /home/.../clang10/bin/../include/c++/v1/map:479:
In file included from /home/.../clang10/bin/../include/c++/v1/__tree:15:
/home/.../clang10/bin/../include/c++/v1/memory:3028:32: error: no matching constructor for initialization of 'Nighthawk::Client::ClientWorkerImpl'
    return unique_ptr<_Tp>(new _Tp(_VSTD::forward<_Args>(__args)...));
                               ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
source/client/process_impl.cc:188:29: note: in instantiation of function template specialization 'std::__1::make_unique<Nighthawk::Client::ClientWorkerImpl, Envoy::Api::Api &, Envoy::ThreadLocal::InstanceImpl &, std::__1::unique_ptr<Envoy::Upstream::ClusterManager, std::__1::default_delete<Envoy::Upstream::ClusterManager> > &, const Nighthawk::Client::BenchmarkClientFactoryImpl &, const Nighthawk::Client::TerminationPredicateFactoryImpl &, const Nighthawk::Client::SequencerFactoryImpl &, const Nighthawk::Client::RequestSourceFactoryImpl &, Envoy::Stats::ThreadLocalStoreImpl &, int &, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000000> > >, std::__1::shared_ptr<Envoy::Tracing::HttpTracer> &, Nighthawk::Client::ClientWorkerImpl::HardCodedWarmupStyle>' requested here
    workers_.push_back(std::make_unique<ClientWorkerImpl>(
                            ^
bazel-out/k8-fastbuild/bin/source/client/_virtual_includes/nighthawk_client_lib/client/client_worker_impl.h:29:3: note: candidate constructor not viable: no known conversion from 'time_point<[...], duration<[...], ratio<[...], 1000000000>>>' to 'const time_point<[...], duration<[...], ratio<[...], 1000000>>>' for 10th argument
  ClientWorkerImpl(Envoy::Api::Api& api, Envoy::ThreadLocal::Instance& tls,
  ^
bazel-out/k8-fastbuild/bin/source/client/_virtual_includes/nighthawk_client_lib/client/client_worker_impl.h:25:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 12 were provided
class ClientWorkerImpl : public WorkerImpl, virtual public ClientWorker {
      ^
1 error generated.
Target //source/client:nighthawk_client_lib failed to build
```

After executing:
```
bazel build //source/client:nighthawk_client_lib --config=libc++
```

The cause seems to be the different resolution of system_clock by llvm and gcc:
- llvm uses a [milliseconds](https://github.com/llvm/llvm-project/blob/61d4b342d15a82d5e841d9e72a52d77eaa5b5097/libcxx/include/chrono#L1574) resolution;
- gcc uses a [nanoseconds](https://github.com/gcc-mirror/gcc/blob/a30d4fc5199ba16cec39fd3f9cca878a9699cf4e/libstdc%2B%2B-v3/include/std/chrono#L1103) resolution.
and the nighthawk [code](https://github.com/envoyproxy/nighthawk/blob/525c7965b1eee1728c112b590e0a5fdcdffbd2d0/source/client/process_impl.cc#L182) requires a nanoseconds resolution.

Possible ways to solve:
1. Revert, and change back to `MonotonicTime`.
1. Use `MonotonicTime` in this part of the code.
1. Change the required resolution in this code to milliseconds instead of nanoseconds.
1. Change Envoy's definition of [SystemTime](https://github.com/envoyproxy/envoy/blob/5330d1d2bea34724bdc20df0ae1a8c67f2b9b22c/include/envoy/common/time.h#L16) to explicitly use a nanoseconds, i.e.:
`using SystemTime = std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>;`, and update the necessary places in Envoy's code to correctly cast when system_clock is used (`std::chrono::time_point_cast<std::chrono::system_clock::duration>(time)`).

@oschaaf Am I missing anything? What's your opinion about this?

oschaaf:
Adds a new schedule option to the proto api., which allows specifying a date/time
at which execution should start.

Should also fix #569 by reducing reliance on `Envoy::SystemTime` to a minimum:
Based on the value schedule option, a monotonic time will be computed
and propagated. This makes us use `Envoy::MonotonicTime` again in places
where sub ms resolution is a must have. (The fix will need confirmation, as I
couldn't get `--config=libc++` to work for other reasons.)

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package nighthawk.client;
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
 import "envoy/config/core/v3/base.proto";
 import "envoy/config/metrics/v3/stats.proto";
@@ -219,4 +220,7 @@ message CommandLineOptions {
   // "emit_previous_request_delta_in_response_header" to record elapsed time between request
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
+  // Provide an execution starting date and time. Optional, any value specified must be in the
+  // future.
+  google.protobuf.Timestamp schedule = 105;
 }
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -222,5 +222,5 @@ message CommandLineOptions {
   google.protobuf.StringValue latency_response_header_name = 36;
   // Provide an execution starting date and time. Optional, any value specified must be in the
   // future.
-  google.protobuf.Timestamp schedule = 105;
+  google.protobuf.Timestamp scheduled_start = 105;
 }
--- a/include/nighthawk/client/output_collector.h
+++ b/include/nighthawk/client/output_collector.h
@@ -3,6 +3,7 @@
 #include <memory>
 
 #include "envoy/common/pure.h"
+#include "envoy/common/time.h"
 
 #include "nighthawk/common/statistic.h"
 
--- a/include/nighthawk/client/output_collector.h
+++ b/include/nighthawk/client/output_collector.h
@@ -26,6 +26,7 @@ public:
    * @param statistics Reference to a vector of statistics to add to the output.
    * @param counters Reference to a map of counter values, keyed by name, to add to the output.
    * @param execution_duration Execution duration associated to the to-be-added result.
+   * @param first_acquisition_time Timing of the first rate limiter acquisition.
    */
   virtual void addResult(absl::string_view name, const std::vector<StatisticPtr>& statistics,
                          const std::map<std::string, uint64_t>& counters,
--- a/include/nighthawk/client/output_collector.h
+++ b/include/nighthawk/client/output_collector.h
@@ -6,6 +6,8 @@
 
 #include "nighthawk/common/statistic.h"
 
+#include "absl/types/optional.h"
+
 namespace Nighthawk {
 namespace Client {
 
@@ -26,7 +28,8 @@ public:
    */
   virtual void addResult(absl::string_view name, const std::vector<StatisticPtr>& statistics,
                          const std::map<std::string, uint64_t>& counters,
-                         const std::chrono::nanoseconds execution_duration) PURE;
+                         const std::chrono::nanoseconds execution_duration,
+                         const absl::optional<Envoy::SystemTime>& first_acquisition_time) PURE;
   /**
    * Directly sets the output value.
    *
--- a/include/nighthawk/common/factories.h
+++ b/include/nighthawk/common/factories.h
@@ -24,7 +24,7 @@ public:
                               const SequencerTarget& sequencer_target,
                               TerminationPredicatePtr&& termination_predicate,
                               Envoy::Stats::Scope& scope,
-                              const Envoy::SystemTime scheduled_starting_time) const PURE;
+                              const Envoy::MonotonicTime scheduled_starting_time) const PURE;
 };
 
 class StatisticFactory {
@@ -46,7 +46,7 @@ public:
   virtual ~TerminationPredicateFactory() = default;
   virtual TerminationPredicatePtr
   create(Envoy::TimeSource& time_source, Envoy::Stats::Scope& scope,
-         const Envoy::SystemTime scheduled_starting_time) const PURE;
+         const Envoy::MonotonicTime scheduled_starting_time) const PURE;
 };
 
 /**
--- a/include/nighthawk/common/rate_limiter.h
+++ b/include/nighthawk/common/rate_limiter.h
@@ -33,6 +33,12 @@ public:
    * @return Envoy::TimeSource& time_source used to track time.
    */
   virtual Envoy::TimeSource& timeSource() PURE;
+
+  /**
+   * @return absl::optional<Envoy::SystemTime> Time of the first acquisition, if any.
+   */
+  virtual absl::optional<Envoy::SystemTime> firstAcquisitionTime() const PURE;
+
   /**
    * @return std::chrono::nanoseconds elapsed since the first call to tryAcquireOne(). Used by some
    * rate limiter implementations to compute acquisition rate.
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -166,7 +166,7 @@ bool ProcessImpl::requestExecutionCancellation() {
 }
 
 void ProcessImpl::createWorkers(const uint32_t concurrency,
-                                const absl::optional<Envoy::SystemTime>& schedule) {
+                                const absl::optional<Envoy::SystemTime>& scheduled_start) {
   // TODO(oschaaf): Expose kMinimalDelay in configuration.
   const std::chrono::milliseconds kMinimalWorkerDelay = 500ms + (concurrency * 50ms);
   ASSERT(workers_.empty());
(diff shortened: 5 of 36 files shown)


dubious90:
@qqustc Please review and assign back to me once ready.

include/nighthawk/client/output_collector.h
```diff
@@ -26,7 +29,8 @@ class OutputCollector {
    */
   virtual void addResult(absl::string_view name, const std::vector<StatisticPtr>& statistics,
                          const std::map<std::string, uint64_t>& counters,
-                         const std::chrono::nanoseconds execution_duration) PURE;
+                         const std::chrono::nanoseconds execution_duration,
+                         const absl::optional<Envoy::SystemTime>& first_acquisition_time) PURE;
```
qqustc, line 33: Add comment for first_acquisition_time?

qqustc (dismissed):
Thanks Otto!

qqustc:
> @qqustc Please review and assign back to me once ready.

LGTM, assigned back to @dubious90 for review.

api/client/options.proto
```diff
@@ -219,4 +220,7 @@ message CommandLineOptions {
   // "emit_previous_request_delta_in_response_header" to record elapsed time between request
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
+  // Provide an execution starting date and time. Optional, any value specified must be in the
+  // future.
+  google.protobuf.Timestamp schedule = 105;
```
dubious90, line 225: I feel like schedule here implies the ability to create recurring requests.  How do you feel about scheduled_start?

include/nighthawk/common/factories.h
```diff
@@ -24,7 +24,7 @@ class SequencerFactory {
                               const SequencerTarget& sequencer_target,
                               TerminationPredicatePtr&& termination_predicate,
                               Envoy::Stats::Scope& scope,
-                              const Envoy::SystemTime scheduled_starting_time) const PURE;
+                              const Envoy::MonotonicTime scheduled_starting_time) const PURE;
```
dubious90, line 27: Not actionable, just making sure I'm understanding - this is where we're moving back to MonotonicTime to prevent #569 ?

include/nighthawk/client/output_collector.h
dubious90, line 33: will likely be addressed when the comment is added, but I'm not sure i understand what acquisition means in this context.

source/client/process_impl.cc
```diff
@@ -179,7 +181,10 @@ void ProcessImpl::createWorkers(const uint32_t concurrency) {
   // TODO(oschaaf): Arguably, this ought to be the job of a rate limiter with awareness of the
```
dubious90, line 181: I think this TODO has become out of date with our ideal, given this restructure.  Do we still want a global rate limiter to do this work anymore, given what we now know?  If not, can we remove this TODO?

source/client/process_impl.cc
```diff
@@ -179,7 +181,10 @@ void ProcessImpl::createWorkers(const uint32_t concurrency) {
...
   // global status quo, which we do not have right now. This has been noted in the
   // track-for-future issue.
-  const auto first_worker_start = time_system_.systemTime() + kMinimalWorkerDelay;
+  const Envoy::MonotonicTime monotonic_now = time_system_.monotonicTime();
+  const std::chrono::nanoseconds offset =
```
dubious90, line 185: 2. I might be misunderstanding but this ternary seems off to me.  kMinimalWorkerDelay has the worker offsets baked into it, correct?  So, if we are using a scheduled start time, and we aren't using minimal worker delay at all, then we aren't including any worker offsets from each other.

So now, batching behavior of nighthawk acts differently depending on whether or not we are using a schedule which doesn't seem right.

source/client/process_impl.cc
```diff
@@ -179,7 +181,10 @@ void ProcessImpl::createWorkers(const uint32_t concurrency) {
   // TODO(oschaaf): Arguably, this ought to be the job of a rate limiter with awareness of the
   // global status quo, which we do not have right now. This has been noted in the
   // track-for-future issue.
-  const auto first_worker_start = time_system_.systemTime() + kMinimalWorkerDelay;
+  const Envoy::MonotonicTime monotonic_now = time_system_.monotonicTime();
```
dubious90, line 184: Three comments here:
1.  This has gotten complex enough (especially with the long comments above to explain what's going on), that I think it might warrant its own privately scoped function.  CalculateWorkerStartTime_ or something?

source/client/process_impl.cc
```diff
@@ -179,7 +181,10 @@ void ProcessImpl::createWorkers(const uint32_t concurrency) {
...
+  const std::chrono::nanoseconds offset =
+      schedule.has_value() ? schedule.value() - time_system_.systemTime() : kMinimalWorkerDelay;
```
dubious90, line 186: 3. We are heavily explaining the workerdelay behavior here in the comment above.  Can we add a one-line comment for how the schedule behavior fits in?  For instance, if I am wrong about point 2 above, we should say why the scheduled workers don't need an offset delay.

include/nighthawk/common/rate_limiter.h
```diff
@@ -33,6 +33,12 @@ class RateLimiter {
    * @return Envoy::TimeSource& time_source used to track time.
    */
   virtual Envoy::TimeSource& timeSource() PURE;
+
+  /**
+   * @return absl::optional<Envoy::SystemTime> Time of the first acquisition, if any.
+   */
+  virtual absl::optional<Envoy::SystemTime> firstAcquisitionTime() const PURE;
```
dubious90, line 40: I might have just missed it in this pull request (if so, please point me to the right file), but while I'm seeing code that writes this field or passes it around, I'm not finding where it's actually used functionally.  What is its intended purpose?

test/process_test.cc
```diff
@@ -178,6 +180,85 @@ TEST_P(ProcessTest, NoFlushWhenCancelExecutionBeforeLoadTestBegin) {
...
+      const bool result = process->run(collector);
+      process->shutdown();
+      verify_callback(result, collector.toProto());
+    });
+
+    // We introduce real-world sleeps to give the executing ProcessImpl
+    // an opportunity to observe passage of simulated time. We increase simulated
+    // time in three steps, to give it an opportunity to start at the wrong time
+    // in case there is an error in the scheduling logic it implements.
+    // Note that these sleeps may seem excessively long, but sanitizer runs may need that.
```
dubious90, line 208: These comments are very helpful.  Thank you.

dubious90 (changes requested):

api/client/options.proto
oschaaf: done here and all other applicable spots.

include/nighthawk/common/rate_limiter.h
oschaaf: This gets consumed over at https://github.com/envoyproxy/nighthawk/pull/573/files#diff-e528c3973715873e2ec7fcac6a842e5d9a59d6523a77bf1219d295dd4c7285cdR539 and ends up in the output as `execution_start`.

include/nighthawk/common/factories.h
oschaaf: Yes, that is right.

oschaaf:
@dubious90 Thanks for the feedback, I think I addressed it. Extracted the worker start time & offset calculations into their own methods for clarity. Ready for another round!

source/client/process_impl.cc
oschaaf: This computes the start of the *first* worker, we will still be injecting an offset between each worker below. I refactored this for clarity, hopefully it's easier to read now.

dubious90 (approved):
Looks good.  Thanks for the readability changes!  It's really easy to follow what's happening now.

PR #575, opened 2020-11-16
Update Envoy to 588d9344b31e6544869547c4bcd359b3b0f1d4cf

oschaaf:
- Fixes to accommodate upstream connection pool changes.
- Fixes to accommodate upstream cluster related changes.

### DIFF
--- a/source/server/README.md
+++ b/source/server/README.md
@@ -34,7 +34,7 @@ static_resources:
           port_value: 10000
       filter_chains:
         - filters:
-            - name: envoy.filters.network.http_connection_manager
+            - name: envoy.http_connection_manager
               config:
                 generate_request_id: false
                 codec_type: auto
--- a/source/server/README.md
+++ b/source/server/README.md
@@ -34,7 +34,7 @@ static_resources:
           port_value: 10000
       filter_chains:
         - filters:
-            - name: envoy.http_connection_manager
+            - name: envoy.filters.network.http_connection_manager
               config:
                 generate_request_id: false
                 codec_type: auto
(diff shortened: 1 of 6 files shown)


oschaaf:
This seems to run into a challenge: Envoy deprecated the v2 api, which means we either need to 
- set flags to allow continued use of the v2 bootstrap / configuration format, which I think will be possible up to Q1 2021
- update configurations used in test and published in docs/images
- add code to automatically translate to control timing of these changes

source/server/README.md
```diff
@@ -34,7 +34,7 @@ static_resources:
           port_value: 10000
       filter_chains:
         - filters:
-            - name: envoy.http_connection_manager
+            - name: envoy.filters.network.http_connection_manager
```
mum4k, line 37: Was this automatically generated or does it need a merge? Not sure if we intended to have this change in this PR.

mum4k (approved):
Please mark as waiting for review again once we are ready to merge this in.

source/server/README.md
oschaaf: I merged master into this, it seems that `source/server/README.md` needs an update from v2 to v3. I can do that here or we can move that into a separate PR, please let me know your preference.

mum4k: It is fine to leave it here, I just wasn't sure it was intended since the PR description doesn't mention it. I will update it and merge. Thank you.

mum4k:
Waiting for #584 before we merge this one in.

mum4k (approved):

PR #581, opened 2020-12-01
Fail the integration tests on warnings or errors.

### LINKED ISSUE #577: Integration tests should report warnings/errors in logs
This would have given us an earlier warning about the Envoy v2 API deprecation. An example of an error message that is logged by the nighthawk test server:

```
[2020-12-01 01:39:50.785][114][warning][misc] [external/envoy/source/common/protobuf/utility.cc:294] Configuration does not parse cleanly as v3. v2 configuration is deprecated and will be removed from Envoy at the start of Q1 2021: Unknown field in: {"admin":{"profile_path":"/tmp/2020-12-01-01-39-49/http_h1_connection_management_single_request_per_conn_1_IpVersion.IPV6//nighthawk-test-server.prof","access_log_path":"/tmp/2020-12-01-01-39-49/http_h1_connection_management_single_request_per_conn_1_IpVersion.IPV6//nighthawk-test-server-admin-access.log","address":{"socket_address":{"address":"::/0","port_value":0}}},"static_resources":{"listeners":[{"filter_chains":[{"filters":[{"config":{"generate_request_id":false,"codec_type":"auto","route_config":{"virtual_hosts":[{"name":"service","domains":["*"]}],"name":"local_route"},"http_filters":[{"name":"time-tracking"},{"name":"dynamic-delay"},{"name":"test-server","config":{"response_body_size":10,"response_headers":[{"header":{"key":"x-nh","value":"1"}}]}},{"name":"envoy.router","config":{"dynamic_stats":false}}],"stat_prefix":"ingress_http"},"name":"envoy.http_connection_manager"}]}],"address":{"socket_address":{"port_value":0,"address":"::/0"}}}]}}
[2020-12-01 01:39:50.787][114][warning][misc] [external/envoy/source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.api.v2.listener.Filter Using deprecated option 'envoy.api.v2.listener.Filter.config' from file listener_components.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

Ideally the implementation would support some kind of list to allow us to "acknowledge" or "filter" known warnings if any.

mum4k:
Fails the integration tests if any unknown warnings or errors are found in the logs of the Nighthawk test server.
Includes an ignore list that (for now) accepts known warnings found in the logs, the ignore list can have different content per each test case.

Also:
- moving the `Attributes:` docstring section on `IntegrationTestBase` into its class docstring. `__init__` is just a method and should only have the `Args:` section.

Fixes #577


mum4k:

test/integration/nighthawk_test_server.py
```diff
@@ -40,6 +40,29 @@ def _substitute_yaml_values(runfiles_instance, obj, params):
...
+# Other messages logged at either of these severities fail the test.
+_TEST_SERVER_WARN_ERROR_IGNORE_LIST = frozenset([
+    # TODO(#580) Remove these once our use of Envoy's API is updated to v3.
+    "Configuration does not parse cleanly as v3",
+    "Deprecated field: type envoy.api.v2.Cluster",
+    "Deprecated field: type envoy.api.v2.listener.Filter",
+    "Deprecated field: type envoy.config.filter.network.http_connection_manager.v2.HttpFilter",
+
+    # TODO(mum4k): Identify these and file issues or add explanation as necessary.
+    "Unable to use runtime singleton for feature envoy.http.headermap.lazy_map_min_size",
```
oschaaf, line 55: not actionable, but I think I remember issues being filed for this, it might be a bug in Envoy.

test/integration/BUILD
```diff
@@ -9,6 +9,14 @@ licenses(["notice"])  # Apache 2
 
 envoy_package()
 
+py_test(
+    name = "test_nighthawk_test_server",
```
oschaaf, line 13: I wonder if tests for our test-code should go into a distinct place for clarity (arguably it's not test-code as we offer it for production purposes too, but still)

benchmarks/envoy_proxy.py
```diff
@@ -92,9 +92,9 @@ def setUp(self):
                                                                port=proxy_server.server_port))
     self.proxy_server = proxy_server
 
-  def tearDown(self):
+  def tearDown(self, caplog):
```
oschaaf, line 95: This adds to the benchmarking framework, specifically when Envoy proxy gets injected between the client and test server. Is the plan to add to the other fixtures in a follow up? E.g. https://github.com/envoyproxy/nighthawk/blob/master/test/integration/integration_test_fixtures.py ?

oschaaf (commented):
looks great, just two questions/thoughts.

test/integration/nighthawk_test_server.py
mum4k: Thank you, if you do know the issue number, please let me know. Also your help in identifying background for the remaining ignored warnings would be much appreciated.

test/integration/BUILD
mum4k: Makes sense, how about we create a subdirectory like `integration/unit_tests` ? Open to suggestions.

benchmarks/envoy_proxy.py
mum4k: I thought this is what this PR was doing. We are only adding that parameter to the benchmarking framework, because it inherits from the same fixtures. This PR does modify `integration_test_fixtures.py` and that is where the failure on errors / warnings happens. Is this what you meant or am I missing something?

test/integration/nighthawk_test_server.py
mum4k: Note - we will be addressing some of the "Using deprecated extension name ..." warnings in my next PR. I will leave this in here as is for now, but leave the TODO and come back after the next PR gets merged in to see if additional cleanup is needed. Tracking in #582.

test/integration/BUILD
mum4k: Addressed in 6cd4657596aa0a0d383bfd63d65995a30c9aa84a.

mum4k:
Adding @dubious90 in case he can get to this sooner than @wjuan-AFK.

benchmarks/envoy_proxy.py
oschaaf: You are right, sorry about the noise, please disregard.

oschaaf (dismissed):
LGTM

wjuan-AFK:
LGTM

test/integration/integration_test_fixtures.py
```diff
@@ -134,6 +140,15 @@ def tearDown(self):
         any_failed = True
     assert (not any_failed)
 
+    messages = []
```
wjuan-AFK, line 143: Maybe this name could be slightly clearer about what kind of messages they are? Probably not necessary.

wjuan-AFK (dismissed):

test/integration/integration_test_fixtures.py
mum4k: Addressed in 61821dbaeca4c7cddf6923cc38f0f693b760ca37.

mum4k:
@oschaaf please take another look after addressing feedback and merging from master. Sorry about the force push, I have fat-fingered a bad rebase into this branch and had to rewrite history to fix it. Same commits were re-applies as cherry-picks.

oschaaf (approved):

mum4k:
@oschaaf please take another look after merging from master.

mum4k:
@oschaaf fyi this PR needs bit more work. Now that we do have a test which verifies we pass with `v2` config if the user requests it, we fail because that warning appears in the server logs. Going to update this so that we can filter out messages per specific set of test cases.

mum4k:
@oschaaf this is ready for another review. We can now specify distinct ignore lists based on the name of the executed test case. The test case that purposefully runs with the deprecated Envoy v2 API is now allowed to log the relevant warnings.

mum4k:
@wjuan-AFK significant changes were made since your last review to allow this to pass tests after #584 was merged in. For your convenience, [here is a view](https://github.com/envoyproxy/nighthawk/pull/581/files/2c3597657f93b3253074759391159502a6106ccd..e10c7b368f47bd174dec4f4b37610fe8f0b2a880) that only includes commits related to this change.

oschaaf (approved):
Nice, this will be very useful, LGTM

wjuan-AFK:
LGTM

Issue #580, opened 2020-12-01
Update Nighthawk's use of Envoy APIs to v3


mum4k:
Nighthawk uses now deprecated v2 API in a couple of locations:

- [x] configurations for integration tests
- [x] configuration for the benchmark
- [x] command line parameters
- [ ] locations in our proto API where we use v2 types
- [x] documentation and examples

Envoy v2 API is now deprecated:
https://github.com/envoyproxy/envoy/pull/13950

We should migrate to using v3 in all these places.

mum4k:
There are two locations in our API where we reuse the Envoy's `v2` types:

https://github.com/envoyproxy/nighthawk/blob/f5f88e31539dd7522a6c5f8d0ef095c085343491/api/server/response_options.proto#L23
https://github.com/envoyproxy/nighthawk/blob/f5f88e31539dd7522a6c5f8d0ef095c085343491/api/request_source/service.proto#L42

Going to figure out what is the long-term plan for these types and if deprecating these fields in favor of their `v3` versions is necessary in the short term.

PR #583, opened 2020-12-02
Migrating configs of integration tests to Envoy API v3.

mum4k:
Verified that all integration tests pass after updating Envoy to commit `588d9344b31e6544869547c4bcd359b3b0f1d4cf`, so this PR unblocks #575. Our next steps will be adding a good warning and a compatibility flag for users of Nighthawk. If they do send configs with Envoy API v2, we will break by default, but allow them to continue with the compatibility flag.

Summary of performed changes:
- changing `config` to `typed_config` and listing the correct type.
- migrating from deprecated field `tls_context` to `transport_socket`.
- changing filter names to ones that match extension names in [extensions_build_config.bzl](https://github.com/envoyproxy/nighthawk/blob/master/extensions_build_config.bzl).
- cosmetic changes of enum value from `auto` to `AUTO`.

Also:
- updating README and help displayed by the CLI in regards to passing in the `--tls-context` flag since this behavior is mirrored by one of the edited integration tests.
- Adding the `test_request_source_plugin.py` integration test as a dependency of the `integration_test` py_binary which was forgotten before.

Works on #580

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,8 +198,8 @@ any other value will allow client-side queuing of requests).
 Transport socket configuration in json or compact yaml. Mutually
 exclusive with --tls-context. Example (json):
 {name:"envoy.transport_sockets.tls"
-,typed_config:{"@type":"type.googleapis.com/envoy.extensions.transpor
-t_sockets.tls.v3.UpstreamTlsContext"
+,typed_config:{"@type":"type.googleapis.com/envoy.extensions.transport
+_sockets.tls.v3.UpstreamTlsContext"
 ,common_tls_context:{tls_params:{cipher_suites:["-ALL:ECDHE-RSA-AES128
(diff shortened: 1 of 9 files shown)


mum4k:

eric846:
LGTM

eric846 (approved):
LGTM

oschaaf (approved):
LGTM, but to make sure it's not left out by accident: will https://github.com/envoyproxy/nighthawk/blob/master/benchmarks/configurations/envoy_proxy.yaml go in a follow up?

mum4k:
Confirming that https://github.com/envoyproxy/nighthawk/blob/master/benchmarks/configurations/envoy_proxy.yaml will be addressed in the next PR.

PR #584, opened 2020-12-02
Add a flag to allow for continued use of the deprecated v2 api

oschaaf:
- Adds a new proto/cli option, which allows requesting continued use of the about to be deprecated v2 api's. 
- Adds a new fixture / integration test for the test server, as a sanity check for making sure that when we update Envoy to the 
  revision that deprecated v2 things will keep working as expected, while documenting how to set up the runtime configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Where:
 
 --allow-envoy-deprecated-v2-api
 Set to allow usage of the v2 api. (Not recommended, support will stop
-soon). Default: false
+in Q1 2021). Default: false
 
 --latency-response-header-name <string>
 Set an optional header name that will be returned in responses, whose
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--allow-v2-api]
+bazel-bin/nighthawk_client  [--allow-envoy-deprecated-v2-api]
 [--latency-response-header-name <string>]
 [--stats-flush-interval <uint32_t>]
 [--stats-sinks <string>] ...
@@ -84,8 +84,9 @@ format>
 
 Where:
 
---allow-v2-api
-Set to allow usage of the v2 api. (Not recommended). Default: false
+--allow-envoy-deprecated-v2-api
+Set to allow usage of the v2 api. (Not recommended, support will stop
+soon). Default: false
 
 --latency-response-header-name <string>
 Set an optional header name that will be returned in responses, whose
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ bazel build -c opt //:nighthawk
 ```
 USAGE:
 
-bazel-bin/nighthawk_client  [--latency-response-header-name <string>]
+bazel-bin/nighthawk_client  [--allow-v2-api]
+[--latency-response-header-name <string>]
 [--stats-flush-interval <uint32_t>]
 [--stats-sinks <string>] ...
 [--no-duration] [--simple-warmup]
@@ -82,6 +83,9 @@ format>
 
 
 Where:
+
+--allow-v2-api
+Set to allow usage of the v2 api. (Not recommended). Default: false
 
 --latency-response-header-name <string>
 Set an optional header name that will be returned in responses, whose
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -220,5 +220,5 @@ message CommandLineOptions {
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
   // Set to allow usage of the v2 api (not recommended). Default is false.
-  google.protobuf.BoolValue allow_api_v2 = 38 [deprecated = true];
+  google.protobuf.BoolValue allow_envoy_deprecated_v2_api = 38 [deprecated = true];
 }
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -105,7 +105,7 @@ message H1ConnectionReuseStrategy {
 
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
-// highest unused number is 38
+// highest unused number is 39
 message CommandLineOptions {
   // The target requests-per-second rate. Default: 5.
   google.protobuf.UInt32Value requests_per_second = 1
@@ -219,4 +219,6 @@ message CommandLineOptions {
   // "emit_previous_request_delta_in_response_header" to record elapsed time between request
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
+  // Set to allow usage of the v2 api (not recommended). Default is false.
+  google.protobuf.BoolValue allow_api_v2 = 38 [deprecated = true];
 }
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -219,6 +219,6 @@ message CommandLineOptions {
   // "emit_previous_request_delta_in_response_header" to record elapsed time between request
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
-  // Set to allow usage of the v2 api (not recommended). Default is false.
+  // Set to allow usage of the v2 api. (Not recommended, support will stop in Q1 2021).
   google.protobuf.BoolValue allow_envoy_deprecated_v2_api = 38 [deprecated = true];
 }
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -270,7 +270,7 @@ ProcessImpl::mergeWorkerStatistics(const std::vector<ClientWorkerPtr>& workers) 
   return merged_statistics;
 }
 
-void ProcessImpl::allowApiV2(envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+void ProcessImpl::allowEnvoyDeprecatedV2Api(envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
   auto* admin_layer = bootstrap.mutable_layered_runtime()->add_layers();
   admin_layer->set_name("admin layer");
   admin_layer->mutable_admin_layer();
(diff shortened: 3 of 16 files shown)


mum4k:

source/client/process_impl.cc
```diff
@@ -270,10 +270,31 @@ ProcessImpl::mergeWorkerStatistics(const std::vector<ClientWorkerPtr>& workers)
...
+  auto* admin_layer = bootstrap.mutable_layered_runtime()->add_layers();
+  admin_layer->set_name("admin layer");
+  admin_layer->mutable_admin_layer();
+  envoy::config::bootstrap::v3::RuntimeLayer* runtime_layer =
+      bootstrap.mutable_layered_runtime()->add_layers();
+  runtime_layer->set_name("static_layer");
+  Envoy::ProtobufWkt::Value proto_true;
+  proto_true.set_string_value(std::string("true"));
+  (*runtime_layer->mutable_static_layer()
+        ->mutable_fields())[std::string("envoy.reloadable_features.enable_deprecated_v2_api")] =
```
dubious90, line 283: why does this string need to be wrapped in a std::string here?

In general, is there a way to simplify this line, or break it into multiple lines?  After some reading i've determined that mutable_fields is likely a map of sorts, and envoy.reloadable_features.enable_deprecated_v2_api is a key in it that we need to set to a boolean proto (proto_true), but the line right now is a little unwieldy.

source/client/process_impl.cc
```diff
@@ -270,10 +270,31 @@ ProcessImpl::mergeWorkerStatistics(const std::vector<ClientWorkerPtr>& workers)
...
+      proto_true;
+  (*runtime_layer->mutable_static_layer()
+        ->mutable_fields())[std::string("envoy.features.enable_all_deprecated_features")] =
```
dubious90, line 286: Just for context, do we really need to enable all deprecated features to enable v2 api?  That seems much more far reaching.

If so, I wonder if we should instead create a generic functionality to enable deprecated features, rather than specifically enabling v2 api.

test/integration/integration_test_fixtures.py
```diff
@@ -297,6 +300,25 @@ def getTestServerRootUri(self):
     return super(HttpIntegrationTestBase, self).getTestServerRootUri(False)
 
 
+class HttpIntegrationTestBaseWithV2Bootstrap(IntegrationTestBase):
+  """Base for running plain http tests against the Nighthawk test server.
+
+  NOTE: any script that consumes derivations of this, needs to needs also explictly
```
dubious90, line 306: two typo:  "needs to needs" and sp: explicitly

source/client/process_impl.cc
oschaaf: There was some code in Envoy that inspired that, but wrapping in std::string indeed isn't needed. Removed!

source/client/process_impl.cc
oschaaf: Good catch, we don't need to enable all deprecated features, we can target just what we need (allowing usage of the deprecated v2 api). Done.

test/integration/integration_test_fixtures.py
oschaaf: Updated here and in another spot.

api/client/options.proto
```diff
@@ -219,4 +219,6 @@ message CommandLineOptions {
   // "emit_previous_request_delta_in_response_header" to record elapsed time between request
   // arrivals.
   google.protobuf.StringValue latency_response_header_name = 36;
+  // Set to allow usage of the v2 api (not recommended). Default is false.
```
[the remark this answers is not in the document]
oschaaf: As discussed on Slack, I will follow up on this in #575

test/integration/test_integration_basics.py
```diff
@@ -86,23 +86,26 @@ def test_nighthawk_client_v2_api_explicitly_set(http_test_server_fixture):
...
   counters = http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
-  asserts.assertCounterEqual(counters, "benchmark.http_2xx", 25)
+  asserts.assertCounterEqual(counters, "benchmark.pool_connection_failure", 1)
 
 
 # TODO(oschaaf): This ought to work after the Envoy update.
 def DISABLED_test_nighthawk_client_v2_api_breaks_by_default(http_test_server_fixture):
   """Test that the v2 api breaks us when it's not explicitly requested."""
-  parsed_json, _ = http_test_server_fixture.runNighthawkClient([
+  _, _ = http_test_server_fixture.runNighthawkClient([
```
dubious90, line 101: nit and also mildly unfamiliar language for me, but if we don't need either variable here, can we just run the function without assigning the result to anything?

dubious90 (approved):
LGTM, mum4k to approve

README.md
```diff
@@ -83,6 +84,9 @@ format>
 
 Where:
 
+--allow-v2-api
+Set to allow usage of the v2 api. (Not recommended). Default: false
```
mum4k, line 88: Can we expand the comment indicating the date of the eventual death of v2 support?

README.md
```diff
@@ -83,6 +84,9 @@ format>
 
 Where:
 
+--allow-v2-api
```
mum4k, line 87: Can we include the subject in the flag's name, e.g. Envoy? I would also include the fact that it is deprecated.

Something like: `--allow-envoy-deprecated-v2-api`.

test/integration/configurations/nighthawk_http_origin_v2_api.yaml
```diff
@@ -0,0 +1,38 @@
+admin:
```
mum4k, line 1: Can we add a comment up here saying that this file is intentionally using v2 api because ... ?

test/integration/integration_test_fixtures.py
```diff
@@ -56,13 +56,15 @@ class IntegrationTestBase():
...
 
-  def __init__(self, ip_version, server_config, backend_count=1):
+  def __init__(self, ip_version, server_config, backend_count=1, bootstrap_version_arg=None):
     """Initialize the IntegrationTestBase instance.
 
     Args:
       ip_version: a single IP mode that this instance will test: IpVersion.IPV4 or IpVersion.IPV6
       server_config: path to the server configuration
       backend_count: number of Nighthawk Test Server backends to run, to allow testing MultiTarget mode
+      bootstrap_version_arg: specify a bootstrap cli argument value for the test server binary.
```
mum4k, line 66: Can we specify the type in the comment? I am assuming this is an int, so something like:

```
bootstrap_version_arg: An int, specifies a bootstrap cli argument value for the test ....
```

And once we do that, can we also make the default argument of the same type rather than None? None is of a different type which can lead to unexpected NoneType errors.

test/integration/integration_test_fixtures.py
```diff
@@ -297,6 +301,25 @@ def getTestServerRootUri(self):
...
+  NOTE: any script that consumes derivations of this, needs to also explicitly
+  import server_config, to avoid errors caused by the server_config not being found
+  by pytest.
+  """
+
+  def __init__(self, ip_version, server_config):
+    """See base class."""
+    super(HttpIntegrationTestBaseWithV2Bootstrap, self).__init__(ip_version,
+                                                                 server_config,
+                                                                 bootstrap_version_arg="2")
```
mum4k, line 316: This actually looks like we are making it a string containing a number, can we use an int if we carry a numerical value?

test/integration/integration_test_fixtures.py
```diff
@@ -378,6 +401,19 @@ def http_test_server_fixture(request, server_config):
   f.tearDown()
 
 
+@pytest.fixture(params=determineIpVersionsFromEnvironment())
+def http_test_server_fixture_v2_api(request, server_config):
```
mum4k, line 405: Can we include the word "envoy" here also? And maybe "deprecated" too? E.g. `http_test_server_fixture_envoy_deprecated_v2_api`.

test/integration/integration_test_fixtures.py
```diff
@@ -378,6 +401,19 @@ def http_test_server_fixture(request, server_config):
...
+  """Fixture for setting up a test environment with http server configuration that uses v2 configuration.
+
+  Yields:
+      HttpIntegrationTestBaseWithV2Bootstrap: A fully set up instance. Tear down will happen automatically.
+  """
+  f = HttpIntegrationTestBaseWithV2Bootstrap(request.param, server_config)
```
mum4k, line 411: I would suggest we do the same name change for the class name here and above.

test/integration/nighthawk_test_server.py
```diff
@@ -63,6 +70,7 @@ def __init__(self, server_binary_path, config_template_path, server_ip, ip_versi
         server_binary_config_path_arg (str): Specify the name of the CLI argument the test server binary uses to accept a configuration path.
         parameters (dict): Supply to provide configuration template parameter replacement values.
         tag (str): Supply to get recognizeable output locations.
+        bootstrap_version_arg (str, optional): specify a bootstrap cli argument value for the test server binary.
```
mum4k, line 73: Same comment about making it an int and making the default value a numeric rather than a None.

We can still keep the meaning, e.g. the default could be zero and the code below can still do:
if self._bootstrap_version_arg:
  ...

Since zero evaluates to false. We can specify here in the comment what the use of zero value causes.

test/integration/test_integration_basics.py
```diff
@@ -69,6 +68,46 @@ def test_http_h1(http_test_server_fixture):
   asserts.assertEqual(len(counters), 12)
 
 
+@pytest.mark.parametrize(
+    'server_config',
+    ["nighthawk/test/integration/configurations/nighthawk_http_origin_v2_api.yaml"])
+def test_nighthawk_test_server_v2_api(http_test_server_fixture_v2_api):
```
mum4k, line 74: Comment about including "envoy" and "deprecated" applies here and below too.

Consider renaming the config file also.

mum4k (changes requested):

test/integration/integration_test_fixtures.py
oschaaf: Changing this to `int`, but still allowing `None`:  this will end up as an optional CLI argument value, where one can indicate that the this cli argument should be omitted altogether gets expressed by passing `None`. Let me know if you are comfortable with that.

test/integration/nighthawk_test_server.py
oschaaf: Ah, I see, that makes sense. Following up.

oschaaf:
Oh, this needs a merge / conflict resolution, let me fix that.

test/integration/integration_test_fixtures.py
mum4k: That makes sense, thank you for the explanation.

mum4k: As discussed offline, based on the explanation if actually made more sense to default the parameter to `None` if its meaning is **unset**. Sorry about the back and forth.

oschaaf:
@mum4k pushed 15512ef to resolve the `0` vs `None` issue with the bootstrap version argument.

mum4k (approved):

PR #586, opened 2020-12-03
Updating documentation to use Envoy API v3.

mum4k:
Verified that an Envoy built after `588d9344b31e6544869547c4bcd359b3b0f1d4cf` can load the configuration from the documents.

Also:
- fixing one typo in the documentation.
- cosmetic changes of enum values to uppercase form.

Works on #580

### DIFF
--- a/source/server/README.md
+++ b/source/server/README.md
@@ -34,10 +34,11 @@ static_resources:
           port_value: 10000
       filter_chains:
         - filters:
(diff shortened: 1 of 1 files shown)


mum4k:
@oschaaf I am a bit suspicious of the fact that I only found a single document location that needed an update. Please point me to others if I missed any.

oschaaf (approved):
Great, thanks!

PR #588, opened 2020-12-04
Deprecate Envoy API v2 primitives in configuration for our Envoy filters.

mum4k:
Adds a new field `v3_response_headers` alongside the old v2 `response_headers` in `api/server/response_options.proto` and marks the old field as deprecated. The old field retains its functionality for backward compatibility.

Since these are repeated fields, we cannot use a oneof. Instead validation is added to ensure the filters report an error on configuration that has both the v2 and the v3 fields set. `Envoy::EnvoyException` is thrown on validation errors to comply with the `Envoy::Server::Configuration::NamedHttpFilterConfigFactory` interface.

Also:
- sorting imports in `response_options.proto ` alphabetically.
- adding missing anonymous namespace in `http_dynamic_delay_filter_integration_test.cc`.

Works on #580.

### DIFF
--- a/api/server/response_options.proto
+++ b/api/server/response_options.proto
@@ -2,10 +2,11 @@ syntax = "proto3";
 
 package nighthawk.server;
 
+import "envoy/api/v2/core/base.proto";
+import "envoy/config/core/v3/base.proto";
+import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 import "validate/validate.proto";
-import "envoy/api/v2/core/base.proto";
-import "google/protobuf/duration.proto";
 
 message ConcurrencyBasedLinearDelay {
   // Minimal delay to add to replies.
@@ -20,7 +21,15 @@ message ConcurrencyBasedLinearDelay {
 // configuration will override.
 message ResponseOptions {
   // List of additional response headers.
-  repeated envoy.api.v2.core.HeaderValueOption response_headers = 1;
+  //
+  // Envoy deprecated its v2 API, prefer to use v3_response_headers instead.
+  // Mutually exclusive with v3_response_headers.
+  repeated envoy.api.v2.core.HeaderValueOption response_headers = 1 [deprecated = true];
+
+  // List of additional response headers.
+  // Mutually exclusive with response_headers.
+  repeated envoy.config.core.v3.HeaderValueOption v3_response_headers = 7;
+
   // Number of 'a' characters in the the response body.
   uint32 response_body_size = 2 [(validate.rules).uint32 = {lte: 4194304}];
   // If true, then echo request headers in the response body.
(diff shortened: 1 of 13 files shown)


mum4k:

api/server/response_options.proto
```diff
@@ -20,7 +21,15 @@ message ConcurrencyBasedLinearDelay {
 // configuration will override.
 message ResponseOptions {
   // List of additional response headers.
-  repeated envoy.api.v2.core.HeaderValueOption response_headers = 1;
+  //
+  // Envoy deprecated its v2 API, prefer to use v3_response_headers instead.
+  // Mutually exclusive with v3_response_headers.
+  repeated envoy.api.v2.core.HeaderValueOption response_headers = 1 [deprecated = true];
```
oschaaf, line 27: Similar thought here as in the other PR: the actual definitions for the HeaderValueOption fields don't seem to change. I wonder if we can leverage that somehow?

test/server/http_time_tracking_filter_integration_test.cc
```diff
@@ -37,6 +39,22 @@ class HttpTimeTrackingIntegrationTest
 INSTANTIATE_TEST_SUITE_P(IpVersions, HttpTimeTrackingIntegrationTest,
                          testing::ValuesIn(Envoy::TestEnvironment::getIpVersionsForTest()));
 
+TEST_P(HttpTimeTrackingIntegrationTest,
+       DiesWhenBothEnvoyApiV2AndV3ResponseHeadersAreSetInConfiguration) {
+  const std::string invalid_configuration = R"EOF(
+  name: time-tracking
+  typed_config:
+    "@type": type.googleapis.com/nighthawk.server.ResponseOptions
+    response_headers:
```
oschaaf, line 48: Here a user needs to put the exact same configuration contents into a field with a different name. 
I wonder, will protocol buffers even be able to distinct between the two types as it deserializes them? If it does, can we bully it into mechanically converting it, based on our knowledge that the messages are actually defined identically?

oschaaf (approved):
Similar thought as in #589 that I'm bring up for consideration, other then that looks good.

jiajunye (commented):
LGTM

api/server/response_options.proto
mum4k: Our goal with this and the other PR is to avoid forcing our users to import and use deprecated Envoy v2 API primitives if they don't want to. Since we exposed the v2 primitives on our API, this does seem like the best way forward. If this becomes a burden, we can start thinking about hiding the Envoy API primitives. Considering that would mean more maintenance and the fact that Envoy isn't going to perform an update like this often, we are probably better off with the current solution.

test/server/http_time_tracking_filter_integration_test.cc
mum4k: Thank you for the offline discussion. The summarize the conclusion - considering our goal is to allow users to use v3 if they want to, performing any sort of serialization / deserialization trick wouldn't achieve it. On top of that even if the discussed approach would work, it doesn't follow the good practices set for proto messages which might cause unexpected results in the future.

PR #589, opened 2020-12-04
Deprecate Envoy API v2 primitives in the service proto for Request Source.

### LINKED ISSUE #580: Update Nighthawk's use of Envoy APIs to v3

mum4k:
Converting the current field into a `oneof` and marking the `v2` primitive as deprecated. The old field retains its functionality for backward compatibility.

Also:
- sorting imports in `api/request_source/service.proto`.
- migrating our dummy request source to using the `v3` primitive.
- improving error message emitted on test failures in `test/request_stream_grpc_client_test.cc`, the current message just dumps byte representation of the two compared header objects.

Fixes #580.

Signed-off-by: Jakub Sobon mumak@google.com

### DIFF
--- a/api/request_source/service.proto
+++ b/api/request_source/service.proto
@@ -2,8 +2,9 @@ syntax = "proto3";
 
 package nighthawk.request_source;
 
-import "google/protobuf/wrappers.proto";
 import "envoy/api/v2/core/base.proto";
+import "envoy/config/core/v3/base.proto";
+import "google/protobuf/wrappers.proto";
 
 // Used to request a RequestStreamResponse.
 message RequestStreamRequest {
@@ -37,9 +38,18 @@ message RequestSpecifier {
   // Request content length. The client will transfer the number of bytes specified here for the
   // request body.
   google.protobuf.UInt32Value content_length = 4;
-  // Request header replacements. Any existing header(s) with the same name will be removed
-  // before setting.
-  envoy.api.v2.core.HeaderMap headers = 5;
+
+  oneof oneof_headers {
+    // Request header replacements. Any existing header(s) with the same name will be removed
+    // before setting.
+    //
+    // Envoy deprecated its v2 API, prefer to use v3_headers instead.
+    envoy.api.v2.core.HeaderMap headers = 5 [deprecated = true];
+
+    // Request header replacements. Any existing header(s) with the same name will be removed
+    // before setting.
+    envoy.config.core.v3.HeaderMap v3_headers = 6;
+  }
   // TODO(oschaaf): nice to have
   // google.protobuf.StringValue sni_hostname = 10;
 }
--- a/test/request_stream_grpc_client_test.cc
+++ b/test/request_stream_grpc_client_test.cc
@@ -1,4 +1,9 @@
+#include "envoy/api/v2/core/base.pb.h"
+#include "envoy/config/core/v3/base.pb.h"
+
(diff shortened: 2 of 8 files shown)


mum4k:

test/request_stream_grpc_client_test.cc
```diff
@@ -42,14 +52,34 @@ TEST_F(ProtoRequestHelperTest, ExplicitFields) {
...
+  // always replace.
+  headers->add_headers()->MergeFrom(*header_2);
+  expected_header_ =
+      Envoy::Http::TestRequestHeaderMapImpl{{"header1", "value1"}, {"header2", "value2"}};
+  translateExpectingEqual();
+}
+
+// Test the generic header api we offer in the proto api using Envoy API v3
+// primitives.
+TEST_F(ProtoRequestHelperTest, GenericHeaderFieldsUsingDeprecatedEnvoyV3Api) {
```
dubious90, line 76: Remove word deprecated here?

dubious90 (approved):
Looks good modulo the one copy-paste error

api/request_source/service.proto
```diff
@@ -37,9 +38,18 @@ message RequestSpecifier {
   // Request content length. The client will transfer the number of bytes specified here for the
   // request body.
   google.protobuf.UInt32Value content_length = 4;
-  // Request header replacements. Any existing header(s) with the same name will be removed
-  // before setting.
-  envoy.api.v2.core.HeaderMap headers = 5;
+
+  oneof oneof_headers {
```
oschaaf, line 42: Comparing `HeaderMap` [v2](https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/core/base.proto) versus [v3](https://github.com/envoyproxy/envoy/blob/master/api/envoy/config/core/v3/base.proto), it seems that there aren't any actual changes to the message types that we depend upon, definitions seem identical to me.

That leads me to think we might be able to get away with a mechanical on the fly conversion if needed? The v3 messages have a `previous_message_type` attached in their definition, can we access that at runtime?

The on the fly v2/v3 conversion will not be free, but for the gRPC case I suspect that it won't bottleneck, so maybe this is worth considering.

oschaaf (approved):
LGTM, but one thought I wanted to float for consideration. LGTM if you lean towards keeping this as is.

test/request_stream_grpc_client_test.cc
mum4k: Done.

api/request_source/service.proto
mum4k: As discussed offline, this was on my radar when working on this PR, but considering the added complexity and deps for the reflection that would load and convert this message mechanically and the size of the message in question (just two fields), it didn't seem like the mechanical solution would pay for itself.

With that we are choosing to keep this as is since the current code is simple to understand and maintain.

mum4k:
@oschaaf @dubious90 Could one of you take another look and re-approve after addressing the comments? Thank you.

oschaaf (approved):

PR #598, opened 2020-12-16
Optimize calls to std::string::find() and friends for a single char.

yanavlasov:
The character literal overload is more efficient.

Signed-off-by: Yan Avlasov <yavlasov@google.com>


oschaaf (approved):

mum4k (approved):
Thank you.

Issue #181, opened 2019-10-28
Bubble up error detail in gRPC service


oschaaf:
Currently the gRPC service relies on `ProcessImpl` for performing executions, and that will result in a boolean to indicate success/failure. 
However, for the gRPC service it would be very useful to be able to be more specific when an execution is marked as a failure (especially when the termination predicate feature lands). Bubble up the error message so it can be forwarded in the error detail of gRPC replies.

PR #600, opened 2020-12-30
Improve Unknown Error 13 messages

eric846:
Split error conditions as much as possible and add user guidance for further troubleshooting.

Nighthawk Service will now return a more specific error description for the following cases, instead of "unknown," for error 13.

- URI exception, with user guidance to check for:
  - Malformed URI
  - DNS error
- Failure predicate, with user guidance to check the output counters to narrow down:
  - Connection failures
    - Bad port
    - Bad address family (--address-family v4 or v6) especially when using 'localhost' or DNS
  - Stream resets (http/https, --h2)
  - 4xx or 5xx received
  - gRPC RequestSource error

The CLI output will also include the same new text.

#181 #613


mum4k:
@eric846 looks like this is still work in progress, but please let me know if I misread it.

eric846:
> @eric846 looks like this is still work in progress, but please let me know if I misread it.

I don't plan any other changes to this code, but I'm still struggling to get the CI to pass. There was some trouble with different unit test behavior on my local machine and CI (filed #605 ).

mum4k:
@eric846 the CI seems to be showing failures still. Would you like to look at it before we commence the code review?

eric846:
Hi all,
What do you think of adding a default failure predicate that exits if there were any stream resets? We encounter stream resets a lot if the protocol is wrong (e.g. http/https mixup). Detecting stream resets could be helpful for troubleshooting the test setup, same as the motivation for the other default failure predicates. But if stream resets occur naturally while traffic is flowing, or users might be depending on tolerating stream resets by default, we can't do it.

oschaaf:
> Hi all, What do you think of adding a default failure predicate that exits if there were any stream resets? 
+1

eric846:
I'll reopen this after figuring out the test failures locally.

PR #602, opened 2021-01-05
Signal handler: don't fire on destruction

oschaaf:
Fixes a bug observed while working on horizontal scaling support:
SignalHandler would fire on destruction.
Fix & add tests.

Note: pre-emptive, I don't think this is  an issue in the current
state of master. But some of the CI errors in #600 seemed familiar
and hopefully merging this in there helps.

(Split out from https://github.com/oschaaf/nighthawk/tree/horizontal-scaling)

/cc @eric846

### DIFF
--- a/source/common/signal_handler.cc
+++ b/source/common/signal_handler.cc
@@ -23,7 +23,9 @@ SignalHandler::SignalHandler(const std::function<void()>& signal_callback) {
     RELEASE_ASSERT(close(pipe_fds_[0]) == 0, "read side close failed");
     RELEASE_ASSERT(close(pipe_fds_[1]) == 0, "write side close failed");
     pipe_fds_.clear();
-    signal_callback();
+    if (!destructing_) {
+      signal_callback();
+    }
   });
 
   signal_handler_delegate = [this](int) { onSignal(); };
@@ -32,6 +34,7 @@ SignalHandler::SignalHandler(const std::function<void()>& signal_callback) {
 }
 
 SignalHandler::~SignalHandler() {
+  destructing_ = true;
   initiateShutdown();
   if (shutdown_thread_.joinable()) {
     shutdown_thread_.join();
--- a/source/common/signal_handler.h
+++ b/source/common/signal_handler.h
@@ -69,6 +69,7 @@ private:
   // the read side will handle the actual shut down of the gRPC service without having to worry
   // about signal-safety.
   std::vector<int> pipe_fds_;
+  bool destructing_{false};
 };
 
 using SignalHandlerPtr = std::unique_ptr<SignalHandler>;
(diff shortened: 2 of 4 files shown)


mum4k:

source/common/signal_handler.cc
```diff
@@ -32,6 +34,7 @@ SignalHandler::SignalHandler(const std::function<void()>& signal_callback) {
 }
 
 SignalHandler::~SignalHandler() {
```
eric846, line 36: If there was a real signal, does the process die and prevent this desturctor from being called? If so, makes perfect sense.

eric846 (dismissed):
LGTM

qqustc (dismissed):

mum4k (approved):
Looks good, please address outstanding comments before we merge this in.

source/common/signal_handler.cc
oschaaf: So what this tries to fix is that without this change, the `DestructDoesNotFireHandler` test would fail, because here we would call the user-provided callback during destruction, even though no signal was raised. 

(This isn't a practical problem today when/where this happens because the callbacks themselves are resilient or called at a time where its effects no longer matter, but it is rather fragile)

So having said that, it sounds like you may have spotted a different tricky angle, but I'm not sure I follow; could you elaborate or rephrase to help me there?

eric846: Sorry, I think I had a completely wrong view of how this works. I assumed there was some system magic that doesn't actually exist. I mainly missed the fact that you call `initiateShutdown()` in the destructor. 

It's definitely a good thing to skip the callback.

So to make sure I understand, is the following how it works in two possible scenarios? Hopefully there's no longer any interesting angle to speak of. If not, totally LGTM.

- Nighthawk process finishes without getting a signal
  - Destructor is called
    - The destructor calls `initiateShutdown()` which wakes up the shutdown thread via the pipe
    - The shutdown thread is smart enough to skip the callback (as of this PR)
    - Pipes get deleted
- Nighthawk process gets a signal like SIGTERM
  - Signal handler calls `initiateShutdown()` which wakes up the shutdown thread via the pipe
    - Callback gets called
    - Pipes get deleted
  - Destructor is called (?)
    - Destructor calls `initiateShutdown()`
      - But now the pipes are gone and nothing happens (?)

oschaaf: That is all correct.  Revisiting this, this class could use a little love (code level docs, member renamings for clarity), it requires too much brainpower to follow as it stands. Thanks for persisting there, I'll file an issue, and your analysis here might then serve as a class level comment to guide others.

oschaaf: (Filed https://github.com/envoyproxy/nighthawk/issues/603)

mum4k (approved):

Issue #593, opened 2020-12-11
CI/ASAN build failure with -fundefined


oschaaf:
As of https://github.com/envoyproxy/nighthawk/pull/587 the asan build fails when `-fundefined` is passed with the following error:

```
ld.lld: error: undefined symbol: __muloti4
```

This seems related to https://github.com/envoyproxy/envoy/issues/13973, which was introduced via an update to
`absl`, and eventually was resolved in Envoy by bumping `absl`. Yet this error persist in Nighthawk. 

I've spend some time trying to figure out how to address this, without coming up with a good solution yet. 
To unblock, `-fundefined` gets dropped from [.bazelrc](https://github.com/envoyproxy/nighthawk/blob/master/.bazelrc) in https://github.com/envoyproxy/nighthawk/pull/587.

Filing this issue to track status of re-enabling `-fundefined`

eric846:
With this customization removed, ASAN tests failed repeatedly on 2023-09-10: https://github.com/envoyproxy/nighthawk/pull/1022.

PR #604, opened 2021-01-12
Update Envoy to c2522c69f (Jan 11th 2021).

mum4k:
- `Envoy::StreamInfo::StreamInfoImpl` no longer contains methods that set local and remote socket addresses. It instead requires a `Envoy::Network::SocketAddressProviderSharedPtr` with the addresses already set. Pass in a `Envoy::Network::SocketAddressSetterImpl` so that we can continue setting the remote address for tests.
- Copied `.bazelversion` and `.bazelrc` from Envoy. 


mum4k:
@qqustc please review and assign back to me when done.

oschaaf (commented):
The code changes LGTM, but checking I noticed that Envoy's https://github.com/envoyproxy/envoy/blob/master/.bazelversion and https://github.com/envoyproxy/envoy/blob/master/.bazelrc have fairly recent changes.
Do we want to mirror those in this update?

mum4k:
Thanks for pointing that out @oschaaf, copied `.bazelversion` and `.bazelrc` from Envoy.

.bazelrc
```diff
@@ -60,8 +60,8 @@ build:asan --config=sanitizer
 # ASAN install its signal handler, disable ours so the stacktrace will be printed by ASAN
 build:asan --define signal_trace=disabled
 build:asan --define ENVOY_CONFIG_ASAN=1
-build:asan --copt -fsanitize=address
-build:asan --linkopt -fsanitize=address
+build:asan --copt -fsanitize=address,undefined
```
oschaaf, line 63: Ugh, sorry, this is annoying: this particular difference is here because of https://github.com/envoyproxy/nighthawk/issues/593. I should have added a comment about that.

oschaaf (approved):
LGMT modulo my last remark

qqustc (dismissed):

.bazelrc
mum4k: Thanks, removed `undefined` and added. a comment.

mum4k:
@oschaaf please take another look and approve when ready.

oschaaf (approved):
LGTM

PR #607, opened 2021-01-25
Update Envoy to 9753819 (Jan 24th 2021).

mum4k:
- sync `.bazelrc` from envoy.
- `tryCreateNewConnection` now returns a `ConnPoolImplBase::ConnectionResult` instead of a `bool`.


oschaaf (approved):
LGTM, assuming that the PR that updates the docker version we use in CI resolves the CI glitch.

PR #609, opened 2021-01-25
Updating repository paths in README under benchmarks.

mum4k:
Changing github repository paths from oschaaf to envoyproxy.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -26,7 +26,7 @@ It will run a selection of an example [benchmarks](test/test_discovery.py)
 extracted from `/benchmarks`, which injects Envoy between the benchmark client and test server.
 
 ```bash
-git clone https://github.com/oschaaf/nighthawk.git benchmark-test
+git clone https://github.com/envoyproxy/nighthawk.git benchmark-test
 cd benchmark-test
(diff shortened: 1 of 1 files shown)


mum4k:
@oschaaf there is also a reference to `oschaaf/benchmark-dev:latest` in the same readme. Is that the correct path for us to use?

oschaaf (approved):
Ah, yeah, good idea.

mum4k:
@oschaaf please see my question about the reference to `oschaaf/benchmark-dev:latest` in the same file.

oschaaf:
> @oschaaf there is also a reference to `oschaaf/benchmark-dev:latest` in the same readme. Is that the correct path for us to use?

Oh, sorry I missed that. I think that for the time being it's the best option, because we don't push the image in CI to a proper location. There's https://hub.docker.com/r/envoyproxy/nighthawk-dev, maybe we should try to also obtain https://hub.docker.com/r/envoyproxy/nighthawk-benchmark-dev (or some such) as a location to push to.

oschaaf:
(Filed https://github.com/envoyproxy/nighthawk/issues/611 to track)

mum4k:
Thank you @oschaaf!

PR #619, opened 2021-02-05
Experiment - revert envoy update to attempt to fix integration tests

dubious90:
Integration tests failing on main branch.  Seeing if reverting the latest envoy update is the root cause.


oschaaf:
Hmm. is the tracking of https cipher usage broke (again)?
I think there's been a recent change in Envoy that *could* be related. 
But then it is really weird that CI didn't catch this before we merged.

oschaaf:
If my hunch is correct (which might very well be not!) the upstream changes that broke us are in these lines:
https://github.com/envoyproxy/envoy/blob/c16a033ddf81f5366cc0e590c67326e6696d5249/source/extensions/transport_sockets/tls/context_impl.cc#L329-L344

dubious90:
> Hmm. is the tracking of https cipher usage broke (again)?
> I think there's been a recent change in Envoy that _could_ be related.
> But then it is really weird that CI didn't catch this before we merged.

Tbh I haven't fully investigated yet.  The Docker failure threw a small wrench in my investigation efforts.  But the failures started shortly after I merged in the latest envoy update, so it was my first suspect.

Agree that it's weird that the CI didn't catch the problem.  Let's see how the system operates after we merge in the docker fix

mum4k:
Are we planning to merge this PR, or is it just an experiment to execute the CI? Asking, because merging it would prevent the next Envoy update that is already in review:
https://github.com/envoyproxy/nighthawk/pull/621

If it is just an experiment, we could mark it as draft or say something like "DO NOT MERGE" in the PR description.

mum4k:
Fyi, this is what fixed the last docker issue, this one looks related:

https://github.com/envoyproxy/nighthawk/pull/610

mum4k:
Sorry, reading emails in sequence - now I saw that we already merged a fix. Assuming we will be discarding this PR, however if merging it is necessary - let's discuss our options first.

PR #620, opened 2021-02-05
CI: Fix broken docker tasks

oschaaf:
I think the root cause is that the alpine-glibc image updates,
and through that dependency requirements can no longer be met.

This PR
- pins the alpine image at the latest version
- changes the dependency versions so they work again

Note: the `>=` for python was changed to `=` as otherwise we'd
write a file to disk mentioning the version, this seemed contra
intent.

There's a complaint [1] emitted during the image build, so we need to
see if everything is OK when CI finishes.

[1] /usr/glibc-compat/sbin/ldconfig: /usr/glibc-compat/lib/ld-linux-x86-64.so.2 is not a symbolic link

### DIFF
--- a/ci/docker/Dockerfile-nighthawk
+++ b/ci/docker/Dockerfile-nighthawk
@@ -1,4 +1,4 @@
-FROM frolvlad/alpine-glibc
+FROM frolvlad/alpine-glibc:alpine-3.13_glibc-2.32
 
 ADD nighthawk_client /usr/local/bin/nighthawk_client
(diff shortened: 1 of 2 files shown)


oschaaf:
Thanks to @abaptiste for help with assessing the root cause.

dubious90:
Thanks Otto!  This looks good. Once I confirm the CI passes, I will merge this through

dubious90 (approved):

oschaaf:
Allright, CI all green here.

PR #621, opened 2021-02-06
Update Envoy to cfe0391 (Feb 5th 2021).

mum4k:
- synchronizing `.bazelrc` from Envoy repository.
- the API of `getJsonStringFromMessage` changed, it now returns `ProtobufUtil::StatusOr` instead of `std::string`. Switching our code over to `getJsonStringFromMessageOrDie` which continues to return `std::string`.
- Increased the `no_output_timeout` parameter to 30m for test_gcc which was timing out.

### DIFF
--- a/source/client/output_formatter_impl.cc
+++ b/source/client/output_formatter_impl.cc
@@ -143,7 +143,7 @@ std::string ConsoleOutputFormatterImpl::statIdtoFriendlyStatName(absl::string_vi
 }
 
 std::string JsonOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
-  return Envoy::MessageUtil::getJsonStringFromMessage(output, true, true);
+  return Envoy::MessageUtil::getJsonStringFromMessageOrDie(output, true, true);
 }
 
 std::string YamlOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
@@ -320,7 +320,7 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
   if (statistic != nullptr) {
     fortio_output.mutable_headersizes()->CopyFrom(renderFortioDurationHistogram(*statistic));
   }
-  return Envoy::MessageUtil::getJsonStringFromMessage(fortio_output, true, true);
+  return Envoy::MessageUtil::getJsonStringFromMessageOrDie(fortio_output, true, true);
 }
 
 const nighthawk::client::DurationHistogram FortioOutputFormatterImpl::renderFortioDurationHistogram(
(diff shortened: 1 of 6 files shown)


source/client/output_formatter_impl.cc
```diff
@@ -143,7 +143,7 @@ std::string ConsoleOutputFormatterImpl::statIdtoFriendlyStatName(absl::string_vi
 }
 
 std::string JsonOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
-  return Envoy::MessageUtil::getJsonStringFromMessage(output, true, true);
+  return Envoy::MessageUtil::getJsonStringFromMessageOrDie(output, true, true);
```
oschaaf, line 146: When does this die? Is that new,  and does it affect us?

oschaaf (dismissed):
It's almost as if CI became a bit slower to execute in the last couple of days somehow. The CI tasks we have that have historically been sensitive to that all started flaking.  Here test_gcc ends with `Too long with no output (exceeded 10m0s): context deadline exceeded`.  That seems like something that should resolve on a retry.

One question about the method that transforms messages to json for us; otherwise LGTM assuming that we just need to update the method names.

source/client/output_formatter_impl.cc
mum4k: The change can be examined in this Envoy commit:

https://github.com/envoyproxy/envoy/commit/de8c5059baf575aaafdcc42572f7b019824987ba

The way I understand it, the rename doesn't introduce any new risk into our codebase. We may want to eventually switch to the version that returns the status and propagate that status back to the caller. But considering that would likely also result in a crash, just elsewhere in the stack - it probably is just cosmetic.

Please take a look at the linked commit and let me know if your interpretation differs.

mum4k:
The `test_gcc` problem reoccured after a retry. Attempting to increase the `no_output_timeout` to see if it is just CI getting slower, or if there is another root cause.

source/client/output_formatter_impl.cc
oschaaf: Agreed. Also, any crash here would be in a CLI, and not a service, so while not beautiful it's not as critical.

FWIW, there's one crash path in this areay that I know if today before the rename here raised awareness of other paths: the Fortio output formatter asserts when there's no global result. We could do better probably, as we could generate Fortio output which its UI will interpret as an error. Maybe it's worth filing an issue tagged as good-first-issue to put solidifying these paths as a whole on the (low priority) backlog.

oschaaf (approved):

source/client/output_formatter_impl.cc
mum4k: Thank you @oschaaf, agreed with your assessment about filing issue. If you get a few, would you mind filing it and including the additional background you have on those paths?

mum4k:
Thank you for filing the issue @oschaaf.

PR #623, opened 2021-02-06
Stats: Hdr histogram serialization/deserialization

oschaaf:
Being able to serialize and deserialize statistics serves several goals:
- Allow over-the-wire transportation in horizontally scaled setups.
- Facilitate storage of high res data.
- Could be used as a means to decouple our workers from the main thread,
  by avoiding reliance on a shared process namespace to merge statistics.
  (not a goal right now)

This adds an abstraction to Statistic, implements it for HdrHistogram, and
adds some tests.

Part of horizontal scaling effort, split out from:
Split out from https://github.com/oschaaf/nighthawk/tree/horizontal-scaling

### DIFF
--- a/include/nighthawk/common/statistic.h
+++ b/include/nighthawk/common/statistic.h
@@ -11,6 +11,7 @@
 
 #include "api/client/output.pb.h"
 
+#include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
 
 namespace Nighthawk {
@@ -115,6 +116,23 @@ public:
    * @param id The id that should be set for the Statistic instance.
    */
   virtual void setId(absl::string_view id) PURE;
+
+  /**
+   * Build a string representation of this Statistic instance.
+   *
+   * @return absl::StatusOr<std::unique_ptr<std::istream>> Status or a stream that will yield
+   * a serialized representation of this Statistic instance.
+   */
+  virtual absl::StatusOr<std::unique_ptr<std::istream>> serializeNative() const PURE;
+
+  /**
+   *  Reconstruct this Statistic instance using the serialization delivered by the input stream.
+   *
+   * @param input_stream Stream that will deliver a serialized representation.
+   * @return absl::Status Status indicating success or failure. Upon success the statistic
+   * instance this was called for will now represent what the stream contained.
+   */
+  virtual absl::Status deserializeNative(std::istream& input_stream) PURE;
 };
 
 } // namespace Nighthawk
--- a/source/common/statistic_impl.cc
+++ b/source/common/statistic_impl.cc
@@ -2,8 +2,10 @@
 
 #include <cmath>
 #include <cstdio>
+#include <fstream>
 #include <sstream>
 
+#include "external/dep_hdrhistogram_c/src/hdr_histogram_log.h"
 #include "external/envoy/source/common/common/assert.h"
 #include "external/envoy/source/common/protobuf/utility.h"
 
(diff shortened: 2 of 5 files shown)


mum4k:

jiajunye:
lgtm

include/nighthawk/common/statistic.h
```diff
@@ -115,6 +116,23 @@ class Statistic : Envoy::NonCopyable {
...
    */
   virtual void setId(absl::string_view id) PURE;
+
+  /**
+   * Build a string representation of this Statistic instance.
+   *
+   * @return absl::StatusOr<std::unique_ptr<std::istream>> Status or a stream that will yield
+   * a serialized representation of this Statistic instance.
+   */
+  virtual absl::StatusOr<std::unique_ptr<std::istream>> serializeNative() const PURE;
```
mum4k, line 126: Curious - what advantage do we gain from serializing into a `std::istream` rather than `std::string`? Is it something that helps the reader / user? Since it doesn't seem to be helping writer much.

source/common/statistic_impl.cc
```diff
@@ -236,6 +246,31 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
   return proto;
 }
 
+absl::StatusOr<std::unique_ptr<std::istream>> HdrStatistic::serializeNative() const {
+  char* data;
+  if (hdr_log_encode(histogram_, &data) == 0) {
+    absl::string_view s(data, strlen(data));
+    auto ss = std::make_unique<std::stringstream>();
```
mum4k, line 253: Can we use more descriptive names instead of just `s` and `ss`?

source/common/statistic_impl.cc
```diff
@@ -236,6 +246,31 @@ nighthawk::client::Statistic HdrStatistic::toProto(SerializationDomain domain) c
...
+    free(data);
+    return ss;
+  }
+  ENVOY_LOG(error, "Failed to write HdrHistogram data.");
+  return absl::Status(absl::StatusCode::kInternal, "Failed to write HdrHistogram data");
+}
+
+absl::Status HdrStatistic::deserializeNative(std::istream& stream) {
+  std::string s(std::istreambuf_iterator<char>(stream), {});
+  struct hdr_histogram* new_histogram = nullptr;
```
mum4k, line 264: Can you help me understand the memory here? Who allocates and owns (eventually frees) the memory pointed o by `new_histogram`?

mum4k (changes requested):

include/nighthawk/common/statistic.h
oschaaf: I agree that for this implementation for HdrHistogram we could use `std::string`. 
I was hoping that using streams here leaves some wiggle room to move to chunked reading and writing (and maybe use a stream wrapper to apply compression) if at some point another stats implementation implements this, and needs to write large volumes of data.
At the extreme, I could see that becoming useful if we would want to serialize a raw latency stream at some point.

source/common/statistic_impl.cc
oschaaf: Done in 79bad46

source/common/statistic_impl.cc
oschaaf: Added some comments in 79bad46 to help with that.

include/nighthawk/common/statistic.h
mum4k: Thank you for explaining.

mum4k (approved):

PR #624, opened 2021-02-06
Add Sink concept plus a filesytem based implementation

oschaaf:
This PR has Sink, MockSink, FileSinkImpl & tests.

Sink will serve as a backing store for ExecutionResponse proto representations.

Context:
- Later on a new service will expose storage and lookup methods.
- This in turn can be used in a horizontally scaled system as a means to
realize a central point to transport results to and from.

Part of the horizontal scaling effort.
Split out from https://github.com/oschaaf/nighthawk/tree/horizontal-scaling


mum4k:

test/common/sink_test.cc
```diff
@@ -0,0 +1,138 @@
...
+};
+
+TYPED_TEST_SUITE(TypedSinkTest, SinkTypes);
+
+TYPED_TEST(TypedSinkTest, BasicSaveAndLoad) {
+  TypeParam sink;
+  nighthawk::client::ExecutionResponse result_to_store;
+  *(result_to_store.mutable_execution_id()) = this->executionIdForTest();
+  absl::Status status = sink.StoreExecutionResultPiece(result_to_store);
+  ASSERT_EQ(status.ok(), true);
```
eric846, line 33: nit: If `ASSERT_TRUE` / `ASSERT_FALSE` are available, it seems more standard (here and below).

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+  for (const auto& it :
+       std::filesystem::directory_iterator(filesystem_directory_path, error_code)) {
+    if (error_code.value()) {
+      break;
+    }
+    ::nighthawk::client::ExecutionResponse response;
+    std::ifstream ifs(it.path(), std::ios_base::binary);
+    if (!response.ParseFromIstream(&ifs)) {
+      return absl::Status(absl::StatusCode::kInternal,
+                          fmt::format("Failure reading/parsing '{}'.", it.path()));
```
eric846, line 97: nit: It could be helpful for debugging to mention `ExecutionResponse` in this error message.

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+  }
+  ENVOY_LOG_MISC(trace, "Stored '{}'.", new_name);
+  return absl::Status();
+}
+
+const absl::StatusOr<std::vector<::nighthawk::client::ExecutionResponse>>
+FileSinkImpl::LoadExecutionResult(absl::string_view execution_id) const {
+  absl::Status status = verifyStringIsGuid(execution_id);
+  if (!status.ok()) {
+    return status; //
```
eric846, line 81: nit: Extraneous `//` or forgotten comment?

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+  std::array<char, L_tmpnam> name_buffer;
+  // TODO(oschaaf): tmpname complaint from compiler, dangerous.
+  if (!std::tmpnam(name_buffer.data())) {
+    return absl::Status(absl::StatusCode::kInternal, "Failure creating temp file");
+  }
+  // Next we write to a tmp file, and if that succeeds, we swap it atomically to the target path.
+  {
+    std::ofstream ofs(name_buffer.data(), std::ios_base::out | std::ios_base::binary);
+    if (!response.SerializeToOstream(&ofs)) {
+      return absl::Status(absl::StatusCode::kNotFound, "Failure writing to temp file");
```
eric846, line 63: Not sure if this is status code distinction is intentional.

test/common/sink_test.cc
```diff
@@ -0,0 +1,138 @@
...
+  ASSERT_EQ(status_or_execution_responses.ok(), false);
+  EXPECT_EQ(status_or_execution_responses.status().code(), absl::StatusCode::kInvalidArgument);
+  EXPECT_EQ(
+      status_or_execution_responses.status().message(),
+      "'14e75b2a-3e31-4x62-9279-add1e54091f9' is not a guid: unexpected character encountered.");
+}
+
+TEST(FileSinkTest, CorruptedFile) {
+  FileSinkImpl sink;
+  const std::string execution_id = "14e75b2a-3e31-4162-9279-add1e54091f9";
```
eric846, line 117: Any reason not to use `this->executionIdForTest()` here?

test/common/sink_test.cc
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Nighthawk {
+namespace {
+
+// Future sink implementations register here for testing top level generic sink behavior.
+using SinkTypes = ::testing::Types<FileSinkImpl>;
+
+template <typename T> class TypedSinkTest : public ::testing::Test {
+public:
+  void SetUp() override { uuid_ = random_.uuid(); }
```
eric846, line 18: Could we automatically clean up `"/tmp/nh/" + executionIdForTest() + "/"` in `TearDown`? (if it makes sense)

eric846 (approved):
LGTM with some minor points that are all optional.

test/common/sink_test.cc
oschaaf: done: 00ebf87

source/common/sink_impl.cc
oschaaf: done: 94f7299

source/common/sink_impl.cc
oschaaf: removed in 94f7299

source/common/sink_impl.cc
oschaaf: No it wasn't, good catch. Changed in 94f7299

test/common/sink_test.cc
oschaaf: Well, this test is different from the others as it doesn't have a fixture. (it's not part of the TEST_P ones)

test/common/sink_test.cc
oschaaf: Done in 94f7299. It's a little odd to do it here, as this may execute for other sink implementations when the TEST_P suites iterate over their parameterization, and will not write files to that location. But cleaning up is rather nice, and this won't cause trouble. 
So it seems pragmatic just add the two lines to clean up here.

oschaaf:
Thanks for the review @eric846

test/common/sink_test.cc
eric846: Ohh I see. Do we need to clean up the directory separately here?

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+#include <fstream>
+
+#include "external/envoy/source/common/common/logger.h"
+#include "external/envoy/source/common/common/random_generator.h"
+
+namespace Nighthawk {
+namespace {
+
+absl::Status verifyStringIsGuid(absl::string_view s) {
+  Envoy::Random::RandomGeneratorImpl random;
```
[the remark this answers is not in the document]
oschaaf: Well, my idea here is that the sink implementation will be the constraining factor when it comes to determining what is ok and not ok to specify for an execution_id. When initiating a load test, and not specifying an execution id, it will be auto-generated, and that will be a unique identifier.

In this file backed implementation I'm being a bit overly cautious, as a guid will have a reasonable length and contain no characters that would be illegal in a filename.  I hope to find a way to easily validate a filename and make sure there's no weird stuff in it.

Later on it will also be possible to manually specify an execution id when initiating an execution, and that might be useful when you want an id that is more easily remembered. As it stands, this implementation will reject that.

api/client/service.proto
```diff
@@ -32,6 +32,9 @@ message ExecutionRequest {
 message ExecutionResponse {
   Output output = 1;
   google.rpc.Status error_detail = 7;
+  // Unique id for lookup purposes. This will be taken from CommandLineOptions if set,
+  // if it is not set there it will be auto-generated.
```
mum4k, line 36: Shall we say that it is an opaque identifier whose format can change at any time? I.e. users shouldn't depend on the format in any way.

Just to give ourselves some flexibility if we need it.

include/nighthawk/common/sink.h
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Abstract Sink interface.
+ */
+class Sink {
+public:
+  virtual ~Sink() = default;
+
+  /**
+   * Store an ExecutionResponse instance. Can be called multiple times for the same execution_id to
+   * persist multiple fragments that together will represent results belonging to a single
```
mum4k, line 25: Don't we need to somehow indicate when the last fragment was added? If we don't how will the system know if it got all the fragment or there was a crash in the middle of the execution?

Or is it that we don't know if we are getting all the fragments at this layer by design?

include/nighthawk/common/sink.h
```diff
@@ -0,0 +1,46 @@
...
+  StoreExecutionResultPiece(const ::nighthawk::client::ExecutionResponse& response) const PURE;
+
+  /**
+   * Attempt to load a vector of ExecutionResponse instances associated to an execution id.
+   *
+   * @param execution_id Specify an execution_id that the desired set of ExecutionResponse
+   * instances are tagged with.
+   * @return const absl::StatusOr<std::vector<::nighthawk::client::ExecutionResponse>>
+   */
+  virtual const absl::StatusOr<std::vector<::nighthawk::client::ExecutionResponse>>
```
mum4k, line 42: Can we document the behavior when no fragments are found for the specified ID? Is it an empty response or a NOT_FOUND error?

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+#include <filesystem>
+#include <fstream>
+
+#include "external/envoy/source/common/common/logger.h"
+#include "external/envoy/source/common/common/random_generator.h"
+
+namespace Nighthawk {
+namespace {
+
+absl::Status verifyStringIsGuid(absl::string_view s) {
```
mum4k, line 13: Can you help me understand why does this code depend on the format at all? Are we parsing something from the UUID or do we depend on some other property of the identifier?

I would expect the code to just use the execution id as an opaque identifier and at most validate that it is unique in this instance of the sink.

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+    return status;
+  }
+  std::error_code error_code;
+  std::filesystem::create_directories("/tmp/nh/" + std::string(execution_id) + "/", error_code);
+  // Note error_code will not be set if an existing directory existed.
+  if (error_code.value()) {
+    return absl::Status(absl::StatusCode::kInternal, error_code.message());
+  }
+  std::array<char, L_tmpnam> name_buffer;
+  // This emits a compiler warning: the use of `tmpnam' is dangerous, better use `mkstemp'.
```
mum4k, line 55: This begs a question as to why we don't use the suggested `mkstemp`. If we do have a good reason, can we expand the comment to include that information?

source/common/sink_impl.cc
```diff
@@ -0,0 +1,109 @@
...
+  if (!std::tmpnam(name_buffer.data())) {
+    return absl::Status(absl::StatusCode::kInternal, "Failure creating temp file");
+  }
+  // Next we write to a tmp file, and if that succeeds, we swap it atomically to the target path.
```
mum4k, line 59: Is it important for the implementation to make the swap atomic? Can we explain why?

source/common/sink_impl.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "nighthawk/common/sink.h"
+
+namespace Nighthawk {
+
+class FileSinkImpl : public Sink {
```
mum4k, line 7: Can we add a class comment explaining specifics of this implementation?

include/nighthawk/common/sink.h
mum4k, line 42: Why are we enforcing the const-ness of the return value? This seems to make a copy, is it important to prevent the caller from performing modifications?

mum4k (changes requested):

include/nighthawk/common/sink.h
oschaaf: That is right, the sink doesn't try to understand what values it is storing for a particular key.
Consumers of sink lookup data will be responsible for interpreting it.

source/common/sink_impl.cc
oschaaf: The file sink implementation depends on us constraining the format to this, so it doesn't have to worry about weirdness with respect to directory names it generates internally based on this. We can work to remove this constraint in the future, but for now this seems the simplest solution.
Any other sink implementation probably wouldn't be subject to this.

api/client/service.proto
oschaaf: Done: 8f8a5da

include/nighthawk/common/sink.h
oschaaf: Done 8f8a5da

source/common/sink_impl.cc
oschaaf: Rewritten en 8f8a5da to not rely on mkstemp

source/common/sink_impl.cc
oschaaf: Done in 8f8a5da

source/common/sink_impl.h
oschaaf: Done in 8f8a5da

include/nighthawk/common/sink.h
oschaaf: No need for that, changed in 8f8a5da

include/nighthawk/common/sink.h
mum4k: Thank you for explaining.

source/common/sink_impl.cc
mum4k: That makes sense, can we then adjust the function names and the checks they perform to indicate that the goal is to verify the ID can be a directory name rather than trying to verify it is a Guid? E.g. the function could be called `verifyCanBeUsedAsDirectoryName` or similar and it should only perform a check for the containing characters the are supported by a file system.

mum4k:
Thank you for amending the PR @oschaaf. Just one remaining question in the thread about validating UUIDs.

source/common/sink_impl.cc
oschaaf: Done in 6a31b48

mum4k (approved):

PR #628, opened 2021-02-12
Fix potential hangs on exit & cap drain wait time

### LINKED ISSUE #627: Nighthawk may hang on exit
In tests I noticed that Nighthawk may sometimes wait for a long time, or even hang after reporting results. After reproducing it, I think there are two factors at play. 
- We set up a drain callback on the connection pool regardless of the number of active connections. But when passing, say, `--request-header "connection: close"`, no active connections may exist, and therefore no drain callback will be fired.
- When sending high volume traffic into a proxy, it may buffer up a lot of requests. In practice, this spooling may take a long time to unwind.

oschaaf:
- don't set up a drain callback when there are no active connections
- set up a timer that will cap the amount of time we wait for the pool
  to drain.
- disable latency measurement when commencing the drain procedure, as
  by that time we are no longer interested in it, and in particular
  don't want to hear about any warnings issued by the Statistic
  implementation about recorded values being too large. 

Fixes #627

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -40,7 +40,7 @@ function do_clang_tidy() {
 
 function do_unit_test_coverage() {
     export TEST_TARGETS="//test/... -//test:python_test"
-    export COVERAGE_THRESHOLD=94.0
+    export COVERAGE_THRESHOLD=94.3
     echo "bazel coverage build with tests ${TEST_TARGETS}"
     test/run_nighthawk_bazel_coverage.sh ${TEST_TARGETS}
     exit 0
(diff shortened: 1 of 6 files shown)


oschaaf:
I think I may have an idea for how to add an integration tests for this, but the problem is that this would rely on timing, 
and I'm a bit hesitant to further increase proneness to flakes there. I wonder if we could get our hands on a CI environment that will itself have a more steady speed of execution, some of the integration tests are inherently time sensitive.
The other way to solve this would be to use very long timings, but when we go down that road then the integration tests are going to take a loong time to complete, and it will hurt in cases where that is not needed.

mum4k (approved):

mum4k:
Agreed that the integration test may be brittle. Arguably we don't really have sufficient coverage just from the unit test, since it heavily relies on mocks. Therefore the addition of the integration test is still the right choice. I am assuming it was executed it a good amount of times and it isn't inherently flaky.

Merging in since it is marked as `p0`.

PR #637, opened 2021-03-02
Examples: a place to stash practical use cases

oschaaf:
This captures something I used and for which I had to dig a bit in
the codebase to get it set up. Stashing it here may help others and
lower the bar for consumption.

I anticipate there may be more examples like this that we could
add over time. So ideally this PR will be assessed both on structure
of the initiative (/examples) and contents (MULTIPLE_ENDPOINTS.md)


examples/MULTIPLE_ENDPOINTS.md
```diff
@@ -0,0 +1,29 @@
+# Hitting multiple endpoints with a traffic profile
+
+## Description
+
+Below is an example which will send requests to two endpoints `127.0.0.1:80` and `127.0.0.2:80`, while alternating over two request headers, which contain different paths and hosts.
```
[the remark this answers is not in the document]
dubious90: actually, importantly this isn't using request source generally, it's using the file based request source specifically

examples/MULTIPLE_ENDPOINTS.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Define a traffic profile
```
dubious90, line 7: Unless I'm mistaken you've created this term traffic profile here, and it is not used elsewhere. The way you've structured this makes it look like traffic profile is a broader nighthawk term in a way that I think is misleading.

Could we instead discuss it like:

Place a file called `traffic-profile.yaml` in your current working directory.  This will act as your configuration for the file-based request source.

oschaaf:
@dubious90 thanks for the review, you raise some good points. Also, looking at this again, I wonder if it would be better to move `examples/` into `docs/`.. thoughts?

dubious90:
> @dubious90 thanks for the review, you raise some good points. Also, looking at this again, I wonder if it would be better to move `examples/` into `docs/`.. thoughts?

This change sounds good to me.  `docs/examples/` seems like the right landing location.

oschaaf:
I addressed feedback in e83f615, added a little more polish & moved `examples/` into `doc/root/`. Ready for another round.

docs/root/examples/MULTIPLE_ENDPOINTS.md
```diff
@@ -0,0 +1,43 @@
...
+## Practical use
+
+This example has been useful to test a mesh that exposed multiple endpoints, which in turn would offer access to multiple applications via different hosts/paths.
+
+## Features used
+
+This example illustrates the following features:
+
+- [Request Source](https://github.com/envoyproxy/nighthawk/blob/261abb62c40afbdebb317f320fe67f1a1da1838f/api/request_source/request_source_plugin.proto#L15) (specifically the file-based implementation).
+- [Multi-targetting](https://github.com/envoyproxy/nighthawk/blob/261abb62c40afbdebb317f320fe67f1a1da1838f/api/client/options.proto#L84)
```
dubious90, line 16: sp: targeting

dubious90 (approved):
Oof, sorry for my delay on this.  For some reason, I thought I had already replied to this.

This looks good modulo one typo

docs/root/examples/MULTIPLE_ENDPOINTS.md
oschaaf: Amended in  baf7a5e

dubious90 (approved):

PR #639, opened 2021-03-02
Sink service client & unit tests

oschaaf:
Add the proto definition for the sink service plus a client implementation and test.

### DIFF
--- a/api/client/service.proto
+++ b/api/client/service.proto
@@ -49,7 +49,7 @@ service NighthawkService {
 
 // Encapsulates an ExecutionResponse.
 message StoreExecutionRequest {
-  // Response must contain the effective execution id, which will be used as a lookup key.0
+  // Response contains the effective execution id, which will serve as the lookup key.
   ExecutionResponse execution_response = 1;
 }
 
@@ -73,7 +73,7 @@ service NighthawkSink {
   rpc StoreExecutionResponseStream(stream StoreExecutionRequest) returns (StoreExecutionResponse) {
   }
 
-  // Gets the stored response of an execution, keyed by execution id.
+  // Gets the stored responses associated to an execution, keyed by execution id.
   rpc SinkRequestStream(stream SinkRequest) returns (stream SinkResponse) {
   }
 }
--- a/api/client/service.proto
+++ b/api/client/service.proto
@@ -2,6 +2,7 @@ syntax = "proto3";
 
 package nighthawk.client;
 
+import "google/protobuf/duration.proto";
 import "google/rpc/status.proto";
 import "validate/validate.proto";
 
@@ -43,5 +44,36 @@ service NighthawkService {
   //  test. Returns a stream of ExecutionResponse objects which will contain results for all
   //  Nighthawk runs.
   rpc ExecutionStream(stream ExecutionRequest) returns (stream ExecutionResponse) {
+  }
+}
+
+// Encapsulates an ExecutionResponse.
+message StoreExecutionRequest {
+  // Response must contain the effective execution id, which will be used as a lookup key.0
+  ExecutionResponse execution_response = 1;
+}
+
+// Empty return value message, that serves as a future extension point.
+message StoreExecutionResponse {
+}
+
+message SinkRequest {
+  // Unique id for lookup purposes.
+  string execution_id = 1;
+}
+
+message SinkResponse {
+  // Response associated to the requested execution id.
+  ExecutionResponse execution_response = 1;
+}
+
+service NighthawkSink {
+  // Accepts a stream of execution responses, which is the return value of
+  // NighthawkService.ExecutionStream. Workers can forward their results using this method.
+  rpc StoreExecutionResponseStream(stream StoreExecutionRequest) returns (StoreExecutionResponse) {
+  }
+
+  // Gets the stored response of an execution, keyed by execution id.
+  rpc SinkRequestStream(stream SinkRequest) returns (stream SinkResponse) {
   }
 }
--- a/api/client/service.proto
+++ b/api/client/service.proto
@@ -46,34 +46,3 @@ service NighthawkService {
   rpc ExecutionStream(stream ExecutionRequest) returns (stream ExecutionResponse) {
   }
 }
-
-// Encapsulates an ExecutionResponse.
-message StoreExecutionRequest {
-  // Response contains the effective execution id, which will serve as the lookup key.
-  ExecutionResponse execution_response = 1;
-}
-
-// Empty return value message, that serves as a future extension point.
-message StoreExecutionResponse {
-}
-
-message SinkRequest {
-  // Unique id for lookup purposes.
-  string execution_id = 1;
-}
-
-message SinkResponse {
-  // Response associated to the requested execution id.
-  ExecutionResponse execution_response = 1;
-}
-
-service NighthawkSink {
-  // Accepts a stream of execution responses, which is the return value of
-  // NighthawkService.ExecutionStream. Workers can forward their results using this method.
-  rpc StoreExecutionResponseStream(stream StoreExecutionRequest) returns (StoreExecutionResponse) {
-  }
-
-  // Gets the stored responses associated to an execution, keyed by execution id.
-  rpc SinkRequestStream(stream SinkRequest) returns (stream SinkResponse) {
-  }
-}
--- a/api/sink/sink.proto
+++ b/api/sink/sink.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package nighthawk.sink;
 
 import "api/client/service.proto";
+import "validate/validate.proto";
 
 // Encapsulates an ExecutionResponse.
 message StoreExecutionRequest {
@@ -15,8 +16,8 @@ message StoreExecutionResponse {
 }
 
 message SinkRequest {
-  // Unique id for lookup purposes.
-  string execution_id = 1;
+  // Unique id for lookup purposes. Required.
+  string execution_id = 1 [(validate.rules).string.min_len = 1];
 }
 
 message SinkResponse {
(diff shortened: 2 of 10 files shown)


mum4k:

source/common/nighthawk_sink_client_impl.cc
```diff
@@ -0,0 +1,56 @@
...
+
+  if (!stream->Write(sink_request)) {
+    return absl::UnavailableError("Failed to write request to the Nighthawk Sink gRPC channel.");
+  } else if (!stream->WritesDone()) {
+    return absl::InternalError("WritesDone() failed on the Nighthawk Sink gRPC channel.");
+  }
+
+  bool got_response = false;
+  while (stream->Read(&response)) {
+    RELEASE_ASSERT(!got_response,
```
wjuan-AFK, line 45: I'm not sure I understand how this loop is supposed to work.

oschaaf: Ah yes, this isn't obvious; the way we handle this reflects other places in the code base, e.g:
https://github.com/envoyproxy/nighthawk/blob/3881261ccc76e382109fdf94e5035b1f7f2b71e9/source/common/nighthawk_service_client_impl.cc#L28

Some more context on that:

At the proto api level we support returning a stream of results. The sink service proto api reflects this, and accepts what NighthawkService. ExecutionStream returns as a parameter (though we wrap it in StoreExecutionRequest messages for extensibility purposes). So this implies a stream, and not a single message.

Having said that, today we constrain what we return to a single message in the implementations where this is relevant. That's why we assert here, to make sure that stays put until an explicit choice is made otherwise.

Why do this? The intent of `NighthawkService. ExecutionStream` was to be able to stream intermediate updates some day. So having streams in the api's keeps the door open on streaming intermediary updates, without forcing a change the proto api.

wjuan-AFK: Thanks for the explanation.

api/client/service.proto
```diff
@@ -45,3 +46,34 @@ service NighthawkService {
...
+  // Unique id for lookup purposes.
+  string execution_id = 1;
+}
+
+message SinkResponse {
+  // Response associated to the requested execution id.
+  ExecutionResponse execution_response = 1;
+}
+
+service NighthawkSink {
```
mum4k, line 70: Would you mind moving the sink related service and its messages to a separate proto file for clarity? Say `sink_service.proto`?

api/client/service.proto
```diff
@@ -45,3 +46,34 @@ service NighthawkService {
...
+  // Response contains the effective execution id, which will serve as the lookup key.
+  ExecutionResponse execution_response = 1;
+}
+
+// Empty return value message, that serves as a future extension point.
+message StoreExecutionResponse {
+}
+
+message SinkRequest {
+  // Unique id for lookup purposes.
```
mum4k, line 61: Assuming this is required, can we clearly indicate that it is?

```
// Unique id for lookup purposes. Required.
```

Alternatively if it isn't, can we describe the behavior when an empty string is provided?

include/nighthawk/common/nighthawk_sink_client.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual ~NighthawkSinkClient() = default;
+
+  /**
+   * @brief Store an execution response.
+   *
+   * @param nighthawk_sink_stub Used to open a channel to the sink service.
+   * @param store_execution_request Provide the message that the sink should store.
+   * @return absl::StatusOr<nighthawk::client::StoreExecutionResponse>
+   */
+  virtual absl::StatusOr<nighthawk::client::StoreExecutionResponse> StoreExecutionResponseStream(
```
mum4k, line 26: Why do we need to abstract the client behind another interface? Are we going to have multiple implementations?

If it is just for testing, we could do with just one implementation that has virtual methods? The additional layer of indirection comes with its cost, so trying to understand if we have a good reason to pay it.

include/nighthawk/common/nighthawk_sink_client.h
```diff
@@ -0,0 +1,43 @@
...
+      nighthawk::client::NighthawkSink::StubInterface* nighthawk_sink_stub,
```
mum4k, line 27: Would a reference work better? (since it cannot be null).

Looks like the other method does accept a reference.

source/common/nighthawk_sink_client_impl.cc
mum4k: Can we convert a portion of this thread into an inline comment?

mum4k, line 45: Why did we choose an ASSERT as compared to returning not-ok status? Is crashing the best choice here?

test/common/nighthawk_sink_client_test.cc
```diff
@@ -0,0 +1,304 @@
...
+
+  NighthawkSinkClientImpl client;
+  absl::StatusOr<nighthawk::client::StoreExecutionResponse> response =
+      client.StoreExecutionResponseStream(&mock_nighthawk_sink_stub, {});
+  ASSERT_FALSE(response.ok());
+  EXPECT_EQ(response.status().code(), absl::StatusCode::kPermissionDenied);
+  EXPECT_THAT(response.status().message(), HasSubstr("Finish failure status message"));
+}
+
+// ---
```
mum4k, line 160: Did we mean to leave this in?

mum4k (changes requested):

api/client/service.proto
oschaaf: Done in fe2eac357d600922489e611a8843548b544a16dd. Let me know what you think, happy to iterate more on this.

api/client/service.proto
oschaaf: Done: d89620e32d3b28caf5fd1eab11b8bff794bdcbca

include/nighthawk/common/nighthawk_sink_client.h
oschaaf: Good catch; that was pretty inconsistent. Done in fe2eac357d600922489e611a8843548b544a16dd

oschaaf: Err sorry f64a229b62bcaaa78b1316708fca4cf4324c3a1d

source/common/nighthawk_sink_client_impl.cc
oschaaf: Done in 3734e54217e9a615c117f73b69f213902cd82f25

test/common/nighthawk_sink_client_test.cc
oschaaf: Remove in 78dad245c0fd4e3f907296efba461ecaaf9b7e7d

source/common/nighthawk_sink_client_impl.cc
oschaaf: This is something that really should not ever happen unless we explicitly decided otherwise. Having said that, personally I would lean slightly towards returning not ok and then cover this in a test as a behavioural aspect; but for now I settled for this because I think consistency is outweighing this (as this mirrors  another part of the code base).

include/nighthawk/common/nighthawk_sink_client.h
oschaaf: This makes it easy to create a mock for it; so if `ProcessImpl` starts using this further down the road, this comes in handy when we want to unit test the interaction to ensure that looks like we expect it to. But I think you are right in that chances are pretty slim that we'll see >1  implementations for this :-)
 
If you have a concern about perf aspects, imho that only weighs slightly here since this will not be something that will be repeatedly consumed at a frequency where this would become significant; but I can see that there's potential overhead in maintenance of the interface layer.. But here there's also a consistency argument: other similar code does this too.

Weighing everything, I'm personally slightly leaning towards leaving this as is; but I'll ping you on Slack to discuss some more, as changing this is also not something that I'd strongly disagree with.

api/sink/sink.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+package nighthawk.sink;
```
mum4k, line 3: In our last discussion about namespaces, we ended up on having just a single namespace for Nighthawk's future code. Do we have a good reason to introduce the new namespace here?

include/nighthawk/common/nighthawk_sink_client.h
mum4k: A class with virtual methods should make it as easy to create mocks. The interface tends to only be really needed when there are multiple concrete implementations. That is also how a reader of the code "reads" the existence of a virtual interface. Having it just for mocking may be considered too big of a hammer.

With that said, we don't need to block this PR on it, it is something we can start considering and applying on future code. Let's proceed as is for expediency.

source/common/nighthawk_sink_client_impl.cc
mum4k: Happy to leave this up to you. For the sake of future discussions - this may not be a strong case for consistency. Consistency across the codebase is very important on items like APIs, used libraries, namespaces, etc. Individual functions should be able to make their own choice of whether they crash the code or return an error. I have learned to lean towards error returning, since crashes are harder to document and are often surprising to someone who might end up reusing the library or the function. Additionally crashes are not the best experience for our users.

With that said, I agree with you in this case that since we are making a choice to only return a single message, returning multiple is technically a programming error which does deserve a crash.

mum4k (approved):

mum4k:
@oschaaf just one remaining question about the new namespace.

api/sink/sink.proto
oschaaf: Modified in 5c49c607558f7fb26fb088c787b8cb22dd820e5d

mum4k (approved):

PR #641, opened 2021-03-08
Update Envoy to 124c160 (March 8th 2021)

eric846:
- Sync `.bazelrc`
- Update Docker image version in `.circleci/config.yml`
- New way to declare Bazel pip imports in `WORKSPACE` (the `requirement("...")` side in `test/integration/BUILD` is unchanged)
- `ci/run_envoy_docker.sh` was unchanged
- Code changes
  - **Relax some byte count checks in integration tests**
    - The counters started to come out lower after updating Envoy,  e.g. 1450 decreased to 900. This was because the underlying Envoy infrastructure in the Nighthawk client stopped sending `content-length: 0` in the requests.
    - Updating the test to only check if the count is >=500, rather than checking for a specific value
      - This should still catch failure modes where Nighthawk fails after 1 request or less
      - `test_http_h2` already has such relaxed checks
  - Pass through new `TimeSource` arg in `allocateConnPool()`
  - `factoryForGrpcService()` changed the `skip_cluster_check` boolean to an enum `Envoy::Grpc::AsyncClientFactoryClusterChecks::Skip`
  - Empty `dumpState()` override added to `StreamDecoder`
  - UUID generator for tracing was moved
  - Ignore warning "Unable to use runtime singleton for feature envoy.reloadable_features.header_map_correctly_coalesce_cookies"


test/integration/integration_test_fixtures.py
```diff
@@ -153,6 +153,8 @@ def tearDown(self, caplog):
       for record in caplog.get_records(when):
         if record.levelno not in (logging.WARNING, logging.ERROR):
           continue
+        if "Unable to use runtime singleton for feature" in record.message:
```
mum4k, line 156: We already have a way of skipping some warnings:

https://github.com/envoyproxy/nighthawk/blob/e2f5265b3e4f0abf7d749c3a0988fb57fe50cea6/test/integration/nighthawk_test_server.py#L63-L69

Can we utilize this one?

mum4k (changes requested):
Thank you for putting this together @eric846.

test/integration/integration_test_fixtures.py
eric846: Done.

oschaaf (commented):
Re: "The counters started to come out lower after updating Envoy, e.g. 1450 decreased to 900. Is this expected?"
It might be worth comparing headers to have a look-see at what precisely changed. This seems significant enough to make an actual real difference?

oschaaf (commented):
Yeah, thinking about it more, the strict byte count testing may seem overly strict, but it might just be worth it as it will prompt investigation in cases like this. In headers-only tests we will transfer a significant % less bytes, possibly impacting packet counts required for transmitting a single request; so not requesting changes in this PR right away but I do feel it is worth discussing.

eric846:
> Re: "The counters started to come out lower after updating Envoy, e.g. 1450 decreased to 900. Is this expected?"
> It might be worth comparing headers to have a look-see at what precisely changed. This seems significant enough to make an actual real difference?

Bingo:

### This PR

```
$ nc -l 127.0.0.1 12345
GET / HTTP/1.1
host: 127.0.0.1:12345

```

```
$ bazel-bin/nighthawk_client --rps 1 --duration 5 http://127.0.0.1:12345
...
upstream_cx_tx_bytes_total              123         24.60
...
```


### Main branch
```
$ nc -l 127.0.0.1 12345
GET / HTTP/1.1
host: 127.0.0.1:12345
content-length: 0

```

```
$ bazel-bin/nighthawk_client --rps 1 --duration 5 http://127.0.0.1:12345
...
upstream_cx_tx_bytes_total              180         36.00
...
```

eric846:
> Yeah, thinking about it more, the strict byte count testing may seem overly strict, but it might just be worth it as it will prompt investigation in cases like this. In headers-only tests we will transfer a significant % less bytes, possibly impacting packet counts required for transmitting a single request; so not requesting changes in this PR right away but I do feel it is worth discussing.

Agreed, opened #644 to discuss. Depending on the schedule, we might want to move forward with these relaxed assumptions. There's a chance Envoy could change its header behavior again.

oschaaf (approved):
Approving; while it seems odd to have a content-length header appear in a GET request, apparently it causes no trouble as all integration tests past — modulo the strict byte count check. 
We may want to file an issue upstream though?

mum4k (approved):

PR #646, opened 2021-03-15
Better error handling using absl::StatusOr

### LINKED ISSUE #625: Solidify output formatter implementations
It would be good to go over the output formatter implementations and ensure that there are no ASSERT paths that can be hit when better handling of the condition that triggered it would be available. 
At the time of writing this there are two known ones:

- In some of the OutputFormatter implementations we rely on `getJsonStringFromMessageOrDie`,
which was introduced recently in an Envoy update. There's a variant of this method which doesn't die but instead returns an `absl::Status`, and that could be leveraged to enhance the flow here.
- The Fortio output formatter has a [known path](https://github.com/envoyproxy/nighthawk/blob/796c71832c3a9b44703c5b25805eec8227fb2c5d/source/client/output_formatter_impl.cc#L248) where it asserts when there are no a global result. This is suboptimal, as this can happen with some arguably silly yet valid configurations. Instead it would be better to instruct the Fortio UI to display an error, for which there are flags in its input json.

In general, it should be rare for any output formatter to terminate abnormally, but input can be transformed into output that reflects an error in the target format.

Context: this came up in review of https://github.com/envoyproxy/nighthawk/pull/621
Note: this affects only the CLI, and not the service, therefore we consider it not critical.

Razdeep:
Fixes #625 
Replacing getJsonStringFromMessageOrDie() by getJsonStringFromMessageOrError().

Signed-off-by: Rajdeep Roy Chowdhury <rrajdeeproychowdhury@gmail.com>


oschaaf (changes requested):
Thanks for looking into this; a small issue is that CI flags a formatting issue, that can be resolved via running `ci/do_ci.sh fix_format`.

But I'm also requesting some changes because while this builds I think it effectively doesn't change much, as we don't check the `status_or` to see if everything is ok before consuming it. Let me try to point out some examples that show the pattern of how to best leverage `absl::status_or`:

Generally, I think we need to propagate the return value of `getJsonStringFromMessageOrError` where we consume it, except at near or at the highest level of the program, where it can be handled. So probably the return value of `formatProto()` needs to change.

That looks like something along the lines of this: https://github.com/envoyproxy/nighthawk/blob/e2a0054ab727dc4a273606215a19d2c1a327c896/source/adaptive_load/plugin_loader.cc#L19 

And then, as an example, here's a place where we consume `formatProto()` at the highest level:
https://github.com/envoyproxy/nighthawk/blob/e6f9af70d446ee018f1d72f8e46b8c1a22c186f4/source/client/output_transform_main.cc#L61

There, instead of optimistically assuming all is well and streaming out the result, we should check for `message_or_error.ok()`, and handle the case where there was things ar not ok. Because if there's an error we should exit with a non-zero status code, log what the status was, and not attempt to stream the status_or to the stdout.

source/client/client.cc
```diff
@@ -81,7 +81,12 @@ bool Main::run() {
     result = process->run(output_collector);
   }
   auto formatter = output_formatter_factory.create(options_->outputFormat());
-  std::cout << formatter->formatProto(output_collector.toProto());
+  absl::StatusOr<std::string> formatted_proto = formatter->formatProto(output_collector.toProto());
+  if (formatted_proto.ok()) {
+    std::cout << formatter->formatProto(output_collector.toProto());
+  } else {
+    ENVOY_LOG(error, "An error occured while formatting proto");
```
oschaaf, line 88: should we force result to false here? if this code line get hit, that probably means we should indicate failure.

alternatively, we could consider propagating `absl::status_or` even further up (so we return that instead of the boolean we use return right now) if you'd be up for it, but feel free to not do this, I don't intent to bloat this PR.

oschaaf (commented):
Thanks for pushing these changes, this looks great!

Now to be able to close https://github.com/envoyproxy/nighthawk/issues/625 we need one more thing: the Fortio output formatter still throws exceptions in two [1] spots that I know of; instead of doing that it should also apply the absl::status_or pattern we introduced here.

[1] https://github.com/envoyproxy/nighthawk/blob/5baabbbd3c33204ff94a69ba87dad1229dc9d915/source/client/output_formatter_impl.cc#L217 and https://github.com/envoyproxy/nighthawk/blob/5baabbbd3c33204ff94a69ba87dad1229dc9d915/source/client/output_formatter_impl.cc#L245

source/client/client.cc
Razdeep: What shall be the status code for [1]?

dubious90 (commented):
Thanks for the contribution!  Have a couple comments on returning strings.

Razdeep:
I cannot find a way to intensionally fail `JsonFormatter::formatProto`, `YamlFormatter::formatProto`, `DottedFormatter::formatProto`. Thus the test-coverage is reduced.

source/client/output_formatter_impl.cc
```diff
@@ -203,18 +203,18 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
       ss << fmt::format("{}:{}", prefix, counter.value()) << std::endl;
     }
   }
-  return ss.str();
+  return absl::Status(absl::StatusCode::kOk, ss.str());
```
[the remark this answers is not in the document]
Razdeep: done

source/client/output_formatter_impl.cc
```diff
@@ -114,7 +114,7 @@ std::string ConsoleOutputFormatterImpl::formatProto(const nighthawk::client::Out
     }
   }
 
-  return ss.str();
+  return absl::Status(absl::StatusCode::kOk, ss.str());
```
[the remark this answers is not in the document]
Razdeep: done

ci/do_ci.sh
```diff
@@ -41,7 +41,7 @@ function do_clang_tidy() {
 
 function do_unit_test_coverage() {
     export TEST_TARGETS="//test/... -//test:python_test"
-    export COVERAGE_THRESHOLD=94.2
+    export COVERAGE_THRESHOLD=93.0
```
oschaaf, line 44: This seems like a rather large decrease, looking at the coverage report we could set this set this at `93.9` with the current state.

oschaaf: having said that, looking at https://26453-180498819-gh.circle-artifacts.com/0/coverage/source/client/output_formatter_impl.cc.gcov.html, I think it should be possible to hit some of the failure cases for the Fortio output formatter, by creating a test that provides an input that doesn't include a "global" result. As far as I am concerned this can be moved to a new issue though.

include/nighthawk/client/output_formatter.h
```diff
@@ -19,10 +22,11 @@ class OutputFormatter {
   virtual ~OutputFormatter() = default;
 
   /**
-   * @return std::string serialized representation of output. The specific format depends
-   * on the derived class, for example human-readable or json.
+   * @return absl::StatusOr<std::string> serialized representation of output, if not error.
```
oschaaf, line 25: Consider `s/, if not/or an/` to make line flow a little better.

source/client/client.cc
oschaaf: Consider setting result to false and fall trough, to have the `if(!result) {...` clause handle the logging/error handling. Minimizing the control logic helps keeping coverage up. If we want to specialise the message that gets logged we can declare/set that separately.

source/client/output_formatter_impl.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "common/version_info.h"
 
+#include "absl/status/status.h"
```
oschaaf, line 17: Do we need this include?

source/client/output_formatter_impl.cc
```diff
@@ -142,15 +145,25 @@ std::string ConsoleOutputFormatterImpl::statIdtoFriendlyStatName(absl::string_vi
...
 }
 
-std::string JsonOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
-  return Envoy::MessageUtil::getJsonStringFromMessageOrDie(output, true, true);
+absl::StatusOr<std::string>
+JsonOutputFormatterImpl::formatProto(const nighthawk::client::Output& output) const {
+  auto status_proto = Envoy::MessageUtil::getJsonStringFromMessage(output, true, true);
+  if (status_proto.ok()) {
+    return status_proto.value();
+  } else {
```
oschaaf, line 153: Consider the ternary operator here, less code, less impact on coverage for the lines we can't cover. And personally I like that it makes it slightly more obvious that the method inescapable returns a value at a glance.

source/client/output_formatter_impl.cc
```diff
@@ -292,7 +312,12 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
   fortio_output.set_numthreads(number_of_connections);
 
   // Get the result that represents all workers (global)
-  const auto& nh_global_result = getGlobalResult(output);
+  auto nh_global_result_status = getGlobalResult(output);
+  if (!nh_global_result_status.ok()) {
+    return absl::Status(nh_global_result_status.status().code(),
```
oschaaf, line 317: This could be covered in a test as well, I think

oschaaf (changes requested):
This is great stuff, thanks so much for doing this. Some last comments.

source/client/output_formatter_impl.cc
Razdeep: Even if we create a test with input that doesn't contain a global result, it gets handled at https://github.com/envoyproxy/nighthawk/blob/a8d5fdef25d6039a39adfd9944e1bf294a53b667/source/client/output_formatter_impl.cc#L291

therefore, the lines @oschaaf mentioned remain uncovered.

oschaaf: That makes me wonder if it's even possible to hit this code.. if that's not the case it means this is dead code, and we ought to remove it. IMHO that would make sense and is related to why I figured that throwing an exception in this case didn't make sense. It looks like the Fortio UI would indicate an error on the output we'd produce, which is sufficient:
https://github.com/fortio/fortio/blob/ee098c572115f4149a67d2a812cb3d1ca79e30e1/ui/static/js/fortio_chart.js#L105

We should probably make sure that this really works though.

source/client/output_formatter_impl.cc
Razdeep: if the `else` part of this code doesn't ever execute, how is the idea of simply returning `status_proto.value()`?
We can get rid of the if-else branching.

oschaaf: Hmm, yes. In fact, I think we can just return `status_proto`? E.g. condense the whole method body to a one-liner, `Envoy::MessageUtil::getJsonStringFromMessage(output, true, true);`

oschaaf (approved):
LGTM, very nice. One last nit: the git commit  signoff is missing in the last commit.. as for the changes in here they look good to me; I am deferring to @mum4k & @dubious90 for final approval. Thanks a lot for persisting!

source/client/output_formatter_impl.cc
```diff
@@ -292,7 +304,8 @@ std::string FortioOutputFormatterImpl::formatProto(const nighthawk::client::Outp
   fortio_output.set_numthreads(number_of_connections);
 
   // Get the result that represents all workers (global)
-  const auto& nh_global_result = getGlobalResult(output);
+  auto nh_global_result_status = getGlobalResult(output);
+  const auto& nh_global_result = nh_global_result_status.value();
```
dubious90, line 308: I know that you and Otto had a conversation here about whether or not to include a line here returning the status if it's not `ok()` and deciding not to.  I think I disagree with the conclusion.  Could we please return the status if it's not ok here?

Reasoning:
1. I don't think it's a good idea to code with perfect understanding of a function's error state, because functions aren't static - they change over time.  While it may not be possible for getGlobalResult to actually return an error _yet_, it might be later, and now this code wouldn't operate as expected.

2. Removing lines just because covering them isn't possible is a distorted incentive.  Coverage should never discourage us from doing the right thing.

source/client/output_transform_main.cc
```diff
@@ -58,7 +58,12 @@ uint32_t OutputTransformMain::run() {
   }
   OutputFormatterFactoryImpl factory;
   OutputFormatterPtr formatter = factory.create(translated_format);
-  std::cout << formatter->formatProto(output);
+  auto format_status = formatter->formatProto(output);
```
dubious90, line 61: I wasn't calling it out in other pre-existing locations, but we should try to avoid using auto unless the type is fully obvious from context (which it isn't here).  Can you please declare the full type?

source/client/output_formatter_impl.cc
oschaaf: This comment drew me into looking into this once more; what I think might be appropriate is for `getGlobalResult() is` to return an `absl:::optional<const nighthawk::client::Result>` instead of a status: I can't anticipate failures modes where an exception wouldn't be appropriate, and not having a global result in the output is business as usual in terms of transforming to Fortio output.

source/client/output_formatter_impl.cc
```diff
@@ -206,15 +211,16 @@ DottedStringOutputFormatterImpl::formatProto(const nighthawk::client::Output& ou
   return ss.str();
 }
 
-const nighthawk::client::Result&
+absl::StatusOr<const nighthawk::client::Result>
```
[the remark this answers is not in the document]
Razdeep: This was intentional. Putting a reference inside `absl::StatusOr<>` was causing a compilation error. 
@oschaaf told that it isn't that performance-critical, so can be switch to non-reference.

source/client/client.cc
```diff
@@ -89,7 +89,7 @@ bool Main::run() {
     ENVOY_LOG(error, "An error occured while formatting proto");
     result = false;
   }
```
oschaaf, line 91: Oh I think we need:
```c++
if (!formatted_proto.ok()) {
  // .... <snip> looks ok to me as-is.
} else {
  // only when formatted_proto.ok() == true
  std::cout << *formatted_proto;
}
```
I'm sorry if my earlier comment here was off or caused confusion.

Razdeep:
@dubious90 Please do another round of review

dubious90 (approved):

PR #655, opened 2021-03-18
Add a missing include of api/client/output.pb.h

eric846:


oschaaf (approved):
Looks like I cleaned up too much, LGTM.
How did you catch this? It would be good to know if this is something that can be caught in CI.

PR #656, opened 2021-03-20
Add execution-id & separate the options BUILD targets

oschaaf:
Another prelude to distributed test execution. Being able to
identify executions and associate results is going to come in
handy.

Also, this separates some BUILD targets to allow for a leaner build
process of future things that are going to depend on it.

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -106,7 +106,7 @@ message H1ConnectionReuseStrategy {
 
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
-// highest unused number is 39
+// highest unused number is 40
 message CommandLineOptions {
   // The target requests-per-second rate. Default: 5.
   google.protobuf.UInt32Value requests_per_second = 1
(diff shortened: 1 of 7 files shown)


mum4k:

wjuan-AFK (approved):

api/client/options.proto
```diff
@@ -106,7 +106,7 @@ message H1ConnectionReuseStrategy {
 
 // TODO(oschaaf): Ultimately this will be a load test specification. The fact that it
 // can arrive via CLI is just a concrete detail. Change this to reflect that.
-// highest unused number is 39
+// highest unused number is 40
```
mum4k, line 109: (pre-existing) Is this true? There seem to be id `105` just above it.

api/client/options.proto
```diff
@@ -225,4 +225,10 @@ message CommandLineOptions {
   google.protobuf.Timestamp scheduled_start = 105;
 
   reserved 38; // deprecated
+  // Provide a unique execution id. Will be reflected in the output when set.
```
mum4k, line 228: Is this a good opportunity to also describe the required format? E.g. the identifier must be something that can use used as a directory name on Linux filesystem?

mum4k (changes requested):

api/client/options.proto
oschaaf: See below, `When populating this field, it is recommended...`. There will only be constraints to this value when there is a Sink involved which imposes limitations; our file based sink is picky (at least today) and only accepts unique identifiers. Maybe we can later work to remove that restriction by using a codec. For max key length I was thinking along the same lines, that might vary depending on the implementation specifics of Sinks?

api/client/options.proto
oschaaf: added cc80ec1e

api/client/options.proto
mum4k: Makes sense, we don't want to expose limitations of a single sink up to the public API.

We can make sure that the individual Sink implementations will validate their inputs and return readable error messages when an incompatible `execution_id` will be specified.

mum4k (approved):

PR #657, opened 2021-03-20
Sink: add in memory sink, slightly refactor validation/testing.

oschaaf:
- Adds an in-memory sink. Comes in handy in tests, and could serve
  a future role if we want to converge the way results are handled
  in both distributed and non-distributed tests.
- Slightly refactors validation & testing to make it easier to see
  what is specific to FileSinkImpl, making things a little more
  natural in case someone else wants to  add a new sink type in the
  future.
- Avoid MOCK_METHOD_N in the Sink mock as per recommendation.

### DIFF
--- a/source/sink/sink_impl.cc
+++ b/source/sink/sink_impl.cc
@@ -16,27 +16,25 @@ absl::Status verifyCanBeUsedAsDirectoryName(absl::string_view s) {
   const std::string err_template = "'{}' is not a guid: {}";
 
   if (s.size() != reference_value.size()) {
-    return absl::Status(absl::StatusCode::kInvalidArgument,
-                        fmt::format(err_template, s, "bad string length."));
+    return absl::InvalidArgumentError(fmt::format(err_template, s, "bad string length."));
   }
   for (size_t i = 0; i < s.size(); i++) {
     if (reference_value[i] == '-') {
       if (s[i] != '-') {
-        return absl::Status(
-            absl::StatusCode::kInvalidArgument,
+        return absl::InvalidArgumentError(
             fmt::format(err_template, s, "expectations around '-' positions not met."));
       }
       continue;
     }
     if (!std::isxdigit(s[i])) {
-      return absl::Status(absl::StatusCode::kInvalidArgument,
-                          fmt::format(err_template, s, "unexpected character encountered."));
+      return absl::InvalidArgumentError(
+          fmt::format(err_template, s, "unexpected character encountered."));
     }
   }
   return absl::OkStatus();
 }
 
-absl::Status validateKey(absl::string_view key, const bool validate_as_directory_name) {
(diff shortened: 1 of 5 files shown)


mum4k:

jiajunye:
LGTM

source/sink/sink_impl.cc
```diff
@@ -36,12 +36,22 @@ absl::Status verifyCanBeUsedAsDirectoryName(absl::string_view s) {
   return absl::OkStatus();
 }
 
+absl::Status validateKey(absl::string_view key, const bool validate_as_directory_name) {
```
mum4k, line 39: (nit) Is it important for the bool argument to be const?

source/sink/sink_impl.cc
```diff
@@ -104,4 +113,30 @@ FileSinkImpl::LoadExecutionResult(absl::string_view execution_id) const {
...
+InMemorySinkImpl::StoreExecutionResultPiece(const nighthawk::client::ExecutionResponse& response) {
+  absl::Status status = validateKey(response.execution_id(), false);
+  if (!status.ok()) {
+    return status;
+  }
+  auto iterator =
+      map_.insert({response.execution_id(), std::vector<nighthawk::client::ExecutionResponse>()})
+          .first;
+  iterator->second.push_back(response);
+  return absl::Status();
```
mum4k, line 126: I think the more conventional way of returning OK is:

```cc
return absl::OkStatus();
```

This one also works, but requires the reader to understand that an empty status defaults to OK.

https://abseil.io/blog/2020-091021-status

source/sink/sink_impl.cc
```diff
@@ -104,4 +113,30 @@ FileSinkImpl::LoadExecutionResult(absl::string_view execution_id) const {
...
+InMemorySinkImpl::LoadExecutionResult(absl::string_view execution_id) const {
+  absl::Status status = validateKey(execution_id, false);
+  if (!status.ok()) {
+    return status;
+  }
+  auto iterator = map_.find(execution_id);
+  if (iterator != map_.end()) {
+    return (*iterator).second;
+  }
+  return absl::Status(absl::StatusCode::kNotFound, "Not found");
```
mum4k, line 139: (optional, fyi) Similarly to the comment above, a shorter version is:

```cc
return absl::NotFoundError("...");
```

Feel free to ignore, I know we have a few of these already in.

mum4k, line 139: Can we expand the error message, giving a better idea as to what wasn't found? Seeing an error message like this one might not be very helpful when debugging.

mum4k (changes requested):

oschaaf:
Thanks for the review @mum4k , comments addressed in 135e803b

mum4k (approved):

PR #658, opened 2021-03-20
[DISCUSS] Milestone tracking PoC

oschaaf:
Pushing out a draft of some old work, which I wanted to float for consideration:
This proposes a change to the way we track latencies by introducing a milestone tracker concept.
That can be used to track durations of an operation as well as mark the stages from start to completion.
Having this in place allows us to centralise (and localize) handling of completed milestones (e.g. requests),
which means that we can stop the statistics from being tied to a lot of concepts through constructors, and also allows us to easily track more & different deltas if we want to. Also, it could provide a convenient hook for tagging additional metadata onto results, like connection-id or a header observed in the reply.

- `MilestoneTestWithSimtime` should allow for easy high level evaluation of usage.
- `ClientWorkerImpl::onMilestoneComplete` shows how we would integrate.

This is expected to fail some tests (and even some of the test build), as it would need some more work to land it.
Also, this is not tied any critical path (and disconnected from efforts on horizontal scaling).


mum4k:
@oschaaf I am skipping this PR for now, since it is marked as draft. If that wasn't intended, please respond and I will assign a reviewer.

oschaaf:
> @oschaaf I am skipping this PR for now, since it is marked as draft. If that wasn't intended, please respond and I will assign a reviewer.

This primarily serves as a hook for discussion; if a change along these lines would be deemed valuable I'd be up for following up and driving this through completion. We can do that here, or on Slack, but it isn't fit for review as-is.

mum4k:
I think centralizing tracking of milestones is worthwhile. I don't fully understand the use case yet. Would you be able to provide (either here on Slack) an example of a testing scenario that involves Nighthawk and would utilize tracking of multiple milestones? Ideally from the user's perspective.

oschaaf:
> I think centralizing tracking of milestones is worthwhile. I don't fully understand the use case yet. Would you be able to provide (either here on Slack) an example of a testing scenario that involves Nighthawk and would utilize tracking of multiple milestones? Ideally from the user's perspective.

Well .. this would be a prelude for some things that could be interesting to end users: 
- Configuring which delta's to track in the lifetime of a request 
- Annotated milestones may allow for reports that group on things like connection-id, origin-id, request-path, endpoint.
- In conjunction with a hypothetical new `TopXStatistic`, which internally tracks the top X milestones, efficient raw exports could provide an interesting high-res view of where in some case our signal is (the long tail).

As it stands, it doesn't bring any of that as-is. Yet it does perhaps clean up the code base for a bit by disassociating the histograms from nearly all other concepts: the milestone completion callback will be the single place where we process measurements.

mum4k:
Thank you, I would need more discussion to understand the use cases you mention (or more code examples). Maybe we can organize a meeting at some point and hash that out.

With that said, I see the value in code simplification and separation of histograms, so if you have the cycles - proceeding with this PR would be worth-while.

oschaaf:
> With that said, I see the value in code simplification and separation of histograms, so if you have the cycles - proceeding with this PR would be worth-while.

allright, I'll see if I can drive this to completion then; as for discussing use cases more: sure, you know where to find me :-)

mum4k:
Closing due to inactivity in order to keep the pending PR list easy to review. Feel free to reopen if you start working on this.

PR #659, opened 2021-03-21
SimpleStatistics: implement serialization roundtrip

oschaaf:
Part of the horizontal scaling effort: `SimpleStatistic` and `StreamingStatistic ` get used
for tracking header byte counts, hence we should be able to wire
transfer statistics for those to be able to properly aggregate.


oschaaf:
Perhaps not obvious when looking at the code changes here, but test coverage for this is in place via `TypedStatisticTest. NativeRoundtrip`

mum4k:
@qqustc please review and assign back to me when done.

api/stats/internal.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package nighthawk;
+
+// Internal protos for representing Statistics in a way that allows
+// "lossless" reconstruction. Do not consume / rely on this in any other place
+// then Nighthawk code as this may be subject to change.
```
qqustc, line 7: typo: than

source/common/statistic_impl.cc
```diff
@@ -104,6 +106,38 @@ StatisticPtr SimpleStatistic::combine(const Statistic& statistic) const {
   return combined;
 }
 
+absl::StatusOr<std::unique_ptr<std::istream>> SimpleStatistic::serializeNative() const {
```
qqustc, line 109: when will this function return absl::Status?

oschaaf: Fixed in f7b1269b

api/stats/internal.proto
oschaaf: Fixed in f7b1269b

source/common/statistic_impl.cc
oschaaf: Some statistics that don't implement this return `absl::Status(absl::StatusCode::kUnimplemented, "serializeNative not implemented.");
}`. 
Also the Hdr variant has:
```c++
absl::StatusOr<std::unique_ptr<std::istream>> HdrStatistic::serializeNative() const {
  char* data;
  if (hdr_log_encode(histogram_, &data) == 0) {
    auto write_stream = std::make_unique<std::stringstream>();
    *write_stream << absl::string_view(data, strlen(data));
    // Free the memory allocated by hrd_log_encode.
    free(data);
    return write_stream;
  }
  ENVOY_LOG(error, "Failed to write HdrHistogram data.");
  return absl::Status(absl::StatusCode::kInternal, "Failed to write HdrHistogram data");
}
```

In this case the simplicity is such that I don't see how this would need to return a status instead of a serialized representation; an exception is warranted in that case I guess.

qqustc: ah, thanks Otto for the explaination.

qqustc (dismissed):
LGTM, assign back to @mum4k

api/stats/internal.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
```
mum4k, line 1: Can we rename this proto file to something more specific to its content? E.g. `statistic.proto` or similar? This is to avoid attracting pretty much any internal messages and keep the files focused.

We can create a subdirectory `internal/` for internal protos and we probably don't want that to live under `api/`, since that implies external use. We could create a directory called `internal_proto` or similar somewhere else.

Also if these proto are only meant for internal use, this could be a good use case to introduce a `nighthawk::internal` namespace. Up to you whether you think this is necessary. It would remove the need of prepending all their names with "Internal".

api/stats/internal.proto
```diff
@@ -0,0 +1,25 @@
...
+
+package nighthawk;
+
+// Internal protos for representing Statistics in a way that allows
+// "lossless" reconstruction. Do not consume / rely on this in any other place
+// than Nighthawk code as this may be subject to change.
+
+message InternalSimpleStatistic {
+  uint64 count = 1;
```
mum4k, line 10: Were some of these meant to be oneoff? Or is this message going to carry multiple values?

api/stats/internal.proto
```diff
@@ -0,0 +1,25 @@
...
+  string id = 2;
+  uint64 min = 5;
+  uint64 max = 6;
+  double sum_x = 7;
+  double sum_x_2 = 8;
```
mum4k, line 15: Actually can we document both the fields and the messages? While they are internal, documentation would help other developers to understand what they are.

mum4k (changes requested):

api/stats/internal.proto
mum4k: I since saw how they are used, so resolving this.

api/stats/internal.proto
oschaaf: Pushed 8ce99fc9 to restructure accordingly

api/stats/internal.proto
oschaaf: In 8ce99fc9 I added a comment, which explains what this file is, and defers to `statistics_impl.h`  for further details.
Would that suffice?

mum4k: Thank you, yes that helps direct the reader to the information they are seeking.

mum4k (approved):

PR #662, opened 2021-03-22
Update Envoy to ff3b5e3 (March 22, 2021)

dubious90:
Required changes:
- signature change for how envoy creates files
- a new enum value (ProtocolError) needed to be accounted for in streamResetReasonToResponseFlag
- check_format script in envoy moved
- error message for integer out of range in protos was modified slightly - made it less specific to fix

Other necessary changes not necessarily caused by update
- Fixed an environment-sensitivity issue where time_since_epoch doesn't.count() wasn't reliably nanoseconds
- In some environments (including mine), request source tests were failing because they couldn't find the files - fixed by adding fileset to BUILD


test/process_test.cc
```diff
@@ -245,7 +245,9 @@ TEST_P(ProcessTestWithSimTime, ScheduleAheadWorks) {
       ASSERT_EQ(output.results_size(), 1);
       EXPECT_EQ(Envoy::ProtobufUtil::TimeUtil::TimestampToNanoseconds(
                     output.results()[0].execution_start()),
-                options_->scheduled_start().value().time_since_epoch().count());
+                std::chrono::duration_cast<std::chrono::nanoseconds>(
+                    options_->scheduled_start().value().time_since_epoch())
```
oschaaf, line 249: Ugh :/ this must have been annoying to figure out.. is this clang vs gcc?

oschaaf (approved):

test/process_test.cc
dubious90: This one actually wasn't too bad!  Reading chrono duration documentation wasn't super fun though.

I assume this is clang vs gcc or something like that?  Honestly, I didn't figure this out.  It was broken on my computer but not in docker, so it's something like that, but I didn't investigate what the actual difference was... sorry.

mum4k (approved):

Issue #613, opened 2021-01-28
connection failure with localhost on IPv4 and IPv6


eric846:
- Start an Envoy load balancer listening on 0.0.0.0:12345.
- Try to run a load test to http://localhost:12345.

You may get connection failures on some machines.

Sending the load to http://127.0.0.1:12345 does work. 

You may have to be using a machine that supports both IPv4 and IPv6

Tentative theory:
- Because IPv6 is available, Nighthawk resolves localhost to the IPv6 address ::1.
- The test server is explicitly listening only on IPv4.
- Connecting to ::1 on port 12345 does not reach the test server, which is only listening on all IPv4 addresses.

That is, the issue could be with how Nighthawk resolves "localhost" on a dual IPv4/IPv6 machine.

oschaaf:
Trace level logging should dig up [1] what address get used to validate, but I think your theory will turn out to be correct.
But ... extending the theory for a bit, if 
- ipv6 gets prioritised over ipv4 at the dns resolution level, and 
- Nighthawk is configured to use either of them ..
Then one could say the dns resolution behaved as expected?
If the extended theory holds, then still, I feel this isn't great UX and maybe not that uncommon to run into.
If we could probe using the other address family upon failure of trying the preferred one and notice that there's a
http service running there, maybe we could write a helpful warning message before exiting.

[1] https://github.com/envoyproxy/nighthawk/blob/main/source/common/uri_impl.cc#L81

eric846:
Added some advice about `localhost`, `::1`, `127.0.0.1`, and `--address-family v4` / `v6`. This should be enough to unblock most situations.

PR #666, opened 2021-03-29
Simple "Error 13" message improvement

eric846:
When there is a URI exception or failing failure predicate, log troubleshooting tips and return the tips in the Nighthawk Service error description.

Add a default failure predicate for `benchmark.stream_resets` > 0, since stream resets usually indicate a misconfigured protocol.

NOTE: If you expect stream resets in a benchmark session, run with `--failure-predicate benchmark.stream_resets:1000000000` to avoid early termination.

Related to #181 and #613.


oschaaf (approved):
Nice, LGTM

dubious90:
@wjuan-AFK Please review and assign back to me once done

wjuan-AFK (approved):
LGTM

source/client/process_impl.cc
```diff
@@ -598,7 +598,23 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
   StatisticFactoryImpl statistic_factory(options_);
   collector.addResult("global", mergeWorkerStatistics(workers_), counters,
                       total_execution_duration / workers_.size(), first_acquisition_time);
-  return counters.find("sequencer.failed_terminations") == counters.end();
+  if (counters.find("sequencer.failed_terminations") == counters.end()) {
+    return true;
+  } else {
+    ENVOY_LOG(
```
dubious90, line 604: (optional) How do you feel about splitting this into two log statements - one error log "Terminated early because of a failure predicate" and a second log (info/debug/verbose/something) with all of these extra details?

source/client/service_impl.cc
```diff
@@ -42,7 +42,18 @@ void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionReque
     response.mutable_error_detail()->set_code(grpc::StatusCode::INTERNAL);
     // TODO(https://github.com/envoyproxy/nighthawk/issues/181): wire through error descriptions, so
     // we can do better here.
-    response.mutable_error_detail()->set_message("Unknown failure");
+    response.mutable_error_detail()->set_message(
+        "Unknown failure. See Nighthawk Service logs. Make sure the URI is well formed and the DNS "
+        "name resolves (if applicable). Check the output for problematic counter values. The "
+        "default Nighthawk failure predicates report failure if (1) Nighthawk could not connect to "
```
dubious90, line 48: I know there was a lot of churn in this investigation, so please feel free to push back.

Were we unable to identify where the predicates are resulting in failures?  This feels like it'd be much better if we could instead log at the point of a predicate failing what predicate was failing and what its value was set to.  Is that not possible?

source/client/service_impl.cc
```diff
@@ -42,7 +42,18 @@ void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionReque
...
+        "the target (see 'benchmark.pool_connection_failure' counter; check the address and port "
```
dubious90, line 49: Regardless of whether or not we make the above change, can we create a TODO/issue a bug for moving this to a documentation location for common trouble areas (similar to our internal playbooks)?  This whole log feels like it really belongs in a markdown document rather than in a log statement.

(Afterwards, we can instead recommend that people reference our troubleshooting document, instead of including everything in this log statement)

eric846: Opened #668.

source/client/service_impl.cc
eric846: Termination predicates are based on counters which don't seem to store their own names. See termination_predicate_impl.h. We could log the first time each counter was incremented in benchmark_client_impl.cc, without knowing whether it will trigger a failure predicate. Logging in either place could impact performance unless we add logic to bypass logging on subsequent times. The user would have to check the log for the detailed info since it would take more code to propagate the information to the point where we return from the service.

source/client/process_impl.cc
eric846: Done.

source/client/service_impl.cc
oschaaf: There is `ENVOY_LOG_ONCE` these days... but thinking about this, maybe if we'd allow wiring in some sort of callback when constructing predicates, which would fire when one indicates termination, that could be helpful here. There is one gotcha to that, which is that one part of the chain might conditionally override the return value of another. This is hypothetical with the current predicate implementations, but conceptually it doesn't feel right to me. So thinking out loud some more, maybe this callback should be specific to a new top level predicate implementation, which never indicates termination all by itself - it relies on the next links in the chain to figure that out -- but is just there to serialize what happened in the chain when termination commenced, for the purpose of communicating that upwards. It think it would be useful to be able to communicate this in a structured way, especially in the gRPC flows.
TLDR; the current state seems like a good enhancement to me, perhaps punt improving predicates to do a better job at facilitating UX to an issue and discuss further out of the context of this issue if need be?

dubious90: agree that we can go forward without fixing this in this PR.  Creating an issue for improving termination predicates (with logging or something) sounds like a good idea to me.

dubious90 (approved):

PR #669, opened 2021-04-01
Horizontal scaling: Distributor gRPC service + tests

oschaaf:
Part of the horizontal scaling effort: add gRPC distributor service + tests.

This new gRPC service allows delegation of load test execution initiation messages
to a set of load generator services.


source/distributor/service_impl.cc
```diff
@@ -0,0 +1,117 @@
...
+
+#include "external/envoy/source/common/common/assert.h"
+#include "external/envoy/source/common/protobuf/message_validator_impl.h"
+#include "external/envoy/source/common/protobuf/utility.h"
+
+#include "api/distributor/distributor.pb.validate.h"
+
+namespace Nighthawk {
+
+::grpc::Status NighthawkDistributorServiceImpl::validateRequest(
```
dubious90, line 15: can this be a private function in an anonymous namespace, rather than a private method object?

source/distributor/service_impl.cc
```diff
@@ -0,0 +1,117 @@
...
+    const ::nighthawk::DistributedRequest& request) const {
+  auto& validation_visitor = Envoy::ProtobufMessage::getStrictValidationVisitor();
```
dubious90, line 17: nit: I think this would be more readable with ValidationVisitor& here instead of auto&.  The type inference isn't quite obvious enough for me.

source/distributor/service_impl.cc
```diff
@@ -0,0 +1,117 @@
...
+        stream) {
+  nighthawk::DistributedRequest request;
+  ::grpc::Status status = grpc::Status::OK;
+  while (status.ok() && stream->Read(&request)) {
+    ENVOY_LOG(trace, "Inbound DistributedRequest {}", request.DebugString());
+    status = validateRequest(request);
+    if (status.ok()) {
+      std::tuple<grpc::Status, nighthawk::DistributedResponse> status_and_response =
+          handleRequest(request);
+      status = std::get<0>(status_and_response);
```
dubious90, line 98: if we know status = ok, this line has no purpose.

source/distributor/service_impl.cc
```diff
@@ -0,0 +1,117 @@
...
+  }
+  return ::grpc::Status::OK;
+}
+
+absl::StatusOr<nighthawk::client::ExecutionResponse>
+NighthawkDistributorServiceImpl::handleExecutionRequest(
+    const envoy::config::core::v3::Address& service,
+    const ::nighthawk::client::ExecutionRequest& request) const {
+  RELEASE_ASSERT(service_client_ != nullptr, "service_client_ != nullptr");
+  std::unique_ptr<nighthawk::client::NighthawkService::Stub> stub;
```
dubious90, line 43: can we instead declare this where it's used?

source/distributor/service_impl.cc
```diff
@@ -0,0 +1,117 @@
...
+  std::shared_ptr<::grpc::Channel> channel;
```
dubious90, line 44: nit: remove :: at start of grpc::Channel

api/distributor/distributor.proto
```diff
@@ -16,8 +18,18 @@ message DistributedRequest {
   repeated envoy.config.core.v3.Address services = 3 [(validate.rules).repeated .min_items = 1];
 }
 
+message DistributedResponseFragment {
```
dubious90, line 21: can we add a top-level explanation of the purpose of this object?

Also, looking at the code, I think that the intention here is that one of these is associated with every nighthawk service, and essentially operates as either:
1. We failed to connect to this service
2. Here's the service address.

Given that, is DistributedResponseFragment a good name here?  I don't love the name fragment here, because it implies to me that they are parts of a whole (often without independent meaning outside of that whole).

dubious90 (approved):
Some nits and comments, but in general looks good.  @mum4k to approve/merge.

api/distributor/distributor.proto
oschaaf: Yes this is spot on; I'll think about a better name here. Also open to suggestions.

source/distributor/service_impl.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic warning "-Wunused-parameter"
```
mum4k, line 4: Can we include a comment explaining why this is needed?

source/distributor/service_impl.h
```diff
@@ -0,0 +1,54 @@
...
+  NighthawkDistributorServiceImpl(std::unique_ptr<NighthawkServiceClient> service_client)
+      : service_client_(std::move(service_client)) {}
+
+  grpc::Status DistributedRequestStream(
+      grpc::ServerContext* context,
+      grpc::ServerReaderWriter<nighthawk::DistributedResponse, nighthawk::DistributedRequest>*
+          stream) override;
+
+private:
+  grpc::Status validateRequest(const nighthawk::DistributedRequest& request) const;
```
mum4k, line 44: (optional) Do all of these need to be on the class? Or can some of them just reside in the anonymous namespace in the .cc file without class access (maybe with small amounts of parameters passed in)?

Anonymous namespace being the smallest scope we can give a function requires less attention from any future readers.

mum4k (changes requested):
Just started the review and then noticed that there are still outstanding comments from @dubious90, so the code will change.

Sending what I have and will continue once the existing comments are addressed. But please let me know if I misunderstood and this is blocked on me.

api/distributor/distributor.proto
dubious90: The DistributorRequest references `services`, so for a name how about `DistributedServiceResponse` or just `DistributedResponse`?

I think you were using `Fragment` to indicate that it's not the full resultset, but I don't think that's necessary here, as long as we document the intent.

(Some APIs I have used in the past have referred to this as an AsyncResponse, but I'm not sure how conventional that is).

source/distributor/service_impl.cc
oschaaf: done in 6471b8687332bcae77a637c8b1d86640299e20c6

source/distributor/service_impl.cc
oschaaf: done in 6471b8687332bcae77a637c8b1d86640299e20c6

source/distributor/service_impl.cc
oschaaf: done in 6471b8687332bcae77a637c8b1d86640299e20c6

source/distributor/service_impl.cc
oschaaf: fixed where applicable in 6471b8687332bcae77a637c8b1d86640299e20c6

source/distributor/service_impl.cc
oschaaf: Well this does break the while loop when status is not OK?

source/distributor/service_impl.h
oschaaf: Done in 6471b8687332bcae77a637c8b1d86640299e20c6

api/distributor/distributor.proto
oschaaf: Renamed in 5462844f. I went with `DistributedServiceResponse`, is `DistributedResponse` was already used in:

```proto
message DistributedResponse {
  repeated DistributedServiceResponse service_response = 1;
}
```

source/distributor/service_impl.h
oschaaf: Even better, trying it out, we no longer need this. I looks like one of the dep updates eliminated the need for this, si I removed the pragmas in f3bc7a7aac0579bd9be72d9cde1928fdf3453eba

source/distributor/service_impl.cc
```diff
@@ -0,0 +1,119 @@
...
+                                grpc::InsecureChannelCredentials());
+  std::unique_ptr<nighthawk::client::NighthawkService::Stub> stub =
+      std::make_unique<nighthawk::client::NighthawkService::Stub>(channel);
+  return service_client_->PerformNighthawkBenchmark(stub.get(), request.start_request().options());
+}
+
+// Translates one or more backend response into a single reply message
+std::tuple<grpc::Status, nighthawk::DistributedResponse>
+NighthawkDistributorServiceImpl::handleRequest(const nighthawk::DistributedRequest& request) const {
+  RELEASE_ASSERT(request.has_execution_request(), "request.has_execution_request()");
```
mum4k, line 58: Why do we prefer to have RELEASE_ASSERTs as compared to validating the request and returning an error?

Assuming the asserts crash the code, it may be ok to do that when verifying internal correctness, but we probably don't want to do that for an invalid argument.

mum4k (changes requested):
Looks good, just one question about asserts.

source/distributor/service_impl.cc
oschaaf: This is pre-checked in `validateRequest()`, hence this assert triggering is indicative of something being very wrong.
The flow is covered in a test: https://github.com/envoyproxy/nighthawk/pull/669/files/5462844f69719621185528e77cb583c66f21900a#diff-1b41e54eeb708a749bd8c54a830c84409cecdd778b7e578dbcd45dc575e1416fR87
Given that context, does that address your concern?

oschaaf: Following our offline discussion, removed the asserts is d0fb1b4c. Thanks for persisting @mum4k, this is overly cautious, and not sustainable as a strategy. Thanks for pointing that out.

mum4k: For posterity - discussed offline, the asserts are considered to be too defensive, the unit tests in place already verify the validation occurs before the code handles the request.

Otto please indicate if I misinterpreted anything.

mum4k (approved):

PR #672, opened 2021-04-11
Enable test_gcc on arm64 platform for ci

TrevorTaoARM:
Add aarch64 Bazel build options for GCC toolchain
which would enable the test_gcc for ci script on
arm platform.

Signed-off-by: trevor tao <trevor.tao@arm.com>


oschaaf (approved):
This is great. I think we are lacking the right CI env to test this, correct? Did you test the resulting binaries manually?

TrevorTaoARM:
> This is great. I think we are lacking the right CI env to test this, correct? Did you test the resulting binaries manually?

Yes, I tested it in my local environment. Thanks.

TrevorTaoARM:
> > This is great. I think we are lacking the right CI env to test this, correct? Did you test the resulting binaries manually?
> 
> Yes, I tested it in my local environment. Thanks.
Now the Clang and Gcc test are both ok now.

mum4k (approved):

PR #679, opened 2021-05-06
Optional cooldown between benchmarks done by the adaptive controller.

mum4k:
Allows the user to specify a cooldown period to wait for between individual benchmarks executed by the adaptive load controller. Useful in situations where cooldown is required to allow previously used connections to time out.

Also ensuring all durations in the adative session spec are validated.

### DIFF
--- a/source/adaptive_load/adaptive_load_controller_impl.cc
+++ b/source/adaptive_load/adaptive_load_controller_impl.cc
@@ -13,6 +13,7 @@
 #include "external/envoy/source/common/common/logger.h"
 #include "external/envoy/source/common/common/statusor.h"
 #include "external/envoy/source/common/protobuf/protobuf.h"
+#include "external/envoy/source/common/protobuf/utility.h"
 
 #include "api/adaptive_load/adaptive_load.pb.h"
 #include "api/adaptive_load/benchmark_result.pb.h"
--- a/source/adaptive_load/adaptive_load_controller_impl.cc
+++ b/source/adaptive_load/adaptive_load_controller_impl.cc
@@ -13,6 +13,7 @@
 #include "external/envoy/source/common/common/logger.h"
 #include "external/envoy/source/common/common/statusor.h"
 #include "external/envoy/source/common/protobuf/protobuf.h"
+#include "external/envoy/source/common/protobuf/utility.h"
 
 #include "api/adaptive_load/adaptive_load.pb.h"
 #include "api/adaptive_load/benchmark_result.pb.h"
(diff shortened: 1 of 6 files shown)


mum4k:
Please review and assign to @dubious90 when done.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -84,6 +85,22 @@ StepControllerPtr LoadStepControllerPluginFromSpec(const AdaptiveLoadSessionSpec
...
+ * Converts the duration proto to a number of milliseconds.
+ *
+ * @param duration the duration to convert.
+ *
+ * @return the number of milliseconds or an error if the conversion failed.
+ */
+absl::StatusOr<uint64_t>
+ProtobufDurationToMilliseconds(const Envoy::ProtobufWkt::Duration& duration) noexcept {
+  try {
+    return Envoy::DurationUtil::durationToMilliseconds(duration);
```
eric846, line 98: Elsewhere in this file we are using a conversion function that doesn't throw an exception:
https://github.com/envoyproxy/nighthawk/blob/cf13dd102b8f6b28bec864b8785a62c972168e60/source/adaptive_load/adaptive_load_controller_impl.cc#L175

It may or may not make things simpler here.

eric846 (dismissed):
LGTM with potential alternate conversion function

source/adaptive_load/adaptive_load_controller_impl.cc
mum4k: Done in 4678ada.

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -172,6 +172,18 @@ absl::StatusOr<AdaptiveLoadSessionOutput> AdaptiveLoadControllerImpl::PerformAda
...
 
+    if (spec.has_benchmark_cooldown_duration()) {
+      ENVOY_LOG_MISC(info, "Cooling down before the next benchmark for duration: {}",
+                     spec.benchmark_cooldown_duration());
+      absl::StatusOr<uint64_t> sleep_time_ms =
+          Envoy::Protobuf::util::TimeUtil::DurationToMilliseconds(
+              spec.benchmark_cooldown_duration());
+      if (!sleep_time_ms.ok()) {
+        return sleep_time_ms.status();
+      }
```
eric846, line 184: I think `StatusOr` here is no longer relevant, right?

source/adaptive_load/adaptive_load_controller_impl.cc
```diff
@@ -13,6 +13,7 @@
 #include "external/envoy/source/common/common/logger.h"
 #include "external/envoy/source/common/common/statusor.h"
 #include "external/envoy/source/common/protobuf/protobuf.h"
+#include "external/envoy/source/common/protobuf/utility.h"
```
eric846, line 16: May not be needed if it was only for the previous Duration helper.

eric846 (changes requested):

test/adaptive_load/BUILD
```diff
@@ -62,6 +62,7 @@ envoy_cc_test(
         "//test/mocks/common:mock_nighthawk_service_client",
         "@envoy//source/common/common:assert_lib_with_external_headers",
         "@envoy//source/common/common:statusor_lib_with_external_headers",
+        "@envoy//source/common/common:utility_lib_with_external_headers",
```
eric846, line 65: May not be needed anymore

dubious90 (approved):
LGTM modulo Eric's comments.  Feel free to submit once that is addressed.

source/adaptive_load/adaptive_load_controller_impl.cc
mum4k: Done in b4c29db.

test/adaptive_load/BUILD
mum4k: Done in b4c29db.

mum4k:
@dubious90 please take another look.

eric846:
LGTM

dubious90 (approved):

PR #680, opened 2021-05-06
Improve logging done by the adaptive controller.

mum4k:
Logs a stripped down version of the Nighthawk output for each of the short benchmarks.

The output only contains the global result without any statistics.


mum4k:
@qqustc  please review and assign to @dubious90 when done.

qqustc (approved):
LGTM. No need to be in the scope of current CL, but we should consider to provide a switch/flag to only log global result in general.

dubious90 (approved):

PR #684, opened 2021-05-13
Store the start and end time of each benchmark done by the adaptive load controller.

mum4k:
This will allow users of the adaptive load controller to correlate statistics collected from the system under test with individual benchmark stages (adjusting stage, testing stage).

Done here:
- updated the `FakeIncrementingMonotonicTimeSource` to also support incrementing system time.
- renamed `FakeIncrementingMonotonicTimeSource` to `FakeIncrementingTimeSource`, since it is no longer specific just to the monotonic time.
- added new methods that allow resetting the `FakeIncrementingTimeSource` to a specific epoch value. 
- added start and end time to the `BenchmarkResult` message and populating them while the adaptive controller executes.

### DIFF
--- a/api/adaptive_load/benchmark_result.proto
+++ b/api/adaptive_load/benchmark_result.proto
@@ -6,6 +6,7 @@ syntax = "proto3";
 package nighthawk.adaptive_load;
 
 import "api/client/output.proto";
+import "google/protobuf/timestamp.proto";
 import "google/rpc/status.proto";
 
 // Records the status of a single metric during a benchmark session.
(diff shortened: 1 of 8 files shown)


mum4k:
@wjuan-AFK please review and assign to @dubious90 when done.

wjuan-AFK:
lgtm

wjuan-AFK (approved):

dubious90 (approved):

mum4k:
Thank you for the reviews.

PR #690, opened 2021-06-05
WIP: Update Envoy to 72bf41f (Jun 4th 2021).

mum4k:
**Blocked** until  https://github.com/envoyproxy/envoy/pull/16829 or an alternative gets merged upstream (or an alternative gets implemented here).

- modified include paths to match recent change done in the envoy repository (https://github.com/envoyproxy/envoy/commit/0fae6970ddaf93f024908ba304bbd2b34e997a51), specifically:
  1. Move `include/nighthawk` up one level by running `git mv include/nighthawk nighthawk`.
  1. Use full paths on all `#include` statements referring to packages under the `source/` directory.
  1. Fixing incorrect include paths (pre-existing) that were referring to envoy headers without the `external/` prefix. This was reported by `check_format` after the above changes.
- Changing a `bool` to `envoy::config::core::v3::DnsResolverOptions` in a call to `Envoy::Event::Dispatcher::createDnsResolver` to match a recent Envoy change (https://github.com/envoyproxy/envoy/commit/2c0e1f21837fdf78cc91d2eb8c3735731d510f7e).
- changing how the HTTP connection pool is accessed after it was moved behind a new class `HttpPoolData` in envoy (https://github.com/envoyproxy/envoy/commit/ff2c0e517c8677dd1c553e1c6ef55d2b14cdd996).
- after envoy switched to the true abseil type of string_view (https://github.com/envoyproxy/envoy/commit/72bf41fb0ecc039f196be02f534bfc2c9c69f348), we can no longer assign `absl::string_view` to `std::string`. Changing Nighthawk code accordingly. 


mum4k:
Going to recreate this without the extra commits that made it into the history here.

PR #714, opened 2021-07-26
Update Envoy to b4027cf (June 26th 2021)

jiajunye:
1. Update ENVOY_COMMIT to the latest Envoy's commit
2. Update ENVOY_SHA in bazel/repositories.bzl
3. Pass an extra QuicStatNames argument when creating the cluster_manager_factory_ in process_impl.cc. This is to accommodate the recent change in the constructor of Envoy::Upstream::ProdClusterManagerFactory in https://github.com/envoyproxy/envoy/pull/17222.

Signed-off-by: Jason Ye jiajunye@google.com


source/client/process_impl.cc
```diff
@@ -522,7 +523,8 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
         admin_, Envoy::Runtime::LoaderSingleton::get(), store_root_, tls_,
         dispatcher_->createDnsResolver({}, dns_resolver_options), *ssl_context_manager_,
         *dispatcher_, *local_info_, secret_manager_, validation_context_, *api_, http_context_,
-        grpc_context_, router_context_, access_log_manager_, *singleton_manager_, envoy_options);
+        grpc_context_, router_context_, access_log_manager_, *singleton_manager_, envoy_options,
+        quic_stat_names_);
```
dubious90, line 527: rather than storing quic_stat_names_ on the object (requiring changes here and in the .h file), could we instead just use store_root_.symbolTable() directly here?
As is, I believe we are essentially storing this on our ProcessImpl twice.

dubious90 (changes requested):
For the description:
```
Update ENVOY_COMMIT to the latest Envoy's commit
Update ENVOY_SHA in bazel/repositories.bzl
```
These are unnecessary details imo, as they are just describing how the envoy import happens.

dubious90 (approved):

source/client/process_impl.cc
jiajunye: discussed this offline. the new QuicStatNames argument in the ClusterManagerFactory constructor is a non const reference. so it expects an l-value to be passed in.

Issue #23, opened 2019-04-23
Support QUIC


oschaaf:
It would be awesome to have support for QUIC eventually. It is anticipated that once this feature is landed in Envoy, adding it here will be low hanging fruit.

PR #720, opened 2021-08-07
[QUIC 020]: Moving code that creates the Envoy bootstrap into its own library.

mum4k:
Having the code in its own library not only allows to test it, but it also provides documentation of how various Nighthawk options translate to the public Envoy API.

No functional changes in this PR except for changing the return value to absl::StatusOr (i.e. the code no longer throws).

Done here:

- adding a test utility that compares proto messages and prints diffs.
- moving the bootstrap generation code into its own library.
- adding unit tests for the major behaviors of the bootstrap generation code.

Works on #23

### DIFF
--- a/source/client/process_bootstrap.cc
+++ b/source/client/process_bootstrap.cc
@@ -57,15 +57,15 @@ absl::StatusOr<Bootstrap> createBootstrapConfiguration(const Client::Options& op
       transport_socket->set_name("envoy.transport_sockets.tls");
       envoy::extensions::transport_sockets::tls::v3::UpstreamTlsContext context =
           options.tlsContext();
-      const std::string sni_host = Client::SniUtility::computeSniHost(
-          uris, options.requestHeaders(), options.upstreamProtocol());
+      const std::string sni_host =
+          Client::SniUtility::computeSniHost(uris, options.requestHeaders(), options.protocol());
       if (!sni_host.empty()) {
         *context.mutable_sni() = sni_host;
       }
       auto* common_tls_context = context.mutable_common_tls_context();
-      if (options.upstreamProtocol() == Envoy::Http::Protocol::Http2) {
+      if (options.protocol() == Envoy::Http::Protocol::Http2) {
         common_tls_context->add_alpn_protocols("h2");
-      } else if (options.upstreamProtocol() == Envoy::Http::Protocol::Http3) {
+      } else if (options.protocol() == Envoy::Http::Protocol::Http3) {
         return absl::UnimplementedError("HTTP/3 Quic support isn't implemented yet.");
       } else {
         common_tls_context->add_alpn_protocols("http/1.1");
@@ -78,7 +78,7 @@ absl::StatusOr<Bootstrap> createBootstrapConfiguration(const Client::Options& op
     cluster->set_name(fmt::format("{}", i));
     cluster->mutable_connect_timeout()->set_seconds(options.timeout().count());
     cluster->mutable_max_requests_per_connection()->set_value(options.maxRequestsPerConnection());
-    if (options.upstreamProtocol() == Envoy::Http::Protocol::Http2) {
+    if (options.protocol() == Envoy::Http::Protocol::Http2) {
       auto* cluster_http2_protocol_options = cluster->mutable_http2_protocol_options();
       cluster_http2_protocol_options->mutable_max_concurrent_streams()->set_value(
           options.maxConcurrentStreams());
--- a/source/client/process_bootstrap.cc
+++ b/source/client/process_bootstrap.cc
@@ -8,34 +8,171 @@
 
 #include "external/envoy/source/common/common/statusor.h"
 #include "external/envoy_api/envoy/config/bootstrap/v3/bootstrap.pb.h"
+#include "external/envoy_api/envoy/extensions/transport_sockets/quic/v3/quic_transport.pb.h"
+#include "external/envoy_api/envoy/extensions/upstreams/http/v3/http_protocol_options.pb.h"
 
 #include "source/client/sni_utility.h"
 
 namespace Nighthawk {
 namespace {
 
-using envoy::config::bootstrap::v3::Bootstrap;
+using ::envoy::config::bootstrap::v3::Bootstrap;
+using ::envoy::config::cluster::v3::CircuitBreakers;
+using ::envoy::config::cluster::v3::Cluster;
+using ::envoy::config::core::v3::Http2ProtocolOptions;
+using ::envoy::config::core::v3::Http3ProtocolOptions;
+using ::envoy::config::core::v3::SocketAddress;
+using ::envoy::config::core::v3::TransportSocket;
+using ::envoy::config::endpoint::v3::ClusterLoadAssignment;
+using ::envoy::config::endpoint::v3::LocalityLbEndpoints;
+using ::envoy::config::metrics::v3::StatsSink;
+using ::envoy::extensions::transport_sockets::quic::v3::QuicUpstreamTransport;
+using ::envoy::extensions::transport_sockets::tls::v3::CommonTlsContext;
+using ::envoy::extensions::transport_sockets::tls::v3::UpstreamTlsContext;
 
-// Adds a cluster for the request source into the bootstrap.
-void addRequestSourceCluster(const Client::Options& options, const Uri& uri, int worker_number,
-                             Bootstrap& bootstrap) {
-  auto* cluster = bootstrap.mutable_static_resources()->add_clusters();
-  cluster->mutable_http2_protocol_options();
-  cluster->set_name(fmt::format("{}.requestsource", worker_number));
-  cluster->set_type(
-      envoy::config::cluster::v3::Cluster::DiscoveryType::Cluster_DiscoveryType_STATIC);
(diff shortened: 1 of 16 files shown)


mum4k:

test/process_bootstrap_test.cc
```diff
@@ -0,0 +1,1153 @@
...
+  // Ensure the generated bootstrap is valid.
+  Envoy::MessageUtil::validate(*bootstrap, Envoy::ProtobufMessage::getStrictValidationVisitor());
+}
+
+TEST_F(CreateBootstrapConfigurationTest, CreatesBootstrapSetsMaxRequestToAtLeastOne) {
+  uris_.push_back(std::make_unique<UriImpl>("http://www.example.org"));
+  resolveAllUris();
+
+  std::unique_ptr<Client::OptionsImpl> options = Client::TestUtility::createOptionsImpl(
+      "nighthawk_client --max-pending-requests 0 http://www.example.org");
```
qqustc, line 927: nit: add a comment to highlight that even though the CLI flag has value 0, created bootstrap will still be set to 1. (otherwise it is not obvious)

test/test_common/proto_matchers.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+
+// A custom IgnoreCriteria that can be added to a MessageDifferencer to ignore
+// unknown fields by their field number, regardless of where they appear in a
+// message.
+class IgnoreUnknownFieldsGloballyByNumber
```
qqustc, line 14: Just fyi I remember we may have standard proto comparing util function that can ignore certain fields, why do we want to implement our own?

qqustc (approved):

test/test_common/proto_matchers.h
mum4k: Thank you @qqustc, can you point me to the function you are referring to? I have implemented our own, because I didn't find any.

qqustc: ah, the one I remembered are internal so I guess we have to implement our own

source/client/process_bootstrap.cc
```diff
@@ -0,0 +1,123 @@
...
+
+absl::StatusOr<Bootstrap> createBootstrapConfiguration(const Client::Options& options,
+                                                       const std::vector<UriPtr>& uris,
+                                                       const UriPtr& request_source_uri,
+                                                       int number_of_workers) {
+  Bootstrap bootstrap;
+  for (int i = 0; i < number_of_workers; i++) {
+    auto* cluster = bootstrap.mutable_static_resources()->add_clusters();
+
+    if (uris.empty()) {
```
dubious90, line 51: preexisting - why is this within the for loop?  given that we're basically asserting a parameter's existence, shouldn't this be the first thing this function does?

source/client/process_bootstrap.cc
```diff
@@ -0,0 +1,123 @@
...
+    // We do not support any retrying.
+    thresholds->mutable_max_retries()->set_value(0);
+    thresholds->mutable_max_connections()->set_value(options.connections());
+    // We specialize on 0 below, as that is not supported natively. The benchmark client will track
+    // in flight work and avoid creating pending requests in this case.
+    thresholds->mutable_max_pending_requests()->set_value(
+        options.maxPendingRequests() == 0 ? 1 : options.maxPendingRequests());
+    thresholds->mutable_max_requests()->set_value(options.maxActiveRequests());
+
+    cluster->set_type(
```
dubious90, line 97: preexisting - mangled enum

test/process_bootstrap_test.cc
```diff
@@ -0,0 +1,1153 @@
...
+            }
+          }
+        }
+      }
+    }
+    stats_flush_interval {
+      seconds: 5
+    }
+  )pb");
+  ASSERT_THAT(expected_bootstrap, StatusIs(absl::StatusCode::kOk));
```
dubious90, line 146: nit: ASSERT_OK?  (here and elsewhere)

dubious90 (approved):

test/test_common/proto_matchers.h
mum4k: That was my conclusion as well. Thank you for bringing it up though.

test/process_bootstrap_test.cc
mum4k: I think ASSERT_OK isn't available here.

test/process_bootstrap_test.cc
mum4k: Done in 37a2d10d464f9d197183ba869ed62a7de0dc4335.

source/client/process_bootstrap.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Nighthawk {
+
+/**
+ * Creates Envoy bootstrap configuration.
+ *
+ * The created bootstrap configuration can be used to upstream requests to the
+ * specified uris.
+ *
+ * @param options are the options Nighthawk was started with.
```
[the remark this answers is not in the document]
mum4k: Done in e054512219403459c36ba7596324afc1f94dba7e.

source/client/process_bootstrap.h
```diff
@@ -0,0 +1,34 @@
...
+ * @param uris are the endpoints to which the requests will be upstreamed. At
+ *        least one uri must be specified. It is assumed that all the uris have
+ *        the same scheme (e.g. https). All the uri objects must already be
+ *        resolved.
+ * @param request_source_uri is the address of the request source service to
+ *        use, can be NULL if request source isn't used. If not NULL, the uri
+ *        object must already be resolved.
+ * @param number_of_workers indicates how many Nighthawk workers will be
```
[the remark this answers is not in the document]
mum4k: Electing to not document this here, since it feels like this isn't the right place for such extensive commentary of a Nighthawk-wide behavior.

source/client/process_bootstrap.cc
mum4k: Good point, will address this in the next PR to avoid merge conflicts.

https://github.com/envoyproxy/nighthawk/pull/721

mum4k: Done in bfd032b6adbd820450ea99cf7edbd05f0130cc84.

mum4k:
@oschaaf please review and approve, but do not merge.

oschaaf (approved):
Some questions/thoughts for consideration aside, this LGTM
It's great to see decent test coverage for the bootstrap creation we perform.

source/client/process_impl.cc
```diff
@@ -531,21 +446,21 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const std::vector<UriP
             : Http1PoolImpl::ConnectionReuseStrategy::MRU);
     cluster_manager_factory_->setPrefetchConnections(options_.prefetchConnections());
     if (tracing_uri != nullptr) {
-      setupTracingImplementation(bootstrap, *tracing_uri);
-      addTracingCluster(bootstrap, *tracing_uri);
+      setupTracingImplementation(*bootstrap, *tracing_uri);
```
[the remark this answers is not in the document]
mum4k: While I would naturally lean towards this, our internal style requires a single variable name for StatusOr objects, so I have been muscle memory trained.

I don't think this is in the public style guide, so if you believe this will be better with two variable names for readability, let me know. Dereferencing a StatusOr is different than dereferencing a pointer, it essentially just calls .value() on it.

Going to resolve for now as this was "intended", but do reopen if you think the other approach is more readable.

test/process_bootstrap_test.cc
```diff
@@ -0,0 +1,1155 @@
...
+}
+
+TEST_F(CreateBootstrapConfigurationTest, CreatesBootstrapForH1) {
+  uris_.push_back(std::make_unique<UriImpl>("http://www.example.org"));
+  resolveAllUris();
+
+  std::unique_ptr<Client::OptionsImpl> options =
+      Client::TestUtility::createOptionsImpl("nighthawk_client http://www.example.org");
+
+  absl::StatusOr<Bootstrap> expected_bootstrap = parseBootstrapFromText(R"pb(
```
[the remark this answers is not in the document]
mum4k: This will fail on unknown fields, not sure why, but the textproto parser doesn't like them.

test/process_bootstrap_test.cc
```diff
@@ -0,0 +1,1155 @@
...
+    static_resources {
```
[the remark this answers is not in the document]
mum4k: I would move them to separate files if they were longer. As it stands my motivation was test case readability, since these tests also serve as a documentation of how our bootstrap looks like for various Nighthawk options. When the reader has the generated bootstrap right next to the options, they can see everything on one screen.

Let me know if you don't think this is worth keeping them in.

test/process_bootstrap_test.cc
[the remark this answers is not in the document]
oschaaf: Oh, great, that is the behaviour I was hoping for (if we remove a deprecated field at some point, this will break and point out we need to maintain the json contents). Resolving this.

test/process_bootstrap_test.cc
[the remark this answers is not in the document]
oschaaf: Allright, happy with this as is, that makes sense.

PR #725, opened 2021-08-10
[QUIC 060]: Adding an e2e test that uses HTTP/3 Quic.

mum4k:
Done here:
- Creating a Quic specific test fixture that changes the target IP address used by the Nighthawk client in e2e tests to loopback to avoid the following Quic error:
```
[01:33:37.413994][1494720][I] Client: Error: QUIC_DECRYPTION_FAILURE detail: Unable to decrypt ENCRYPTION_HANDSHAKE header protection (missing key).
```
- Improving how the Nighthawk client arguments are logged in e2e tests to simplify copy&pasting while debugging.
- Defining a single BUILD filegroup that contains all the test server configs and adding `nighthawk_https_origin_quic.yaml` for the HTTP/3 Quic e2e test.
- Adding the `envoy.access_loggers.file` extension to the Nighthawk build, this  extension is required to avoid a deprecation warning about the  `admin/access_log_path` field in the new test server config.

Works on #23.

### DIFF
--- a/test/integration/integration_test_fixtures.py
+++ b/test/integration/integration_test_fixtures.py
@@ -81,7 +81,7 @@ class IntegrationTestBase():
     self.request = request
     self.ip_version = request.param
     assert self.ip_version != IpVersion.UNKNOWN
-    self.server_ip = "::" if self.ip_version == IpVersion.IPV6 else "0.0.0.0"
+    self.server_ip = "::1" if self.ip_version == IpVersion.IPV6 else "127.0.0.1"
     self.server_ip = os.getenv("TEST_SERVER_EXTERNAL_IP", self.server_ip)
     self.tag = ""
     self.parameters = {}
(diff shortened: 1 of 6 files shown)


mum4k:

dubious90 (approved):

mum4k:
@oschaaf please review and approve, but do not merge.

test/integration/integration_test_fixtures.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, request, server_config, backend_count=1):
     self.request = request
     self.ip_version = request.param
     assert self.ip_version != IpVersion.UNKNOWN
-    self.server_ip = "::" if self.ip_version == IpVersion.IPV6 else "0.0.0.0"
+    self.server_ip = "::1" if self.ip_version == IpVersion.IPV6 else "127.0.0.1"
```
oschaaf, line 84: The benchmarking framework consumes this too. Would it be possible and safer to somehow localize this change to a derived class from `IntegrationTestBase`, specifically targeting tests for h/3, and use that?

oschaaf (approved):
One remark we should investigate I think; approving on the assumption that we're either ok or there will be a trivial fix that can be reviewed in the final larger PR this will re-emerge in.

jiajunye:
lgtm

test/integration/integration_test_fixtures.py
mum4k: Good point, done in 0f9949096e739c46a05af41e6a0d881cabb29b4f.

PR #733, opened 2021-08-16
Update Envoy to 295aba4 (Aug 16 2021) 

qqustc:
- Update bazel/repositories.bzl
- .bazelrc already in sync
- Add a bootstrap arg to Envoy::Api::Impl() ctor in ProcessImpl. This is because Nighthawk is broken by Envoy PR https://github.com/envoyproxy/envoy/pull/17562

Signed-off-by: qqustc@gmail.com <qqin@google.com>


source/client/process_impl.h
```diff
@@ -159,6 +159,7 @@ class ProcessImpl : public Process, public Envoy::Logger::Loggable<Envoy::Logger
   Envoy::ThreadLocal::InstanceImpl tls_;
   Envoy::Stats::ThreadLocalStoreImpl store_root_;
   Envoy::Quic::QuicStatNames quic_stat_names_;
+  envoy::config::bootstrap::v3::Bootstrap bootstrap_;
```
oschaaf, line 162: We do construct a bootstrap:
https://github.com/envoyproxy/nighthawk/blob/d05a0c9a4120812a921d01efdb59d1739c639c55/source/client/process_impl.cc#L405
Does the bootstrap that we need to start passing into the `Api` constructor in this PR need to be in sync with that?

qqustc: Ah, thanks Otto. I'm not aware about that, currently Api::Impl does not modify the passed in bootstrap. Anyway, I think it is a good idea to pass in the existing bootstrap.

mum4k:
Closing this PR as the changes were addressed in:
- https://github.com/envoyproxy/nighthawk/pull/734
- https://github.com/envoyproxy/nighthawk/pull/736

PR #734, opened 2021-08-18
Update Envoy to 95038fe (Aug 12th 2021).

mum4k:
- refactoring `ProcessImpl`, creating a named constructor that allows us to
  create bootstrap initially and pass it to `Envoy::Api::Impl()` as required
  since https://github.com/envoyproxy/envoy/pull/17562. Specifically:
  - Moving method `ProcessImpl::determineConcurrency()` out of the ProcessImpl class so that it can be used during its construction.
  - Moving code that extracts URIs from `process_impl.cc` into `process_bootstrap.cc`.
  - Adding a previously missing test case `CreatesBootstrapForH1RespectingPortInUri` into `process_bootstrap_test.cc`.
  - Removing a TODO that incorrectly indicated URI DNS resolution is optional. Envoy [requires](https://github.com/envoyproxy/envoy/blob/716ee8abc526d51f07ed6d3c2a5aa8a3b2481d9d/api/envoy/config/core/v3/address.proto#L64-L67) resolved IPs in the Bootstrap for cluster of type STATIC.
  - Creating a named constructor for `ProcessImpl` that creates the `Envoy::Api::Api` with an empty Bootstrap that is then replaced with the one generated. See an inline comment for explanation.
  - Moving callers onto the named constructor and making the original constructor of `ProcessImpl` private.
- no changes to `.bazelrc`, `.bazelversion`, `run_envoy_docker.sh`.


oschaaf (dismissed):
LGTM

source/client/process_impl.cc
```diff
@@ -153,6 +191,35 @@ ProcessImpl::ProcessImpl(const Options& options, Envoy::Event::TimeSystem& time_
...
+    return bootstrap.status();
+  }
+
+  // Ideally we would create the bootstrap first and then pass it to the
+  // constructor of Envoy::Api::Api. That cannot be done because of a circular
+  // dependency:
+  // 1) The constructor of Envoy::Api::Api requires an instance of Bootstrap.
+  // 2) The bootstrap generator requires an Envoy::Event::Dispatcher to resolve
+  //    URIs to IPs required in the Bootstrap.
+  // 3) The constructor of Envoy::Event::Dispatcher requires Envoy::Api::Api.
```
dubious90, line 213: Can we create an issue to potentially remove this circular dependency?  I wonder if one of these things (dispatcher?) doesn't really need what it current requires.

dubious90 (approved):

source/client/process_impl.cc
oschaaf: I was on the fence of making a similar comment earlier... +1 from me.

dubious90 (approved):

source/client/process_impl.cc
mum4k: Opened https://github.com/envoyproxy/nighthawk/issues/735.

PR #736, opened 2021-08-19
Update Envoy to dee7021 (Aug 18th 2021).

mum4k:
- Envoy now demands that `Envoy::Grpc::AsyncClient` is created, used and destroyed on the same thread. (See
  https://github.com/envoyproxy/envoy/pull/17619). Nighthawk creates the `Envoy::Grpc::AsyncClient` on the worker thread, but destroys it on the main which triggers an assertion failure. Fixing this by:
  - Adding a new method `RequestSource::destroyOnThread()` which complements the pre-existing `RequestSource::initOnThread()`. While the latter gets called when a worker thread starts, the former now gets called when a worker thread is shutting down.
  - Implementing `RequestSource::destroyOnThread()` on the only request source implementation that uses `Envoy::Grpc::AsyncClient`, i.e. `Nighthawk::RemoteRequestSourceImpl`, the method destroys the object that
    owns `Envoy::Grpc::AsyncClient` when called.
- no changes to `.bazelrc`, `.bazelversion`, `run_envoy_docker.sh`.
- lowering the coverage threshold to `93.2` to accommodate the fact that this code path is only covered in integration tests. The unit-tests use a mock.


oschaaf (dismissed):
the coverage test seems to indicate we started trending low, I'd be surprised if this PR pushed it over the edge, could that be a flake somehow? 
That aside, this LGTM

mum4k:
Adjusted the coverage threshold and included explanation in the PR description.

PTAL

oschaaf (approved):

PR #763, opened 2021-11-29
Fix bad import.

tomjzzhang:

### DIFF
--- a/source/common/termination_predicate_impl.h
+++ b/source/common/termination_predicate_impl.h
@@ -3,6 +3,8 @@
 #include "envoy/common/time.h"
 #include "envoy/stats/store.h"
 
+#include "external/envoy/source/common/common/assert.h"
+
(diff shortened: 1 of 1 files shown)


source/common/termination_predicate_impl.h
```diff
@@ -5,6 +5,8 @@
 
 #include "nighthawk/common/termination_predicate.h"
 
+#include "external/envoy/source/common/common/assert.h"
```
mum4k, line 8: Should we also modify the BUILD file to include the necessary target as a dep?

tomjzzhang: Please correct me if I'm wrong but I believe the dependency already is included in the BUILD file under https://github.com/envoyproxy/nighthawk/blob/main/source/common/BUILD#L153.

mum4k (approved):

PR #768, opened 2021-12-01
Architecture diagrams of Nighthawk.

### LINKED ISSUE #249: High level documentation: create a block diagram
In https://github.com/envoyproxy/nighthawk/pull/246 we insert a placeholder image.
That should be replaced with a real block diagram.

mum4k:
- replacing the draft architecture diagram with two diagrams (an overall architecture and architecture of a client worker).
- qualifying mentions of "pool" as "connection pool".
- adding text descriptions alongside the new diagrams.
- moving the "Notable upcoming changes" section to the end of the overview document.

Fixes #249.


mum4k:
@jiajunye please review and assign to @dubious90 once done.

docs/root/overview.md
```diff
@@ -1,60 +1,75 @@
...
-### Streaming parameterization and output stats
-
-Once we have phases, the gRPC service, and perhaps the CLI, would be natural
-candidates to follow up with to allow dynamic phase injection, as well as send
-back reports per phase.
+## Architecture
+
+The following diagram outlines the overall architecture of Nighthawk.
+
+![Nighthawk architecture](images/nighthawk_architecture.png)
```
oschaaf, line 7: I'm running github with the dark theme; the arrows are invisible in that case as they merge with the background.. perhaps use a white background instead of a transparant one? 

image

docs/root/overview.md
```diff
@@ -1,60 +1,75 @@
...
+
+Nighthawk exposes its API either via a [command line
+interface](https://github.com/envoyproxy/nighthawk/blob/main/source/exe/client_main_entry.cc)
```
oschaaf, line 10: Are we using fully qualified links on purpose? Or would it make sense to convert them to relative?

docs/root/overview.md
oschaaf: transparancy aside, do we want to show that:
- we can test multiple endpoints? 
- we can stream stats optionally to a separate backend during execution?

oschaaf (approved):
two additional thoughts/suggestions, regardless LGTM

docs/root/overview.md
mum4k: Added white background in 188c83057665ec6db290f9fbb0f9750b6eee3db1.

As for the other two questions - I would lean towards not doing that in these diagrams. The goal here isn't to outline all the features Nighthawk has, but to help understand its overall architecture. The more details we add here, the more distracted we get from that goal. These features can be found in the command line options, or alternatively we could add a separate document listing Nighthawk's features.

Choosing not to address thee here, but please let me know if you feel strongly.

docs/root/overview.md
mum4k: Changed links to relative in 3708abb045021068becfa7382a3112d51f28251c.

mum4k:
@oschaaf PTAL.

jiajunye:
LGTM

oschaaf (dismissed):

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+[Sequencer](../../source/common/sequencer_impl.h)
+which is responsible for maintaining the pace of requests that was specified by
+the user. To maintain the pace, the **Sequencer** queries a
+[RateLimiter](../../source/common/rate_limiter_impl.h)
+for [request-release timing](terminology.md#request-release-timing). When it
+is time to release a request, the
+[BenchmarkClient](../../source/client/benchmark_client_impl.h)
+is requested to do so by the **Sequencer**.
+
+**BenchMarkClient** retrieves data for the next request from the configured
```
dubious90, line 53: preexisting: let's be consistent and make this Benchmark Client (no capital m)

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+[RequestSource](../../source/common/request_source_impl.h)
+and asks its underlying **Connection pool** to create a
+[StreamDecoder](../../source/client/stream_decoder.h)
+which releases the request. The **StreamDecoder** sends the request and emits
+events as it progresses (connection pool ready, completion, etc...).
+
+The timings of each requests along with other statistics are recorded into the
```
dubious90, line 60: nit/grammar: each request

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+
+Once the execution finishes, the [Output
+Collector](../../source/client/output_collector_impl.h)
+provides results collected from individual **Client workers** to the [Output
+formatter](../../source/client/output_formatter_impl.h)
+which represents the results in the format selected by the user.
+
+### Client worker architecture
+
+![Client worker architecture](images/client_worker_architecture.png)
```
dubious90, line 41: For this diagram:

1. Can we indicate that the request source is optional? Maybe with a dotted arrow?
2. Can we add in the user here?  I think some arrows would help there, particularly with making clearer what happens to the things you are calling results

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+
+Each **Client worker** runs a
+[Sequencer](../../source/common/sequencer_impl.h)
+which is responsible for maintaining the pace of requests that was specified by
+the user. To maintain the pace, the **Sequencer** queries a
+[RateLimiter](../../source/common/rate_limiter_impl.h)
+for [request-release timing](terminology.md#request-release-timing). When it
+is time to release a request, the
+[BenchmarkClient](../../source/client/benchmark_client_impl.h)
+is requested to do so by the **Sequencer**.
```
dubious90, line 51: preexisting - passive tense is considered harder to follow for documentation.  (Also we are double-using the word request here).  Can we instead say:

"At appropriate intervals, the Sequencer requests the BenchmarkClient to release requests."

I am also not sure what is meant by the word "release" here.  Does that mean send?

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+
+**BenchMarkClient** retrieves data for the next request from the configured
+[RequestSource](../../source/common/request_source_impl.h)
```
dubious90, line 54: if one is provided?

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
+and asks its underlying **Connection pool** to create a
+[StreamDecoder](../../source/client/stream_decoder.h)
+which releases the request. The **StreamDecoder** sends the request and emits
+events as it progresses (connection pool ready, completion, etc...).
```
dubious90, line 58: It's still difficult to tell what the Stream Decoder is doing here.  In particular, we are saying the stream decoder "sends the request and emits evens as it progress" which begs the question - what about that is decoding?

Also what are the events, which aren't in the diagram?

docs/root/overview.md
```diff
@@ -1,60 +1,70 @@
...
-
-Once we have phases, the gRPC service, and perhaps the CLI, would be natural
-candidates to follow up with to allow dynamic phase injection, as well as send
-back reports per phase.
+## Architecture
+
+The following diagram outlines the overall architecture of Nighthawk.
+
+![Nighthawk architecture](images/nighthawk_architecture.png)
+
```
dubious90, line 8: diagram comments:
1. The output formatter should, I believe, be returning results to the CLI.
2. Feels like we are missing some arrows on the return path - Should we indicate that the client workers are sending the responses to process?  Do they actually put them directly into the output collector?
3. (optional) - Especially if we are going to include return process arrows, can we possibly number or label the arrows?

docs/root/overview.md
mum4k: As discussed offline, RequestSource isn't optional, there just happen to be multiple implementations of it. I am assuming you were talking about the gRPC or plugin based request sources, but even without them there is a built-in request source that generates requests. Note that all the available request sources are outlined later on this page.

docs/root/overview.md
mum4k: Done in 7ba9de7ba336819db5977c8876e69084abe35ea1.

docs/root/overview.md
mum4k: Done in 7ba9de7ba336819db5977c8876e69084abe35ea1.

docs/root/overview.md
mum4k: Done in 7ba9de7ba336819db5977c8876e69084abe35ea1.

Also changed "release" to "send". Release just happens to be the terminology the Nighthawk codebase uses.

docs/root/overview.md
mum4k: Moved the examples of events (the parenthesis) right after the "event" word. Also clarified the role of StreamDecoder.

Done in 7ba9de7ba336819db5977c8876e69084abe35ea1.

docs/root/overview.md
mum4k: 1. added an arrow going back to the user. While technically that goes back to the CLI (or the gRPC), adding it there seems to complicate the diagram. The return arrow crosses the API boundary, so this should be clear enough.
2. added a return arrow. Technically the workers write the stats directly, but this diagram needs a bit of abstraction to remain readable.
3. labeled most of the arrows.

Done in e4079385796871b10bbb9e7b6a0c5dee433af0e8.

docs/root/overview.md
mum4k: As discussed offline, request source isn't optional. If we aren't using the plugin or the gRPC request source, we still need an instance of request source to source the requests, in which case Nighthawk uses the built-in one.

Added a user icon to the diagram.

Done in e4079385796871b10bbb9e7b6a0c5dee433af0e8.

mum4k:
@dubious90 PTAL.

docs/root/overview.md
dubious90: I leave it up to you how to handle this, but I will note the output formatter doesn't do anything if we're using the grpc api.  The output is always a proto there.  That's one of the reasons I clarified that they go to the CLI.  However, if you think this is a minor difference or clear enough, I defer to you.

dubious90 (approved):

mum4k:
Fair point about the output formatter being related just to the CLI. I think we would straggle to highlight this in the architecture diagram, this is something that may better fit into the flow diagram once we get to it. That is where we can better highlight the different execution modes and their flow.

So assuming you agree, I will choose not to address it here.

PR #778, opened 2021-12-03
Avoid compare Protobuf TextFormat with Protobuf DebugString

zhaomoy:
The Protobuf DebugString format is an unstable format and it may be different from the Protobuf TextFormat at some point in the future. To make this unit test more robust,  
we deserialize the text proto (in TextFormat) from the golden file first, and then compare the DebugString of that deserialized proto with the actual value (a DebugString).

### DIFF
--- a/test/adaptive_load/adaptive_load_client_main_test.cc
+++ b/test/adaptive_load/adaptive_load_client_main_test.cc
@@ -4,6 +4,7 @@
 #include "nighthawk/adaptive_load/adaptive_load_controller.h"
 #include "nighthawk/common/exception.h"
 
+#include "third_party/envoy/src/source/common/protobuf/protobuf.h"
 #include "external/envoy/test/mocks/filesystem/mocks.h"
 #include "external/envoy/test/test_common/file_system_for_test.h"
 #include "external/envoy/test/test_common/utility.h"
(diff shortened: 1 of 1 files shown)


test/adaptive_load/adaptive_load_client_main_test.cc
```diff
@@ -312,10 +313,12 @@ TEST(AdaptiveLoadClientMainTest, WritesOutputProtoToFile) {
...
   main.Run();
 
-  std::string golden_output =
+  std::string golden_text =
       Envoy::Filesystem::fileSystemForTest().fileReadToEnd(Nighthawk::TestEnvironment::runfilesPath(
           std::string("test/adaptive_load/test_data/golden_output.textproto")));
-  EXPECT_EQ(actual_outfile_contents, golden_output);
+  nighthawk::adaptive_load::AdaptiveLoadSessionOutput golden_proto;
+  Envoy::Protobuf::TextFormat::ParseFromString(golden_text, &golden_proto);
+  EXPECT_EQ(actual_outfile_contents, golden_proto.DebugString());
```
mum4k, line 321: I wonder, since we do have an actual proto matcher now, could we compare these two in their binary form instead?

https://github.com/envoyproxy/nighthawk/blob/main/test/test_common/proto_matchers.h

test/adaptive_load/adaptive_load_client_main_test.cc
```diff
@@ -4,6 +4,7 @@
 #include "nighthawk/adaptive_load/adaptive_load_controller.h"
 #include "nighthawk/common/exception.h"
 
+#include "external/envoy/src/source/common/protobuf/protobuf.h"
```
mum4k, line 7: This seems to be causing a test failure, looks like we may need to add this as a dep in the BUILD file too.

However, please see mu other comment first.

mum4k (changes requested):
Thank you for contributing the improvement @zhaomoy, leaving a few comments below.

test/adaptive_load/adaptive_load_client_main_test.cc
zhaomoy: The actual value is still a DebugString. Even though current you can deserialize a DebugString just like a textual proto, in the near future this will stop working. The idea here is that we store the textual proto (TextFormat) of the expected proto in the golden file, and create a DebugString of that at run-time. This way no matter how the DebugString format changes (there will be multiple rounds of changes), this test still works.

test/adaptive_load/adaptive_load_client_main_test.cc
zhaomoy: I got a wrong path. Now it should be fixed.

test/adaptive_load/adaptive_load_client_main_test.cc
mum4k: Ah I misunderstood what was going on here, thank you for the explanation.

mum4k (approved):

PR #779, opened 2021-12-04
A collection of fixes that simplify import of Nighthawk into Google.

mum4k:
Done here:

- removing comments next to the licenses in BUILD files, since that triggers a style warning.
- exporting the main LICENSES file.
- fixing bad include prefix and import of Nighthawk's distributor API.
- removing implicit conversions of strings to booleans.

PR #785, opened 2021-12-07
A collection of fixes that simplify import of Nighthawk into Google.

mum4k:
Done here:

- exporting the main LICENSES file.
- not referring to Nighthawk's code as if it was external.
- fixing bad include prefix and import of Nighthawk's distributor API.
- removing implicit conversions of strings to booleans.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -6,6 +6,8 @@ load(
 
 licenses(["notice"])  # Apache 2
 
+exports_files(["LICENSE"])
+
 envoy_package()
 
 filegroup(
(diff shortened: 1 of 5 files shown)


mum4k:
@tomjzzhang sorry had to close #779 since I damaged my branch. Here is a clone, but with less changes. We are no longer removing the comments next to the license statements.

PTAL

tomjzzhang (approved):
LGTM

dubious90 (approved):

PR #5, opened 2019-04-12
Complete transfer from `envoy-perf/nighthawk`

oschaaf:
This PR completes the transfer of `envoy-perf/nighthawk` this repo, by fixing the build, tests, and CI.

Needs https://github.com/envoyproxy/nighthawk/pull/4 to go in first, this adds three distinct steps (commits):
- Move over things we need from `envoy-perf` that live outside of `nighthawk/`
- Adds the Envoy submodule
- Amendments to compensate the source-code moving from `nighthawk/` to `/`

PR #792, opened 2021-12-16
Delete unused file gen_compilation_database.sh

dubious90:


mum4k (approved):

dubious90:
Hey @oschaaf sorry to bother.  Can you take a quick look at this?  As far as I can tell, this file isn't used, and was a little confusing when I was trying to diagnose the ci/do_ci.sh clang-tidy stuff.  Does that seem right to you? The fact that the pipeline fully passed without it seems to be confirmation to me, but just wanted to run this by you quick first.

oschaaf (approved):
My memory fails me, but I am pretty sure that `bazel/gen_compilation_database.sh` is not involved in any critical path. 

When the file was introduced in https://github.com/envoyproxy/nighthawk/pull/5/files#diff-2e4ade9d4be99e055c448e90d99f74786db3f8f2789138c30285333503fd9a45R13 it actually was used in ` tools/gen_compilation_database.py`.  That no longer seems to be the case.

PR #805, opened 2022-02-11
Support specifying a custom upstream bind config.

mum4k:
Allows Nighthawk users to specify a custom upstream bind config, i.e. to indicate the source address, port and socket options to be used when sending requests.

### DIFF
--- a/source/client/options_impl.cc
+++ b/source/client/options_impl.cc
@@ -155,6 +155,15 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "{common_tls_context:{tls_params:{cipher_suites:[\"-ALL:ECDHE-RSA-AES128-SHA\"]}}}",
       false, "", "string", cmd);
 
+  TCLAP::ValueArg<std::string> upstream_bind_config(
+      "", "upstream-bind-config",
+      "BindConfig in json or compact yaml. If specified, this configuration is used to bind newly "
+      "established upstream connections. "
+      "Allows selecting the source address, port and socket options used when sending requests. "
+      "Example (json): "
+      "{source_address:{address:\"127.0.0.1\",port_value:0}}",
+      false, "", "string", cmd);
+
   TCLAP::ValueArg<std::string> transport_socket(
       "", "transport-socket",
       "Transport socket configuration in json or compact yaml. "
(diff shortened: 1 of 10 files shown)


mum4k:
@tomjzzhang please review and assign to @dubious90 once done.

test/integration/test_integration_basics.py
```diff
@@ -334,6 +334,39 @@ def test_h3_quic(quic_test_server_fixture):
...
+      "1",
+      # Envoy doesn't support disabling certificate verification on Quic
+      # connections, so the host in our requests has to match the hostname in
+      # the leaf certificate.
+      "--request-header",
+      "Host:www.lyft.com",
+      "--upstream-bind-config %s" % upstream_bind_config
+  ])
+  counters = quic_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
+  asserts.assertCounterEqual(counters, "benchmark.http_2xx", 25)
```
tomjzzhang, line 366: Unsure if we are having stability issues regarding integration tests but I think asserting that we received at least single 200 is good enough if we are.

tomjzzhang (approved):
LGTM.

dubious90 (dismissed):

source/client/options_impl.cc
```diff
@@ -155,6 +155,15 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "{common_tls_context:{tls_params:{cipher_suites:[\"-ALL:ECDHE-RSA-AES128-SHA\"]}}}",
       false, "", "string", cmd);
 
+  TCLAP::ValueArg<std::string> upstream_bind_config(
+      "", "upstream-bind-config",
+      "BindConfig in json or compact yaml. If specified, this configuration is used to bind newly "
```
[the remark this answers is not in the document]
mum4k: That is a good catch. I found no proof that any of the parameters that mention it do support YAML and I found no unit tests with YAML in them. I am assuming this was added at some point, but not implemented.

Removing the mention of YAML from here and the other locations. That way we only mention what we have test coverage for.

Done in 94785cf27365a3c9daf52d22d30c0599656f2905.

test/integration/test_integration_basics.py
mum4k: This seems to be stable as is, probably because of the termination predicate.

Agreed that the suggested would be a good fix if this started being flaky.

mum4k:
@dubious90 PTAL.

dubious90 (approved):

Issue #808, opened 2022-02-17
crash when run a simple bench with concurrency 2 for nightawk server


wbpcode:
*Title*: *crash when run a simple bench with concurrency 2 for nightawk server*

*Description*:

I created a simple benchmark with nightawk to evaluate perf of Envoy. And I and If I run the nightawk server with concurrency 2 and run the tests, the nightawk server will crash.

*Reproduction steps*:

This problem can be simple reproduced by the following commands.

```
git clone https://github.com/wbpcode/envoy-simple-bench.git
cd envoy-simple-bench
git checkout server-crash
docker-compose up --abort-on-container-exit
```

Ref: https://github.com/wbpcode/envoy-simple-bench/tree/server-crash

*Call Stack*:

```
server_1  | [2022-02-17 03:33:25.432][13][critical][backtrace] [external/envoy/source/server/backtrace.h:104] Caught Segmentation fault, suspect faulting address 0x0
server_1  | [2022-02-17 03:33:25.432][13][critical][backtrace] [external/envoy/source/server/backtrace.h:91] Backtrace (use tools/stack_decode.py to get line numbers):
server_1  | [2022-02-17 03:33:25.432][13][critical][backtrace] [external/envoy/source/server/backtrace.h:92] Envoy version: 0/1.21.0-dev//RELEASE/BoringSSL
server_1  | [2022-02-17 03:33:25.445][13][critical][backtrace] [external/envoy/source/server/backtrace.h:98] #0: [0x7f19d8f36090]
server_1  | [2022-02-17 03:33:25.497][13][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #1: Envoy::Http::FilterManager::decodeData() [0x558cd500267e]
server_1  | [2022-02-17 03:33:25.500][13][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #2: Envoy::Http::ConnectionManagerImpl::ActiveStream::decodeData() [0x558cd4ff036b]
server_1  | [2022-02-17 03:33:25.504][13][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #3: Envoy::Http::Http1::ServerConnectionImpl::onMessageCompleteBase() [0x558cd503c0cc]
server_1  | [2022-02-17 03:33:25.507][13][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #4: Envoy::Http::Http1::ConnectionImpl::onMessageComplete() [0x558cd5039476]
[...]

PR #809, opened 2022-02-22
fix nighthawk crash error when run server with concurrency >1

wbpcode:
Signed-off-by: wbpcode <wbphub@live.com>

To address #808.

The filter config is shared by mutiple filter instances in different workers. In the original logic, the `computeEffectiveConfiguration` may updates the `effective_config_` field of filter config. This operations are unsafe for mutiple threads scenarios.

### DIFF
--- a/source/server/http_dynamic_delay_filter.cc
+++ b/source/server/http_dynamic_delay_filter.cc
@@ -57,7 +57,7 @@ HttpDynamicDelayDecoderFilter::decodeHeaders(Envoy::Http::RequestHeaderMap& head
 
 Envoy::Http::FilterDataStatus
 HttpDynamicDelayDecoderFilter::decodeData(Envoy::Buffer::Instance& buffer, bool end_stream) {
-  if (effective_config_.ok()) {
+  if (!effective_config_.ok()) {
     if (end_stream) {
       config_->validateOrSendError(effective_config_, *decoder_callbacks_);
       return Envoy::Http::FilterDataStatus::StopIterationNoBuffer;
--- a/source/server/http_dynamic_delay_filter.cc
+++ b/source/server/http_dynamic_delay_filter.cc
@@ -40,14 +40,14 @@ void HttpDynamicDelayDecoderFilter::onDestroy() {
 Envoy::Http::FilterHeadersStatus
 HttpDynamicDelayDecoderFilter::decodeHeaders(Envoy::Http::RequestHeaderMap& headers,
                                              bool end_stream) {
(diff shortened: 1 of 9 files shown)


mum4k (approved):
Thank you for finding this race condition and for the fix.

I noticed that the CI didn't execute tests for this PR, not sure if that is a one time issue, or some lack of permissions. The CI doesn't know about this PR at all at the monent.

Would you mind pushing another commit to try and trigger it again? If that won't help, I can try to investigate why the CI isn't running the tests.

wbpcode:
> Thank you for finding this race condition and for the fix.
> 
> I noticed that the CI didn't execute tests for this PR, not sure if that is a one time issue, or some lack of permissions. The CI doesn't know about this PR at all at the monent.
> 
> Would you mind pushing another commit to try and trigger it again? If that won't help, I can try to investigate why the CI isn't running the tests.

get it. I will try to run the ci first and then push a new empty commit.

mum4k:
I have looked into it and the CI is not triggering due to lack of permissions. I will look into triggering it manually from our end.

mum4k:
Despite my attempts, I did not figure out how to fix the CI permission issues. I did open a support ticket, but I expect it will be a few days before we hear back.

In order to not block this fix, I cloned this PR into #812 and am executing the tests there.

mum4k:
@wbpcode, we got a response from CircleCI:

```
Hi,

Thank you for contacting CircleCI Support!

Sorry to hear you are running into this issue with your project's PR.

When checking our logs, I found that we were having trouble authenticating the user who opened that pr.
This happens when a user has a CircleCI account, but hasn't logged in for a while resulting in us having a stale or invalid token stored on our end.

I have gone ahead and cleared the cache for this user's account.

If possible, could you have wbpcode push another commit to the branch the pr was opened from? If this still does not trigger a build, I would recommend that they log into CircleCI again before pushing another commit.

Please let me know if you have any additional questions as well.

Best Regards
```

Would you be willing to give this a try? I.e. could you push another commit now and see if that triggers and potentially re-logging into CircleCI?

wbpcode:
> 

Thanks, i will have a try.

wbpcode:
sad, the gcc test always timeout.

mum4k:
@wbpcode thank you for retrying the CI, I am glad we managed to fix the problem with the trigger.

The test_gcc timeout is an issue we are tracking and are planning to address. You can leave this PR with me, I will work on merging it in.

mum4k (approved):

PR #813, opened 2022-02-24
Simplify locking related to the start of a benchmark.

mum4k:
Simplifies the locking code to make it easier to reason about. This is intended to help when troubleshooting an error where a new benchmark is rejected with `Only a single benchmark session is allowed at a time.`, while the previous benchmark already finished and the client has closed the gRPC stream. I suspect this was related to the use of `TryLock` method on the Mutex.

Done here:
- using a single lock to protect a single boolean `benchmark_in_progress_`.
- simplifying the main method with a few early checks and returns (avoids nested if statements).
- the `benchmark_in_progress_` value now determines if we are busy.
- the `benchmark_in_progress_` value is set to true by the thread that first manages the acquire the lock and the same thread then starts the benchmark using the `future`.
- the `future` sets the the `benchmark_in_progress_` value to false just before it writes the response.

### DIFF
--- a/source/client/service_impl.cc
+++ b/source/client/service_impl.cc
@@ -13,16 +13,6 @@ namespace Nighthawk {
 namespace Client {
 
 void ServiceImpl::handleExecutionRequest(const nighthawk::client::ExecutionRequest& request) {
-  std::unique_ptr<Envoy::Thread::LockGuard> busy_lock;
-  {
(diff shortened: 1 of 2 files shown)


mum4k:
Still work in progress, some unit tests are failing.

mum4k:
This needs a bit more thought, going to send another PR.

Issue #820, opened 2022-03-17
Migrate Nighthawk's CI to machines with more memory


mum4k:
The current allocation only gives us 16GB of memory. `test_gcc` takes a very long time to finish and the integration tests can no longer execute.

We should migrate to machines with more memory.

mum4k:
Nighthawk's CI now executes on AZP.

PR #821, opened 2022-03-17
Defining our first Azure Pipeline.

mum4k:
This is first of a few PRs that will migrate Nighthawk's CI to Azure Pipelines. This is what Envoy uses currently and this is where we can get more resources which will enable the integration tests to pass again.

This defines a single step `build` that will execute in our `nighthawk-presubmit` pipeline.

Works on #820.

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -1,0 +1,23 @@
+trigger:
+  branches:
+    include:
+    - "main"
+
+stages:
+- stage: check
+  dependsOn: []
+  jobs:
+  - job: bazel
+    displayName: "do_ci.sh"
+    dependsOn: []
+    strategy:
+      maxParallel: 1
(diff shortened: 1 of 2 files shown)


.azure-pipelines/pipelines.yml
```diff
@@ -0,0 +1,24 @@
+# See https://docs.microsoft.com/en-us/azure/devops/pipelines/customize-pipeline?view=azure-devops.
+trigger:
+  branches:
+    include:
+    - "main"
+
+- stage: check
```
lizan, line 7: From azure pipeline: 
Encountered error(s) while parsing pipeline YAML:
/.azure-pipelines/pipelines.yml: (Line: 7, Col: 1, Idx: 148) - (Line: 7, Col: 2, Idx: 149): While parsing a block mapping, did not find expected key.

PR #828, opened 2022-04-12
Update envoy to 90fb3c8 (2022-04-11)

dubious90:
- Updated .bazelversion
- Reacted to changes in https://github.com/envoyproxy/envoy/pull/20397. Proto extensions are now looked up by @type rather than name in envoy.  Getting onto this standard is too much work for this nighthawk update, so instead:
  - Added a runtime override for all integration test yaml files
  - Added a runtime override for fake stats sink in the proccess_test unit test


mum4k (approved):

mum4k:
Verified integration_test_coverage passes locally.

mum4k:
Running test_gcc locally as well. Will post results in the morning.

PR #837, opened 2022-04-21
Deprecate CircleCI pipeline and configuration.

mum4k:
Works on #820.

- Deleting CircleCI config.
- Removing instructions related to CircleCI from MAINTAINERS.md.
- Removing reference to CircleCi in benchmarks/README.md.
- Folding previously forked docker scripts.
- Removing repokitteh retest support for CircleCI. Going to send separate PR that will add this functionality for AZP.
- Removing utility that detects CircleCi in integration tests.
- Removing references to CircleCI from ci/do_ci.sh.

Signed-off-by: Jakub Sobon <mumak@google.com>

PR #840, opened 2022-04-22
Fix the docker stage by renaming s/docker_azp/docker.

mum4k:
The argument was renamed in #837, but I forgot to rename it in the pipeline configuration.

Works on #820.


mum4k:

PR #847, opened 2022-05-10
Envoy Update to d3110f4 (May 9, 2022)

tomjzzhang:
- Update bazel/repositories.bzl
- Update .bazelrc. Uncommented platform_mappings override as it is necessary for nighthawk to build. 
- Update .bazelversion
- Update stream_decoder with updated interface (https://github.com/envoyproxy/envoy/pull/20367). Setting to nullptr is fine as an appropriate default is set if nullptr is passed. In addition, tracing is not a large concern for nighthawk.
- Changes due to (https://github.com/envoyproxy/envoy/pull/20862). Due to cipher deprecation, there is only 1 RSA and DSA cipher default left. As a result, to preserve testing coverage, the appropriate certs need to be loaded to the nighthawk instances to test multiple ciphers. As a result, created a new nighthawk configuration file and refactored integration tests. 
- Updated update process documentation to create PR as last step.
- Temporarily disable the `clang-tidy` CI step until #849 gets resolved.


.bazelrc
```diff
@@ -27,7 +27,7 @@ build --incompatible_strict_action_env
 build --host_force_python=PY3
 build --java_runtime_version=remotejdk_11
 build --tool_java_runtime_version=remotejdk_11
-#build --platform_mappings=""                                                               # unique
+build --platform_mappings=""                                                               # unique
```
mum4k, line 30: This line was previously marked as "# unique" and commented out on purpose, since we don't need platform mappings in Nighthawk.

Do we need to uncomment it?

test/integration/configurations/nighthawk_https_origin_dsa.yaml
```diff
@@ -0,0 +1,58 @@
...
+                  - name: envoy.filters.http.router
+                    typed_config:
+                      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+                      dynamic_stats: false
+          transport_socket:
+            name: tls
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext
+              common_tls_context:
+                tls_certificates:
```
mum4k, line 47: (nit) Some trailing spaces here after `tls_certificates:`.

mum4k (approved):
Only two minor comments, otherwise we're good to go.

mum4k:
Going to merge in as is, we can address the minor comments in a follow-up.

.bazelrc
tomjzzhang: It appears that it works this line commented as well. Address in #851

test/integration/configurations/nighthawk_https_origin_dsa.yaml
tomjzzhang: Addressed in #851 851

PR #852, opened 2022-05-16
Update envoy to f1a3fd1 (05/16/2022) 

dubious90:
Required changes:
-


test/integration/configurations/nighthawk_http_origin.yaml
```diff
@@ -29,7 +29,19 @@ static_resources:
               - "*"
           http_filters:
           - name: time-tracking
+            typed_config:
+              "@type": type.googleapis.com/nighthawk.server.TimeTrackingConfiguration
+              response_options:
```
eric846, line 34: Just noting that these yamls have become better example code through the added fields, but technically you might be able to have just `"@type": type.googleapis.com/nighthawk.server.TimeTrackingConfiguration` with no no `response_options:` inside, if there is value in keeping the test behavior identical.

test/integration/configurations/nighthawk_http_origin.yaml
```diff
@@ -29,7 +29,19 @@ static_resources:
               - "*"
           http_filters:
           - name: time-tracking
+            typed_config:
```
eric846, line 32: Does it still work if we delete `name: time-tracking`? I heard but haven't confirmed that the type under `typed_config` type takes precedence over `name`. If that's true, we could avoid confusion by deleting `name` throughout.

eric846 (changes requested):

test/integration/configurations/nighthawk_http_origin.yaml
eric846: (Just saw your note in the description that a full migration might only be possible later.)

tomjzzhang (commented):
Should these be updated?

https://github.com/envoyproxy/nighthawk/blob/main/test/integration/nighthawk_test_server.py#L93

https://github.com/envoyproxy/nighthawk/blob/main/source/server/README.md?plain=1#L52

https://github.com/envoyproxy/nighthawk/blob/main/docs/root/examples/HEADER_BASED_LATENCY.md?plain=1#L26

oschaaf (commented):
As discussed on Slack, at a high level, I think this is probably what we must do if we can't re-instate compatibility for re-using proto configuration types across extensions.  But perhaps we can postpone the api breakage by massaging our inputs, to at least give consumers a heads up and some grace period before we enforce configuration updates.

api/server/response_options.proto
```diff
@@ -48,3 +48,11 @@ message ResponseOptions {
   // responses 3 and 2>.
   string emit_previous_request_delta_in_response_header = 6;
 }
+
+message DynamicDelayConfiguration {
+  ResponseOptions response_options = 1;
```
pamorgan, line 53: Why do we need the  response_options here? Logically I assume test-server filter is performing most of the work configured by response_option. The dynamic delay should only include the delay settings.

pamorgan: As a result oneof_delay_options should be deprecated ?

api/server/response_options.proto
```diff
@@ -48,3 +48,11 @@ message ResponseOptions {
...
+}
+
+message TimeTrackingConfiguration {
+  ResponseOptions response_options = 1;
```
pamorgan, line 57: Same comment, this should be empty.

source/server/http_dynamic_delay_filter.cc
```diff
@@ -13,9 +13,10 @@ namespace Nighthawk {
 namespace Server {
 
 HttpDynamicDelayDecoderFilterConfig::HttpDynamicDelayDecoderFilterConfig(
-    const nighthawk::server::ResponseOptions& proto_config, Envoy::Runtime::Loader& runtime,
-    const std::string& stats_prefix, Envoy::Stats::Scope& scope, Envoy::TimeSource& time_source)
-    : FilterConfigurationBase(proto_config, "dynamic-delay"), runtime_(runtime),
+    const nighthawk::server::DynamicDelayConfiguration& proto_config,
+    Envoy::Runtime::Loader& runtime, const std::string& stats_prefix, Envoy::Stats::Scope& scope,
+    Envoy::TimeSource& time_source)
+    : FilterConfigurationBase(proto_config.response_options(), "dynamic-delay"), runtime_(runtime),
```
pamorgan, line 19: What is the value of inheriting from FilterConfigurationBase. I think this is adding more complexity than needed -  this config should be or inherit from fault.proto since the the filer itself inherits from it.

source/server/http_time_tracking_filter.cc
```diff
@@ -15,8 +15,8 @@ namespace Nighthawk {
 namespace Server {
 
 HttpTimeTrackingFilterConfig::HttpTimeTrackingFilterConfig(
-    const nighthawk::server::ResponseOptions& proto_config)
-    : FilterConfigurationBase(proto_config, "time-tracking"),
+    const nighthawk::server::TimeTrackingConfiguration& proto_config)
+    : FilterConfigurationBase(proto_config.response_options(), "time-tracking"),
```
pamorgan, line 19: same comment.

api/server/response_options.proto
```diff
@@ -48,3 +48,11 @@ message ResponseOptions {
   // responses 3 and 2>.
   string emit_previous_request_delta_in_response_header = 6;
 }
+
+message DynamicDelayConfiguration {
```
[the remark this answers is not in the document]
dubious90: Done. And true

api/server/response_options.proto
dubious90: We've discussed this, and are renaming this to experimental_response_options with the idea of a refactor soon after

source/server/http_dynamic_delay_filter.cc
dubious90: Addressing in a followup refactor

source/server/http_dynamic_delay_filter.h
```diff
@@ -33,10 +33,10 @@ class HttpDynamicDelayDecoderFilterConfig : public FilterConfigurationBase {
    * @param scope Statistics scope to be used by the filter.
    * @param time_source Time source to be used by the filter.
    */
-  HttpDynamicDelayDecoderFilterConfig(const nighthawk::server::ResponseOptions& proto_config,
```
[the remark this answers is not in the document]
dubious90: Done.

test/integration/configurations/nighthawk_http_origin.yaml
dubious90: While it might work to remove the name, I'm not really sure what that would gain us.

dubious90:
@tomjzzhang @pamorgan @eric846 feedback addressed.

test/integration/configurations/nighthawk_http_origin.yaml
dubious90: Missed this on first pass, but updated the two readme files I found that referenced dynamic-delay or time-tracking

eric846 (approved):
LGTM

tomjzzhang (approved):
LGTM.

api/server/response_options.proto
```diff
@@ -48,3 +48,17 @@ message ResponseOptions {
   // responses 3 and 2>.
   string emit_previous_request_delta_in_response_header = 6;
 }
+
+// Configures the dynamic-delay test filter.
+message DynamicDelayConfiguration {
+  // This is a temporary solution to allow this functionality to continue, but will likely be
```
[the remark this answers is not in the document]
dubious90: https://github.com/envoyproxy/nighthawk/issues/854

pamorgan (approved):

PR #855, opened 2022-05-19
Refactor filter configuration to make the different test server filters independent

dubious90:
This is a followup to https://github.com/envoyproxy/nighthawk/pull/852. In that change, we duplicated ResponseOptions onto separate DynamicDelayConfiguration and TimeTrackingConfiguration protos. However, this resulted in many fields on those protos that were not relevant to each filter. In this PR, we make it so that only the relevant fields exist on each proto.

Resulting changes:
1. Moved DynamicDelayConfiguration and TimeTrackingConfiguration into their own files, and moved their fields into those protos directly.
2. In order to make request headers still work, there is now manual logic for taking in a ResponseOptions proto via json and cherrypicking the fields relevant to each filter.
3. The generic method computeEffectiveConfiguration was moved to each filter's anonymous namespace.
4. All configs, tests, and markdown files updated to use the new protos.
5. Some tests have been removed that are no longer necessary, or never provided additional coverage, including some that are less feasible now.
    - HeaderMerge was removed, but everything it tested for was moved into new unit tests of MergeJsonConfig.


mum4k:

qqustc (dismissed):

source/server/http_dynamic_delay_filter.cc
```diff
@@ -4,22 +4,85 @@
...
+// and merges them into the base_config.
+absl::Status cherryPickDynamicDelayConfiguration(absl::string_view header_json,
+                                                 DynamicDelayConfiguration& base_config) {
+  try {
+    ResponseOptions response_options;
+    auto& validation_visitor = Envoy::ProtobufMessage::getStrictValidationVisitor();
+    Envoy::MessageUtil::loadFromJson(std::string(header_json), response_options,
+                                     validation_visitor);
+    switch (response_options.oneof_delay_options_case()) {
+    case ResponseOptions::OneofDelayOptionsCase::kStaticDelay:
```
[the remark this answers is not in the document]
dubious90: Weirdly enough, it seems that this is the way our formatter wants it. I tried to indent it manually and running fix_format undid it

mum4k (changes requested):
Can we as part of this PR add a CHANGELOG.md file and document the breaking change? Specifically we should call out (with an example) what broke and how users are expected to modify their configs to make it work again.

Looks like we introduced two breaks, so we should document both:
1. A break when we moved onto the `experimental_response_options` fields.
2. A break where we removed the `experimental_response_options` fields.

dubious90:
@mum4k Believe I have addressed all your feedback now.

source/server/http_test_server_filter.h
```diff
@@ -18,7 +18,11 @@ class HttpTestServerDecoderFilterConfig : public FilterConfigurationBase {
 public:
   HttpTestServerDecoderFilterConfig(const nighthawk::server::ResponseOptions& proto_config);
 
-  std::shared_ptr<const nighthawk::server::ResponseOptions> getServerConfig();
+  /**
+   * @return std::shared_ptr<const nighthawk::server::TimeTrackingConfiguration> the base
```
mum4k, line 22: It may not be clear to the reader what "base configuration" is. Can we clarify that this is the startup configuration as read from the bootstrap which then later gets updated via in-flight headers?

CHANGELOG.md
```diff
@@ -0,0 +1,99 @@
...
+              response_body_size: 10
+              v3_response_headers:
+              - { header: { key: "x-nh", value: "1"}}
+```
+
+This change does NOT affect how headers update the base configurations.
+
+## 0.4.0
+
+Due to [upstream envoy change](https://github.com/envoyproxy/nighthawk/commit/4919c54202329adc3875eb1bce074af33d54e26d), we modified the dynamic-delay and time-tracking filters' configuration protos to have their own configuration protos wrapping nighthawk.server.ResponseOptions.
```
mum4k, line 57: Can we wrap lines to 80 or 100 chars for make reviews and change history easier to read?

mum4k, line 57: The comment about adding some links applies here as well.

CHANGELOG.md
```diff
@@ -0,0 +1,99 @@
+# Changelog
```
mum4k, line 1: Do we need to change anything in this doc?

https://github.com/envoyproxy/nighthawk/blob/main/source/server/README.md#response-options-config

Could you also check if there are any other README docs or similar that may need updates due to these changes?

CHANGELOG.md
```diff
@@ -0,0 +1,99 @@
...
+              - { header: { key: "x-nh", value: "1"}}
+```
+
+This change does NOT affect how headers update the base configurations.
+
+## 0.4.0
+
+Due to [upstream envoy change](https://github.com/envoyproxy/nighthawk/commit/4919c54202329adc3875eb1bce074af33d54e26d), we modified the dynamic-delay and time-tracking filters' configuration protos to have their own configuration protos wrapping nighthawk.server.ResponseOptions.
+
+For base yaml configuration files, if you previously had:
```
mum4k, line 59: Can we add some more context here? I.e. we are talking about a yaml file that holds the static bootstrap configuration for the test server, with a link to here:

https://github.com/envoyproxy/nighthawk/blob/main/source/server/README.md#nighthawk-test-server

CHANGELOG.md
```diff
@@ -0,0 +1,99 @@
...
+            typed_config:
+              "@type": type.googleapis.com/nighthawk.server.ResponseOptions
+              response_body_size: 10
+              static_delay: 1.33s
+              emit_previous_request_delta_in_response_header: x-origin-request-receipt-delta
+              v3_response_headers:
+              - { header: { key: "x-nh", value: "1"}}
+```
+
+You should now explicitly specify the types and values of the protos as such:
```
mum4k, line 75: This isn't the only change in the diff though. Looks like we also have to wrap the parameters in the new `experimental_response_options` field. Can we call it out in the text here?

CHANGELOG.md
```diff
@@ -0,0 +1,99 @@
+# Changelog
+
+All breaking changes to this project will be documented in this file, most recent changes at the top.
+
+## 0.5.0
+
+In an effort to clean up the previous change and improve it, we re-modified the dynamic-delay and time-tracking filters to extricate their configuration entirely from nighthawk.server.ResponseOptions.
```
mum4k, line 7: Can we enhance this paragraph with some permalinks to the proto definitions?

Ideally we will have:
- a link to the own configuration protos the dynamic-delay and time-tracking filters now have.
- convert nighthawk.server.ResponseOptions to a permalink.

CHANGELOG.md
dubious90: Done. There don't appear to be other docs to change. Did a search for the phrases ResponseOptions and response_options for anything that might reference the configuration, and found only this and a couple other files I had already fixed.

CHANGELOG.md
dubious90: Done. Sorry, thought the fix_format script would do this for me.

CHANGELOG.md
dubious90: Done.

CHANGELOG.md
dubious90: Done.

source/server/http_test_server_filter.h
dubious90: Done. Language changed across all three filters

CHANGELOG.md
dubious90: Created links for all of the proto configurations

CHANGELOG.md
dubious90: Done.

dubious90 (commented):
Addressed feedback.

mum4k (approved):
Thank you.

PR #863, opened 2022-06-21
Update test-server.yaml on README.md

martinezlucas98:
The current test-server.yaml returns the following error "Protobuf message (type envoy.config.bootstrap.v3.Bootstrap reason INVALID_ARGUMENT:static_delay: Cannot find field.) has unknown fields".

By adding "experimental_response_options:" this error is fixed.

This fix is already in multiple docs, but missing on this particular README.

Signed-off-by: Lucas Martinez <57833320+martinezlucas98@users.noreply.github.com>

### DIFF
--- a/source/server/README.md
+++ b/source/server/README.md
@@ -50,7 +50,8 @@ static_resources:
                   - name: dynamic-delay
                     typed_config:
                       "@type": type.googleapis.com/nighthawk.server.DynamicDelayConfiguration
-                      static_delay: 0.5s
+                      experimental_response_options:
+                        static_delay: 0.5s
                   - name: test-server # before envoy.router because order matters!
                     typed_config:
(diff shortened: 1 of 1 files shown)


dubious90:
Hi @martinezlucas98, thanks for helping us with this. The changes were made hastily to adapt to changes in Envoy, and you're correct we missed this.  However, experimental_response_options was a temporary solution that no longer exists in this codebase. Is it possible you just need to update to the latest version of Nighthawk?  Rather than a documentation change, if we're still seeing experimental_response_options required somewhere, that would be a significant bug.

You can see the https://github.com/envoyproxy/nighthawk/blob/main/CHANGELOG.md for more info on these changes.

mum4k:
Closing this as per the comment from @dubious90. However please feel free to reopen and reach out if the problem persists at the latest commit.

PR #864, opened 2022-06-21
Add example for output format transformation

martinezlucas98:
The way to use the "output transformation command" is not very intuitive. 
I believe this example will facilitate the understanding of its use for future interested developers.

Signed-off-by: Lucas Martinez <lucasoctaviom@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,15 @@ Displays usage information and exits.
 L7 (HTTP/HTTPS/HTTP2) performance characterization transformation tool.
 
 ```
+
+**Example:** transform json output to fortio compatible format
+
+> Notice that the default output format for `nighthawk_client` is "human", therefore to produce a json output you must run `nighthawk_client` with `--output-format json`. This json output is the one that can be transformed to the different formats as shown in the example below.
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -486,6 +486,13 @@ Displays usage information and exits.
 L7 (HTTP/HTTPS/HTTP2) performance characterization transformation tool.
 
 ```
+
+**Example:** transform json output to fortio compatible format
```
dubious90, line 490: This is a good improvement to the documentation. Thanks!

Could we add a line saying how we got `/your/json/output/file.json`?  I'm assuming it was produced by `nighthawk_client` with `--output-format json`?

dubious90:
As another note, I think that to pass the presubmit checks, you just need to run:
`tools/update_cli_readme_documentation.sh --mode fix`

README.md
martinezlucas98: Good idea, I'll add it!

> I'm assuming it was produced by `nighthawk_client` with `--output-format json`

Exactly as you said

dubious90 (approved):

PR #875, opened 2022-07-15
Don't verify the precise output of error messages from `json_util`.

mum4k:
This change is intended to remove test assertions that verify the precise output of error messages from Protobuf's `json_util.h` library. This will unblock potential migration to another JSON parser.


mum4k:
@dubious90 please review and merge if it looks good.

dubious90 (approved):

PR #877, opened 2022-07-16
Fix construction order of `ServiceImpl` and `RequestSourceServiceImpl`.

### LINKED ISSUE #872: do_ci.sh test_gcc CI target fails
*Title*: *do_ci.sh test_gcc CI target fails*

*Description*:
>When updating to [the latest envoy](https://github.com/envoyproxy/envoy/commit/0f0f675c22de944889298e0787e107265367b367), the `test_gcc` CI target fails with the following error.

```
E         [02:08:19.944607][11620][I] Nighthawk grpc service exits
E         pure virtual method called
E         terminate called without an active exception
```

https://dev.azure.com/cncf/envoy/_build/results?buildId=112587&view=logs&j=6d188cbb-0fce-5ab7-d058-3dafa1306ac4&t=3254ec1c-a4e0-51e2-2720-602bcbfef0a6

To not block the import, we have disabled these tests in https://github.com/envoyproxy/nighthawk/pull/869. We should reenable these tests ASAP.

*Reproduction steps*:
> Run do_ci.sh test_gcc in CI. Local test passes.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Logs*:
>Include the Nighthawk logs.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Call Stack*:
> If the Envoy binary is crashing, a call stack is **required**.
Please refer to the [Bazel Stack trace documentation](https://github.com/envoyproxy/envoy/tree/master/bazel#stack-trace-symbol-resolution).

mum4k:
Also re-enable `text_gcc`.

Fixes #872.

**Details**:
- `ServiceImpl` creates `Envoy::Logger::Context` which requires `Envoy::Thread::MutexBasicLockable`.
- The `Envoy::Thread::MutexBasicLockable` was constructed after the `Envoy::Logger::Context`, so `Envoy::Logger::Context` became invalid during destruction of `ServiceImpl`.
- this PR ensures that `Envoy::Thread::MutexBasicLockable` lives longer than `Envoy::Logger::Context`.
- Fixing the same problem in the construction of `RequestSourceServiceImpl`.

**Reason for `test_gcc` failure:**
- https://github.com/envoyproxy/envoy/pull/21884 triggered this bug, because it added a logging call initiated indirectly from the destructor of `Envoy::ProcessWide`.
- This logging call happened during the destruction of `ServiceImpl` when the `Envoy::Thread::MutexBasicLockable` instance was already destroyed.
- As a result the PURE virtual methods on its parent object `Envoy::Thread::BasicLockable` were called.
- `test_gcc` reported it because it places a special function meant to detect this behavior in the `vtbl` of unimplemented pure virtual methods.

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -61,6 +61,23 @@ stages:
       parameters:
         ciTarget: $(CI_TARGET)
 
+- stage: test_gcc
+  dependsOn: ["check"]
+  pool: "x64-large"
+  jobs:
+  - job: test_gcc
+    displayName: "do_ci.sh"
+    strategy:
+      maxParallel: 1
+      matrix:
+        test_gcc:
+          CI_TARGET: "test_gcc"
+    timeoutInMinutes: 120
+    steps:
+    - template: bazel.yml
+      parameters:
+        ciTarget: $(CI_TARGET)
+
 - stage: sanitizers
   dependsOn: ["test"]
   pool: "x64-large"
@@ -102,6 +119,7 @@ stages:
 - stage: release
   dependsOn:
   - "clang_tidy"
+  - "test_gcc"
   - "sanitizers"
   - "coverage"
   condition: eq(variables['PostSubmit'], true)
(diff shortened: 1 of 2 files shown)


mum4k:

oschaaf (approved):

Issue #819, opened 2022-03-15
clean up Python dependencies


eric846:
- An error `TypeError: required field "lineno" missing from alias` appeared in #817 
  - In these tests:
    - `//test/integration/unit_tests:test_nighthawk_test_server`
    - `//test/integration/unit_tests:test_utility`
    - `//test:python_test`
  - Worked around by setting `--assert=plain` on the pytest command line
- Attempting to change `pytest==6.0.1` to `pytest>=6.0.1,<7` in requirements.txt led to `ERROR: REDACTED/nighthawk/test/integration/BUILD:32:11: no such package '@python_pip_deps//pypi__more_itertools': BUILD file not found in directory 'pypi__more_itertools' of external repository @python_pip_deps. Add a BUILD file to a directory to mark it as a package. and referenced by '//test/integration:integration_test_base_lean'`
- Some other errors and warnings: https://gist.github.com/eric846/51e6db632eae0760961681a65cb06eca

mum4k:
Once this is fixed we should also undo the workaround (the addition of `--assert=plain` pytest argument) added in https://github.com/envoyproxy/nighthawk/pull/817.

PR #883, opened 2022-08-01
Update python dependencies to their latest version.

dubious90:
Also, set it up so that we take in non-major version updates.

Updating pytest and request required adding in new dependencies.

This address some of the problems in https://github.com/envoyproxy/nighthawk/issues/819.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ requests==2.24.0
 pytest==6.0.1
 pytest-dependency==0.5.1
 pytest-xdist==1.34.0
-apipkg==1.5
-pyyaml==5.4
-zipp==3.1.0
-importlib_metadata==1.7.0
+apipkg>=3.0.1, <4.0.0
+pyyaml>=6.0.0, <7.0.0
+zipp>=3.8.1, <4.0.0
+importlib_metadata>=4.12.0, <5.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,11 @@
-requests==2.24.0
-pytest==6.0.1
-pytest-dependency==0.5.1
-pytest-xdist==1.34.0
+requests>=2.28.1, <3.0.0
+pytest>=7.1.2, <8.0.0
+pytest-dependency>=0.5.1, <0.6.0 # versions below 1.0.0 are unstable, so we should only take in bugfixes automatically
+pytest-xdist>=2.5.0, <3.0.0
 apipkg>=3.0.1, <4.0.0
 pyyaml>=6.0.0, <7.0.0
 zipp>=3.8.1, <4.0.0
 importlib_metadata>=4.12.0, <5.0.0
+more_itertools>=8.13.0, <9.0.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,8 +1,11 @@
-requests==2.24.0
-pytest==6.0.1
-pytest-dependency==0.5.1
-pytest-xdist==1.34.0
-apipkg==1.5
-pyyaml==5.4
-zipp==3.1.0
-importlib_metadata==1.7.0
+requests>=2.28.1, <3.0.0
```
mum4k, line 1: What would happen if we simply don't put the upper limit and take in all updates automatically? Considering  how short the list is here, wouldn't that be a better approach?

Or to spin it the other way - our goal is to detect breakages early, so that they can be addressed in the regular update cycle (or at least noticed and the work scheduled). If we put the upper limit, we will be blind to breaking changes in future major versions. What is our plan to ensure that gap won't be too large and we get notified?

mum4k (changes requested):

requirements.txt
mum4k: Found the documented procedure in the other PR.

However it might still be good for us to discuss here which approach we are choosing and why.

dubious90: I think it's safer for us to include the upper bounds, which are currently set to just major release updates. While this prevents us from bring in updates, it also protects our users from being the first ones to experience this.  If we have our own process, which we can iterate on (e.g. frequency) in the other PR, then we can bring in these changes at our own timing without negatively impacting non-maintainer contributors to nighthawk.

However, I'm not sure what you mean by `blind to breaking changes in future major versions`. Could you clarify what kind of scenarios you're worried about? I'm just not sure I'm answering your concern.

mum4k: By `blind to breaking changes in future major versions` I meant we won't know whether there are any breaking changes in the new versions until we do try to run the update process.

You did answer my question, thank you and I agree with your reasoning. Let's just make sure the update process is executed regularly at some frequency. We can address this in #884.

mum4k (approved):

mum4k:
@tomjzzhang please review and assign back to me once done.

requirements.txt
```diff
@@ -1,8 +1,11 @@
...
+pytest>=7.1.2, <8.0.0
+pytest-dependency>=0.5.1, <0.6.0 # versions below 1.0.0 are unstable, so we should only take in bugfixes automatically
+pytest-xdist>=2.5.0, <3.0.0
+apipkg>=3.0.1, <4.0.0
+pyyaml>=6.0.0, <7.0.0
+zipp>=3.8.1, <4.0.0
+importlib_metadata>=4.12.0, <5.0.0
+more_itertools>=8.13.0, <9.0.0
+chardet>=5.0.0, <6.0.0
+six>=1.16.0, <2.0.0
```
tomjzzhang, line 11: nit: Is it possible to put this in alphabetical order?

tomjzzhang (approved):
LGTM.

PR #884, opened 2022-08-01
Create documentation for finding major dependency updates

dubious90:
Documentation includes:
1. Added step to check for python dependencies once a month when updating envoy.
2. Included documentation for how to find python dependency updates.
3. Included a section on identifying a class of errors as needing a new dependency, which was part of what made #819 more confusing.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -73,6 +73,9 @@ important maintenance task. When performing the update, follow this procedure:
    [Envoy's version](https://github.com/envoyproxy/envoy/blob/main/tools/gen_compilation_database.py) to
    update our build configurations. Be sure to retain our local modifications,
    all lines that are unique to Nighthawk are marked with comment `# unique`.
+1. If [requirements.txt](requirements.txt) has not been updated in the last month (based on comment at top
+   of file), check for major dependency updates. See
+   [Finding python dependencies](#finding-python-dependencies) below for instructions.
 1. Run `ci/do_ci.sh test`. Sometimes the dependency update comes with changes
    that break our build. Include any changes required to Nighthawk to fix that
    in the same PR.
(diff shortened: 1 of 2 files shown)


MAINTAINERS.md
```diff
@@ -84,6 +84,44 @@ important maintenance task. When performing the update, follow this procedure:
 1. Create a PR with a title like `Update Envoy to 9753819 (Jan 24th 2021)`,
    describe all performed changes in the PR's description.
 
+## Finding python dependencies
```
mum4k, line 87: Great, thank you for documenting this. How often are we going to execute this procedure? It feels a bit disconnected now, since it isn't listed form the regular update procedure, so it might not be run at all?

mum4k (changes requested):

MAINTAINERS.md
dubious90: I agree. Wanted to figure out an appropriate cadence for that.  It feels unrelated to the envoy update, so including it there didn't feel like the right place, plus I think weekly is probably overkill for this procedure.

I'm not sure if this is out of scope for the current work, but could we have an issue auto-created once a month?  Or could we have something automated run the commands I created?

mum4k: I lean towards including it in the regular update actually. How about we include a step to do this if it wasn't done for the last 4 weeks? We can include a timestamp in the file and have the person doing the regular update check how old it is.

dubious90: I did my best to include this idea here. I made it within the month instead of last 4 weeks, which I think is just visually easier, but very similar.  Please take a look

MAINTAINERS.md
```diff
@@ -84,6 +87,61 @@ important maintenance task. When performing the update, follow this procedure:
...
+   deactivate
+   rm -rf pip_update_env
+   ```
+
+### Bazel Python Error
+
+If you encounter an error that looks like:
+
+```
+RROR: REDACTED/nighthawk/test/integration/BUILD:32:11: no such package '@python_pip_deps//pypi__more_itertools':
```
tomjzzhang, line 135: nit: ERROR

tomjzzhang (dismissed):
LGTM.

MAINTAINERS.md
dubious90: Good catch, fixed.

mum4k (approved):

PR #885, opened 2022-08-02
Add new GetMetricByNameWithReportingPeriod function

### LINKED ISSUE #798: Adaptive load controller should provide metric plugins with the time of test
The current implementation of the adaptive load controller [executes](https://github.com/envoyproxy/nighthawk/blob/ba3b5d0bcc57c334f34e1c323c6bb564781caa60/source/adaptive_load/adaptive_load_controller_impl.cc#L142-L144) Nighthawk and only then [requests](https://github.com/envoyproxy/nighthawk/blob/ba3b5d0bcc57c334f34e1c323c6bb564781caa60/source/adaptive_load/adaptive_load_controller_impl.cc#L154-L156) metric values from the metric plugins.

This means that the metrics are requested well after the load stopped hitting the system under test. This works well for the [builtin metrics](https://github.com/envoyproxy/nighthawk/blob/ba3b5d0bcc57c334f34e1c323c6bb564781caa60/source/adaptive_load/metrics_plugin_impl.cc#L123-L136), since the data source for those is the result of the Nighthawk run, but doesn't work well for any metric retrieved from external systems (e.g. CPU usage of the system under test).

We should update the adaptive load infrastructure, so that the [GetMetricByName](https://github.com/envoyproxy/nighthawk/blob/ba3b5d0bcc57c334f34e1c323c6bb564781caa60/include/nighthawk/adaptive_load/metrics_plugin.h#L29) method of the metric plugins also receives information identifying the time range when the load was hitting the system under test. This could be propagated all the way from the Nighthawk output.

It is important to note that this time range should strictly be the time when Nighthawk's workers were sending the load and shouldn't include any time Nighthawk spent aggregating, formatting and serializing the results.

tomjzzhang:
Create new interface function to provide plugins with appropriate measuring period.

Provide default implementation to avoid breaking any existing clients. 

Try providing measuring period first, if it is unimplemented, fallback to old implementation. This way, external plugins can easily migrate to the new interface.

Fixes #798

### DIFF
--- a/include/nighthawk/adaptive_load/metrics_plugin.h
+++ b/include/nighthawk/adaptive_load/metrics_plugin.h
@@ -53,7 +53,7 @@ public:
    *
    * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
    * @param reporting_period the time period the Nighthawk test iteration is sending the intended
-   * load (i.e. the time period in which the metrics are of interest.)
+   * load (i.e. the time period in which the metrics are of interest).
    *
    * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
    * unavailable.
--- a/include/nighthawk/adaptive_load/metrics_plugin.h
+++ b/include/nighthawk/adaptive_load/metrics_plugin.h
@@ -27,6 +27,23 @@ public:
    * unavailable.
    */
   virtual absl::StatusOr<double> GetMetricByName(absl::string_view metric_name) PURE;
+
+  /**
+   * Obtains the numeric metric with the given name, usually by querying an outside system. Provides
+   * start_time and duration of measurement period for the plugin to utilitize.
+   *
+   * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
+   * @param start_time start_time with duration indicates when the metric is relevant.
+   * @param duration start_time with duration indicates when the metric is relevant.
+   *
+   * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
+   * unavailable.
+   */
+  virtual absl::StatusOr<double>
+  GetMetricByNameWithTime(absl::string_view metric_name,
+                          const google::protobuf::Timestamp& start_time,
+                          const google::protobuf::Duration& duration);
+
   /**
    * All metric names implemented by this plugin, for use in input validation.
    *
--- a/include/nighthawk/adaptive_load/metrics_plugin.h
+++ b/include/nighthawk/adaptive_load/metrics_plugin.h
@@ -11,11 +11,18 @@
 
 namespace Nighthawk {
 
-// Describes the period in which a Metrics plugin should measure.
-struct MeasuringPeriod {
-  // The beginning of the time period where nighthawk is measuring.
+// Describes the period of time where the Nighthawk test iteration is sending the intended load.
+// Metric Plugins should report metrics relevant to this time period.
+// For example, if a plugin is tracking the peak memory usage of a system under test. When given
+// this data, it should filter the memory usage samples to only include data points in this time
+// period and then calculate the peak usage out of those data points.
+struct ReportingPeriod {
+  // start time of the latest (current) iteration of Nighthawk test in the adaptive stage. See
+  // https://github.com/envoyproxy/nighthawk/blob/main/docs/root/adaptive_load_controller.md#the-adaptive-load-controller
+  // for more information on adaptive load testing.
   google::protobuf::Timestamp start_time;
-  // The duration of the time period nighthawk is measuring.
+
+  // The duration of the time where nighthawk is sending the intended load in the adaptive stage.
   google::protobuf::Duration duration;
 };
 
@@ -34,21 +41,26 @@ public:
    * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
    * unavailable.
    */
+  ABSL_DEPRECATED("Use GetMetricByNameWithReportingPeriod instead.")
   virtual absl::StatusOr<double> GetMetricByName(absl::string_view metric_name) PURE;
 
   /**
    * Obtains the numeric metric with the given name, usually by querying an outside system. Provides
-   * measuring_period to allow plugins to determine what time period is relevant.
+   * reporting_period to allow plugins to determine what metrics to consider report.
+   * For example, if a plugin is tracking the peak memory usage of a system under test. When given
+   * this data, it should filter the memory usage samples to only include data points in this time
+   * period and then calculate the peak usage out of those data points.
    *
    * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
-   * @param measuring_period the time period when the metric is relevant.
+   * @param reporting_period the time period the Nighthawk test iteration is sending the intended
+   * load (i.e. the time period in which the metrics are of interest.)
    *
    * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
    * unavailable.
    */
   virtual absl::StatusOr<double>
-  GetMetricByNameWithMeasuringPeriod(absl::string_view metric_name,
-                                     const MeasuringPeriod& measuring_period);
+  GetMetricByNameWithReportingPeriod(absl::string_view metric_name,
+                                     const ReportingPeriod& reporting_period);
 
   /**
    * All metric names implemented by this plugin, for use in input validation.
--- a/include/nighthawk/adaptive_load/metrics_plugin.h
+++ b/include/nighthawk/adaptive_load/metrics_plugin.h
@@ -38,7 +38,7 @@ public:
 
   /**
    * Obtains the numeric metric with the given name, usually by querying an outside system. Provides
-   * start_time and duration of measurement period for the plugin to utilitize.
+   * measuring_period to allow plugins to determine what time period is relevant.
    *
    * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
    * @param measuring_period the time period when the metric is relevant.
--- a/include/nighthawk/adaptive_load/metrics_plugin.h
+++ b/include/nighthawk/adaptive_load/metrics_plugin.h
@@ -11,6 +11,14 @@
 
 namespace Nighthawk {
 
+// Describes the period in which a Metrics plugin should measure.
+struct MeasuringPeriod {
+  // The beginning of the time period where nighthawk is measuring.
+  google::protobuf::Timestamp start_time;
+  // The duration of the time period nighthawk is measuring.
+  google::protobuf::Duration duration;
+};
+
 /**
  * An interface for plugins that retrieve platform-specific metrics from outside data sources.
  * Connection info is passed via a plugin-specific config proto.
@@ -33,16 +41,14 @@ public:
    * start_time and duration of measurement period for the plugin to utilitize.
    *
    * @param metric_name The name of the metric to retrieve. Must be supported by the plugin.
-   * @param start_time start_time with duration indicates when the metric is relevant.
-   * @param duration start_time with duration indicates when the metric is relevant.
+   * @param measuring_period the time period when the metric is relevant.
    *
    * @return StatusOr<double> The metric value, or an error status if the metric was unsupported or
    * unavailable.
    */
   virtual absl::StatusOr<double>
-  GetMetricByNameWithTime(absl::string_view metric_name,
-                          const google::protobuf::Timestamp& start_time,
-                          const google::protobuf::Duration& duration);
+  GetMetricByNameWithMeasuringPeriod(absl::string_view metric_name,
+                                     const MeasuringPeriod& measuring_period);
 
   /**
    * All metric names implemented by this plugin, for use in input validation.
--- a/source/adaptive_load/metrics_evaluator_impl.cc
+++ b/source/adaptive_load/metrics_evaluator_impl.cc
@@ -91,8 +91,6 @@ MetricsEvaluatorImpl::AnalyzeNighthawkBenchmark(
                         nighthawk_response.error_detail().message());
   }
 
-  // nighthawk_response.output.timestamp
-  // nighthawk_response.output.options.duration
   nighthawk::adaptive_load::BenchmarkResult benchmark_result;
   *benchmark_result.mutable_nighthawk_service_output() = nighthawk_response.output();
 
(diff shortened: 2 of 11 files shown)


source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -97,7 +115,8 @@ MetricsEvaluatorImpl::AnalyzeNighthawkBenchmark(
     absl::StatusOr<nighthawk::adaptive_load::MetricEvaluation> evaluation_or =
         EvaluateMetric(*spec_threshold_pair.first,
                        *name_to_plugin_map[spec_threshold_pair.first->metrics_plugin_name()],
-                       spec_threshold_pair.second);
+                       spec_threshold_pair.second, nighthawk_response.output().timestamp(),
```
tomjzzhang, line 118: I'm unsure whether this timestamp is the start time of the measuring period.

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -15,16 +15,32 @@ using ::nighthawk::adaptive_load::MetricSpec;
...
 
+// Utility function for GetMetric functionality with fallback logic.
+absl::StatusOr<double> GetMetric(MetricsPlugin& metrics_plugin, absl::string_view metric_name,
+                                 const google::protobuf::Timestamp& start_time,
+                                 const google::protobuf::Duration& duration) {
+  absl::StatusOr<double> metric_value_or =
+      metrics_plugin.GetMetricByNameWithTime(metric_name, start_time, duration);
+  // If the metric plugin does not support WithTime implementation (i.e. is using the default
+  // implementation), default to GetMetricByName
+  if (metric_value_or.status().code() == absl::StatusCode::kUnimplemented) {
```
tomjzzhang, line 26: Do we want to create a follow up task to eventually phase out the old interface?

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -15,16 +15,42 @@ using ::nighthawk::adaptive_load::MetricSpec;
...
+  }
+  // Assume the first result is a good representation for the entire run.
+  MeasuringPeriod time_interval;
+  time_interval.start_time = output.results(0).execution_start();
+  time_interval.duration = output.results(0).execution_duration();
+  return time_interval;
+}
+
+// Utility function for GetMetric functionality with fallback logic.
+absl::StatusOr<double> GetMetric(MetricsPlugin& metrics_plugin, absl::string_view metric_name,
```
dubious90, line 31: can we move this to an anonymous namespace?

dubious90 (dismissed):

include/nighthawk/adaptive_load/metrics_plugin.h
```diff
@@ -11,6 +11,14 @@
 
 namespace Nighthawk {
 
+// Describes the period in which a Metrics plugin should measure.
```
mum4k, line 14: Since this is the public API of the metric plugins, we should expand this doc comment. We need to provide enough information here, so that any developers of metric plugins understand exactly what they need to do.

The current comment just says that this is the period plugin should measure. However we aren't trying to dictate when plugin measures. I think we are trying to tell the plugin over which period should the metric be reported.

With that - maybe `ReportingPeriod` would be a better name?

Bonus points if we also provide an example of how this is used to document it even better.

include/nighthawk/adaptive_load/metrics_plugin.h
```diff
@@ -11,6 +11,14 @@
...
+struct MeasuringPeriod {
+  // The beginning of the time period where nighthawk is measuring.
```
mum4k, line 16: We should expand these comments as well. It isn't clear what "where nighthawk is measuring" means, since the measuring is done by the plugin.

We should explain that this is the start time of the latest (current) iteration of Nighthawk test in the adaptive stage (maybe link to [the docs](https://github.com/envoyproxy/nighthawk/blob/main/docs/root/adaptive_load_controller.md) for the adaptive load controller).

include/nighthawk/adaptive_load/metrics_plugin.h
```diff
@@ -11,6 +11,14 @@
...
+  google::protobuf::Timestamp start_time;
+  // The duration of the time period nighthawk is measuring.
```
mum4k, line 18: Similarly, let's expand this comment too.

include/nighthawk/adaptive_load/metrics_plugin.h
```diff
@@ -27,6 +35,21 @@ class MetricsPlugin {
    * unavailable.
    */
   virtual absl::StatusOr<double> GetMetricByName(absl::string_view metric_name) PURE;
```
mum4k, line 37: We should add a note next to `GetMetricByName` indicating that it is deprecated and why.

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -15,16 +15,42 @@ using ::nighthawk::adaptive_load::MetricSpec;
 using ::nighthawk::adaptive_load::MetricSpecWithThreshold;
 using ::nighthawk::adaptive_load::ThresholdSpec;
 
+// Extract estimated measuring period from Nighthawk output proto.
+absl::StatusOr<MeasuringPeriod> GetMeasuringPeriod(const nighthawk::client::Output& output) {
+  if (output.results().empty()) {
+    return absl::Status{absl::StatusCode::kInvalidArgument, "output.results cannot be empty."};
+  }
+  // Assume the first result is a good representation for the entire run.
```
mum4k, line 23: I remember we discussed sorting these and choosing one in particular.

Why is it safe to assume the first is a good representation? Since this is concurrent, I don't think there are any guarantees.

source/adaptive_load/metrics_evaluator_impl.cc
mum4k: Yes please, could you file an issue describing what needs to be done?

source/adaptive_load/metrics_evaluator_impl.cc
```diff
@@ -15,16 +15,42 @@ using ::nighthawk::adaptive_load::MetricSpec;
...
+  time_interval.duration = output.results(0).execution_duration();
+  return time_interval;
+}
+
+// Utility function for GetMetric functionality with fallback logic.
+absl::StatusOr<double> GetMetric(MetricsPlugin& metrics_plugin, absl::string_view metric_name,
+                                 const MeasuringPeriod& measuring_period) {
+  absl::StatusOr<double> metric_value_or =
+      metrics_plugin.GetMetricByNameWithMeasuringPeriod(metric_name, measuring_period);
+  // If the metric plugin does not support WithTime implementation (i.e. is using the default
```
mum4k, line 35: Looks like this comment refers to the old name when it contained the words "WithTime".

test/adaptive_load/fake_plugins/fake_metrics_plugin/fake_metrics_plugin.cc
```diff
@@ -41,6 +41,18 @@ absl::StatusOr<double> FakeMetricsPlugin::GetMetricByName(absl::string_view metr
   return value_or_error_from_name_[metric_name];
 }
 
+FakeMetricsPluginWithMeasuringPeriod::FakeMetricsPluginWithMeasuringPeriod(
+    const nighthawk::adaptive_load::FakeMetricsPluginConfig& config)
+    : FakeMetricsPlugin(config) {}
+
+absl::StatusOr<double> FakeMetricsPluginWithMeasuringPeriod::GetMetricByNameWithMeasuringPeriod(
```
mum4k, line 48: Does this return `absl::InternalError` unconditionally?

If yes, how are we testing the case where a plugin has the `WithMeasuringPeriod` method implemented and returns value?

test/adaptive_load/fake_plugins/fake_metrics_plugin/fake_metrics_plugin.h
```diff
@@ -35,6 +35,22 @@ class FakeMetricsPlugin : public MetricsPlugin {
...
+  /**
+   * Initializes the fake plugin with a FakeMetricsPluginConfig proto.
+   *
+   * @param config FakeMetricsPluginConfig proto for setting the fixed metric value.
+   */
+  explicit FakeMetricsPluginWithMeasuringPeriod(
+      const nighthawk::adaptive_load::FakeMetricsPluginConfig& config);
+
+  absl::StatusOr<double>
+  GetMetricByNameWithMeasuringPeriod(absl::string_view metric_name,
```
mum4k, line 50: If this implementation always fails, we should document that here.

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -365,5 +377,59 @@ TEST(AnalyzeNighthawkBenchmark, UsesBuiltinMetricsPluginForUnspecifiedPluginName
   EXPECT_EQ(result_or.value().metric_evaluations()[0].metric_value(), kExpectedSendRate);
 }
 
+TEST(AnalyzeNighthawkBenchmark, CorrectlyPrioritizesWithMeasuringPeriodImplementation) {
+  nighthawk::adaptive_load::AdaptiveLoadSessionSpec spec;
```
mum4k, line 381: I think we should add a few more test cases for complete coverage.

We have at least the following behaviors:
- a plugin that didn't implement the `WithMeasuringPeriod` method.
- a plugin that did implement the `WithMeasuringPeriod`, but fails when getting the metric.
- a plugin that did implement the `WithMeasuringPeriod` and successfully gets the metric.

mum4k (changes requested):

source/adaptive_load/metrics_evaluator_impl.cc
tomjzzhang: https://github.com/envoyproxy/nighthawk/issues/889

include/nighthawk/adaptive_load/metrics_plugin.h
tomjzzhang: Done.

include/nighthawk/adaptive_load/metrics_plugin.h
tomjzzhang: Done.

include/nighthawk/adaptive_load/metrics_plugin.h
tomjzzhang: Done.

include/nighthawk/adaptive_load/metrics_plugin.h
tomjzzhang: Done.

test/adaptive_load/metrics_evaluator_test.cc
tomjzzhang: Done, with the exclusion of the first case as this is implicitly tested with the current tests.

test/adaptive_load/fake_plugins/fake_metrics_plugin/fake_metrics_plugin.cc
tomjzzhang: Opted for a Mock approach instead.

source/adaptive_load/metrics_evaluator_impl.cc
tomjzzhang: Good catch. Done.

source/adaptive_load/metrics_evaluator_impl.cc
tomjzzhang: Changed to a more accurate approach. I originally opted for using the first one as the deviation seemed to be not large and any result seemed like a good enough representation.

test/adaptive_load/fake_plugins/fake_metrics_plugin/fake_metrics_plugin.h
tomjzzhang: Opted for mocking approach, so this comment no longer applies.

include/nighthawk/adaptive_load/metrics_plugin.h
mum4k: Thank you, the updated doc comment are detailed and easy to understand.

source/adaptive_load/metrics_evaluator_impl.cc
mum4k: I think the current approach will work well. Thank you.

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -365,5 +395,145 @@ TEST(AnalyzeNighthawkBenchmark, UsesBuiltinMetricsPluginForUnspecifiedPluginName
...
+  MetricsEvaluatorImpl evaluator;
+  absl::StatusOr<BenchmarkResult> result_or = evaluator.AnalyzeNighthawkBenchmark(
+      nighthawk_response, spec, name_to_custom_metrics_plugin_map);
+  ASSERT_FALSE(result_or.ok());
+  EXPECT_EQ(result_or.status().code(), absl::StatusCode::kInvalidArgument);
+  EXPECT_THAT(result_or.status().message(), HasSubstr("output.results cannot be empty."));
+}
+
+// Utility to set the execution time and execution duration of a results object.
+void SetTimeAndDurationOnResult(nighthawk::client::Result* result, int64_t time_seconds,
```
mum4k, line 478: Nice, this makes the test cases easier to read.

test/adaptive_load/metrics_evaluator_test.cc
```diff
@@ -365,5 +395,145 @@ TEST(AnalyzeNighthawkBenchmark, UsesBuiltinMetricsPluginForUnspecifiedPluginName
...
+                                int64_t duration_seconds) {
+  result->mutable_execution_start()->set_seconds(time_seconds);
+  result->mutable_execution_start()->set_nanos(0);
+  result->mutable_execution_duration()->set_seconds(duration_seconds);
+  result->mutable_execution_duration()->set_nanos(0);
+}
+
+TEST(AnalyzeNighthawkBenchmark, FailsWithDurationWhereAllWorkersAreActive) {
```
mum4k, line 486: Should the name of the test case be in the negative, e.g. `FailsWithDurationWhereAllWorkersAreNeverActive` or `FailsWithoutDurationWhereAllWorkersAreActive`.

mum4k (changes requested):
Looks great, just one last nit on a test case name, assuming I understood the goal of the test right.

test/adaptive_load/metrics_evaluator_test.cc
tomjzzhang: Done.

mum4k (approved):

PR #887, opened 2022-08-03
Use proto matcher instead of debug_string matching for protos.

dubious90:
In some environments, whitespace differences cause these tests to fail.
String matching is less stable than the proto matchers that we now have.

Works on #433 


dubious90:
@tomjzzhang please review and assign to @mum4k when done.

tomjzzhang (approved):
LGTM.

mum4k (approved):

PR #894, opened 2022-08-09
Refactor subprocess mixin to control setting up and tearing down chil…

KBaichoo:
The Subprocess mixin can be used to control setting up and tearing down child process. 
The `TestServerBase` will leverage this and it will be re-used by a xDS configuration generator.

### DIFF
--- a/test/integration/subprocess_mixin.py
+++ b/test/integration/subprocess_mixin.py
@@ -10,7 +10,7 @@ class SubprocessMixin(ABC):
   """Mixin used to manage launching subprocess using a separate Python thread."""
 
   def __init__(self):
-    """Creates SubprocessMixin."""
+    """Create SubprocessMixin."""
     self._server_thread = Thread(target=self._serverThreadRunner)
     self._server_process = None
     self._has_launched = False
--- a/test/integration/subprocess_mixin.py
+++ b/test/integration/subprocess_mixin.py
@@ -1,3 +1,5 @@
+"""Mixin for managing child subprocess with an additional thread."""
+
 from threading import Thread, Condition
 import subprocess
 import logging
@@ -8,6 +10,7 @@ class SubprocessMixin(ABC):
   """Mixin used to manage launching subprocess using a separate Python thread."""
 
   def __init__(self):
+    """Creates SubprocessMixin."""
     self._server_thread = Thread(target=self._serverThreadRunner)
     self._server_process = None
     self._has_launched = False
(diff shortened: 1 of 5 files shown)


mum4k:

qqustc (dismissed):

test/integration/subprocess_mixin.py
```diff
@@ -0,0 +1,74 @@
...
+class SubprocessMixin(ABC):
+  """Mixin used to manage launching subprocess using a separate Python thread."""
+
+  def __init__(self):
+    """Create SubprocessMixin."""
+    self._server_thread = Thread(target=self._serverThreadRunner)
+    self._server_process = None
+    self._has_launched = False
+    self._has_launched_cv = Condition()
+    self.stdout = b''
```
mum4k, line 18: Do we have to store the stderr and stdout as class attributes? The lifetime feels questionable, since the class could potentially be reused for multiple runs.

The `_serverThreadRunner` returns the same, can we just return them from a public method without storing them?

test/integration/subprocess_mixin.py
```diff
@@ -0,0 +1,74 @@
...
+    self.stderr = b''
+
+  @abstractmethod
+  def _argsForSubprocess(self) -> list[str]:
+    """Return the args to launch the subprocess."""
+    pass
+
+  def _serverThreadRunner(self):
```
mum4k, line 26: Can we document this method? The child class seems to be overriding it, so a doc comment can help future maintainers / implementers.

test/integration/subprocess_mixin.py
```diff
@@ -0,0 +1,74 @@
+"""Mixin for managing child subprocess with an additional thread."""
+
+from threading import Thread, Condition
+import subprocess
+import logging
+from abc import ABC, abstractmethod
+
+
+class SubprocessMixin(ABC):
+  """Mixin used to manage launching subprocess using a separate Python thread."""
```
mum4k, line 10: Can we expand the class docstring with an example use, so it is obvious in what order are the methods supposed to be called?

mum4k (changes requested):
This is a good improvement, thank you!

test/integration/subprocess_mixin.py
KBaichoo: makes sense, done.

test/integration/subprocess_mixin.py
KBaichoo: done.

test/integration/subprocess_mixin.py
KBaichoo: makes sense, I added a note to look at the particular tests since they provide a clear example and won't go stale as if they're broken CI will complain

mum4k (approved):
Thank you for iterating.

PR #898, opened 2022-08-22
Add base CDS churn component and test.

KBaichoo:
Currently it only works on the single test endpoint, but will be expanded. Will follow up tagged TODOs in a follow up as this PR is fairly large already.


mum4k:
@KBaichoo please tag me on the PR when it is ready for review. Looks like it is failing some CI tests.

KBaichoo:
/assign @mum4k 

CI is green now, PTAL, thank!

mum4k:
@tomjzzhang please review and assign back to me once done.

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,196 @@
...
+        return
+
+      random.shuffle(self.inactive_clusters)
+
+      for i in range(num_clusters_to_add):
+        self.active_config.resources.add().Pack(self.inactive_clusters.pop())
+
+    elif action == self.__class__.Action.REMOVE:
+      resources = self.active_config.resources
+      num_clusters_to_remove = random.randrange(0, len(resources) + 1)
```
tomjzzhang, line 136: I assume it is intentional that remove and add have a chance of doing nothing?

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,196 @@
...
+
+      # All clusters inactive or 0 clusters choosen, do nothing.
+      if not resources or not num_clusters_to_remove:
+        return
+
+      for i in range(num_clusters_to_remove):
+        any_resource = resources.pop(random.randrange(0, len(resources)))
+        unpacked_cluster = cluster_pb2.Cluster()
+        any_resource.Unpack(unpacked_cluster)
+        self.inactive_clusters.append(unpacked_cluster)
```
tomjzzhang, line 146: optional: extracting REMOVE and ADD into functions might make this more readable.

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,196 @@
...
+  parser = argparse.ArgumentParser(
+      description='Provide a base64 encoded protobuf string of the configuration.')
+  parser.add_argument('--config', required=True, help='base64 encoded protobuf string.')
+  args = parser.parse_args()
+
+  base64_decoded = base64.b64decode(args.config, validate=True)
+  settings = cluster_manager_pb2.DynamicClusterConfigManagerSettings()
+  settings.ParseFromString(base64_decoded)
+
+  mgr = DynamicClusterConfigManager(settings)
```
tomjzzhang, line 194: nit: can we use manager instead of the abbreviation for clarify?

test/integration/nighthawk_test_server.py
```diff
@@ -23,26 +23,6 @@
 from test.integration.subprocess_mixin import SubprocessMixin
 
 
-def _substitute_yaml_values(runfiles_instance, obj, params):
```
tomjzzhang, line 26: +1 to extracting this out

test/integration/nighthawk_test_server.py
```diff
@@ -283,6 +273,27 @@ def stop(self):
     os.remove(self._admin_address_path)
     return self.stopSubprocess()
 
+  @staticmethod
```
tomjzzhang, line 276: I think my own preference would be for this to be in a utility file with supporting unit tests.  Understand this is out of scope for your PR, so happy with a TODO for now.

tomjzzhang (dismissed):
Overall LGTM. Just some questions and minor quality improvements.

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -0,0 +1,100 @@
...
+
+  def __init__(self, request, server_config, proxy_config, dynamic_config_settings):
+    """Initialize an InjectDynamicHttpProxyIntegrationTestBase.
+
+    Arguments:
+      request: The pytest `request` test fixture used to determine information
+        about the currently executing test case.
+      server_config: Path to the server configuration.
+      proxy_config: Path to the proxy configuration.
+      dynamic_config_settings: Settings for the dynamic configuration component.
```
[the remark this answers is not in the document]
KBaichoo: Added type annotations to much of the code in this file. Since we can do this incrementally, and there's no precedent will just incrementally add.

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Assert that both started successfully, and return afterwards.
+    """
+    super(InjectDynamicHttpProxyIntegrationTestBase, self).setUp()
+
+    output_file = os.path.join(self.test_server.tmpdir, self.dynamic_config_settings.output_file)
+    self.dynamic_config_settings.output_file = output_file
+    logging.info(f"Injecting dynamic configuration. Output file: {output_file}")
+
+    # TODO(kbaichoo): we only hardcode single endpoint, but will expand on this.
```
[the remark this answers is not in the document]
KBaichoo: done

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -0,0 +1,100 @@
...
+    self.dynamic_config_controller.stop()
+
+
+@pytest.fixture(params=determineIpVersionsFromEnvironment())
+def inject_dynamic_envoy_http_proxy_fixture(request, server_config, proxy_config,
+                                            dynamic_config_settings, caplog):
+  """Injects a dynamically configured Envoy proxy in front of the test server.
+
+  Arguments:
+    request: supplies the ip version.
```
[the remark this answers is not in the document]
KBaichoo: done.

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,196 @@
...
+      clusters.append(new_cluster)
+
+    # Randomly assign clusters to either inactive or active.
+    while clusters:
+      add_as_active_cluster = random.choice([True, False])
+      if add_as_active_cluster:
+        self.active_config.resources.add().Pack(clusters.pop())
+      else:
+        self.inactive_clusters.append(clusters.pop())
+    self.output_file = config.output_file if config.output_file else "new_cds.pb"
```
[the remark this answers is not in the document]
KBaichoo: that's a good idea, will add that in follow up (added todo here)

KBaichoo (commented):
Thanks for the review!

benchmarks/static_test/test_discovery.py
```diff
@@ -5,16 +5,12 @@
 integration test framework to run benchmark executions.
```
mum4k, line 5: I think the rename of `benchmarks/test` to `benchmarks/static_test` also breaks Salvo. There is reference to the path in its configuration.

I would recommend we try to avoid changing the pre-existing layout and target names, since the `/benchmarks` subdirectory technically is public and part of Nighthawk's API.

benchmarks/utilities.py
```diff
@@ -0,0 +1,32 @@
...
+
+import logging
+import json
+from rules_python.python.runfiles import runfiles
+import os
+from shutil import copyfile
+
+
+def output_benchmark_results(json_results, fixture):
+  """Output the benchmark results."""
```
mum4k, line 11: Can we expand this and document the arguments?

We could also add a few more details into the docstring indicating what this utility does, since it isn';t immediately obvious.

dynamic_config/BUILD
```diff
@@ -0,0 +1,54 @@
+load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
```
mum4k, line 1: Can we also add a top level `README.md` indicating what this directory is for?

Additionally would you mind moving it under `test/integration` where most of test utilities and fixtures live? Making this a top level directory doesn't seem to be the right move, since it doesn't have anything to do with Nighthawk itself, just with the test suite built around it.

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,195 @@
+"""Component that handles dynamic configuration of the Envoy proxy."""
```
mum4k, line 1: Can we be a bit more specific indicating that this is the Envoy proxy which is the system under test and not the one used by Nighthawk as the test server?

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,195 @@
...
+
+from google.protobuf import json_format
+import argparse
+import time
+import base64
+import logging
+import os
+
+
+class DynamicConfigManager(ABC):
```
mum4k, line 25: It is hard for me to review an interface and justify its existence if we only have one implementation. Can you help me understand why this can't be a single class without the inheritance hierarchy?

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,195 @@
...
+  """Base class for Dynamic configuration components."""
```
mum4k, line 26: If we do have a good reason to keep this as an interface, can we expand the commentary here outlining what the interface is for and what rules (if any) should the implementations follow?

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,195 @@
...
+  """
+  cluster_template = json_format.Parse(
+      json.dumps(yaml.load(cluster_template_string, Loader=yaml.FullLoader)), cluster_pb2.Cluster())
+
+  mutate_actions = list(Action)
+
+  def __init__(self, config: cluster_manager_pb2.DynamicClusterConfigManagerSettings):
+    """Create a DynamicClusterConfigManager."""
+    DynamicConfigManager.__init__(self)
+    self.active_config = discovery_pb2.DiscoveryResponse()
```
mum4k, line 89: Can we document all the public attributes in the class docstring (or make them private if they don't need to be public)?

dynamic_config/dynamic_config_manager.py
mum4k: I tend to agree with @tomjzzhang, we don't want to extract them for the purpose of reuse, but to make the high level business logic of mutate more readable. That way the reader doesn't have to delve into the details immediately.

dynamic_config/dynamic_config_manager.py
```diff
@@ -0,0 +1,195 @@
...
+    self.active_config.version_info = "1"
+    self.inactive_clusters = []
+    clusters = []
+    for service_config in config.clusters:
```
mum4k, line 93: Let's extract some of this low level code into helper private functions or method that will be well named and will improve readability of the initializer.

dynamic_config/dynamic_config_server.py
```diff
@@ -0,0 +1,51 @@
...
+class DynamicConfigController(SubprocessMixin):
+  """Controller for the Dynamic Configuration Manager."""
+
+  def __init__(self, dynamic_config_settings):
+    """Create DynamicConfigController."""
+    super().__init__()
+    runfiles_instance = runfiles.Create()
+    self.binary_path = runfiles_instance.Rlocation(
+        'nighthawk/dynamic_config/dynamic_config_manager')
+    self.settings = dynamic_config_settings
```
mum4k, line 21: Let's document the public attributes (or make them private).

dynamic_config/dynamic_config_server.py
```diff
@@ -0,0 +1,51 @@
...
+from rules_python.python.runfiles import runfiles
+import base64
+import os
+from os.path import exists
+import time
+import logging
+
+
+class DynamicConfigController(SubprocessMixin):
+  """Controller for the Dynamic Configuration Manager."""
```
mum4k, line 13: Can we expand this and the other class docstrings with enough information so that a new user / reader knows 1) what the class does and 2) how to use it?

test/integration/nighthawk_test_server.py
```diff
@@ -283,6 +273,28 @@ def stop(self):
     os.remove(self._admin_address_path)
     return self.stopSubprocess()
 
+  # TODO(kbaichoo): migrate to a utility file.
+  @staticmethod
```
mum4k, line 277: Use of staticmethod should be avoided. Can we make this a plain function (public if need be)?

https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

mum4k (changes requested):
Looks like the file `benchmarks/configurations/dummy_cds.yaml` is empty and it doesn't seem to be referenced anywhere. What do we need it for?

If we need an empty config file, can we call it `empty_cds.yaml` instead?

benchmarks/utilities.py
KBaichoo: done.

test/integration/nighthawk_test_server.py
KBaichoo: done, plain fn in module for now.

benchmarks/static_test/test_discovery.py
KBaichoo: shifted static_test, targets, etc to prior.

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -0,0 +1,102 @@
...
+    Arguments:
+      request: The pytest `request` test fixture used to determine information
+        about the currently executing test case.
+      server_config: Path to the server configuration.
+      proxy_config: Path to the proxy configuration.
+      dynamic_config_settings: Settings for the dynamic configuration component.
+    """
+    super(InjectDynamicHttpProxyIntegrationTestBase, self).__init__(request, server_config,
+                                                                    proxy_config)
+    self.dynamic_config_settings = dynamic_config_settings
```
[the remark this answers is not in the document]
KBaichoo: done.

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -0,0 +1,102 @@
+"""Customized fixture to dynamically configure an Envoy Proxy."""
+
+import pytest
+import logging
+import os
+from typing import Generator
+
+from test.integration.integration_test_fixtures import determineIpVersionsFromEnvironment
+from envoy_proxy import EnvoyProxyServer, InjectHttpProxyIntegrationTestBase
```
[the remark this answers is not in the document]
KBaichoo: how do you want to handle pytest fixtures conflicting with this rule.

e.g. 

```
\@pytest.fixture()
def my_fixture()
   yield someFooFixture()
```

but in another module we have to import this function otherwise pytest would pick it up.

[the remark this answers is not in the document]
KBaichoo: done for added code apart from the conflict mentioned above.

dynamic_config/BUILD
KBaichoo: Added a readme.

I don't think `test/integration` is the right place since the components can be used outside of those fixtures e.g. similar to how you can run nighthawk outside of these fixtures, you can use these components for an Envoy there without any of the integration test stuff.

KBaichoo (commented):
Thanks for the review @mum4k , think I've address all comments. PTAL.

benchmarks/dynamic_config_envoy_proxy.py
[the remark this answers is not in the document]
mum4k: Fair point about pytest fixtures, I think it is fine to exclude those from this rule.

Sorry for not realizing that, I am not used to pytest yet.

dynamic_config/BUILD
mum4k: That is fair, if you think this belongs to the top level directory, can we add a README.md outlining what this is and how can one start and use the component?

mum4k: My bad, just noticed that the README is already present. Could you expand it with an example of use though?

Ok to go with a TODO and postpone to a later PR if you prefer.

mum4k (changes requested):
Looks good, can you please look into the CI failures before we proceed?

dynamic_config/BUILD
KBaichoo: added a TODO here.

mum4k (approved):
Thank you Kevin.

PR #904, opened 2022-08-31
CDS churn with multiple clusters with traffic

KBaichoo:
Add a test that has 5 clusters, 3 endpoints each. Use `request-source-plugin-config` to configure client to target different backends.

### DIFF
--- a/benchmarks/dynamic_config_envoy_proxy.py
+++ b/benchmarks/dynamic_config_envoy_proxy.py
@@ -5,7 +5,7 @@ import logging
 import os
 from typing import Generator
 
-from test.integration import integration_test_fixtures, common
+from test.integration import integration_test_fixtures, common, utility
 import envoy_proxy
 from dynamic_config import dynamic_config_server
 from nighthawk.api.configuration import cluster_config_manager_pb2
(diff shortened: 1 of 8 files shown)


mum4k:

qqustc (dismissed):

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -68,17 +64,45 @@ def setUp(self):
...
 
     assert (self._dynamic_config_controller.start())
     logging.info("dynamic configuration running")
 
+  def _assignEndpointsToClusters(self, clusters):
+    """Process all backend endpoints, round robin assigning to the given clusters."""
+    available_endpoints = []
+
+    for endpoint in self.getAllTestServerRootUris():
+      ip_and_port = endpoint.split('/')[2]
```
mum4k, line 92: Can we add a comment here indicating an example of the string that we are splitting here? That can help the readers understand the code.

benchmarks/dynamic_config_envoy_proxy.py
```diff
@@ -68,17 +64,45 @@ def setUp(self):
...
+      available_endpoints.append(ip_and_port)
+
+    for cluster in clusters:
+      new_cluster = self._dynamic_config_settings.clusters.add()
+      new_cluster.name = cluster
+    cluster_idx = 0
+    for endpoint in available_endpoints:
+      new_endpoint = self._dynamic_config_settings.clusters[cluster_idx].endpoints.add()
+      ip, port = endpoint.rsplit(':', maxsplit=1)
```
mum4k, line 101: Can we extract the logic that parses the ip / port and the one that calculates the cluster index and unit test it to improve test coverage?

mum4k (changes requested):
Just two minor comments.

benchmarks/dynamic_config_envoy_proxy.py
KBaichoo: done.

benchmarks/dynamic_config_envoy_proxy.py
KBaichoo: extracted the utility for ip/port and added unit test for it

mum4k (approved):

PR #906, opened 2022-09-06
Update Envoy to c1744d4 (Sep 6th 2022)

jiajunye:
- Update Nighthawk to the latest Envoy.
- Adjust the 2xx count assertion from exact matching to GreaterOrEqual in `test_h3_quic_with_custom_http3_protocol_options` to pass the test. Nighthawk doesn't guarantee the exact number of responses that will be sent.
- Update the latest update timestamp of the `requirements.txt`. There is no outdated dependency for the dependencies in `requirements.txt`.

Signed-off-by: jiajunye [jiajunye@google.com](mailto:jiajunye@google.com)


mum4k (changes requested):
Since the last update of Python requirements was more than a month ago, we should perform that step as well:

https://github.com/envoyproxy/nighthawk/blob/2f568d40fa11bcd75a87d1b17ce378174d3760c5/MAINTAINERS.md?plain=1#L80

mum4k (approved):

PR #907, opened 2022-09-12
Create API layer for a plugin system that collects user-defined data.

dubious90:
Creates the API layer (unimplemented) for a new plugin framework that:
1. Receives response headers and data, allowing them to collect their own statistics
2. Allows them to define their own output proto that will be appended to the Nighthawk Result for each worker.

Also creates a fake plugin implementation to make sure the api compiles.


test/user_defined_output/fake_plugin/fake_user_defined_output_test.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "gtest/gtest.h"
+
+namespace Nighthawk {
+namespace {
+
+using ::nighthawk::FakeUserDefinedOutput;
+using ::nighthawk::FakeUserDefinedOutputConfig;
+using ::testing::HasSubstr;
+
+TEST(FakeUserDefinedOutputPluginFactory, CreateEmptyConfigProtoCreatesCorrectType) {
```
dubious90, line 20: More tests should probably be created for this fake plugin, but I wanted to wait until we actually agree on the API first before making those tests.

dubious90:
@eric846 Please review and assign to mum4k when done.

include/nighthawk/user_defined_output/user_defined_output_plugin.h
```diff
@@ -0,0 +1,103 @@
...
+   *   Might return:
+   *     {int_value: 3, array_value: ["a","b","c"]}
+   * 
+   * @param per_worker_outputs List of the outputs that every per-worker instance of the User
+   *    Defined Output Plugin created.
+   * @param global_output Any object to pack the aggregated output into to add to the global
+   *    Result.
+   * @return absl::Status
+   */
+  virtual absl::Status AggregateGlobalOutput(
```
eric846, line 99: This is a surprising place to do any kind of processing, but it's also a convenient place if we need each kind of plugin to implement something that isn't down at the worker level. To be more traditional we could have two different factory methods here, something like:
- `createUserDefinedPerWorkerOutputHandler()`
- `createUserDefinedOutputAggregator()`

and introduce an output aggregator interface.

But it seems like there could be a tradeoff between clarity and added boilerplate.

test/user_defined_output/fake_plugin/fake_user_defined_output.cc
```diff
@@ -0,0 +1,82 @@
...
+absl::Status FakeUserDefinedOutputPluginFactory::AggregateGlobalOutput(
+  absl::Span<const google::protobuf::Any> per_worker_outputs,
+  google::protobuf::Any& global_output_any) {
+  FakeUserDefinedOutput global_output;
+  global_output.set_worker_name("global");
+  int data_called = 0;
+  int headers_called = 0;
+  for (const google::protobuf::Any& any : per_worker_outputs) {
+    FakeUserDefinedOutput output;
+    any.UnpackTo(&output);
```
eric846, line 67: Is the caller going to guarantee that the `per_worker_outputs` will only come from this plugin type, so we don't need to check that the unpacking succeeded?

test/user_defined_output/fake_plugin/fake_user_defined_output.cc
```diff
@@ -0,0 +1,82 @@
...
+  output.set_headers_called(headers_called_);
+  output.set_worker_name(worker_metadata_.name);
+
+  output_any.PackFrom(output);
+
+  return absl::OkStatus();
+}
+
+
+// BROKEN BELOW HERE
```
eric846, line 40: Leftover comment?

test/user_defined_output/fake_plugin/fake_user_defined_output.proto
```diff
@@ -0,0 +1,27 @@
...
+
+  // Configures handleResponseHeaders to fail after header_failure_countdown calls.
+  bool fail_headers = 3;
+  int32 header_failure_countdown = 4;
+
+  // Causes the plugin to fail on getPerWorkerOutput step.
+  bool fail_per_worker_output = 5;
+}
+
+message FakeUserDefinedOutput {
```
eric846, line 22: Should have comments for the message and fields

eric846 (changes requested):

eric846 (dismissed):
LGTM

include/nighthawk/user_defined_output/user_defined_output_plugin.h
```diff
@@ -0,0 +1,107 @@
...
+   *     )
+   *   Might return:
+   *     {int_value: 3, array_value: ["a","b","c"]}
+   * 
+   * @param per_worker_outputs List of the outputs that every per-worker instance of the User
+   *    Defined Output Plugin created.
+
+   * @return global_output Any-packed aggregated output to add to the global Result.
+   */
+  virtual absl::StatusOr<google::protobuf::Any> AggregateGlobalOutput(
```
mum4k, line 104: Here as well - can we indicate what happens to the Nighthawk execution if this fails?

test/user_defined_output/fake_plugin/fake_user_defined_output.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "api/client/options.pb.h"
+#include "api/client/service.grpc.pb.h"
+
+#include "test/user_defined_output/fake_plugin/fake_user_defined_output.pb.h"
+
+namespace Nighthawk {
+
+/**
+ * UserDefinedOutputPlugin for testing: ...
```
mum4k, line 17: Can we expand the `...` and finish the comment?

Ideally we should describe the behaviors of the fake plugin, either here or on the individual methods. The comment here could focus on how the plugin should be used.

include/nighthawk/user_defined_output/user_defined_output_plugin.h
```diff
@@ -0,0 +1,107 @@
...
+struct WorkerMetadata {
+  std::string name;
+};
+
+/**
+ * An interface for the UserDefinedOutputPlugin that receives responses and allows users to
+ * attach their own custome output to each worker Result.
+ *
+ */
+class UserDefinedOutputPlugin {
```
mum4k, line 26: Where are we documenting the thread-safety requirements for plugin implementations?

test/user_defined_output/fake_plugin/fake_user_defined_output.proto
```diff
@@ -0,0 +1,30 @@
...
+
+package nighthawk;
+
+// Configuration for FakeUserDefinedOutputPlugin (plugin name: "nighthawk.fake_user_defined_output")
+// that counts the number of times each api is called. Can also be programmed to return a proto
+// validation failure or return an error after some number of each step.
+message FakeUserDefinedOutputConfig {
+  // Configures handleResponseData to fail after data_failure_countdown calls.
+  bool fail_data = 1;
+  int32 data_failure_countdown = 2;
```
mum4k, line 11: Can we also document and explain the *_countdown fields?

test/user_defined_output/fake_plugin/fake_user_defined_output_test.cc
mum4k: The API looks good. Can we expand the tests coverage for the plugin?

mum4k (changes requested):

api/client/options.proto
```diff
@@ -302,4 +302,8 @@ message CommandLineOptions {
   // compatibility purposes. In the future, this field may be auto-populated when left unset and
   // circumstances mandate so (distributed load test execution).
   google.protobuf.StringValue execution_id = 106;
+
+  // Optional set of plugin config to be parsed by a UserDefinedOutputPlugin and used to parse 
+  // responses to provide additional output to each Result.
+  repeated envoy.config.core.v3.TypedExtensionConfig user_defined_plugin_configs = 112;
```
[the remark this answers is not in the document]
dubious90: Done. Envoy throws an exception when that happens, but we'd likely convert it to a Status or something. As you state, we'll fail the execution if a plugin isn't registered properly.

test/user_defined_output/fake_plugin/fake_user_defined_output_test.cc
dubious90: Done.

oschaaf (commented):
Leaving some code-level comments/questions. 
Also, I am curious if there would be value in wiring in some generic/optional way for existing transforms to handle output produces by plugins (for example, consider the human readable output). 

Having said that, this looks good to me overall

include/nighthawk/user_defined_output/user_defined_output_plugin.h
```diff
@@ -0,0 +1,142 @@
...
+   * non-ok status will be logged and increment a counter that will be added to the worker Result.
+   * Callers can also provide a failure predicate for this counter that will abort the request
+   * after n plugin failures.
+   *
+   * Must be thread safe.
+   *
+   * @param headers
+   */
+  virtual absl::Status
+  handleResponseHeaders(const Envoy::Http::ResponseHeaderMapPtr&& headers) PURE;
```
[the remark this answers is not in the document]
mum4k: The latest version seems to be passing it by value. We likely don't want to do that - since this is a smart pointer this would mean that Nighthawk would lose the ownership of the object. This should probably become a raw pointer now.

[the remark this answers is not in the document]
dubious90: Yep, had meant to make it a raw ptr (*) and corrected it incorrectly to Ptr without thinking. Thanks

api/client/output.proto
```diff
@@ -44,14 +45,30 @@ message Statistic {
   }
 }
 
+// An output generated by a UserDefinedOutput plugin.
+message UserDefinedOutput {
+  // The official name of the plugin as registered through Envoy.
```
mum4k, line 50: Sorry, can we clarify this a bit more, as it stands this might be read as there is some Envoy instance the plugins have to register with. But I think what we wanted to say is that the plugins register through the Envoy's plugin registration mechanism.

include/nighthawk/user_defined_output/user_defined_output_plugin.h
```diff
@@ -0,0 +1,141 @@
...
+  // Identifies which worker instantiated the plugin instance.
+  int worker_number;
+};
+
+/**
+ * An interface for the UserDefinedOutputPlugin that receives responses and allows users to
+ * attach their own custom output to each worker Result.
+ *
+ * All UserDefinedOutputPlugins must be thread safe, as it may receive multiple responses
+ * simultaneously. In addition, handleResponseData and handleResponseHeaders may be called in any
```
mum4k, line 26: optional nit: s/simultaneously/concurrently/. I think saying concurrently may be more accurate.

mum4k (dismissed):
Last two comment nits, feel free to merge once addressed.

mum4k (approved):

Issue #913, opened 2022-09-21
fix_format script always silently passes without formatting anything


dubious90:
*Description*:
See title.

*Reproduction steps*:
Create an obvious formatting error.
Then run:
`ci/do_ci.sh fix_format`
Code will pass, despite your new format error

Issue #915, opened 2022-09-21
Investigate moving off of direct google::protobuf types


dubious90:
When running fix_format script, it complains:
```
Please note: `tools/code_format/check_format.py` no longer checks API `.proto` files, please use `tools/proto_format/proto_format.sh` if you are making changes to the API files
ERROR: From ./../nbperry-nighthawk/source/adaptive_load/metrics_evaluator_impl.cc
ERROR: ./../nbperry-nighthawk/source/adaptive_load/metrics_evaluator_impl.cc:27: unexpected direct dependency on google.protobuf, use the definitions in common/protobuf/protobuf.h instead.
ERROR: ./../nbperry-nighthawk/source/adaptive_load/metrics_evaluator_impl.cc:28: unexpected direct dependency on google.protobuf, use the definitions in common/protobuf/protobuf.h instead.
ERROR: From ./../nbperry-nighthawk/include/nighthawk/adaptive_load/metrics_plugin.h
ERROR: ./../nbperry-nighthawk/include/nighthawk/adaptive_load/metrics_plugin.h:23: unexpected direct dependency on google.protobuf, use the definitions in common/protobuf/protobuf.h instead.
ERROR: ./../nbperry-nighthawk/include/nighthawk/adaptive_load/metrics_plugin.h:26: unexpected direct dependency on google.protobuf, use the definitions in common/protobuf/protobuf.h instead.
ERROR: check format failed. diff has been applied'
```

For now, we are excluding all of Nighthawk from this check. Can we fix this without making a backwards incompatible change?

Issue #914, opened 2022-09-21
Stop using deprecated mock method macros


dubious90:
When running fix_format, it will give a complaint like the following:
```
ERROR: From ./../nbperry-nighthawk/test/test_common/mock_stream.h
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:15: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:18: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:21: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:31: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:34: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:35: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:38: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:41: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
ERROR: ./../nbperry-nighthawk/test/test_common/mock_stream.h:42: The MOCK_METHODn() macros should not be used, use MOCK_METHOD() instead
```

I considered this outside the scope of fixing formatting, but as a followup we should probably move to MOCK_METHOD.

PR #916, opened 2022-09-21
Make the fix_format and check_format scripts work on Nighthawk code

dubious90:
Fixes https://github.com/envoyproxy/nighthawk/issues/913

Direct changes:
- Modified config.yaml to not exclude the current directory, which was excluding all of NH
- Overrode some settings in config.yaml to move past some code changes
    - Filed https://github.com/envoyproxy/nighthawk/issues/914 and https://github.com/envoyproxy/nighthawk/issues/915 to address these overrides in the future
- Moved include_dir_order from check_format.sh over to the config.yaml where it now needs to live.

Also ran the format script to fix all current code formatting errors.

### DIFF
--- a/tools/check_format.sh
+++ b/tools/check_format.sh
@@ -15,7 +15,6 @@ TO_CHECK="${2:-$FULL_CHECK}"
 bazel run @envoy//tools/code_format:check_format.py -- \
   --skip_envoy_build_rule_check  --namespace_check Nighthawk \
   --build_fixer_check_excluded_paths=$TO_CHECK \
-  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,distributor,sink,grpcpp,request_source,test_common,test \
   $1 $TO_CHECK
 
 # The include checker doesn't support per-file checking, so we only
--- a/tools/code_format/config.yaml
+++ b/tools/code_format/config.yaml
@@ -35,7 +35,8 @@ suffixes:
 paths:
   # Path prefixes to exclude from checking
   excluded:
-  - .
+  # this exclusion prevents all of our files from being formatted              # unique
+  # - .                                                                        # unique
   - bazel/external/http_parser/
   - bazel/toolchains/configs/
   - bazel-
@@ -119,6 +120,8 @@ paths:
     - ci/prebuilt
     - source/common/protobuf
     - test/extensions/bootstrap/wasm/test_data
+    - .                                                                        # unique
+    # TODO(915): Remove this inclusion when all direct references removed      # unique
 
   # Files that are allowed to use try without main thread assertion.
   raw_try:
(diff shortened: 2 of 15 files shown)


dubious90:
@qqustc Please review and assign to @mum4k when done.

qqustc (approved):

tools/code_format/config.yaml
```diff
@@ -215,14 +218,22 @@ paths:
...
-  - test
+dir_order:                                                                     # unique
+  - envoy                                                                      # unique
+  - nighthawk                                                                  # unique
+  - external/source/envoy                                                      # unique
+  - external                                                                   # unique
+  - api                                                                        # unique
+  - common                                                                     # unique
+  - source                                                                     # unique
+  - exeserver                                                                  # unique
```
mum4k, line 229: Should these be two entries `exe` and `server`?

tools/check_format.sh
```diff
@@ -15,7 +15,6 @@ TO_CHECK="${2:-$FULL_CHECK}"
 bazel run @envoy//tools/code_format:check_format.py -- \
   --skip_envoy_build_rule_check  --namespace_check Nighthawk \
   --build_fixer_check_excluded_paths=$TO_CHECK \
-  --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,distributor,sink,grpcpp,request_source,test_common,test \
```
mum4k, line 18: Can we say something about the removal of this argument (its migration to the yaml) in the description of the PR for posterity?

tools/code_format/config.yaml
```diff
@@ -215,14 +218,22 @@ paths:
...
-  - source
-  - exe
-  - server
-  - extensions
-  - test
+dir_order:                                                                     # unique
+  - envoy                                                                      # unique
+  - nighthawk                                                                  # unique
+  - external/source/envoy                                                      # unique
+  - external                                                                   # unique
```
mum4k, line 225: I am trying to verify if this list contains all the directories it should, but I am having a hard time figuring out how these map into the list of our actual directories here:

https://github.com/envoyproxy/nighthawk

This seems to be a list of top level and sub-level directories. Can you help me understand what this list should contain and how it is used?

E.g. the following directories are missing here, is that intentional?
- benchmarks (no c++ code, so probably ok)
- adaptive_load

mum4k (changes requested):

tools/check_format.sh
dubious90: This was in the PR, but I had accidentally called it header_dir_order. Fixed that typo

tools/code_format/config.yaml
dubious90: Yes, fixed.

tools/code_format/config.yaml
dubious90: The short answer is I'm not 100% sure what this list's intention is. I copied this directly from:
   --include_dir_order envoy,nighthawk,external/source/envoy,external,api,common,source,exe,server,client,distributor,sink,grpcpp,request_source,test_common,test 

which did not contain adaptive_load or benchmarks.

My understanding of this list is that we probably have more things here than we actually need. Anything omitted I believe gets bundled into one section and alphabetized within that section. This list is ordered, so will ensure that (for example) external/source/envoy appears before anything else in external. I assume that for some envoy-related thing that header order matters, but for most files/subdirectories it shouldn't, and it would probably be fine to have them bundled in one section.

I don't remember what it is, but we've typically added directories to this list upon some mysterious error that we have encountered on occasion, but I forget what it is, or if it is still relevant after this yaml migration.

mum4k: Thank you, feel free to merge in assuming the files in the adaptive_load directories do get formatted even though they aren't listed.

mum4k (approved):

tools/code_format/config.yaml
dubious90: We talked about this offline, but after running an experiment, I realized that we are in fact missing some features in those directories. I am going to follow up separately adding directories and fixing issues

PR #920, opened 2022-10-04
Pass user defined output plugin through to options.

dubious90:
Adds in the scaffolding pieces of passing around user defined plugin options.

Deferring the actual use of the plugins to a later PR, in an effort to keep these PRs small/readable.

### DIFF
--- a/source/client/options_impl.cc
+++ b/source/client/options_impl.cc
@@ -362,6 +362,16 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       "Default: \"\"",
       false, "", "string", cmd);
 
+  // TODO(nbperry): Use better example when one exists.
+  TCLAP::MultiArg<std::string> user_defined_output_plugin_configs(
+      "", "user-defined-plugin-config",
+      "WIP - will throw unimplemented error. Optional configurations for plugins that collect data "
+      "about responses received by NH and attach a corresponding UserDefinedOutput to the Result. "
+      "Example (json): {name:\"nighthawk.fake_user_defined_output\",typed_config:{"
+      "\"@type\":\"type.googleapis.com/nighthawk.FakeUserDefinedOutputConfig\","
+      "fail_data:\"false\"}}",
+      false, "string", cmd);
+
   Utility::parseCommand(cmd, argc, argv);
 
   if (h2_use_multiple_connections.isSet()) {
@@ -644,6 +654,18 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       throw MalformedArgvException(e.what());
     }
   }
+  if (!user_defined_output_plugin_configs.getValue().empty()) {
+    for (std::string plugin_config_string : user_defined_output_plugin_configs.getValue()) {
+      try {
+        envoy::config::core::v3::TypedExtensionConfig typed_config;
+        Envoy::MessageUtil::loadFromJson(plugin_config_string, typed_config,
+                                         Envoy::ProtobufMessage::getStrictValidationVisitor());
+        user_defined_output_plugin_configs_.push_back(typed_config);
+      } catch (const Envoy::EnvoyException& e) {
+        throw MalformedArgvException(e.what());
+      }
+    }
+  }
 
   validate();
 }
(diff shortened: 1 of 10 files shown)


dubious90:
@jiajunye please review and assign to @mum4k when done.

source/client/options_impl.h
```diff
@@ -178,6 +183,7 @@ class OptionsImpl : public Options, public Envoy::Logger::Loggable<Envoy::Logger
   std::string latency_response_header_name_;
   absl::optional<Envoy::SystemTime> scheduled_start_;
   absl::optional<std::string> execution_id_;
+  std::vector<envoy::config::core::v3::TypedExtensionConfig> user_defined_output_plugin_configs_{};
```
jiajunye, line 186: Why do we need the curly braces?

source/client/process_impl.cc
```diff
@@ -425,6 +425,13 @@ absl::StatusOr<ProcessPtr> ProcessImpl::CreateProcessImpl(
                                                        std::move(typed_dns_resolver_config),
                                                        process_wide));
 
+  if (options.userDefinedOutputPluginConfigs().size() > 0) {
```
jiajunye, line 428: nit: empty() is slightly preferred over size() http://g3doc/devtools/cymbal/clang_tidy/g3doc/checks/readability-container-size-empty

test/process_test.cc
```diff
@@ -204,6 +204,22 @@ TEST_P(ProcessTest, NoFlushWhenCancelExecutionBeforeLoadTestBegin) {
...
+      "{name:\"nighthawk.fake_user_defined_output\",typed_config:"
+      "{\"@type\":\"type.googleapis.com/nighthawk.FakeUserDefinedOutputConfig\"}}";
+  OptionsPtr options =
+      TestUtility::createOptionsImpl(fmt::format("foo --user-defined-plugin-config {} https://{}/",
+                                                 user_defined_output_plugin, loopback_address_));
+  envoy::config::core::v3::TypedExtensionConfig typed_dns_resolver_config;
+  Envoy::Network::DnsResolverFactory& dns_resolver_factory =
+      Envoy::Network::createDefaultDnsResolverFactory(typed_dns_resolver_config);
+
+  absl::StatusOr<ProcessPtr> process_or_status = ProcessImpl::CreateProcessImpl(
```
jiajunye, line 218: nit: omit "_or_status". consider calling it process or process_ptr?

test/process_test.cc
```diff
@@ -204,6 +204,22 @@ TEST_P(ProcessTest, NoFlushWhenCancelExecutionBeforeLoadTestBegin) {
   EXPECT_EQ(numFlushes, 0);
 }
 
+TEST_P(ProcessTest, FailsIfUserDefinedOutputPluginSpecified) {
+  const std::string user_defined_output_plugin =
+      "{name:\"nighthawk.fake_user_defined_output\",typed_config:"
+      "{\"@type\":\"type.googleapis.com/nighthawk.FakeUserDefinedOutputConfig\"}}";
+  OptionsPtr options =
```
jiajunye, line 211: Would std::unique_ptr<OptionsPtr> options(TestUtility::createOptionsImpl...()) more readable?

jiajunye:
LGTM

source/client/options_impl.cc
```diff
@@ -644,6 +654,18 @@ OptionsImpl::OptionsImpl(int argc, const char* const* argv) {
       throw MalformedArgvException(e.what());
     }
   }
+  if (!user_defined_output_plugin_configs.getValue().empty()) {
+    for (const std::string& plugin_config_string : user_defined_output_plugin_configs.getValue()) {
```
mum4k, line 658: Can we add unit tests (or TODOs to add them later) that:

- verify the behavior where the user supplied config does parse, but does not validate.
- verify the behavior where multiple configs are passed by the user.

mum4k (changes requested):

source/client/options_impl.cc
dubious90: Added a test for the validation failure. Added a TODO for multiple plugins, which isn't possible until we have a second kind of User Defined Output Plugin.

test/process_test.cc
jiajunye: Oops, I was about to say `std::unique_ptr<Options> options(TestUtility::createOptionsImpl...())`.

I didn't know the Envoy style is to use OptionsPtr. Your current code LGTM then.

source/client/process_impl.cc
jiajunye: I see and agreed.

source/client/options_impl.h
jiajunye: Got it. Good to know this defensive way to initialize private members. Sounds like this could apply to all types, not just non-primitive types?

mum4k (approved):

PR #927, opened 2022-10-19
Create means of calling getPerWorkerOutput for UserDefinedOutputPlugins.

dubious90:
Add apis to Client Worker and Benchmark Client to bring these through. The Client Worker method getAdditionalOutput will be called from Process.

Also fix getPerWorkerOutput to be a const method.


dubious90:
@eric846 please review and assign to @mum4k when done.

eric846 (approved):
LGTM

include/nighthawk/client/benchmark_client.h
```diff
@@ -60,9 +60,18 @@ class BenchmarkClient {
...
   virtual bool shouldMeasureLatencies() const PURE;
+
+  /**
+   * Get any additional output that this benchmark client may have stored while sending and
+   * receiving requests.
+   *
+   * @return a vector of protos. Not guaranteed for all of the Any protobufs to have the same
+   * underlying type.
+   */
+  virtual std::vector<Envoy::ProtobufWkt::Any> getAdditionalOutput() const PURE;
```
mum4k, line 71: This is named in a generic way, but the implementation only deals with the user defined output plugins.

Can we make that more obvious in the function name, e.g. by calling it `getUserDefinedOutputPluginsResults`.

I know this is a mouthful, but it gives the reader a better idea what this additional output is. Open to other naming suggestions.

Other names to consider would be names that would closely resemble the counter name we are adding int this PR: `user_defined_plugin_per_worker_output_failure`.

mum4k (changes requested):

include/nighthawk/client/benchmark_client.h
dubious90: Do we consider our abstract class and implementations tightly coupled?  I agree that in the implementation, getUserDefinedOutputPluginResults is the best name, but there's nothing in the abstract class that necessitates the benchmark client implementation use the plugin system I've made.

I'm open to changing it as you suggest, but just want to give my reasoning for why I did it this way first. I very intentionally didn't mention the user defined output plugin framework except in the implementation .h file

mum4k: Thank you for explaining your reasoning. On a private API like this one I would recommend prioritizing readability (easy of code comprehension) over generality (accounting for future theoretical use cases).

If other use cases come up later, we can decide then if we rename the methods, add new methods or otherwise.

mum4k (approved):

PR #929, opened 2022-10-24
Pass around plugin names along with plugins and output

dubious90:
This change does the following three broad things:
1. Passes plugin names whenever the plugin pointers are passed.
2. getUserDefinedOutputResults now returns UserDefinedOutput instead of Any, so that it can include the name of the plugin the config goes with
3. created using statements for a couple std::pairs that are long and used often


dubious90:
Hey @qqustc please review and assign to @mum4k when done.

qqustc (approved):

mum4k (approved):

PR #930, opened 2022-10-27
Add user defined outputs to the nighthawk::Result.

dubious90:
Modifies output collector to accept a new parameter, a vector of UserDefinedOutputs, which it will add onto the Result.


dubious90:
Hi @fei-deng please review and assign to @mum4k when done.

fei-deng (dismissed):
LGTM

api/client/output.proto
```diff
@@ -63,8 +63,8 @@ message Result {
   google.protobuf.Duration execution_duration = 4;
   google.protobuf.Timestamp execution_start = 5;
 
-  // Outputs generated by User Defined Output Plugins. Plugin outputs will be returned in the same
-  // order the user_defined_plugin_configs are provided in the request.
+  // Outputs generated by User Defined Output Plugins. Plugin outputs are not guaranteed to be
```
mum4k, line 66: Why did we decide to change this?

And if the order isn't guaranteed, how will a user of Nighthawk map the outputs to the respective plugins?

dubious90: 2 Reasons to change. but first answering your other question -- there are actually two better ways to map the output to their plugin:
- UserDefinedOutput.plugin_name
- UserDefinedOutput.typed_output is an any, so it has its proto's type as well.

It is unnecessary for users to keep track of order, and I don't think we should make them think that it is a good idea

The main reason to change
If AggregateGlobalOutput fails, we don't provide any User Defined Output, so not only does the order not match, we can't even guarantee the number of plugins represented matches. There's actually a similar problem with the per worker outputs.

Second weaker reason:
In code that is coming up and was originally a part of this PR before I split it into two, I'm introducing a HashMap to allow combining all of the plugins together into the global output. Is it actually worth using an ordered map to guarantee this?  If this was the only barrier, I'd probably keep the requirement and go for it, but given the existing problem, I'm less inclined to do so.

mum4k: Thank you for explaining, that sounds good. Can we extend the comment in the proto indicating to users how to correctly map outputs to plugins, since they should not depend on the order? This is likely the place where they would be looking for such information.

dubious90: Added a comment about using the UserDefinedOutput.plugin_name

mum4k (approved):

PR #934, opened 2022-10-28
Update python dependencies to new major versions

dubious90:
Integration tests have been failing with an error that looks like:

```
2022-10-28T16:21:42.2524307Z INTERNALERROR>     from py.log import Producer
2022-10-28T16:21:42.2525457Z INTERNALERROR> ModuleNotFoundError: No module named 'py.log'; 'py' is not a package
```

Updating dependencies and adding in a dependency on `py` appears to address this failure.


eric846 (approved):
LGTM

PR #936, opened 2022-10-31
Update Envoy to 6b7c95d.

dubious90:
Specifics:
- gen_compilation_database updated, but only in ways that we override
- https://github.com/envoyproxy/envoy/commit/ba81ae537e722ad99906599f5d3ad367ca854433 updates rules_python to 0.13.0
  - Deprecates pip_install
  - Requires pinning of everything in requirements.txt
  - Now requires us to now call install_deps when workspace is loaded.

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -90,7 +90,7 @@ function do_integration_test_coverage() {
     export TEST_TARGETS="//test:python_test"
     # TODO(#830): Raise the integration test coverage.
     # TODO(nbperry): Raise back to 73 when User Defined Output plugin completed
-    export COVERAGE_THRESHOLD=72.2
+    export COVERAGE_THRESHOLD=72.1
     echo "bazel coverage build with tests ${TEST_TARGETS}"
     test/run_nighthawk_bazel_coverage.sh ${TEST_TARGETS}
     exit 0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,26 @@
-# Last updated 2022-10-03
-apipkg>=3.0.1, <4.0.0
-chardet>=5.0.0, <6.0.0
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,13 +1,26 @@
...
-more_itertools>=9.0.0, <10.0.0
-py>=1.11.0, <2.0.0
-pytest>=7.1.2, <8.0.0
-pytest-dependency>=0.5.1, <0.6.0 # versions below 1.0.0 are unstable, so we should only take in bugfixes automatically
-pytest-xdist>=3.0.2, <4.0.0
-pyyaml>=6.0.0, <7.0.0
-requests>=2.28.1, <3.0.0
-six>=1.16.0, <2.0.0
-zipp>=3.8.1, <4.0.0
+# Last updated 2022-10-31
```
mum4k, line 1: Does this change or affect our documented update procedure for requirements.txt ?

https://github.com/envoyproxy/nighthawk/blob/e610970a3ac4bbbb8e6cdacba2cd289af3110879/MAINTAINERS.md?plain=1#L94

ci/do_ci.sh
```diff
@@ -90,7 +90,7 @@ function do_integration_test_coverage() {
     export TEST_TARGETS="//test:python_test"
     # TODO(#830): Raise the integration test coverage.
     # TODO(nbperry): Raise back to 73 when User Defined Output plugin completed
-    export COVERAGE_THRESHOLD=72.2
+    export COVERAGE_THRESHOLD=72.1
```
mum4k, line 93: Do we understand why this is necessary? This PR doesn't seem to be changing the coverage.

mum4k (changes requested):

mum4k (approved):

PR #937, opened 2022-11-04
Minor fix: Correct int size for threshold, clarify readme.

KBaichoo:

In the README, proto it accepts uint64_t but we were using uint32_t for parsing -> out of range error for valid values


dubious90 (dismissed):
Looks good, but can you run:
```
tools/update_cli_readme_documentation.sh --mode fix
```

This will make the CI green so we can submit

dubious90 (approved):

PR #938, opened 2022-11-05
Log and re-throw on NigthhawkException same as on EnvoyException.

mum4k:
Our process code can throw NigthhawkException (e.g. incorrect request source config). The code currently doesn't log any information about it. This PR adds a logging statement for these cases.


mum4k:
@fei-deng please review and assign to @dubious90 when done.

fei-deng (approved):
LGTM

PR #942, opened 2022-11-11
Change signature for createUserDefinedOutputPlugin to allow validation and prevent unnecessary casting

dubious90:
Originally we were under a mistaken impression that this function had to have a specific signature, but we actually have full control over it.  Removing the unnecessary dynamic_cast from all of the functions, and allowing absl::StatusOr for up front config validation.

This PR also modifies the users of these methods to use the validation appropriately.


dubious90:
@qqustc please review and assign to @mum4k when done.

qqustc (dismissed):

source/user_defined_output/log_response_headers_plugin.cc
```diff
@@ -128,13 +122,16 @@ Envoy::ProtobufTypes::MessagePtr LogResponseHeadersPluginFactory::createEmptyCon
...
 }
 
-UserDefinedOutputPluginPtr LogResponseHeadersPluginFactory::createUserDefinedOutputPlugin(
-    const Envoy::Protobuf::Message& message, const WorkerMetadata& worker_metadata) {
-  const auto& any = dynamic_cast<const Envoy::ProtobufWkt::Any&>(message);
+absl::StatusOr<UserDefinedOutputPluginPtr>
+LogResponseHeadersPluginFactory::createUserDefinedOutputPlugin(
+    const Envoy::ProtobufWkt::Any& message, const WorkerMetadata& worker_metadata) {
   LogResponseHeadersConfig config;
+  Envoy::MessageUtil::unpackTo(message, config);
```
mum4k, line 129: Would `MessageUtil::unpackToNoThrow` be a better match now, since it returns status instead of throwing?

source/user_defined_output/user_defined_output_plugin_creator.cc
```diff
@@ -19,10 +19,16 @@ std::vector<UserDefinedOutputNamePluginPair> createUserDefinedOutputPlugins(
...
+    if (!plugin.ok()) {
+      return plugin.status();
+    }
+
     UserDefinedOutputNamePluginPair name_plugin_pair;
     name_plugin_pair.first = factory->name();
-    name_plugin_pair.second =
-        factory->createUserDefinedOutputPlugin(config.typed_config(), metadata);
+    name_plugin_pair.second = std::move(*plugin);
+    ;
```
mum4k, line 31: Looks like this is an extra `;`.

mum4k (changes requested):

mum4k (approved):

PR #945, opened 2022-11-18
Upload coverage report as a pipeline artifact.

mum4k:
The PR splits the coverage AZP stage into two, so that we can publish two artifacts with unique names.

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -4,102 +4,101 @@ trigger:
     - "main"
 
 stages:
-#- stage: check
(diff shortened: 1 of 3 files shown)


mum4k:
@tomjzzhang please review and assign to dubious90 when done.

tomjzzhang (approved):
LGTM.

dubious90 (approved):
Very excited for this. Downloaded a sample coverage report, and there's some weird stuff I don't fully understand yet, but I'll assume for now that it's an actual coverage problem rather than a problem with the report storing.

PR #948, opened 2022-11-28
Create full integration of User Defined Output Plugin Framework

dubious90:
- Process now calls AggregateGlobalOutput
- Process no longer rejects User Defined Output configurations with Unimplemented
- Full integration tests for User Defined Output framework
  - Requires moving to a separate testonly binary
  - Restore coverage threshold to 73


dubious90:
@chipmunkie could you please review and assign to @mum4k when done?

test/integration/test_user_defined_output_plugins.py
```diff
@@ -0,0 +1,185 @@
...
+      "--termination-predicate", "benchmark.http_2xx:24", "--rps", "100",
+      "--user-defined-plugin-config", fake_plugin_config
+  ])
+  counters = http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
+  asserts.assertCounterGreaterEqual(counters, "benchmark.http_2xx", 50)
+
+  user_defined_outputs_by_result = getUserDefinedOutputsFromJson(parsed_json)
+  asserts.assertEqual(len(user_defined_outputs_by_result), 3)
+
+  for result_name in user_defined_outputs_by_result:
```
chipmunkie, line 43: Optional: 
The key and value can be retrieved at the same time from a dictionary if so desired.
  _for k, v in dict.items():
     ..._

Applies below as well.

test/integration/test_user_defined_output_plugins.py
```diff
@@ -0,0 +1,185 @@
...
+    user_defined_outputs = user_defined_outputs_by_result[result_name]
+    asserts.assertEqual(len(user_defined_outputs), 1)
+    user_defined_output = user_defined_outputs[0]
+    asserts.assertEqual(user_defined_output["plugin_name"], "nighthawk.fake_user_defined_output")
+    if result_name == "global":
```
chipmunkie, line 48: These checks could be condensed to (and easier to follow):
      _asserts.assertGreaterEqual(user_defined_output["typed_output"]["data_called"], expected_output)
      asserts.assertGreaterEqual(user_defined_output["typed_output"]["headers_called"], expected_output)_ 
if we defined _expected_output_ somewhere above based on _result_name_.

test/integration/test_user_defined_output_plugins.py
```diff
@@ -0,0 +1,185 @@
...
+  ])
+  counters = http_test_server_fixture.getNighthawkCounterMapFromJson(parsed_json)
+  asserts.assertCounterGreaterEqual(counters, "benchmark.http_2xx", 25)
+
+  user_defined_outputs_by_result = getUserDefinedOutputsFromJson(parsed_json)
+  asserts.assertEqual(len(user_defined_outputs_by_result), 1)
+
+  global_user_defined_outputs = user_defined_outputs_by_result['global']
+  asserts.assertEqual(len(global_user_defined_outputs), 2)
+  for user_defined_output in global_user_defined_outputs:
```
chipmunkie, line 82: Same comment as above on iterating over dictionaries.

test/integration/test_user_defined_output_plugins.py
```diff
@@ -0,0 +1,185 @@
...
+  asserts.assertEqual(len(user_defined_outputs_by_result), 1)
+
+  global_user_defined_outputs = user_defined_outputs_by_result["global"]
+  asserts.assertEqual(len(global_user_defined_outputs), 1)
+  user_defined_output = global_user_defined_outputs[0]
+  asserts.assertEqual(user_defined_output["plugin_name"], "nighthawk.fake_user_defined_output")
+  asserts.assertGreaterEqual(user_defined_output["typed_output"]["data_called"], 20)
+  asserts.assertGreaterEqual(user_defined_output["typed_output"]["headers_called"], 20)
+
+  expected_failure_counter = user_defined_output["typed_output"]["headers_called"] - 9
```
chipmunkie, line 119: Optional:
Should this operation for calculating the expected value have a comment? Deducting the value 9 feels a bit magical for someone who is unfamiliar with the codebase.

chipmunkie (approved):

test/integration/test_user_defined_output_plugins.py
dubious90: Done. Thanks for the python knowledge!

source/client/process_impl.cc
```diff
@@ -338,6 +338,71 @@ getUserDefinedFactoryConfigPairs(const Options& options) {
...
+    absl::flat_hash_map<std::string, std::vector<nighthawk::client::UserDefinedOutput>>*
+        user_defined_outputs_by_plugin,
+    const std::vector<nighthawk::client::UserDefinedOutput>& worker_user_defined_outputs) {
+  for (const nighthawk::client::UserDefinedOutput& user_defined_output :
+       worker_user_defined_outputs) {
+    std::vector<nighthawk::client::UserDefinedOutput> cross_worker_results_for_plugin =
+        (user_defined_outputs_by_plugin->contains(user_defined_output.plugin_name())
+             ? user_defined_outputs_by_plugin->find(user_defined_output.plugin_name())->second
+             : std::vector<nighthawk::client::UserDefinedOutput>{});
+    cross_worker_results_for_plugin.emplace_back(user_defined_output);
```
mum4k, line 360: Does calling `emplace_back` instead of `push_back` have any meaningful impact when passing in an object that was already created?

Assuming it doesn't, can we use `push_back` to make the code more obvious?

source/client/process_impl.cc
```diff
@@ -338,6 +338,71 @@ getUserDefinedFactoryConfigPairs(const Options& options) {
...
+ * into, maps plugin name to the set of results for that plugin.
+ * @param worker_user_defined_outputs The per worker results to collect and organize.
+ */
+void collectUserDefinedOutputs(
+    absl::flat_hash_map<std::string, std::vector<nighthawk::client::UserDefinedOutput>>*
+        user_defined_outputs_by_plugin,
+    const std::vector<nighthawk::client::UserDefinedOutput>& worker_user_defined_outputs) {
+  for (const nighthawk::client::UserDefinedOutput& user_defined_output :
+       worker_user_defined_outputs) {
+    std::vector<nighthawk::client::UserDefinedOutput> cross_worker_results_for_plugin =
```
mum4k, line 356: This is a bit long for me to easily parse the ternary expression. Would you mind expanding it into an if<->else? It will take more space horizontally, but should be easier to follow.

(optional) Consider taking the portion that finds a previous output by plugin name into its own private function. That would improve the readability here further.

source/client/process_impl.cc
```diff
@@ -338,6 +338,71 @@ getUserDefinedFactoryConfigPairs(const Options& options) {
...
+        (user_defined_outputs_by_plugin->contains(user_defined_output.plugin_name())
+             ? user_defined_outputs_by_plugin->find(user_defined_output.plugin_name())->second
+             : std::vector<nighthawk::client::UserDefinedOutput>{});
+    cross_worker_results_for_plugin.emplace_back(user_defined_output);
+    user_defined_outputs_by_plugin->insert_or_assign(user_defined_output.plugin_name(),
```
mum4k, line 361: If I understand the code correctly, we make a local copy of the vector from the user_defined_outputs_by_plugin, then push a new entry at its back. Then we assign it back to the map, throwing away the original vector it used to hold. That is going to be suboptimal.

The bigger problem is that it feels unintuitive, it took me a considerable amount of time to understand this code (assuming I got what it does).

Can we try to simplify this? If I understand what the code wants to do, something like this (pseudocode) would be more idiomatic:

```cc
if (!user_defined_outputs_by_plugin.contains(plugin_name)) {
  user_defined_outputs_by_plugin[plugin_name] = {user_defined_output}
} else {
  user_defined_outputs_by_plugin[plugin_name].push_back(user_defined_output)
}
```

source/client/process_impl.cc
```diff
@@ -338,6 +338,71 @@ getUserDefinedFactoryConfigPairs(const Options& options) {
...
+    if (it != user_defined_outputs_by_plugin.end()) {
+      absl::StatusOr<Envoy::ProtobufWkt::Any> global_output_any =
+          factory->AggregateGlobalOutput(it->second);
+      if (global_output_any.ok()) {
+        *global_output.mutable_typed_output() = *global_output_any;
+      } else {
+        *global_output.mutable_error_message() = global_output_any.status().ToString();
+      }
+    } else {
+      throw NighthawkException(
```
mum4k, line 398: Why are we choosing to throw an exception and terminate Nighthawk, instead of storing this as an error_message and allowing it to continue?

If we choose to continue, the user will get the benefit of receiving the measurements collected from the execution.

source/client/process_impl.cc
```diff
@@ -338,6 +338,71 @@ getUserDefinedFactoryConfigPairs(const Options& options) {
...
+          "No per worker outputs found for a factory when performing aggregation");
+    }
+    global_outputs.emplace_back(global_output);
```
mum4k, line 401: Same as above, `push_back` is likely to be clearer here.

mum4k (changes requested):

mum4k (approved):

Issue #950, opened 2022-11-30
In User Defined Output plugin framework, handleResponseData called twice per request


dubious90:
It is possible that it is being called on both the request path and the response path. More investigation is required here.

When called in integration tests, it is always first empty data followed by the expected response.

dubious90:
I've confirmed that we are _not_ calling this on the request path, only the response path. New theory is that this is just envoy sending response data in chunks, though more investigation is required as to why there is an empty data buffer sent ahead of each actual response.

PR #951, opened 2022-11-30
Fix issues found when creating integration tests

dubious90:
The resulting PR was too large, so splitting these fixes out from the full integration PR:
1. FakeUserDefinedOutputPlugin is now thread safe.
2. AggregateGlobalOutput takes in UserDefinedOutput so that it can account for errors
  - Previously there was no way for the global aggregate to know about missing outputs which could have created confusion for users
3. FakeUserDefinedOutput - handleResponseData now ignores empty data responses
  - We uncovered an issue where handleData gets called twice per request, once with empty data - more investigation required in https://github.com/envoyproxy/nighthawk/issues/950 
4. fake_user_defined_output plugin can no longer be test only because of a dependency chain that leads to non-test code, such as benchmarks and dynamic_config. It is too complex to break this dependency chain at the moment.


dubious90:
@fei-deng please review and assign to @mum4k when ready

fei-deng (approved):
LGTM. Nice change Nate.

test/user_defined_output/fake_plugin/fake_user_defined_output.h
```diff
@@ -49,8 +50,10 @@ class FakeUserDefinedOutputPlugin : public UserDefinedOutputPlugin {
   absl::StatusOr<Envoy::ProtobufWkt::Any> getPerWorkerOutput() const override;
 
 private:
-  int data_called_ = 0;
-  int headers_called_ = 0;
+  mutable Envoy::Thread::MutexBasicLockable data_lock_;
```
mum4k, line 53: Using multiple locks tends to be error prone (makes the code harder to understand). Assuming we don't have a good reason to do this, can we do with a single mutex instead?

test/user_defined_output/fake_plugin/fake_user_defined_output.cc
```diff
@@ -23,18 +24,25 @@ FakeUserDefinedOutputPlugin::handleResponseHeaders(const Envoy::Http::ResponseHe
   return absl::OkStatus();
 }
 
-absl::Status FakeUserDefinedOutputPlugin::handleResponseData(const Envoy::Buffer::Instance&) {
+absl::Status FakeUserDefinedOutputPlugin::handleResponseData(const Envoy::Buffer::Instance& data) {
+  Envoy::Thread::LockGuard guard(data_lock_);
+  if (data.toString().empty()) {
+    // handleResponseData seemingly gets called twice per request, once always empty, once with the
```
mum4k, line 30: Consider adding a TODO pointing to an issue if we filed any.

test/user_defined_output/fake_plugin/fake_user_defined_output.cc
```diff
@@ -69,28 +77,33 @@ FakeUserDefinedOutputPluginFactory::createUserDefinedOutputPlugin(
...
+      absl::Status status = Envoy::MessageUtil::unpackToNoThrow(any, output);
+      if (status.ok()) {
+        data_called += output.data_called();
+        headers_called += output.headers_called();
+      } else {
+        return status;
+      }
     } else {
-      return status;
+      return absl::InvalidArgumentError(
```
mum4k, line 97: Is this what we expect the real plugin to do? Or should it aggregate the error messages and return the output to make some results from the run useful?

Related - since we need to aggregate partial errors into the global error - should the proto carry a list strings (errors) instead of just a single error?

mum4k (changes requested):

mum4k (approved):

PR #978, opened 2023-03-15
Copy built binaries to a directory when running in Docker.

mum4k:
Allows for extraction of binaries built when running inside a Docker container. The `BUILD_DIR` will be mounted to a directory on the host.

Also removes an extra space in the script's help output.


chipmunkie (approved):
LGTM

PR #980, opened 2023-03-22
Use RBE when Building Nighthawk.

mum4k:
This PR only moves the `build` stage onto RBE and AZP hosted VM. The other stages remain non-RBE and self-hosted for now, since there is more work needed to make them pass with RBE.

The additional stages will be moved in future PRs.

Done here:
- brought the `setup_cache.sh` script in from Envoy.
- converted bash variables that hold build options to bash arrays to be compatible with Envoy scripts.
- configured the pipeline to accept a new `rbe` parameter that enables RBE builds when set.
- split the `format` step into its own stage, since it currently does not work on an AZP hosted VM.

### DIFF
--- a/.azure-pipelines/bazel.yml
+++ b/.azure-pipelines/bazel.yml
@@ -3,6 +3,11 @@ parameters:
   displayName: "CI target"
   type: string
   default: build
+- name: rbe
+  # If set to true, build is performed using the RBE service.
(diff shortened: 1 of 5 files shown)


mum4k:
Shelving this until we resolve current RBE issues in the Envoy CI.

PR #986, opened 2023-04-17
Update Envoy to e7543a0 (Apr 17th 2023)

tomjzzhang:
- Update the ENVOY_COMMIT and ENVOY_SHA in [bazel/repositories.bzl](https://github.com/envoyproxy/nighthawk/blob/main/bazel/repositories.bzl) to the latest Envoy's commit.
- Update [.bazelrc](https://github.com/envoyproxy/nighthawk/blob/main/.bazelrc) to match Envoy commit https://github.com/envoyproxy/envoy/pull/26776
- Update [ci/run_envoy_docker.sh](https://github.com/envoyproxy/nighthawk/blob/main/ci/run_envoy_docker.sh) to match Envoy commit https://github.com/envoyproxy/envoy/pull/26776
- Update [requirements.txt](https://github.com/envoyproxy/nighthawk/blob/main/requirements.txt]) with latest python dependencies


test/integration/integration_test_fixtures.py
```diff
@@ -154,6 +154,9 @@ def tearDown(self, caplog):
       for record in caplog.get_records(when):
         if record.levelno not in (logging.WARNING, logging.ERROR):
           continue
+        # Temporary hack until fixed upstream. Log does not need to be an error.
```
mum4k, line 157: We already have a mechanism for ignoring known messages from the test server's log here:

- https://github.com/envoyproxy/nighthawk/blob/f4de666db853b48a1f241ef2eaf263c2d0d3bac6/test/integration/nighthawk_test_server.py#L182
- https://github.com/envoyproxy/nighthawk/blob/f4de666db853b48a1f241ef2eaf263c2d0d3bac6/test/integration/nighthawk_test_server.py#L44-L83

If the test server is the source of the error, let's add it there. If it is the Nighthawk itself, can we implement a matching mechanism in the corresponding class instead?

mum4k (changes requested):
Thank you @tomjzzhang. An excellent and clear PR description!

test/integration/integration_test_fixtures.py
tomjzzhang: The problem has been fixed upstream so I opted to avoid this all together.

mum4k (approved):

PR #991, opened 2023-05-19
Implement NighthawkServerInstance::statsConfig

pradeepcrao:
The deferred initialisation of stats (https://github.com/envoyproxy/envoy/pull/23921) requires server instances to always implement `statsConfig()`. This PR adds a dummy implementation for `NighthawkServerInstance`.

Tested with `Bazel test //test/...`

### DIFF
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -105,6 +105,24 @@ public:
   }
 };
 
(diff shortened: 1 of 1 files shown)


mum4k (approved):
Thank you @pradeepcrao !

PR #1004, opened 2023-07-12
Update Envoy to 77e5189 (Jul 10th 2023).

chipmunkie:
Updates Envoy to commit  77e5189. 
This is the last good commit before 65273b2 which causes unresolved linker errors  for which debugging is ongoing.

Also, preventively fixes test/output_formatter_test.cc which is broken by commit 65273b2 due to protobuf field reordering.


mum4k:
@fei-deng please review and assign back to me once done.

fei-deng (approved):

mum4k (approved):

PR #1005, opened 2023-07-14
Temporarily disable coverage_integration test.

chipmunkie:


chipmunkie:
https://github.com/envoyproxy/envoy/pull/28075 introduces linking errors  caused by an increased binary size when executing the converage_integration test. 
We are temporarily disabling this check to unblock the following Nighthawk updates until a solution is checked in.

ci/do_ci.sh
```diff
@@ -314,11 +314,11 @@ case "$1" in
         do_unit_test_coverage
         exit 0
     ;;
-    coverage_integration)
-        setup_clang_toolchain
-        do_integration_test_coverage
-        exit 0
-    ;;
+#    coverage_integration)
```
mum4k, line 317: Commenting it out here will likely make CI fail, since it will still execute this step, but the function won't be defined.

Let's comment it out in the pipeline instead:
https://github.com/envoyproxy/nighthawk/blob/6a05300799498b06e57969ba149e898d1c262573/.azure-pipelines/pipelines.yml#L123

And let's also add a comment there, linking to the issue you filed.

mum4k (changes requested):

mum4k (approved):

PR #1014, opened 2023-08-22
CI Bisect for nighthawk/pull/1013

tomjzzhang:
Benchmarks are failing on the azure pipelines. Setting up a simple PR to validate they can still pass on those machines.


tomjzzhang:
Issue appears to be a flake probably with Azure infrastructure. No need to bisect anymore closing. Noting --experimental_ui_max_stdouterr_bytes=-1 as a useful knob for seeing test outputs in CI.

PR #1025, opened 2023-09-25
Update Envoy to 91fd294 (Sep 25, 2023)

tomjzzhang:
- Update the ENVOY_COMMIT and ENVOY_SHA in bazel/repositories.bzl to the latest Envoy's commit.
- Update to .bazelrc to https://github.com/envoyproxy/envoy/pull/29732
- Update ci/run_envoy_docker.sh to https://github.com/envoyproxy/envoy/pull/29772
- Update requirements.txt to up to date dependencies
- Update source/client/process_impl.cc to support pure virtual function change introduced in https://github.com/envoyproxy/envoy/pull/29173
- Update tools/code_format/config.yaml to https://github.com/envoyproxy/envoy/pull/29497

### DIFF
--- a/ci/run_envoy_docker.sh
+++ b/ci/run_envoy_docker.sh
@@ -117,8 +117,6 @@ fi
 docker run --rm \
        "${ENVOY_DOCKER_OPTIONS[@]}" \
        "${VOLUMES[@]}" \
-       -e AZP_BRANCH \
-       -e AZP_SHA1 `# unique` \
        -e HTTP_PROXY \
        -e HTTPS_PROXY \
        -e NO_PROXY \
@@ -129,6 +127,8 @@ docker run --rm \
        -e BAZEL_FAKE_SCM_REVISION \
        -e BAZEL_REMOTE_CACHE \
        -e BAZEL_STARTUP_EXTRA_OPTIONS \
+       -e CI_BRANCH \
+       -e CI_SHA1 \
        -e CI_TARGET_BRANCH \
        -e DOCKERHUB_USERNAME \
        -e DOCKERHUB_PASSWORD \
@@ -150,6 +150,7 @@ docker run --rm \
        -e ENVOY_HEAD_REF \
        -e ENVOY_PUBLISH_DRY_RUN \
        -e ENVOY_REPO \
+       -e ENVOY_TARBALL_DIR \
        -e SYSTEM_PULLREQUEST_PULLREQUESTNUMBER \
        -e GCS_ARTIFACT_BUCKET \
        -e GITHUB_REF_NAME \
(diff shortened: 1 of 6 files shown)


ci/run_envoy_docker.sh
```diff
@@ -117,8 +117,6 @@ fi
 docker run --rm \
        "${ENVOY_DOCKER_OPTIONS[@]}" \
        "${VOLUMES[@]}" \
-       -e AZP_BRANCH \
-       -e AZP_SHA1 `# unique` \
```
mum4k, line 121: I believe our docker push depends on this variable:

https://github.com/envoyproxy/nighthawk/blob/9e68bd31dfeaf71fd47f6921d4a6b2cd0b74a08f/.azure-pipelines/bazel.yml#L19
https://github.com/envoyproxy/nighthawk/blob/9e68bd31dfeaf71fd47f6921d4a6b2cd0b74a08f/ci/docker/docker_azp_push.sh#L25

Are we removing it by accident, or do we need to make some changes to support it again?

ci/run_envoy_docker.sh
```diff
@@ -117,8 +117,6 @@ fi
 docker run --rm \
        "${ENVOY_DOCKER_OPTIONS[@]}" \
        "${VOLUMES[@]}" \
-       -e AZP_BRANCH \
```
mum4k, line 120: We depend on AZP_BRANCH, which is used to determine if we should skip some tests:

https://github.com/envoyproxy/nighthawk/blob/9e68bd31dfeaf71fd47f6921d4a6b2cd0b74a08f/ci/do_ci.sh#L116

https://github.com/envoyproxy/nighthawk/blob/9e68bd31dfeaf71fd47f6921d4a6b2cd0b74a08f/test/integration/utility.py#L113

https://github.com/envoyproxy/nighthawk/blob/9e68bd31dfeaf71fd47f6921d4a6b2cd0b74a08f/test/integration/test_integration_basics.py#L146

mum4k (changes requested):

mum4k:

ci/run_envoy_docker.sh
tomjzzhang: Good catch.  This was removed in https://github.com/envoyproxy/envoy/pull/29772 so I copied it over but it involves some additional CI changes. In particular, we need to understand the intention of https://github.com/envoyproxy/envoy/pull/29646. To unblock the import I've kept the variables as I look on how to integrate similar changes.

ci/run_envoy_docker.sh
tomjzzhang: See above response.

eric846 (approved):
LGTM

mum4k (approved):

PR #1026, opened 2023-09-29
Update Envoy to 7479b6c (Sep 29, 2023)

fei-deng:
- Update the ENVOY_COMMIT and ENVOY_SHA in bazel/repositories.bzl to the latest Envoy's commit.
- Update to .bazelrc to https://github.com/envoyproxy/envoy/pull/29874
- Update ci/run_envoy_docker.sh to https://github.com/envoyproxy/envoy/pull/29811
- Update source/client/process_impl.cc to adopt deprecation of `Envoy::Http::Context& httpContext()` in envoy ServerFactoryContext introduced in https://github.com/envoyproxy/envoy/pull/29779
- Update source/adaptive_load/adaptive_load_client_main.cc and a few tests which use `fileReadToEnd` to adopt return type change of `fileReadToEnd` introduced in https://github.com/envoyproxy/envoy/pull/29700

### DIFF
--- a/source/adaptive_load/adaptive_load_client_main.cc
+++ b/source/adaptive_load/adaptive_load_client_main.cc
@@ -101,15 +101,14 @@ AdaptiveLoadClientMain::AdaptiveLoadClientMain(int argc, const char* const* argv
 
 uint32_t AdaptiveLoadClientMain::Run() {
   ENVOY_LOG(info, "Attempting adaptive load session: {}", DescribeInputs());
-  absl::StatusOr<std::string> spec_textproto_or_error = filesystem_.fileReadToEnd(spec_filename_);
-  if (!spec_textproto_or_error.ok()) {
+  absl::StatusOr<std::string> spec_textproto = filesystem_.fileReadToEnd(spec_filename_);
+  if (!spec_textproto.ok()) {
     throw Nighthawk::NighthawkException("Failed to read spec textproto file \"" + spec_filename_ +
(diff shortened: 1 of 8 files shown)


source/adaptive_load/adaptive_load_client_main.cc
```diff
@@ -101,13 +101,13 @@ AdaptiveLoadClientMain::AdaptiveLoadClientMain(int argc, const char* const* argv
 
 uint32_t AdaptiveLoadClientMain::Run() {
   ENVOY_LOG(info, "Attempting adaptive load session: {}", DescribeInputs());
-  std::string spec_textproto;
-  try {
-    spec_textproto = filesystem_.fileReadToEnd(spec_filename_);
-  } catch (const Envoy::EnvoyException& e) {
+  absl::StatusOr<std::string> spec_textproto_or_error = filesystem_.fileReadToEnd(spec_filename_);
```
eric846, line 104: nit: the StatusOr variable should just be named `spec_textproto`, and on line 112 we can just say `*spec_textproto`. 

(However, we can just keep `.value()` in the unit tests because `.value()` either crashes or throws an exception if the `StatusOr` is not OK, whereas `*` is undefined behavior when not OK.)

eric846 (changes requested):

eric846 (approved):
LGTM

source/adaptive_load/adaptive_load_client_main.cc
fei-deng: thank you. feedback addressed.

Issue #1032, opened 2023-10-23
Update Nighthawk's use of Python to Hermetic Python


mum4k:
The format check was disabled in https://github.com/envoyproxy/nighthawk/pull/1030, because it depends on host Python that is no longer available in the Docker image (since https://github.com/envoyproxy/envoy-build-tools/pull/290).

Additionally the check hasn't been really running on the NH codebase since the --path argument was added here: https://github.com/envoyproxy/envoy/pull/29397.

What needs to happen:
- switch the python tools to hermetic python
- reformat all files
- re-enable the CI format check

mum4k:
More than just the format check is affected here:

https://dev.azure.com/cncf/envoy/_build/results?buildId=153576&view=results

None of our Python targets can execute on the new Docker build. Instead of disabling the format check I am freezing the Docker image to the last one that still had host Python.

We need to do all mentioned above and then advance to the latest docker image.

PR #1033, opened 2023-10-25
Moving our Python tools from on-host Python to Hermetic Python.

mum4k:
- updated the bazel `WORKSPACE` to use hermetic Python.
- bazel-ified `format_python_tools` and `flake8` executions.
- migrated our Python requirements management to using a lockfile. Not updating
  the requirements in this PR to keep it small, will do in a follow-up. Just merging all
  the requirements and still using `==` pins in this PR.
- removing the pin on the `envoyproxy/envoy-build-ubuntu`, we are once again
  using the same version as Envoy.
- disabling our `clang-tidy` CI check. It doesn't work with the new Docker
  image and hermetic Python. Envoy is working on a new solution upstream which
  we will adopt once available.
- passing the correct directory to the `check_format` and `format_python_tools`
  script, so that it actually checks our files for formatting. This was a
  pre-existing issue which resulted in multiple format violations. Making the
  format check information-only at this PR. Will fix formatting of all files in
  a follow-up.
- rewrote instructions on how to update Python dependencies. These documents
  will be updated once more in a follow-up PR once I remove pins (`==`) from
  `requirements.in`.

Works on #1032


.bazelrc
```diff
@@ -527,8 +526,7 @@ build:rbe-envoy-engflow --grpc_keepalive_time=30s
 build:rbe-envoy-engflow --remote_timeout=3600s
 build:rbe-envoy-engflow --bes_timeout=3600s
 build:rbe-envoy-engflow --bes_upload_mode=fully_async
-# TODO(#1032): Change to the latest. Don't update the hash until the bug is fixed. # unique
-build:rbe-envoy-engflow --remote_default_exec_properties=container-image=docker://docker.io/envoyproxy/envoy-build-ubuntu:94e5d873c145ae86f205117e76276161c9af4806@sha256:8d3763e19d5b71fdc95666d75073ce4581e566ce28ca09106607b6a3ef7ba902
+build:rbe-envoy-engflow --remote_default_exec_properties=container-image=docker://docker.io/envoyproxy/envoy-build-ubuntu:fdd65c6270a8507a18d5acd6cf19a18cb695e4fa@sha256:3c8a3ce6f90dcfb5d09dc8f79bb01404d3526d420061f9a176e0a8e91e1e573e
```
phlax, line 529: im guessing you dont need this - otherwise, lmk as it would require some setup

bazel/python_dependencies.bzl
```diff
@@ -0,0 +1,10 @@
+load("@rules_python//python:pip.bzl", "pip_parse")
+load("@python3_11//:defs.bzl", "interpreter")
+
+def nighthawk_python_dependencies():
+    pip_parse(
+        name = "python_pip_deps",
```
phlax, line 6:
```suggestion
        name = "nh_pip_deps",
```

our pipset is named `@base_pip3` which is meaningless externally, so im thinking to rename -> `@envoy_pip3`

docs/root/updating_envoy_dependency.md
```diff
@@ -283,62 +281,28 @@ head -1 requirements.txt
...
+[tools/base/requirements.in](/tools/base/requirements.in). This file specifies
+version constraints and it is our goal to use the latest but still compatible
+compatible version of every package. Ideally all constraints are in the `>=`
+format. If an incompatibility is found, you can pin a package by specifying a
+`<=` constraint. These should always be accompanied with a comment explaining
+them. Avoid using `==` constraint to the extent possible.
+
+First attempt to remove all existing pins in
+[tools/base/requirements.in](/tools/base/requirements.in) to see if they are
+still necessary. Once done editting
```
phlax, line 295:
```suggestion
still necessary. Once done editing
```

tools/BUILD
```diff
@@ -16,3 +18,23 @@ py_binary(
...
+    data = [
+        ":.style.yapf",
+    ],
+    deps = [
+        requirement("yapf"),
+    ],
+)
+
+py_binary(
+    name = "flake8",
```
phlax, line 34: you can use an `entry_point` for this

tools/base/BUILD
```diff
@@ -0,0 +1,16 @@
...
+
+envoy_package()
+
+compile_pip_requirements(
+    name = "requirements",
+    extra_args = [
+        "--allow-unsafe",
+        "--generate-hashes",
+        "--reuse-hashes",
+        "--resolver=backtracking",
```
phlax, line 14: some of these might not be necessary - we have them in envoy but i think they might duplicate

tools/flake8.py
```diff
@@ -0,0 +1,12 @@
...
+Exists, so that we can define a Bazel py_binary target that combines flake8 and
+any flake8 plugins we might need.
+
+"""
+import flake8.main.cli
+
+import sys
+
+if __name__ == '__main__':
+  sys.exit(flake8.main.cli.main())
```
phlax, line 12: you seem to be having intendation issues here 8/

phlax: argh - just realized the rest has google indents

**request changes!!!**

tools/format_python_tools.py
```diff
@@ -12,19 +12,20 @@
...
     any directories in the EXCLUDE_DIRECTORIES constant.
   """
   # TODO: Add ability to collect a specific file or files.
   matches = []
   path_parts = os.getcwd().split('/')
-  dirname = '.'
-  if path_parts[-1] == 'tools':
-    dirname = '/'.join(path_parts[:-1])
-  for root, dirnames, filenames in os.walk(dirname):
+  for root, dirnames, filenames in os.walk(directory):
```
phlax, line 28: as a general point `pathlib` is a nicer/more explicit interface to work with

i resisted for a long time being old-school, but it really does improve things

phlax (commented):
hopefully helpful review, comments inline ...

docs/root/updating_envoy_dependency.md
```diff
@@ -283,62 +281,28 @@ head -1 requirements.txt
...
-```
-
-```bash
-vi requirements.txt
-```
+The Python dependencies need to be updated regularly. The list of packages
+the Nighthawk codebase uses is listed in
+[tools/base/requirements.in](/tools/base/requirements.in). This file specifies
+version constraints and it is our goal to use the latest but still compatible
+compatible version of every package. Ideally all constraints are in the `>=`
```
fei-deng, line 288:
```suggestion
version of every package. Ideally all constraints are in the `>=`
```

fei-deng (approved):

tools/BUILD
mum4k: I cannot, there is no way of including the plugin `flake8-docstrings` with an `entry_point`. This is the only workaround I found.

.bazelrc
mum4k: It is likely we don't need everything in the .bazelrc here, but for ease of maintenance we sync it from Envoy in its entirety. Otherwise we diverge too much and increase maintenance cost.

tools/BUILD
phlax: fwiw - we have `envoy_entry_point` to workaround this specific issue

https://github.com/bazelbuild/rules_python/issues/600

tools/format_python_tools.py
mum4k: I agree. I don't think it is important to change it in this PR though. Please let me know if you disagree.

phlax: no - totally general point

bazel/python_dependencies.bzl
phlax: i think this is worth address - probs `@nh_pip3` would be better

tools/flake8.py
mum4k: Ignoring and walking away slightly entertained.

tools/base/BUILD
mum4k: Removed `--allow-unsafe` and `--resolver=backtracking` which are both the default now according to their docs:

https://pip-tools.readthedocs.io/en/latest/#deprecations

Done in 9a7a196db825aa10991f122512b5b3661eb0a395.

bazel/python_dependencies.bzl
mum4k: Thank you, done in 6349726d8423ceee216164ed7e411c5beae9b377.

mum4k:
@phlax PTAL

phlax (approved):
lgtm, thanks @mum4k

PR #1034, opened 2023-10-26
Delete requirementx.txt that is no longer in use.

mum4k:
Not used since https://github.com/envoyproxy/nighthawk/pull/1033.


chipmunkie (approved):
LGTM

PR #1035, opened 2023-10-27
Updating Python dependencies.

### LINKED ISSUE #935: Use a dependency tool for allowing us to not force pin dependencies manually
`rules_python` is now forcing everything in requirements.txt to be pinned to one version. From their [readme](https://github.com/bazelbuild/rules_python#installing-third_party-packages):

>Note: The pip_install rule is deprecated. pip_parse offers identical functionality and both pip_install and pip_parse now have the same implementation. The name pip_install may be removed in a future version of the rules. The maintainers have taken all reasonable efforts to faciliate a smooth transition, but some users of pip_install will need to replace their existing requirements.txt with a fully resolved set of dependencies using a tool such as pip-tools or the compile_pip_requirements repository rule.

`compile_pip_requirements` is their own solution, while `pip-tools` appears to be a separate pip repository.

Note that envoy appears to use a third tool pip-compile: https://github.com/envoyproxy/envoy/blob/main/tools/base/requirements.txt

### LINKED ISSUE #1032: Update Nighthawk's use of Python to Hermetic Python

mum4k:
- removing all unnecessary version pins (which happens to be all the pins).
- updating the lockfile by running `bazel run //tools/base:requirements.update`.

Fixes #1032 
Fixes #935

### DIFF
--- a/tools/base/requirements.in
+++ b/tools/base/requirements.in
@@ -1,35 +1,23 @@
-# Last updated 2023-10-20
-apipkg==3.0.2
-chardet==5.2.0
-flake8-docstrings==1.7.0
-flake8==6.1.0
-importlib_metadata==6.8.0
-mccabe==0.7.0
-more_itertools==10.1.0
-platformdirs==3.11.0
-py==1.11.0
-pycodestyle==2.11.1
-pydocstyle==6.3.0
-pyflakes==3.1.0
-pytest-dependency==0.5.1 # versions below 1.0.0 are unstable, so we should only take in bugfixes automatically
-pytest-xdist==3.3.1
-pytest==7.4.2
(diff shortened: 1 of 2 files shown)


mum4k:
@eric846 please review and merge if all good.

eric846 (approved):

PR #1059, opened 2024-01-05
Fix procedure for python dependency updates

chipmunkie:
Our current python dependency update command doesn't update the requirements.txt file when the file has contents.
Removing the contents allows for the it to be written with updated dependencies.

### DIFF
--- a/docs/root/updating_envoy_dependency.md
+++ b/docs/root/updating_envoy_dependency.md
@@ -293,7 +293,8 @@ These should always be accompanied with a comment explaining them. Avoid using
 First attempt to remove all existing pins in
 [tools/base/requirements.in](/tools/base/requirements.in) to see if they are
 still necessary. Once done editing
-[tools/base/requirements.in](/tools/base/requirements.in), update the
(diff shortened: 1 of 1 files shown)


docs/root/updating_envoy_dependency.md
```diff
@@ -293,7 +293,8 @@ These should always be accompanied with a comment explaining them. Avoid using
 First attempt to remove all existing pins in
 [tools/base/requirements.in](/tools/base/requirements.in) to see if they are
 still necessary. Once done editing
-[tools/base/requirements.in](/tools/base/requirements.in), update the
+[tools/base/requirements.in](/tools/base/requirements.in), delete the contents
+of [tools/base/requirements.txt](/tools/base/requirements.txt) and update the
 dependencies by running:
 
 ```bash
```
mum4k, line 300: Let's also add the command that deletes the content:

```
echo > tools/base/requirements.txt
```

chipmunkie: Done.

mum4k (approved):

PR #1078, opened 2024-02-06
Migrate Nighthawk off the proto DebugString method.

mum4k:
The `DebugString` method will be deprecated shortly. Migrating to `absl::StrCat` on the proto messages instead.


mum4k:

qqustc (approved):

PR #1110, opened 2024-03-25
Update Envoy to 38c5c86 (Mar 25, 2024)

eric846:
- `.bazelrc` updated from upstream
- `tools/code_format/config.yaml` updated from upstream
- In an incredibly confusing refactor:
  - Envoy's `FaultFilterConfig` started taking a `CommonFactoryContext` instead of a `Runtime` and `TimeSource`. 
      - Updated `HttpDynamicDelayDecoderFilterConfig` to store a `CommonFactoryContext&` instead of a `TimeSource&`, so that it has a `CommonFactoryContext&` to use when constructing a `FaultFilterConfig`.
        - How to get a `CommonFactoryContext` when creating the `HttpDynamicDelayDecoderFilterConfig`? `HttpDynamicDelayDecoderFilterConfigFactory` has access to a `Envoy::Server::Configuration::FactoryContext`, which has a `serverFactoryContext()` method. `ServerFactoryContext` is a subclass of `CommonFactoryContext`.
  - Envoy has started calling `serverFactoryContext()` on the `NighthawkServerInstance`, which had a `PANIC` placeholder there. This was discovered in the integration tests.
    - Gave `NighthawkServerInstance` a reference to a `NighthawkServerFactoryContext` to return in  `serverFactoryContext()`, replacing the `PANIC` placeholder.
    - In order to remove a circular dependency, took away `NighthawkServerFactoryContext`'s reference to a  `NighthawkServerInstance`, to which it delegated many calls. Now `NighthawkServerFactoryContext` stores its own references to the underlying objects formerly wrapped in the `NighthawkServerInstance`.
  - I originally assumed the above two things were related, but it appears that both arose here by coincidence.

### DIFF
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -130,20 +130,20 @@ private:
 // when Nighthawk is running are implemented.
 class NighthawkServerInstance : public Envoy::Server::Instance {
 public:
-  NighthawkServerInstance(Envoy::OptRef<Envoy::Server::Admin> admin, Envoy::Api::Api& api,
-                          Envoy::Event::Dispatcher& dispatcher,
-                          Envoy::AccessLog::AccessLogManager& log_manager,
-                          Envoy::Server::Options& options, Envoy::Runtime::Loader& runtime,
-                          Envoy::Singleton::Manager& singleton_manager,
-                          Envoy::ThreadLocal::Instance& tls,
-                          Envoy::LocalInfo::LocalInfo& local_info,
-                          Envoy::ProtobufMessage::ProdValidationContextImpl& validation_context,
-                          Envoy::Grpc::Context& grpc_context,
-                          Envoy::Router::Context& router_context)
+  NighthawkServerInstance(
+      Envoy::OptRef<Envoy::Server::Admin> admin, Envoy::Api::Api& api,
+      Envoy::Event::Dispatcher& dispatcher, Envoy::AccessLog::AccessLogManager& log_manager,
+      Envoy::Server::Options& options, Envoy::Runtime::Loader& runtime,
+      Envoy::Singleton::Manager& singleton_manager, Envoy::ThreadLocal::Instance& tls,
+      Envoy::LocalInfo::LocalInfo& local_info,
+      Envoy::ProtobufMessage::ProdValidationContextImpl& validation_context,
+      Envoy::Grpc::Context& grpc_context, Envoy::Router::Context& router_context,
+      Envoy::Server::Configuration::ServerFactoryContext& server_factory_context)
       : admin_(admin), api_(api), dispatcher_(dispatcher), log_manager_(log_manager),
         options_(options), runtime_(runtime), singleton_manager_(singleton_manager), tls_(tls),
         local_info_(local_info), validation_context_(validation_context),
-        grpc_context_(grpc_context), router_context_(router_context) {}
+        grpc_context_(grpc_context), router_context_(router_context),
+        server_factory_context_(server_factory_context) {}
 
   void run() override { PANIC("NighthawkServerInstance::run not implemented"); }
   Envoy::OptRef<Envoy::Server::Admin> admin() override { return admin_; }
@@ -227,15 +227,8 @@ public:
   envoy::config::bootstrap::v3::Bootstrap& bootstrap() override {
     PANIC("NighthawkServerInstance::bootstrap not implemented");
   }
-  void setServerFactoryContext(
-      Envoy::Server::Configuration::ServerFactoryContext& server_factory_context) {
-    server_factory_context_ = &server_factory_context;
-  }
   Envoy::Server::Configuration::ServerFactoryContext& serverFactoryContext() override {
-    RELEASE_ASSERT(server_factory_context_ != nullptr,
-                   "Mutual pointers between NighthawkServerInstance and "
-                   "NighthawkServerFactoryContext were not correctly set");
-    return *server_factory_context_;
+    return server_factory_context_;
   }
   Envoy::Server::Configuration::TransportSocketFactoryContext&
   transportSocketFactoryContext() override {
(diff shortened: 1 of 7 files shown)


source/client/process_impl.cc
```diff
@@ -227,8 +227,15 @@ class NighthawkServerInstance : public Envoy::Server::Instance {
   envoy::config::bootstrap::v3::Bootstrap& bootstrap() override {
     PANIC("NighthawkServerInstance::bootstrap not implemented");
   }
+  void setServerFactoryContext(
```
mum4k, line 230: Having a `setFoo` method implies that one chan call it multiple times (i.e. swap the Foo instance during its lifetime).

I am assuming that we don't have such use case and also that such usecase would not be suppoted (thread safety, etc.).

Can we keep the API minimal by setting the factory context at construction time?

source/client/process_impl.cc
```diff
@@ -261,6 +268,8 @@ class NighthawkServerInstance : public Envoy::Server::Instance {
   Envoy::Grpc::Context& grpc_context_;
   Envoy::Router::Context& router_context_;
   StatsConfigImpl stats_config_;
+  // Lifetime managed by ProcessImpl
```
mum4k, line 271: Do we need to retain the raw pointer? The parent unique pointer will go out of scope at some point, making this a dangling reference.

If we can - let's aintain the same lifetime on both the nighthawk_server and the server_factory_context by limiting them both to the stack of the single function that uses them.

source/server/http_dynamic_delay_filter.h
```diff
@@ -66,10 +66,18 @@ class HttpDynamicDelayDecoderFilterConfig : public FilterConfigurationBase {
...
+   */
+  Envoy::Server::Configuration::ServerFactoryContext& server_factory_context() {
+    return server_factory_context_;
+  }
+
   /**
    * @return Envoy::TimeSource& to be used by filter instantiations associated to this.
    */
-  Envoy::TimeSource& time_source() { return time_source_; }
+  // Envoy::TimeSource& time_source() { return server_factory_context_.timeSource(); }
```
mum4k, line 80: Did you mean to leave this commented out?

mum4k (changes requested):

source/client/process_impl.cc
eric846: I don't see a way to achieve this at construction time with the current design, because the two classes both need to have references to each other.
-  The `NighthawkServerInstance` is passed to the constructor of the `NighthawkServerFactoryContext`, implying `NighthawkServerInstance` needs to be created first. 
- Now adding `setServerFactoryContext` to come back around and provide the `NighthawkServerInstance` to the `NighthawkServerFactoryContext` since we didn't have it at construction time.

What if we also have an assertion that only allows setFactoryContext to be called once? And maybe rename it `initializeFactoryContext`?

Another possibility might be an even bigger super-object that implements both `ServerFactoryContext` and `ServerInstance`.

eric846: This just in: `Envoy::Server::Configuration::ServerFactoryContext` methods are exactly a subset of `Envoy::Server::Configuration::Instance` methods. I will see what I can do -- the super-object we seek may just be `Envoy::Server::Configuration::Instance` itself.

mum4k: Good news, thank you for looking deeper.

eric846: I found a way to eliminate the circular dependency.

The job of `NighthawkServerInstance` and `NighthawkServerFactoryContext` is to return references to various Envoy objects in Nighthawk like the `Api` object.

Because of recent Envoy changes, `NighthawkServerInstance` needed to start implementing `serverFactoryContext()` instead of the `PANIC` placeholder. The easiest way to implement `serverFactoryContext()` is to store a reference to a prebuilt `ServerFactoryContext`.

Unfortunately in the existing code,  `NighthawkServerFactoryContext` was already storing a reference to `NighthawkServerInstance` and delegating a lot of method calls to it. So there would be mutual references between and  `NighthawkServerFactoryContext` and `NighthawkServerInstance`. I don't know a way to initialize two classes that mutually reference each other, without opening the door to setters and pointers.

Fortunately the delegation in `NighthawkServerFactoryContext` was unnecessary. I updated `NighthawkServerFactoryContext` to store its own references to the objects it had been getting from inside the `NighthawkServerInstance`, and eliminated its reference to the `NighthawkServerInstance`.

source/client/process_impl.cc
eric846: This has now become a reference.

source/server/http_dynamic_delay_filter.h
eric846: Now deleted this method since it's no longer called.

source/client/process_impl.cc
```diff
@@ -363,7 +362,19 @@ class NighthawkServerFactoryContext : public Envoy::Server::Configuration::Serve
...
+  Envoy::AccessLog::AccessLogManager& log_manager_;
+  Envoy::Server::Options& options_;
+  Envoy::Runtime::Loader& runtime_;
+  Envoy::Singleton::Manager& singleton_manager_;
+  Envoy::ThreadLocal::Instance& tls_;
+  Envoy::LocalInfo::LocalInfo& local_info_;
+  Envoy::ProtobufMessage::ProdValidationContextImpl& validation_context_;
+  Envoy::Grpc::Context& grpc_context_;
+  Envoy::Router::Context& router_context_;
+  StatsConfigImpl stats_config_;
```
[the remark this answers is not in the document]
eric846: 👀🦅 

- I checked using the integration tests, and it turns out one of the `statsConfig()` is never called. I changed the implementation of that one to `PANIC` and got rid of that unnecessary `stats_config_`.
- For the other `stats_config_`, the `statsConfig()` does get called. I added a comment on the field to confirm the default value from construction time is intentional.

mum4k (approved):

PR #1114, opened 2024-03-29
Update Envoy to 1f52a8e (Mar 29, 2024)

eric846:
- No changes to .bazelrc etc.
- Implemented `lifecycleNotifier()` on `NighthawkServerInstance` and `NighthawkServerFactoryContext` (H/T @eziskind for the advance warning and known-good fix)
- Implemented `regexEngine()` on `NighthawkServerFactoryContext`
- Reacted to https://github.com/envoyproxy/envoy/pull/32775
  - Renamed `unpackTo` to `unpackToOrThrow`
  - Renamed `unpackToNoThrow` to `unpackTo` (the version that returns `Status`)
- `fix_format` reordered some bzl imports and removed unused ones

### DIFF
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -129,10 +129,16 @@ private:
 
 // A fake ServerLifecycleNotifier. Because it does nothing, it's safe to just create
 // multiple instances of it, rather than manage the lifetime of a single one.
-class LifecycleNotifierImpl : public Envoy::Server::ServerLifecycleNotifier {
+class NighthawkLifecycleNotifierImpl : public Envoy::Server::ServerLifecycleNotifier {
 public:
-  HandlePtr registerCallback(Stage, StageCallback) override { return nullptr; }
-  HandlePtr registerCallback(Stage, StageCallbackWithCompletion) override { return nullptr; }
+  HandlePtr registerCallback(Stage, StageCallback) override {
+    PANIC(
+        "NighthawkLifecycleNotifierImpl::registerCallbacki(Stage, StageCallback) not implemented");
+  }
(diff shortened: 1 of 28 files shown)


source/client/process_impl.cc
```diff
@@ -126,6 +127,14 @@ class StatsConfigImpl : public Envoy::Server::Configuration::StatsConfig {
   const std::chrono::milliseconds flush_interval_;
 };
 
+// A fake ServerLifecycleNotifier. Because it does nothing, it's safe to just create
+// multiple instances of it, rather than manage the lifetime of a single one.
+class LifecycleNotifierImpl : public Envoy::Server::ServerLifecycleNotifier {
+public:
+  HandlePtr registerCallback(Stage, StageCallback) override { return nullptr; }
```
mum4k, line 134: Are these methods called in Nighthawk's lifecycle? If they are, returning a nullptr could create an unexpected issue for the future and maybe we need another solution.

If they aren't expected to be called, I would recommend we use the same approach as in many other methods that are not supposed to be called - PANIC when they get called. That way we get a much simpler error to debug, i.e. we know that a method started being called and we know where to start investigating.

eziskind: These methods should not be called so putting a PANIC in the method body sounds good.

eric846: Changed the bodies to PANIC, and the integration tests didn't have any panics.

mum4k: Thank you for looking into it.

mum4k (approved):

qqustc (approved):

PR #1118, opened 2024-04-02
Update Envoy to 690d3aa (Apr 01, 2024)

fei-deng:
- update [bazel/repositories.bzl](https://github.com/fei-deng/nighthawk/blob/update_apr/bazel/repositories.bzl) with new ENVOY_COMMIT.
- update [.bazelrc](https://github.com/fei-deng/nighthawk/blob/update_apr/.bazelrc) with explicitly set --host_action_env.
- update [tools/code_format/config.yaml](https://github.com/fei-deng/nighthawk/blob/update_apr/tools/code_format/config.yaml)


.bazelrc
```diff
@@ -40,10 +40,13 @@ build --enable_platform_specific_config
 build --action_env=CC --host_action_env=CC
 build --action_env=CXX --host_action_env=CXX
 build --action_env=LLVM_CONFIG --host_action_env=LLVM_CONFIG
-# Do not pass through PATH however.
-# It tends to have machine-specific values, such as dynamically created temp folders.
-# This would make it impossible to share remote action cache hits among machines.
-# build --action_env=PATH --host_action_env=PATH
+# Do not pass machine-specific values through PATH however.
```
mum4k, line 43: This file is kept in sync with the Envoy's version.

Let's only modify the lines we have to modify. It is ok to add additional comment. Let's also mark all lines we are modifying as unique.

mum4k (changes requested):

.bazelrc
fei-deng: Updated comment.  Now it is in sync with Envoy's version. Added lines with `#unique` marks.

mum4k (approved):

PR #1131, opened 2024-04-12
Update Envoy to 1ae6fb2 (Apr 8th, 2024)

yanavlasov:
- Adjust code to changes in the createTagProducer return value.

### DIFF
--- a/source/client/process_impl.cc
+++ b/source/client/process_impl.cc
@@ -22,7 +22,6 @@
 #include "external/envoy/source/common/common/cleanup.h"
 #include "external/envoy/source/common/common/regex.h"
 #include "external/envoy/source/common/common/statusor.h"
-#include "external/envoy/source/common/config/stats_utility.h"
 #include "external/envoy/source/common/config/utility.h"
 #include "external/envoy/source/common/event/dispatcher_impl.h"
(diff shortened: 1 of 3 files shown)


source/client/process_impl.cc
```diff
@@ -848,8 +848,14 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const UriPtr& tracing_
     // Needs to happen as early as possible (before createWorkers()) in the instantiation to preempt
     // the objects that require stats.
     if (!options_.statsSinks().empty()) {
-      store_root_.setTagProducer(
-          Envoy::Config::StatsUtility::createTagProducer(bootstrap_, envoy_options_.statsTags()));
+      auto producer_or_error = Envoy::Stats::TagProducerImpl::createTagProducer(
```
mum4k, line 851: Would you mind listing the full type here instead of the `auto`? The type isn't obvious to me from the context.

mum4k (approved):
Thank you!

source/client/process_impl.cc
yanavlasov: Done

mum4k: Thank you.

mum4k (approved):

PR #1133, opened 2024-04-15
Update Envoy to d417618 2024/04/15

qqustc:
- sync files
- add `# unique` in .bazelrc to avoid CI failure https://dev.azure.com/cncf/4684fb3d-0389-4e0b-8251-221942316e06/_apis/build/builds/168119/logs/57
- add `cluster_manager_->initialize(bootstrap_);` to fix a CI crash which is introduced by https://github.com/envoyproxy/envoy/pull/33221


qqustc:
retest/

qqustc:
retest\

.bazelrc
```diff
@@ -141,11 +142,11 @@ build:asan --copt -O1
 build:asan --copt -fno-optimize-sibling-calls
 
 # Clang ASAN/UBSAN
-build:clang-asan --config=clang
-build:clang-asan --config=asan
-build:clang-asan --linkopt -fuse-ld=lld
-build:clang-asan --linkopt --rtlib=compiler-rt
-build:clang-asan --linkopt --unwindlib=libgcc
+build:clang-asan --config=clang                                                             # unique
```
mum4k, line 145: As discussed offline, please wait for and pull https://github.com/envoyproxy/envoy/pull/33605 in. Let's keep this file as close to the Envoy's version as possible.

source/client/process_impl.cc
```diff
@@ -902,6 +902,7 @@ bool ProcessImpl::runInternal(OutputCollector& collector, const UriPtr& tracing_
     }
     ENVOY_LOG(debug, "Computed configuration: {}", absl::StrCat(bootstrap_));
     cluster_manager_ = cluster_manager_factory_->clusterManagerFromProto(bootstrap_);
+    THROW_IF_NOT_OK(cluster_manager_->initialize(bootstrap_));
```
mum4k, line 905: Let's not throw here. The pattern in this function is to handle errors in the following way:

```cc
if (!status.ok()) {
      ENVOY_LOG(error, "error message");
      return false;
}
```

mum4k (changes requested):

.bazelrc
qqustc: Thanks.

source/client/process_impl.cc
qqustc: Done

mum4k (approved):

PR #1176, opened 2024-05-31
fix the discrepant usage description

zhangbo1882:
The correct command is
nighthark-client --failure-predicate benchmark.http_4xx10:10 instead of benchmark.http_4xx,10


zhangbo1882:
@eric846， could you help review this PR?  Thanks.

eric846 (approved):
Thank you for fixing that!

PR #1223, opened 2024-09-23
Update Envoy to a93ea37 (Sep 23, 2024)

sebastianavila5:
- Updated `.bazelrc` and `tools/code_format/config.yaml`
- no changes in `ci/run_envoy_docker.sh.`, `.bazelversion`,  or `tools/gen_compilation_database.py`
- Changed external dependencies -> dependencies
- Changed `Envoy::Network::DnsResolver::ResolutionStatus::Success` -> `Envoy::Network::DnsResolver::ResolutionStatus::Completed`
- Added ignore to `internal_address_config is not configured` warning for Test Server

Signed-off-by: Sebastian Avila <sebastianavila@google.com>

### DIFF
--- a/test/integration/nighthawk_test_server.py
+++ b/test/integration/nighthawk_test_server.py
@@ -78,6 +78,7 @@ _TEST_SERVER_WARN_ERROR_IGNORE_LIST = frozenset([
 
             # Logged for normal termination, not really a warning.
             "caught ENVOY_SIGTERM",
+            "internal_address_config is not configured. The existing default behaviour will trust RFC1918 IP addresses, but this will be changed in next release. Please explictily config internal address config as the migration step.",
         ),
     ),
 ])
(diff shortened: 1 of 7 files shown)


mum4k (approved):

test/integration/nighthawk_test_server.py
```diff
@@ -78,6 +78,7 @@ class _TestCaseWarnErrorIgnoreList(
 
             # Logged for normal termination, not really a warning.
             "caught ENVOY_SIGTERM",
+            "internal_address_config is not configured. The existing default behaviour will trust RFC1918 IP addresses, but this will be changed in next release. Please explictily config internal address config as the migration step.",
```
mum4k, line 81: Did we try to find a way to explicitly configure this behavior? If we ignore it here, it will stop working in a non-obvious way in one of the future updates.

What would it cost to fix it now?

sebastianavila5: Yes, we have to add a field to the HTTPConnectionManagerConfig instances. I will do so in a separate PR. A quick search thorugh nighthawk codebase shows we have ~5 configs so I'm not sure if we have to change just one or all.

mum4k: Thank you, please open a Github issue to track it and add a TODO next to that override, so we don't lose track of it.

mum4k (approved):
Thank you.

PR #1225, opened 2024-09-24
Explicitly config internal address config in HTTPConnectionManager (Sep 24, 2024)

sebastianavila5:
Explicitly config internal address config in HTTPConnectionManager

In the next Envoy release, no IP addresses will be considered trusted. 

Therefore, we specify internal address config for each HTTPConnectionManager used in the Nighthawk test servers.

Signed-off-by: Sebastian Avila <sebastianavila@google.com>


mum4k (approved):
Thank you!

PR #1241, opened 2024-10-18
Update Envoy to 4210340 (Oct 18, 2024)

fei-deng:
- Update Envoy commit number and SHA.
- Update tools/code_format/config.yaml.
- Re-tune test_http_h2_mini_stress_test_without_client_side_queueing. The knee rps is between 9.5k-10k, set it to 7k.


test/integration/test_integration_basics.py
```diff
@@ -149,7 +149,7 @@ def test_http_h2_mini_stress_test_without_client_side_queueing(http_test_server_
 def test_http_h1_mini_stress_test_open_loop(http_test_server_fixture):
   """Run an H1 open loop stress test. We expect higher pending and overflow counts."""
   counters = _mini_stress_test(http_test_server_fixture, [
-      http_test_server_fixture.getTestServerRootUri(), "--rps", "10000", "--max-pending-requests",
+      http_test_server_fixture.getTestServerRootUri(), "--rps", "1000", "--max-pending-requests",
```
mum4k, line 152: How many overflows is this causing now approximately? Is it significantly more than 10?

A 10 fold decrease is fine, but want to make sure we're not going to make the test flaky.

fei-deng:
re-tuned and set rps at 70% of the value around the knee.

test/integration/test_integration_basics.py
fei-deng: re-tuned and set rps at 70% of the value around the knee.

mum4k: Thank you!

mum4k (approved):

PR #1248, opened 2024-11-01
WIP: NH on Github Actions.

mum4k:


.github/workflows/nighthawk_checks.yml
```diff
@@ -0,0 +1,20 @@
+name: Nighthawk Envoy CI/CD
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, reopened]
```
phlax, line 8:
```suggestion
```

i think this is not necessary

.github/workflows/nighthawk_checks.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  check:
+    runs-on: ubuntu-latest  # Equivalent to your 'envoy-x64-large' pool
```
phlax, line 12: this will give you the default github-hosted vm - you will need to use the label i posted `envoy-x64-medium` for larger machine

also - as a general point - best not to use `-latest` as its better to pin to a version and change it explicitly

.github/workflows/nighthawk_checks.yml
```diff
@@ -0,0 +1,20 @@
...
+    strategy:
```
phlax, line 13: `fail-fast: false` ?

phlax (commented):
opinionated format pushback - use least indent for yaml lists

mum4k:
Thank you, comments addressed when adding the stub workflow. Closing this one.

PR #1251, opened 2024-11-01
rbe: Testing

phlax:


.github/workflows/nighthawk-checks.yml
```diff
@@ -8,6 +8,9 @@ on:
 
 jobs:
   check:
+    permissions:
+      contents: read
+      packages: read
```
phlax, line 13: `packages: read` permission required for remote access

.github/workflows/nighthawk-checks.yml
```diff
@@ -17,4 +20,9 @@ jobs:
     - uses: actions/checkout@v3
     - name: Run CI script
       run: |
-        echo "Hello github"
+        bazel run \
+            --config=remote-envoy-engflow \
+            --@envoy//tools/cat:target=//:hello \
+            @envoy//tools/cat
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
```
phlax, line 28: github token needs to be made available

.github/workflows/nighthawk-checks.yml
```diff
@@ -17,4 +20,9 @@ jobs:
     - uses: actions/checkout@v3
     - name: Run CI script
       run: |
-        echo "Hello github"
+        bazel run \
+            --config=remote-envoy-engflow \
```
phlax, line 24: for rbe+cache+bes use the `--config=remote-envoy-engflow` flag

test/integration/integration_test.py
```diff
@@ -3,26 +3,30 @@
...
+import multiprocessing
 import pytest
 
 from test.integration import utility
 
 if __name__ == '__main__':
   path = os.path.dirname(os.path.realpath(__file__))
   test_selection_arg = sys.argv[1] if len(sys.argv) > 1 else ""
+  num_cores = multiprocessing.cpu_count()
+  num_workers = max(1, min(num_cores - 1, 4 if utility.isSanitizerRun() else num_cores))
```
phlax, line 15: assuming that you dont want to actually test the kernels abitlity to context switch - this shouldnt be set to more than the actual cores available

.bazelrc
```diff
@@ -70,11 +70,15 @@ build --incompatible_config_setting_private_default_visibility
 build --incompatible_enforce_config_setting_visibility
 
 test --test_verbose_timeout_warnings
-test --experimental_ui_max_stdouterr_bytes=11712829 #default 1048576
+test --experimental_ui_max_stdouterr_bytes=120000000 #default 1048576
```
phlax, line 73: not sure if this is needed

.bazelrc
```diff
@@ -70,11 +70,15 @@ build --incompatible_config_setting_private_default_visibility
...
 
 # Allow tags to influence execution requirements
 common --experimental_allow_tags_propagation
 
+# Test configuration flags
+# Enable stress tests (expensive tests that are skipped by default)
```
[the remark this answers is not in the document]
phlax: in which case we want to set this to false in .bazelrc i think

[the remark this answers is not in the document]
phlax: or just make the default false, which i think is better anyway in that case

.bazelrc
phlax: lets remove it can be upped if needed

.github/workflows/nighthawk.yml
```diff
@@ -0,0 +1,80 @@
...
+        - test_gcc
+    uses: ./.github/workflows/_ci.yml
+    with:
+      task: ${{ matrix.target }}
+
+  sanitizers:
+    permissions:
+      contents: read
+      packages: read
+    needs: test
```
phlax, line 50: @mum4k i would just fold these into the test - it will speed everything up no end

mum4k: +1 if we have the resources to run these together, it will be much better.

.bazelrc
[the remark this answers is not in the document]
mum4k: Right, the intention is to run the stress tests when running in Github actions and avoid running them on local executions.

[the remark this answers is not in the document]
phlax: ah - i misread this - this is actually correct/good - it enables iff `--config=stress` - ill update the README

mum4k (approved):

test/integration/integration_test.py
```diff
@@ -3,26 +3,28 @@
...
+
   r = pytest.main(
       [
           "--rootdir=" + path,
-          "-vvvv",
-          "--showlocals",  # Don't abbreviate/truncate long values in asserts.
```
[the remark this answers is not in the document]
phlax: ive never used it, and i use pytest a lot - just dont think its necessary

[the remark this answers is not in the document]
phlax: yeah - the `-vvvv` defo is intentional - but i dont think the showlocals flags helps either

PR #1270, opened 2025-01-01
Use StrCat for string concatenation

yurykats:
This makes the code compatible with a future Protobuf release where various methods return absl::string_view instead of std::string or const std::string&.


mum4k:
Thank you @yurykats, looks like the CI is complaining about formatting, would you mind running `ci/do_ci.sh fix_format`?

mum4k (approved):

PR #1286, opened 2025-02-03
Add new field for customizable JSON request body

Pelinthecoder:
Add json_body field to RequestOptions to support dynamic request payloads.

This allows Nighthawk to generate structured request bodies for testing scenarios requiring specific JSON payloads.

### DIFF
--- a/api/client/options.proto
+++ b/api/client/options.proto
@@ -20,6 +20,8 @@ message RequestOptions {
   // Our StreamDecoder depends on bounding the size here, so if this changes, an amendment
   // to that is needed as well.
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
+  // New field for customizable JSON request body
+  string json_body = 4;
 }
 
 // Used for providing multiple request options, especially for RequestSourcePlugins.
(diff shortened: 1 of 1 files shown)


api/client/options.proto
```diff
@@ -20,6 +20,8 @@ message RequestOptions {
   // Our StreamDecoder depends on bounding the size here, so if this changes, an amendment
   // to that is needed as well.
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
+  // New field for customizable JSON request body
```
jiajunye, line 23: I would remove New field from the comment. How about "Customizable JSON request body"?

jiajunye:
LGTM

api/client/options.proto
```diff
@@ -20,6 +20,8 @@ message RequestOptions {
   // Our StreamDecoder depends on bounding the size here, so if this changes, an amendment
   // to that is needed as well.
   google.protobuf.UInt32Value request_body_size = 3 [(validate.rules).uint32 = {lte: 4194304}];
+  // Customizable JSON request body
```
mum4k, line 23: Let's expand the comment a bit to make it clearer how it should be used and what Nighthawk does when this is set. I suggest the following:

```proto
  // Optional body of the HTTP request.
  // If this is provided, Nighthawk sets the "Content-Type" header to "application/json"
  // and issues the request with the body provided here.
  // If this isn't provided, Nighthawk sends its built-in request body (the character 'a'
  // repeated n times to the specified request size).
  string json_body = 4;
```

mum4k (changes requested):

mum4k (approved):

PR #1289, opened 2025-02-10
Ensure request body is correctly propagated in Nighthawk Request Source.

Pelinthecoder:
Previously, requests generated by `OptionsListRequestSource` did not correctly set the request body, leading to empty payloads when using `json_body`. This CL ensures that `json_body` is correctly propagated from `RequestOptionsList` to `RequestImpl`, aligning request behavior with expectations.

### Changes:
- **Updated Request interface (`request.h`)**: Added `body()` as a pure virtual method.
- **Modified `RequestImpl` (`request_impl.h` & `request_impl.cc`)**:
  - Updated the constructor to accept `json_body_` as an argument.
  - Implemented `body()` to return `json_body_` instead of an empty string.
- **Fixed `OptionsListRequestSource::get()`**:
  - Now correctly assigns `json_body` when creating `RequestImpl`.
  - Ensures request bodies are set properly instead of being empty.
- **Fixed `RemoteRequestSourceImpl::get()`**:
  - Ensures that request bodies from gRPC sources are also correctly populated.
- **Updated `request_source_plugin_test.cc`**:
  - Modified tests to verify that `request3->body()` is correctly set.
  - Ensured `request3` is not `nullptr` and contains the expected `json_body`.

### **Why This Change?**
Without this fix, request bodies remain empty despite being set in the configuration (`RequestOptionsList`). This caused inconsistencies in request generation. This CL ensures that the body is always populated correctly, improving reliability.

### **Testing:**
- Verified that `json_body` is correctly loaded from the YAML file.
- Confirmed `request3->body()` now returns the expected JSON payload.

This change ensures that request generation works as expected and aligns with the intended behavior.


eric846:
The description refers to changes in `OptionsListRequestSource` but `request_options_list_plugin_impl.cc` isn't modified in this PR. Are there changes that should be brought in here?

source/common/request_source_impl.cc
```diff
@@ -80,7 +82,17 @@ void RemoteRequestSourceImpl::destroyOnThread() {
 }
 
 RequestGenerator RemoteRequestSourceImpl::get() {
-  return [this]() -> RequestPtr { return grpc_client_->maybeDequeue(); };
+  return [this]() -> RequestPtr {
```
eric846, line 85: It doesn't seem like we should have to modify the RemoteRequestSourceImpl if the default body value in request_impl.h is working as intended. `maybeDequeue()` already constructs a RequestImpl (getting the default json_body "") at https://github.com/envoyproxy/nighthawk/blob/b5771920b5840ed405bad60c9609514f3f4b317d/source/common/request_stream_grpc_client_impl.cc#L83

test/request_source/test_data/test-config-ab.yaml
```diff
@@ -1,11 +1,13 @@
 options:
   - request_method: 1
     request_body_size: 10
+    json_body: '{"message": "hello"}'
```
eric846, line 4: Could we set different body values in the two requests and check both of them in the test, along with checking /a and /b?

source/common/request_impl.h
```diff
@@ -1,18 +1,23 @@
 #pragma once
 
+#include <string>
+#include <utility>
 #include "envoy/http/header_map.h"
-
 #include "nighthawk/common/request.h"
 
 namespace Nighthawk {
```
Pelinthecoder, line 9: /retest

source/common/request_source_impl.cc
```diff
@@ -1,11 +1,13 @@
 #include "source/common/request_source_impl.h"
 
 #include <chrono>
+#include <memory>
 
 #include "envoy/common/exception.h"
 
 #include "external/envoy/source/common/common/assert.h"
 
+#include "source/common/request.h"
```
fei-deng, line 10: This file does not exist, which caused build failure.

source/common/request_source_impl.cc
Pelinthecoder: Thanks for the question Eric. I checked maybeDequeue() and confirmed that it calls ProtoRequestHelper::messageToRequest(). However, that function does not propagate json_body—it only sets headers and metadata. Since RequestImpl is constructed with just the header, the request body defaults to an empty string ("").

So I believe the change in RemoteRequestSourceImpl::get() is needed to ensure that if maybeDequeue() returns a request, we correctly propagate its body when constructing RequestImpl. Otherwise, gRPC-generated requests will always have an empty body.

Let me know if you have any other insights on this. Thanks

source/common/request_impl.h
```diff
@@ -1,18 +1,23 @@
...
 
 class RequestImpl : public Request {
 public:
-  RequestImpl(HeaderMapPtr header) : header_(std::move(header)) {}
+  RequestImpl(HeaderMapPtr header, std::string json_body = "")
+      : header_(std::move(header)), json_body_(std::move(json_body)) {}
```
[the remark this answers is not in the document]
Pelinthecoder: Thanks for catching that.

source/common/request_source_impl.cc
eric846: `RemoteRequestSourceImpl` will only support customizing the request body in the future when a body field is added to another proto (the `RemoteRequestSource`-specific `RequestSpecifier` proto) and plumbed to the `RequestImpl` here: https://github.com/envoyproxy/nighthawk/blob/b5771920b5840ed405bad60c9609514f3f4b317d/source/common/request_stream_grpc_client_impl.cc#L85-L124. Until that is done, `request->body()` will always be "" anyway.

Updating the remote request source is out of scope for this project, so it's ok for the `Request` coming out of `maybeDequeue()` to have an empty body.

source/request_source/request_options_list_plugin_impl.cc
eric846, line 1: Thank you for bringing this file, this makes sense.

source/common/request_impl.h
```diff
@@ -1,18 +1,23 @@
 #pragma once
 
+#include <string>
```
Pelinthecoder, line 3: /retest

test/request_source/test_data/test-config-ab.yaml
```diff
@@ -1,11 +1,13 @@
 options:
   - request_method: 1
-    request_body_size: 10
+    request_body_size: 23
```
fei-deng, line 3: A general question here :
for Json_body, if the user already provided the content already, do we need the user to provide the request_body_size?

source/common/request_source_impl.cc
Pelinthecoder: Than you for all the insights. That part is reverted.

eric846 (approved):

test/request_source/test_data/test-config-ab.yaml
fei-deng: 1. agree to address the `request_body_size` and `json_body` conflict in a separate PR.
2. in the next PR, we need to fix the second issue. Use a separate yaml and test case to test Json_body.

Pelinthecoder: Thanks Eric, Appreciate the insights! +1 to introduce a dedicated yaml file for testing.

PR #1290, opened 2025-02-11
Fix possible crash while using user defined plugins

tomjzzhang:
The crash can occur when onComplete is called before response headers are available, but it tries to dereference the pointer. onComplete can be called when the client or server reset the stream / connection, which is a common case, especially under heavy load.

A more robust fix would be changing StreamDecoderCompletionCallback::onComplete to take Envoy::OptRef for the header, and actually check on whether it has value.

The fix currently works because there is only one call site that might have nullptr dereference, which happens to set `success` to false: https://github.com/envoyproxy/nighthawk/blob/b5771920b5840ed405bad60c9609514f3f4b317d/source/client/stream_decoder.cc#L99


eric846:
Addressed in https://github.com/envoyproxy/nighthawk/pull/1308

PR #1292, opened 2025-02-13
Update stream_decoder.cc

Pelinthecoder:


source/client/stream_decoder.cc
```diff
@@ -118,20 +118,26 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
   stream_info_.upstreamInfo()->upstreamTiming().onFirstUpstreamTxByteSent(
       time_source_); // XXX(oschaaf): is this correct?
   const Envoy::Http::Status status =
-      encoder.encodeHeaders(*request_headers_, request_body_size_ == 0);
+      encoder.encodeHeaders(*request_headers_, false);
```
fei-deng, line 121: Why do you hard coded request_body_size==0 to false  here? This changes the current behavior and assumes that every request has the request_body_size != 0. Is that by intention?

source/client/stream_decoder.cc
```diff
@@ -118,20 +118,26 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
...
                           *request_headers_);
   }
-  if (request_body_size_ > 0) {
+  std::string json_body = "";
+  auto json_body_header = request_headers_->get(Envoy::Http::LowerCaseString("x-json-body"));
+  if (json_body_header.empty()) {
+    json_body = json_body_header[0]->value().getStringView();
+  }
+  
+  if (json_body.empty()) {
```
fei-deng, line 134: this stream_decoder not only handle Jason payload case. Please don't change its behavior by assuming all requests have json_body.

source/client/stream_decoder.cc
```diff
@@ -118,20 +118,26 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
...
     ENVOY_LOG_EVERY_POW_2(error,
                           "Request header encoding failure. Might be missing one or more required "
                           "HTTP headers in {}.",
                           *request_headers_);
   }
-  if (request_body_size_ > 0) {
+  std::string json_body = "";
+  auto json_body_header = request_headers_->get(Envoy::Http::LowerCaseString("x-json-body"));
+  if (json_body_header.empty()) {
+    json_body = json_body_header[0]->value().getStringView();
```
fei-deng, line 131: I'm not sure I understand it. Where do we set the "x-json-body" header and how can you get json_body from it? 

We should let StreamDecoder to construct with request_body information (like request_headers_ or request_body_size_) and use it in the `onPoolReady` method to replace the static content (staticUploadContent().data()).

fei-deng: not sure I understand what value we expect to get from an empty json_body_header.

PR #1295, opened 2025-02-16
Propagates JSON Body to Stream Decoder

Pelinthecoder:
This change modifies the StreamDecoder to correctly handle JSON payloads provided through RequestOptions.

### DIFF
--- a/source/client/benchmark_client_impl.cc
+++ b/source/client/benchmark_client_impl.cc
@@ -191,8 +191,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
       dispatcher_, api_.timeSource(), *this, std::move(caller_completion_callback),
       *statistic_.connect_statistic, *statistic_.response_statistic,
       *statistic_.response_header_size_statistic, *statistic_.response_body_size_statistic,
-      *statistic_.origin_latency_statistic, request->header(), shouldMeasureLatencies(),
-      content_length, generator_, tracer_, latency_response_header_name_);
+      *statistic_.origin_latency_statistic, request->header(), request->body(),
+      shouldMeasureLatencies(), content_length, generator_, tracer_, latency_response_header_name_);
   requests_initiated_++;
   pool_data.value().newStream(*stream_decoder, *stream_decoder,
                               {/*can_send_early_data_=*/false,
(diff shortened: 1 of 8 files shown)


source/client/benchmark_client_impl.cc
```diff
@@ -191,8 +191,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
       dispatcher_, api_.timeSource(), *this, std::move(caller_completion_callback),
       *statistic_.connect_statistic, *statistic_.response_statistic,
       *statistic_.response_header_size_statistic, *statistic_.response_body_size_statistic,
-      *statistic_.origin_latency_statistic, request->header(), shouldMeasureLatencies(),
-      content_length, generator_, tracer_, latency_response_header_name_);
+      *statistic_.origin_latency_statistic, request->header(), request->body(),
+      shouldMeasureLatencies(), content_length, generator_, tracer_, latency_response_header_name_);
```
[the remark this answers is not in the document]
Pelinthecoder: /retest

Pelinthecoder, line 195: /retest

source/client/stream_decoder.cc
```diff
@@ -132,15 +132,21 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
...
-    // Revisit this when we have non-uniform request distributions and on-the-fly reconfiguration in
-    // place. The string size below MUST match the cap we put on RequestOptions::request_body_size
-    // in api/client/options.proto!
-    auto* fragment = new Envoy::Buffer::BufferFragmentImpl(
-        staticUploadContent().data(), request_body_size_,
-        [](const void*, size_t, const Envoy::Buffer::BufferFragmentImpl* frag) { delete frag; });
-    Envoy::Buffer::OwnedImpl body_buffer;
-    body_buffer.addBufferFragment(*fragment);
-    encoder.encodeData(body_buffer, true);
+    if (request_body_.empty()) {
```
eric846, line 135: Currently we will only reach this point if `request_body_size_ > 0`, which is the case only if the user set the `Content-Length` header explicitly, or the `OptionsListRequestSource` set the `Content-Length` header from `request_body_size` in `RequestOptions` proto.

The outer `if` statement could be updated to

`if (request_body_size_ > 0 || !request_body_.empty())` 

so that the json body takes effect independent of the legacy `request_body_size_`/`Content-Length` mechanism.

source/client/stream_decoder.cc
```diff
@@ -125,22 +125,28 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
                           "HTTP headers in {}.",
                           *request_headers_);
   }
-  if (request_body_size_ > 0) {
```
Pelinthecoder, line 128: /retest

PR #1296, opened 2025-02-18
Feature/propagate jsonbody to streamdecoder3

Pelinthecoder:
pushing to run the CI pipeline


source/client/benchmark_client_impl.cc
```diff
@@ -191,8 +191,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
       dispatcher_, api_.timeSource(), *this, std::move(caller_completion_callback),
       *statistic_.connect_statistic, *statistic_.response_statistic,
       *statistic_.response_header_size_statistic, *statistic_.response_body_size_statistic,
-      *statistic_.origin_latency_statistic, request->header(), shouldMeasureLatencies(),
-      content_length, generator_, tracer_, latency_response_header_name_);
+      *statistic_.origin_latency_statistic, request->header(), request->body(),
+      shouldMeasureLatencies(), content_length, generator_, tracer_, latency_response_header_name_);
```
Pelinthecoder, line 195: /retest

Pelinthecoder, line 195: /retest

source/client/stream_decoder.cc
```diff
@@ -125,23 +125,29 @@ void StreamDecoder::onPoolReady(Envoy::Http::RequestEncoder& encoder,
                           "HTTP headers in {}.",
                           *request_headers_);
   }
-  if (request_body_size_ > 0) {
+  if (request_body_size_ > 0 || !request_body_.empty()) {
```
eric846, line 128: It looks like the `addBytesReceived` call below should do something different in the two cases.

source/client/stream_decoder.h
```diff
@@ -45,8 +45,8 @@ class StreamDecoder : public Envoy::Http::ResponseDecoder,
                 OperationCallback caller_completion_callback, Statistic& connect_statistic,
                 Statistic& latency_statistic, Statistic& response_header_sizes_statistic,
                 Statistic& response_body_sizes_statistic, Statistic& origin_latency_statistic,
-                HeaderMapPtr request_headers, bool measure_latencies, uint32_t request_body_size,
-                Envoy::Random::RandomGenerator& random_generator,
+                HeaderMapPtr request_headers, std::string request_body, bool measure_latencies,
```
eric846, line 48: In a followup we should reevaluate the lifetime of the request body and consider string_view or string& parameters in various places.

Currently we are doing a very safe option, both passing strings by value and storing the value in the StreamDecoder, but there may be some avoidable copies that can carefully be eliminated. For now we can proceed as-is despite the small overhead of copying "".

test/stream_decoder_test.cc
```diff
@@ -167,13 +170,61 @@ TEST_F(StreamDecoderTest, LatencyIsMeasured) {
...
+  delete decoder;
+}
+
+TEST_F(StreamDecoderTest, NonEmptyRequestBodyWithNonZeroRequestBodySize) {
+  std::string json_body = R"({"Message": "Hello"})";
+  Envoy::Buffer::OwnedImpl json_buf(json_body);
+  auto decoder = new StreamDecoder(
+      *dispatcher_, time_system_, *this, [](bool, bool) {}, connect_statistic_, latency_statistic_,
+      response_header_size_statistic_, response_body_size_statistic_, origin_latency_statistic_,
+      request_headers_, json_body, false, 20, random_generator_, tracer_, "");
```
eric846, line 203: IIUC the 20 can now become 0 and it should still work, because the behavior is determined by json_body being nonempty.

eric846 (changes requested):

test/stream_decoder_test.cc
fei-deng: +1 we better set it to 0.

source/client/benchmark_client_impl.cc
```diff
@@ -191,8 +191,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
       dispatcher_, api_.timeSource(), *this, std::move(caller_completion_callback),
       *statistic_.connect_statistic, *statistic_.response_statistic,
       *statistic_.response_header_size_statistic, *statistic_.response_body_size_statistic,
```
Pelinthecoder, line 193: /retest

eric846 (dismissed):

source/client/benchmark_client_impl.cc
```diff
@@ -191,8 +191,8 @@ bool BenchmarkClientHttpImpl::tryStartRequest(CompletionCallback caller_completi
...
-      *statistic_.origin_latency_statistic, request->header(), shouldMeasureLatencies(),
-      content_length, generator_, tracer_, latency_response_header_name_);
```
Pelinthecoder, line 195: /retest

eric846 (approved):

test/stream_decoder_test.cc
```diff
@@ -167,13 +170,61 @@ TEST_F(StreamDecoderTest, LatencyIsMeasured) {
...
+      .Times(1)
+      .WillOnce(testing::DoAll(SaveArg<0>(&captured_encoder_body_input_buf)));
+  decoder->onPoolReady(stream_encoder, ptr, stream_info,
+                       {} /*absl::optional<Envoy::Http::Protocol> protocol*/);
+  decoder->decodeHeaders(std::move(test_header_), false);
+  EXPECT_EQ(captured_encoder_body_input_buf.toString(), expected_body);
+  delete decoder;
+}
+
+TEST_F(StreamDecoderTest, NonEmptyRequestBodyWithNonZeroRequestBodySize) {
```
fei-deng, line 197: We should update the name of the test case.

PR #1299, opened 2025-02-20
address content type and content length for json payload

fei-deng:
- If json_body is provided, we should use it to set the content length in request header.
- If json_body is provided, we should set the content type as "application/json" no matter if it is provided in the header.


Pelinthecoder (approved):
Lokks good to me, thanks for the changes!

eric846 (approved):

PR #1300, opened 2025-02-21
Updates test names to reflect the removal of request_body_size

Pelinthecoder:


test/stream_decoder_test.cc
```diff
@@ -170,7 +170,7 @@ TEST_F(StreamDecoderTest, LatencyIsMeasured) {
   EXPECT_EQ(1, stream_decoder_export_latency_callbacks_);
 }
 
-TEST_F(StreamDecoderTest, EmptyRequestBodyWithNonZeroRequestBodySize) {
+TEST_F(StreamDecoderTest, EmptyRequestBody) {
```
fei-deng, line 173: The old name looks totally fine and explains the test case well, so let's keep the original one.

Pelinthecoder: Makes sense, pushed the changes accordingly.

fei-deng (approved):

Issue #1312, opened 2025-03-12
error while loading shared libraries: libatomic.so.1


zhaohuabing:
docker run envoyproxy/nighthawk-dev:latest nighthawk_client       

```             
nighthawk_client: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
```

eric846:
Sorry about that. The -lrt -latomic was needed for the CI build to work after updating to the latest Envoy. But these flags were actually not needed to build on my local Debian.

I would like to find a way to enable those flags only when building on CI, so we could remove the flags by default, including when building for Docker.

There are signs that these libraries are complicated, if you search for "In particular" here: https://github.com/envoyproxy/envoy/issues/1069

eric846:
Meanwhile if it's possible to edit the docker file locally before you run it, you might be able to fix it with:

apk add libatomic

zirain:
@eric846 there's no image base on tag, we can send a workaround fix with `apk add libatomic` in Dockerfile for short term?

eric846:
I see what you mean about the tags on docker hub. The most recent tag is 3 years old.

Unfortunately I tried the `apk add libatomic` workaround locally and it doesn't seem to work:

```
sudo docker run --rm -it envoyproxy/nighthawk-dev:latest ash

/ # nighthawk_client
nighthawk_client: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

/ # apk add libatomic
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
(1/1) Installing libatomic (10.2.1_pre1-r3)
Executing glibc-bin-2.33-r0.trigger
/usr/glibc-compat/sbin/ldconfig: /usr/glibc-compat/lib/ld-linux-x86-64.so.2 is not a symbolic link

OK: 17 MiB in 18 packages

/ # nighthawk_client
nighthawk_client: /usr/lib/libatomic.so.1: no version information available (required by nighthawk_client)
Inconsistency detected by ld.so: dl-lookup.c: 105: check_match: Assertion `version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!
/ # 
```


Please let me know if you find a workaround with the image. We will keep trying to find a solution.

eric846:
[...]

PR #1314, opened 2025-03-13
Fix Docker container after adding -latomic

eric846:
In https://github.com/envoyproxy/nighthawk/pull/1309 we added `-latomic -lrt`, but this caused https://github.com/envoyproxy/nighthawk/issues/1312. The binaries in the Docker image can no longer start, failing to find `libatomic.so.1`.

`apk add libatomic` in the Docker container does not work because of unknown Alpine Linux issues.

Instead we now build separate static versions of the Nighthawk binaries, solely for the Nighthawk in Docker.

Statically linking may lead to problems: https://groups.google.com/g/envoy-users/c/XK9M-RhWrxg/m/bgvUqtG2BwAJ.

Confirmed that the static Nighthawk binaries solve the `libatomic.so.1` error and apparently fully start:
```
ci/do_ci.sh docker
...
docker run envoyproxy/nighthawk-dev:latest nighthawk_client
Bad argument: A URI or --multi-target-* options must be specified.
```

It is worth a try publishing this version, since it can't get worse than the status quo where the Nighthawk Docker image doesn't work.

If the statically linked version fails, another possibility is to switch to Ubuntu as in https://github.com/envoyproxy/envoy/blob/main/ci/Dockerfile-envoy.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,11 +21,36 @@ filegroup(
     ],
 )
 
+filegroup(
+    name = "nighthawk_static",
+    srcs = [
+        ":nighthawk_adaptive_load_client_static",
+        ":nighthawk_client_static",
+        ":nighthawk_output_transform_static",
+        ":nighthawk_service_static",
+        ":nighthawk_test_server_static",
+    ],
+)
+
 envoy_cc_binary(
     name = "nighthawk_adaptive_load_client",
     linkopts = [
         "-latomic",
         "-lrt",
+    ],
+    repository = "@envoy",
+    deps = [
+        "//source/exe:adaptive_load_client_entry_lib",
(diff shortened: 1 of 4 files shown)


zirain:
@eric846 can you add a check for the CI?

```
docker run envoyproxy/nighthawk-dev:latest nighthawk_client --help
```

eric846:
Now that we have a known-good Ubuntu workaround in https://github.com/envoyproxy/nighthawk/issues/1312, I can try to migrate the container to Ubuntu, which is simpler.

eric846:
Closing in favor of https://github.com/envoyproxy/nighthawk/pull/1315.

PR #1369, opened 2025-06-18
Update Nightawk to Envoy 82ccdbc0f4a78441e776a9fbb793b59831077f77

ebyerly:
Following https://github.com/envoyproxy/nighthawk/blob/main/docs/root/updating_envoy_dependency.md


mum4k:
Thank you for getting this started @ebyerly. It was finalized in https://github.com/envoyproxy/nighthawk/pull/1372, so the NH update gets in this week without creating friction for the Envoy import into Google next week.

PR #1372, opened 2025-06-19
Update Envoy to 9ceb376 (Jun 19, 2025)

eric846:
- Major .bazelrc update after https://github.com/envoyproxy/envoy/pull/39755
  - Not taking the upstream update `build:linux --action_env=BAZEL_LINKOPTS=-lm:-fuse-ld=gold` which breaks my local `ci/do_ci.sh build` with `clang-18: error: invalid linker name in argument '-fuse-ld=gold'`
- `Http1PoolImpl` started taking an `OverloadManager` argument; created a `NullOverloadManager` in `NighthawkServerInstance` to be used for that


.bazelrc
```diff
@@ -430,29 +407,26 @@ build:docker-sandbox --experimental_enable_docker_sandbox
...
-build:docker-tsan --config=rbe-toolchain-clang-libc++
-build:docker-tsan --config=rbe-toolchain-tsan
+build:docker-tsan --config=rbe-toolchain-clang
+build:docker-tsan --config=tsan
 
 # CI configurations
 build:remote-ci --config=ci
 build:remote-ci --remote_download_minimal
-
 build:remote-ci-download --config=ci  # unique
```
mum4k, line 430: Do we still need these two unique lines? Nothing seems to be using `build:remote-ci-download`.

.bazelrc
```diff
@@ -367,27 +345,25 @@ build:remote-windows --remote_download_toplevel
 build:remote-clang --config=remote
 build:remote-clang --config=rbe-toolchain-clang
```
mum4k, line 347: Ithink adding one additional newline here would alliminate a diffe between ours and Envoy's `.bazelrc`. Bringing us to a state where all unique lines are marked unique.

.bazelrc
```diff
@@ -34,6 +34,9 @@ build --tool_java_runtime_version=remotejdk_11
 # build --platform_mappings=""                                                               # unique
```
mum4k, line 34: For these lines above, since `clang-asan` no longer exists and the tests are passing, can we remove the unique lines and bring this closer to Envoy?

```
build:clang-asan --test_timeout=900                                                         # unique
build:clang-tsan --test_timeout=900                                                          # unique
```

mum4k (changes requested):
I am pretty sure we need to modify `ci/do_ci.sh` here, since the configs `clang-asan` and `clang-tsan` no longer exist in `.bazelrc`. So the builds pass, but they are no longer sanitizing:

https://github.com/envoyproxy/nighthawk/blob/cfb23207741025fa8fa0849455fb479620e982a2/ci/do_ci.sh#L334-L342

The right thing seems to be renaming them to `asan` and `tsan`, so that bazel is executed as follows:

```
For a tsan run:
bazel ... --config=tsan --config=libc++

For an asan run:
bazel ... --config=asan --config=libc++
```

Current execution runs:

```
run_on_build_parts: running command run_bazel build  --verbose_failures  --experimental_generate_json_trace_profile --config=remote-ci --config=remote-envoy-engflow --test_env=UBSAN_OPTIONS=print_stacktrace=1 --cache_test_results=no --test_output=errors --test_env=ENVOY_IP_TEST_VERSIONS=v4only  -c dbg --config=clang-asan --config=libc++ -- //source/server/...
```

Where `-config=clang-asan` has no effect.

From: https://github.com/envoyproxy/nighthawk/actions/runs/15748386960/job/44389479296

.bazelrc
mum4k: Ironically I think these two lines are the reason why the CI is still "passing" even though the sanitizers aren't running anymore. Envoy removed `clang-tsan` configs, but we don't get an error about missing configs even though our CI script still refers to them, because these to liones make `clang-tsan` and `clang-asan` into valid configs.

.bazelrc
eric846: Removed.

.bazelrc
eric846: That's how they get you! Removed those.

.bazelrc
eric846: Had to bring these back, docker was using them.

eric846:
Confirmed that the CI is now doing such commands as

`run_on_build_parts: running command run_bazel build  --verbose_failures  --experimental_generate_json_trace_profile --config=remote-ci --config=remote-envoy-engflow --test_env=UBSAN_OPTIONS=print_stacktrace=1 --cache_test_results=no --test_output=errors --test_env=ENVOY_IP_TEST_VERSIONS=v4only  -c dbg --config=asan --config=libc++ -- //api/...`

Source:
https://github.com/envoyproxy/nighthawk/actions/runs/15783992837/job/44496405869?pr=1372

eric846:
> Based on seeing tsan time out once after 600s, restored the 900s timeouts under the new name:
> 
> ```
> build:asan --test_timeout=900                                                         # unique
> build:tsan --test_timeout=900                                                         # unique       
> ```

Indeed in a passing run `tsan` was reported to take 16m total, clearly not all of which was in bazel. So 900s may be a good start.

PR #1379, opened 2025-06-28
CONNECT/CONNECT-UDP Encapsulation Support

asingh-g:
To be able to load test Terminating HTTP CONNECT/CONNECT-UDP Proxies, I've implemented encapsulation by spinning up a second envoy as a subprocess and routing nighthawk's generated traffic through it. 

Presently CONNECT-UDP is only supported for HTTP/2 in envoy, and CONNECT is supported for all of H1, H2 and H3 QUIC.
Therefore only following combinations are supported
H1,H2 over H1,H2,H3
and
H3 over H2
using a combination of the --protocol and --tunnel-protocol flags.


api/client/options.proto
```diff
@@ -162,6 +162,34 @@ message CommandLineOptions {
...
+  // Options for routing requests via a proxy. if set, requests
+  // are encapsulated and forwarded to a terminating proxy running at
+  // tunnel_uri.
+  // When the oneof_protocol field is set to H1 or H2, an HTTP CONNECT
+  // tunnel is established with the proxy
+  // When the oneof_protocol field is set to H3, a CONNECT-UDP
+  // upgrade is used instead
+  message TunnelOptions {
+    // URI to the proxy.
+    string tunnel_uri = 115;
```
mum4k, line 174: (nit) This being its own message, can we number its fields starting from 1?

api/client/options.proto
```diff
@@ -162,6 +162,34 @@ message CommandLineOptions {
...
+    // the top level protocol.
+    Protocol tunnel_protocol = 116;
+    // TLS context for the proxy.
+    // TLS configuration is required for HTTP/3 tunnels.
+    envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext tunnel_tls_context = 117;
+    // Protocol options for when HTTP/3 tunnels are used
+    envoy.config.core.v3.Http3ProtocolOptions tunnel_http3_protocol_options = 118;
+    // Concurrency of the encapsulation server.
```
mum4k, line 182: Curious for your reasons to expose this as a knob. I would naively think that setting it to the same concurrency as the other Nighthawk process should be sufficient. Do the initial performance data show that the encapsulating process is more resource heavy than the Nighthawk that generates load?

The laod generating process is usually more resource heavy than anything else on the data plane chain, due to it's need to serialize / deserialize and record statistics.

source/client/options_impl.cc
```diff
@@ -25,6 +32,71 @@ using ::nighthawk::client::Protocol;
 #define TCLAP_SET_IF_SPECIFIED(command, value_member)                                              \
   ((value_member) = (((command).isSet()) ? ((command).getValue()) : (value_member)))
 
+// Obtains an available TCP or UDP port. Throws an exception if one cannot be
+// allocated.
+uint16_t OptionsImpl::GetAvailablePort(bool udp) {
```
mum4k, line 37: Do we know if there is an Envoy / other library that implements this that we could reuse? If we do need to implement it, would you mind placing it in its own library with a couple of unit tests? That might help with supporting it going forward.

source/client/process_impl.cc
```diff
@@ -38,6 +38,7 @@
 #include "external/envoy/source/common/singleton/manager_impl.h"
 #include "external/envoy/source/common/stats/tag_producer_impl.h"
 #include "external/envoy/source/common/thread_local/thread_local_impl.h"
+#include "external/envoy/source/exe/main_common.h"
```
mum4k, line 41: Additional ask for test coverage if we may. The `process_impl` library has existing unit tests. Can we expand them to cover the new behaviors, so that we don't depend only on the integratiom test for test coverage?

mum4k (commented):
Thank you @asingh-g, glad to see this landing!

I note this is still marked as Draft, leaving a few observations below to get us started. Please mark it as ready for review once ready. I will be ooo for two weeks, but @eric846 will be able to help with review and merge.

source/client/process_impl.cc
```diff
@@ -94,6 +95,13 @@ class BootstrapFactory : public Envoy::Logger::Loggable<Envoy::Logger::Id::main>
     // affinity is set / we don't have affinity with all cores, we should default to autoscale.
     // (e.g. we are called via taskset).
     uint32_t concurrency = autoscale ? cpu_cores_with_affinity : std::stoi(options.concurrency());
+    if (!options.tunnelUri().empty() && options.tunnelConcurrency() == "auto") {
+      // Divide concurrency in half
+      concurrency = concurrency / 2;
```
asedeno, line 100: nit: `concurrency /= 2;`

asedeno (commented):
I still need to read over `source/client/process_impl.cc` and the tests, but here's some comments.

api/client/options.proto
asingh-g: I see. I don't have any reason for the time being. I can remove the options for concurrency and protocol options for now.

source/client/options_impl.cc
asingh-g: The only option I've found is in the [https://github.com/envoyproxy/envoy/blob/bf2c671933942ba04ba41f6b761644421207dbc2/test/test_common/network_utility.cc#L22](test source of envoy)

I'll move mine to a lib and write tests as suggested

source/client/process_impl.cc
asingh-g: Sure thing

source/client/options_impl.cc
```diff
@@ -25,6 +32,71 @@ using ::nighthawk::client::Protocol;
...
+  int family = (address_family_ == nighthawk::client::AddressFamily::V4) ? AF_INET : AF_INET6;
+  int sock = socket(family, udp ? SOCK_DGRAM : SOCK_STREAM, udp ? 0 : IPPROTO_TCP);
+  if (sock < 0) {
+    throw NighthawkException(absl::StrCat("could not create socket: ", Envoy::errorDetails(errno)));
+    return 0;
+  }
+
+  // Reuseaddr lets us start up a server immediately after it exits
+  int one = 1;
+  if (setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one)) < 0) {
```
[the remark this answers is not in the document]
asingh-g: I don't see why not

source/client/process_impl.cc
asingh-g: Removed with the other concurrency options

source/client/process_impl.cc
```diff
@@ -94,6 +95,13 @@ class BootstrapFactory : public Envoy::Logger::Loggable<Envoy::Logger::Id::main>
     // affinity is set / we don't have affinity with all cores, we should default to autoscale.
     // (e.g. we are called via taskset).
     uint32_t concurrency = autoscale ? cpu_cores_with_affinity : std::stoi(options.concurrency());
+    if (!options.tunnelUri().empty() && options.tunnelConcurrency() == "auto") {
+      // Divide concurrency in half
```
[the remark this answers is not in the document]
asingh-g: Removed with the concurrency options

source/client/process_bootstrap.cc
```diff
@@ -1,5 +1,7 @@
 #include "source/client/process_bootstrap.h"
 
+#include <semaphore.h>
```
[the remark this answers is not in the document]
asingh-g: Yes, I'll switch it over

eric846 (dismissed):

eric846 (approved):

PR #1390, opened 2025-07-30
Envoy incremental check workflow

ebyerly:
check-envoy-commits.yml introduces a new Github Actions workflow. It runs every N hours.

It checks whether an Envoy commit after the current Nighthawk Bazel dependency requires a non-trivial merge. It does this by: looking for modifications to shared files, running the CI test process at the latest Envoy commit, and using `git bisect` if the latest Envoy commit introduces a test break.

It manages creating Github issues tracking Envoy commits that will require a non-trivial merge. It closes those issues once obsolete the Nighthawk Bazel dependency commit is a descendant of the issue commit.

The goal is to automate what is currently a manual bisection process during the weekly merge. The first extension would be to include managing a PR to increment Nighthawk's Bazel dependency to the latest possible trivial merge. The second extension would be to use an LLM to attempt to manage non-trivial merges.


ebyerly:
@phlax - added you as a reviewer because I see you've been digging into how Nighthawk manages its automation. I didn't want to spend more than a few hours poking at this first draft of a Github Action without pulling in someone with more familiarity for feedback to make sure I'm not wandering down an obviously terrible path. Feel encouraged to bounce this back if you have no cycles!

PR #1419, opened 2025-09-17
ci: push nighthawk images tagged with Github SHA

fristonio:
When using nighthawk docker image it would be helpful for downstream consumers to have a stable tag for reference instead of rely on `latest`.


eric846 (approved):

fristonio:
Ah, I didn't realize there is already an issue to track this. Thanks for quick merge.

Issue #1103, opened 2024-03-19
support request trailers in the test server


eric846:
Based on looking at the code, my guess is that the test server doesn't respond when the request has trailers. I haven't tried it though.

We only do `sendReply()` when `end_stream` is true in `decodeHeaders` or `decodeBody`: 
https://github.com/envoyproxy/nighthawk/blob/main/source/server/http_test_server_filter.cc

If the request has trailers, my guess is is:

- end_stream won't be true in decodeHeaders
- end_stream won't be true in decodeBody
- end_stream will be true in decodeTrailers, but it doesn't send a reply

The integration test fixture we're using doesn't appear to support adding request trailers.

PR #1548, opened 2026-06-16
fix: handle request trailers in the test server

immanuwell:
Description

This PR is related to #1103

If a request ends with trailers, the test server can miss the local reply and just hang.
Right now it only replies from `decodeHeaders(..., end_stream=true)` or `decodeData(..., end_stream=true)`.

This patch sends the reply from `decodeTrailers()` too, adds a regression test, and notes the behavior in docs.

Repro on `main`:
1. Build and run `bazel-bin/nighthawk_test_server`
2. Send `printf 'POST / HTTP/1.1\r\nHost: 127.0.0.1:10000\r\nTransfer-Encoding: chunked\r\nTrailer: x-test-trailer\r\n\r\n1\r\na\r\n0\r\nx-test-trailer: yep\r\n\r\n' | nc 127.0.0.1 10000`
3. The request hangs, no reply comes back.

With this patch it returns `200` with the configured body. tiny fix, real gap.

Notes for Reviewers

No API changes.

### DIFF
--- a/source/server/http_test_server_filter.cc
+++ b/source/server/http_test_server_filter.cc
@@ -72,20 +72,26 @@ void HttpTestServerDecoderFilter::sendReply(const ResponseOptions& options) {
       absl::nullopt, "");
 }
 
+bool HttpTestServerDecoderFilter::maybeSendReply() {
+  if (config_->validateOrSendError(effective_config_.status(), *decoder_callbacks_)) {
+    return false;
+  }
+  sendReply(*effective_config_.value());
+  return true;
+}
+
 Envoy::Http::FilterHeadersStatus
 HttpTestServerDecoderFilter::decodeHeaders(Envoy::Http::RequestHeaderMap& headers,
                                            bool end_stream) {
   effective_config_ =
       computeEffectiveConfiguration(config_->getStartupFilterConfiguration(), headers);
   if (end_stream) {
-    if (!config_->validateOrSendError(effective_config_.status(), *decoder_callbacks_)) {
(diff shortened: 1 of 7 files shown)


source/server/http_test_server_filter.cc
```diff
@@ -72,37 +72,42 @@ void HttpTestServerDecoderFilter::sendReply(const ResponseOptions& options) {
...
       computeEffectiveConfiguration(config_->getStartupFilterConfiguration(), headers);
   if (end_stream) {
-    if (!config_->validateOrSendError(effective_config_.status(), *decoder_callbacks_)) {
-      if (effective_config_.value()->echo_request_headers()) {
-        std::stringstream headers_dump;
-        headers_dump << "\nRequest Headers:\n" << headers;
-        request_headers_dump_ = headers_dump.str();
-      }
-      sendReply(*effective_config_.value());
+    if (effective_config_.ok() && effective_config_.value()->echo_request_headers()) {
```
eric846, line 89: I'm not sure, but I think there might be another bug that we might be able to address at the same time:

If there are trailers, that might mean decodeHeaders is always called with end_stream=false. If so, we would fail to dump the request headers to the response body.

To fix that, maybe we should just lift the `if (effective_config.ok()...` to before `if (end_stream)`. 

If you can't quickly determine if that would solve the problem correctly (possibly with an extra testcase), feel free to skip, and we can work on the case of header dumping with request trailers later.

eric846 (changes requested):

PR #1590, opened 2026-08-25
Introduce Interactive mode to Nighthawk Update Script

tomjzzhang:
Augment the update script with interactive mode. On failure, it will give the user the ability to retry, proceed or exit. 

Made this behavior default over bisecting which is a niche feature.


fei-deng (approved):
Thanks for iterating on it.

Issue #24, opened 2019-04-23
Support gRPC


oschaaf:
Add the capability to load-test the gRPC protocol.

DelusionalOptimist:
Hey there @oschaaf! :wave: 
We're looking forward to add support for gRPC load generation using Nighthawk in Meshery.
I'd like to know what's the current status on this. Is there a way which will let me generate any load on a gRPC service using Nighthawk?

mum4k:
Hi @DelusionalOptimist and thank you for reaching out. The gRPC support has been de-prioritized in favor of other Nighthawk features for now. While this remains on our roadmap, we don't plan to start working on it in the near future.

With that said, assuming you are interested in contributing this, I would be happy to support your efforts via discussions, code reviews, or other means you might find useful. Is this what you had in mind / would be interested in?

gyohuangxin:
@mum4k  Is there any updates here? I'm trying to use Nighthawk as the load-generator and test server. 
After setting up the simple nighthawk gRPC service refer to https://github.com/envoyproxy/nighthawk#nighthawk-grpc-service, and I failed to use nighthawk client to send request to this service and here is the error log:
```
./nighthawk_client -v debug 127.0.0.1:8443
[00:07:24.240580][2049658][D] Unable to use runtime singleton for feature envoy.restart_features.use_apple_api_for_dns_lookups
[00:07:24.240726][2049658][D] create DNS resolver type: envoy.network.dns_resolver.cares
[00:07:24.240792][2049658][I] Starting 1 threads / event loops. Time limit: 5 seconds.
[00:07:24.240801][2049658][I] Global targets: 100 connections and 5 calls per second.
[00:07:24.244664][2049658][D] dns resolution for 127.0.0.1 started
[00:07:24.328722][2049658][D] dns resolution for 127.0.0.1 completed with status 0
[00:07:24.328751][2049658][D] DNS resolution complete for 127.0.0.1 (1 entries, using 127.0.0.1:8443).
[00:07:24.329832][2049658][D] Unable to use runtime singleton for feature envoy.http.headermap.lazy_map_min_size
[00:07:24.344398][2049658][D] Computed configuration: static_resources {
  clusters {
    name: "0"
    type: STATIC
    connect_timeout {
      seconds: 30
    }
    circuit_breakers {
      thresholds {
        max_connections {
          value: 100
        }
        max_pending_requests {
          value: 1
        }
        max_requests {
          value: 100
        }
        max_retries {
        }
[...]

PR #1601, opened 2026-09-07
client: add --request-body-file to send a file's bytes as the request body

bpalermo:
**Description**

This PR is related to #1608

Nighthawk can only send a synthetic body (`'a'` repeated `--request-body-size` times) or, through a request source plugin, a `json_body`, which is a proto3 `string` and therefore must be UTF-8. That rules out sending an arbitrary payload such as a serialized protobuf.

This adds `--request-body-file <path>`: the file's bytes are sent verbatim on every request (binary safe). Over the gRPC service API the body travels in the new `RequestOptions.request_body` (`bytes`) field. `Content-Length` is set from the body size; no `Content-Type` is assumed, pass one with `--request-header`. `--request-body-size` behavior is unchanged, and the two are mutually exclusive.

**Notes for Reviewers**

- `OptionsImpl` reads the file at parse time (size cap, clear errors for unreadable files); `StaticRequestSourceImpl` gains an optional body; `RequestSourceFactoryImpl` passes it through. Plugin and remote request sources are unchanged.
- Testing: `//test:options_test` (verbatim binary content and proto round trip, missing file, mutual exclusion), `//test:factories_test` (body, `Content-Length`, no `Content-Type`), `test/integration/test_request_body_file.py` (full run posting a binary file to the test server). README usage regenerated; version history updated.
- Prerequisite for #1602 (gRPC unary mode, #24).

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ bazel-bin/nighthawk_client  [--user-defined-plugin-config <string>] ...
 <uint32_t>] [--transport-socket <string>]
 [--upstream-bind-config <string>]
 [--tls-context <string>]
+[--request-body-file <string>]
 [--request-body-size <uint32_t>]
 [--request-header <string>] ...
 [--request-method <GET|HEAD|POST|PUT|DELETE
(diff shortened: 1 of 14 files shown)


eric846 (dismissed):

bpalermo:
@eric846 gentle ping after fixing conflicts :pray: ?

eric846 (approved):

